PR #789, opened 2015-01-06
Making SmartStore's backing DB access more stable

bhariharan:
This ensures every access to the DB goes through ```DBOpenHelper```, which ensures we always access a DB instance that hasn't been closed.


bhariharan:
The original pull request came from the S1 team.

wmathurin:
What if the db is closed after the call to dbOpenHelper.getWritableDatabase() and before it is used?

bhariharan:
The helper is closed only on the ```deleteDb``` call, so that should never happen.

khawkins:
Looks fine to me.

wmathurin:
Is the helper returning a different db on every call?
Is there a performance impact getting a new db object every time?

bhariharan:
No - the only thing that's different here is that the DB lifecycle is in sync with the lifecycle of ```DBOpenHelper```, which ensures that a stale, already closed DB object is never hanging around.

wmathurin:
Makes sense.

bhariharan:
Any other comments/concerns before I merge?

wmathurin:
No, go ahead.

PR #791, opened 2015-01-07
Custom permissions + authentication dependent on MDM managed device/app settings

wmathurin:
Reading custom permissions from identity response / storing them alongside the custom settings
Erring out authentication if manage app perm is on but app is not managed

No tests yet
Custom settings and custom permissions are stored together in the same pref file, so if the same keys are used, they will conflict


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AdminPrefsManager.java
```diff
@@ -83,27 +78,18 @@ public void setPrefs(JSONObject attribs, UserAccount account) {
...
-	        e.commit();
-	    }
+        setPrefs(new JSONObject(attribs), account);
 	}
 
-	/**
+    private SharedPreferences getSharedPreferences(UserAccount account) {
+        String sharedPrefPath = ADMIN_PREFS;
+        if (account != null) {
+            sharedPrefPath = ADMIN_PREFS + account.getOrgLevelFilenameSuffix();
```
bhariharan: Are the ```custom_permissions``` user level or org level settings?

bhariharan:
Looks good otherwise.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AdminPrefsManager.java
wmathurin: There are user level.

bhariharan: Then we need to use ```account.getUserLevelFilenameSuffix()``` here. And if we're doing this across the board, we'll need a one time upgrade step to migrate an existing org level admin pref file to user level :-(

wmathurin: Either that, or maybe I should use other pref files for the custom perms (and they would use the user level filename prefix).

wmathurin: Maybe I should create a AdminPermsManager to store custom permissions?

bhariharan: Yeah, that's probably easier to maintain. Works for me.

wmathurin:
Refactored AdminPrefsManager / added AdminPermsManager - see https://github.com/wmathurin/SalesforceMobileSDK-Android/commit/cb6b945b9ef310c0d89a0885efcbcc720cc7f690

bhariharan:
Looks good to me. :+1:

PR #794, opened 2015-01-07
Fix for account decryption errors

bhariharan:
There doesn't appear to be any good reason to run in the ```:auth``` process. Android doesn't require it either.


wmathurin:
:+1:

PR #796, opened 2015-01-12
New option on sync down to leave alone "dirty" records

wmathurin:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -262,88 +269,100 @@ private void syncUp(SyncState sync, SyncUpdateCallback callback) throws Exceptio
 		String soupName = sync.getSoupName();
 		SyncOptions options = sync.getOptions();
 		List<String> fieldlist = options.getFieldlist();
-		QuerySpec querySpec = QuerySpec.buildExactQuerySpec(soupName, LOCAL, "true", 2000); // XXX that could use a lot of memory
```
wmathurin: New approach on sync up. Getting all the ids (looping through pages if needed) then doing retrieve and save to server in a loop one record at a time. (unrelated to the story)

native/SampleApps/SmartSyncExplorer/res/layout/list_item.xml
```diff
@@ -41,7 +41,6 @@
     </LinearLayout>
 
     <ImageView android:id="@+id/sync_status_view"
-        android:layout_toRightOf="@id/obj_layout"
```
wmathurin: Icons were not right justified (on nexus 7).

native/SampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/ui/MainActivity.java
```diff
@@ -292,6 +297,7 @@ private void handleSyncUpdate(SyncState sync) {
 				Toast.makeText(MainActivity.this,
 						"Sync down successful!",
 						Toast.LENGTH_LONG).show();
+                refreshList();
```
wmathurin: Without the refresh call, the list would remain empty until the app was paused/resumed.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTest.java
```diff
@@ -100,44 +102,54 @@ public void testGetSyncStatusForInvalidSyncId() throws JSONException {
...
-		}
-		JSONTestHelper.assertSameJSONObject("Wrong data in db", new JSONObject(idToNames), idToNamesFromDb);
+        checkDb(idToNames);
 	}
 
-	/**
+    /**
+     * Sync down the test accounts, make some local changes, sync down again with merge mode LEAVE_IF_CHANGED then sync down with merge mode OVERWRITE
+     */
+    public void testSyncDownWithoutOverwrite() throws Exception {
```
wmathurin: New test for story.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncState.java
```diff
@@ -252,4 +256,11 @@ public SyncState copy() throws JSONException {
     }
 
 
+    /**
+     * Enum for merge modes
+     */
+    public enum MergeMode {
```
wmathurin: More merge modes to come in the future (we have 4 in smartsync.js).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -437,23 +459,59 @@ private void syncDownSosl(SyncState sync, SyncUpdateCallback callback) throws Ex
...
+        }
+        return set;
+    }
 	
-	private void saveRecordsToSmartStore(String soupName, JSONArray records)
+	private void saveRecordsToSmartStore(String soupName, JSONArray records, MergeMode mergeMode)
 			throws JSONException {
-		// Save to SmartStore
+        // Gather ids of dirty records
+        Set<String> idsToSkip = null;
```
wmathurin: Approach: we get the ids of ALL the dirty records and skip them when saving records fetched from the server.
We don't do a:
select id where dirty and id in (ids fetched from server) 
because the list of ids could be to long and we would have to chunk up the query etc. 

We don't expect to many records changed locally, so the full list should be "small" most of the time.

bhariharan:
Looks good to me. :+1:

khawkins:
Yep, looks like a good first pass.

PR #805, opened 2015-01-23
Falling back on string value, if string array is not provided

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -185,23 +186,42 @@ public void reset() {
...
-	 * Returns the list of login servers from runtime configuration (from MDM provider) if any
-	 * 
+	 * Returns the list of login servers from runtime configuration
+	 * (from MDM provider), if any.
+	 *
 	 * @return List of login servers or null.
 	 */
 	public List<LoginServer> getLoginServersFromRuntimeConfig() {
 		RuntimeConfig runtimeConfig = RuntimeConfig.getRuntimeConfig(ctx);
-		String[] mdmLoginServers = runtimeConfig.getStringArray(ConfigKey.AppServiceHosts);
```
wmathurin: Maybe we should modify the RuntimeConfig:getStringArray method to do that:
- if exception, get the string and return a new array with the single string value

bhariharan: I thought about that, but I feel like that's a bit misleading and a little too broad? Its like asking a ```Bundle``` to check for multiple types.

khawkins:
:+1:

PR #812, opened 2015-01-28
Delta sync down for Android

wmathurin:


bhariharan:
Looks fine.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/phonegap/SmartSyncPlugin.java
```diff
@@ -89,6 +90,7 @@ public void run() {
 		        		  case syncUp:             syncUp(args, callbackContext); break;
 		        		  case syncDown:		   syncDown(args, callbackContext); break;
 		        		  case getSyncStatus:	   getSyncStatus(args, callbackContext); break;
+                          case reSync:             reSync(args, callbackContext); break;
```
khawkins: I think we need some code format templates for our IDEs. :)

khawkins:
Looks fine. :+1:

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/phonegap/SmartSyncPlugin.java
wmathurin: I've to check my tabs/spaces settings in Android Studio.

khawkins: It's an issue we've all participated in at one time or another. :)  Perhaps we can move to something line the [IntelliJ Android code style](https://code.google.com/p/google-styleguide/source/browse/trunk/intellij-java-google-style.xml) when we cut over to Android Studio officially?  Cc: @bhariharan

PR #813, opened 2015-01-28
Fixing branch for Cordova plugin

bhariharan:
It was incorrectly pointing to ```unstable```.

### DIFF
--- a/node/forcedroid.js
+++ b/node/forcedroid.js
@@ -42,7 +42,7 @@ var fs = require('fs');
 var cordovaHelper = require('../external/shared/node/cordovaHelper');
 var miscUtils = require('../external/shared/node/utils');
 
-var version = '3.1.0';
+var version = '3.1.1';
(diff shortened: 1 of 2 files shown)


node/forcedroid.js
```diff
@@ -42,7 +42,7 @@ var fs = require('fs');
 var cordovaHelper = require('../external/shared/node/cordovaHelper');
 var miscUtils = require('../external/shared/node/utils');
 
-var version = '3.1.0';
```
wmathurin: Do we need to bump up the version number?

bhariharan: This is only the ```npm``` package version number. We could do a force override, but that's not preferable (I don't know what will happen to the users who have already downloaded ```3.1.0```).

wmathurin: Sounds good.

wmathurin:
Looks fine.

PR #818, opened 2015-01-30
A couple of tests were broken when getAuthorizationUrl was refactored

wmathurin:
As of 3.0, no scope parameter was put in (even for refresh token) if the scopes array passed in was null or empty
Going back to 3.0 behavior

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -26,22 +26,11 @@
  */
 package com.salesforce.androidsdk.auth;
 
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -150,14 +155,14 @@ public static URI getAuthorizationUrl(URI loginServer, String clientId,
...
-        sb.append(AND).append(SCOPE).append(EQUAL).append(Uri.encode(computeScopeParameter(scopes)));
+        if (scopes != null && scopes.length > 0) sb.append(AND).append(SCOPE).append(EQUAL).append(Uri.encode(computeScopeParameter(scopes)));
         sb.append(AND).append(REDIRECT_URI).append(EQUAL).append(callbackUrl);
         return URI.create(sb.toString());
     }
 
     private static String computeScopeParameter(String[] scopes) {
         final List<String> scopesList = Arrays.asList(scopes == null ? new String[]{} : scopes);
-        Set<String> scopesSet = new HashSet<String>(scopesList);
+        Set<String> scopesSet = new TreeSet<String>(scopesList); // sorted set to make tests easier
```
wmathurin: Using a tree set so that we know in what order to find the scopes (useful in tests)

khawkins:
:+1:

bhariharan:
:+1:

PR #819, opened 2015-01-30
Background sync implementation for SmartSyncExplorer

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
```diff
@@ -354,6 +360,15 @@ private HttpURLConnection createHttpUrlConnection(URI uri, String requestMethod)
     	if (uri != null) {
     		URL url = uri.toURL();
     		if (url != null) {
+
+    			/*
+    			 * FIXME: PATCH has been added to the latest OkHttp library,
```
bhariharan: Something unrelated that I recently discovered. Made a note here, so I won't forget to fix it in the future.

native/SampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/loaders/ContactListLoader.java
```diff
@@ -88,4 +111,55 @@ public ContactListLoader(Context context, UserAccount account) {
 		}
 		return contacts;
 	}
+
+	/**
+	 * Pushes local changes up to the server.
+	 */
+	public synchronized void syncUp() {
```
bhariharan: I've moved these methods from ```MainActivity``` here, since they're better suited to the ```Loader``` (not main thread). This enables us to trigger sync operations from a background ```Service```, since loaders are not on UI thread.

native/SampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/sync/ContactProvider.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * A dummy content provider that doesn't really do anything. Sync adapters
+ * on Android require a content provider, which indirectly forces us to
+ * work with cursors and use a cursor loader. Since we use a custom loader,
+ * we simply declare this provider in the manifest, but use the loader
+ * for all sync operations.
+ *
+ * @author bhariharan
+ */
+public class ContactProvider extends ContentProvider {
```
bhariharan: Dummy provider - does nothing at all! Unfortunately, content providers are a requirement for background sync :-(

wmathurin:
Looks fine.

native/SampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/sync/ContactProvider.java
khawkins: 76 lines o' nuthin.

khawkins:
Seems good over all. :+1:

PR #822, opened 2015-02-03
First stab at merge mode LEAVE_IF_CHANGED implementation for sync up

bhariharan:
What's Left:

* Tests.
* Other odds and ends.
* Maybe an optimization to the fetch of ```LAST_MODIFIED_DATE``` - TBD.

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -364,7 +364,7 @@ public class SyncManager {
                 }
             }
         }
-        if (serverLastModified < lastModifiedDate) {
+        if (serverLastModified <= lastModifiedDate) {
         	isNewer = true;
         }
     	return isNewer;
@@ -390,13 +390,21 @@ public class SyncManager {
         // Getting type and id
         final String objectType = (String) SmartStore.project(record, Constants.SOBJECT_TYPE);
         final String objectId = record.getString(Constants.ID);
-        final long lastModifiedDate = record.optLong(SmartStore.SOUP_LAST_MODIFIED_DATE, UNCHANGED);
+        long lastModifiedDate = UNCHANGED;
+        final String lastModStr = record.optString(Constants.LAST_MODIFIED_DATE);
+        if (!TextUtils.isEmpty(lastModStr)) {
+        	try {
+        		lastModifiedDate = TIMESTAMP_FORMAT.parse(lastModStr).getTime();
+        	} catch (ParseException e) {
+        		Log.e("SmartSyncManager:syncUpOneRecord", "Error during date parsing", e);
+        	}
+        }
 
         /*
          * Checks if we are attempting to update a record that has been updated
-         * on the server AFTER the client update. If the merge mode passed in
-         * tells us to leave the record alone under these circumstances, we will
-         * do nothing and return here.
+         * on the server AFTER the client's last sync down. If the merge mode
+         * passed in tells us to leave the record alone under these
+         * circumstances, we will do nothing and return here.
          */
         if (mergeMode == MergeMode.LEAVE_IF_CHANGED &&
         		(action == Action.update || action == Action.delete) &&
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, salesforce.com, inc.
+ * Copyright (c) 2014-2105, salesforce.com, inc.
  * All rights reserved.
  * Redistribution and use of this software in source and binary forms, with or
  * without modification, are permitted provided that the following conditions
@@ -54,6 +54,7 @@ import org.json.JSONObject;
 
 import java.io.IOException;
 import java.text.DateFormat;
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
@@ -65,12 +66,11 @@ import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
-
-
 /**
  * Sync Manager
-*/
+ */
 public class SyncManager {
+
     // Constants
     public static final int PAGE_SIZE = 2000;
     private static final int UNCHANGED = -1;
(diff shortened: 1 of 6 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -337,7 +335,44 @@ private void syncUp(SyncState sync, SyncUpdateCallback callback) throws Exceptio
         }
 	}
 
-    private boolean syncUpOneRecord(String soupName, List<String> fieldlist, JSONObject record) throws JSONException, IOException {
+    private boolean isNewerThanServer(String objectType, String objectId,
```
bhariharan: Maybe there's a better way of doing this, rather than fetch it per record. The best I can think of is batching them per record type. If there's a better suggestion, I'm open to ideas.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncOptions.java
```diff
@@ -62,23 +62,32 @@ public static SyncOptions fromJSON(JSONObject options) throws JSONException {
...
 	}
-	
+
 	/**
 	 * @param fieldlist
 	 * @return
 	 */
 	public static SyncOptions optionsForSyncUp(List<String> fieldlist) {
-		return new SyncOptions(fieldlist, null);
+		return new SyncOptions(fieldlist, MergeMode.OVERWRITE);
```
bhariharan: For backward compat.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -346,15 +381,30 @@ else if (record.getBoolean(LOCALLY_CREATED))
...
             // Nothing to do for this record
             return true;
         }
 
         // Getting type and id
-        String objectType = (String) SmartStore.project(record, Constants.SOBJECT_TYPE);
-        String objectId = record.getString(Constants.ID);
+        final String objectType = (String) SmartStore.project(record, Constants.SOBJECT_TYPE);
+        final String objectId = record.getString(Constants.ID);
+        final long lastModifiedDate = record.optLong(SmartStore.SOUP_LAST_MODIFIED_DATE, UNCHANGED);
```
bhariharan: Looks like I forgot to commit some local changes. There's a bug here with the last mod date. I'll add the fix in my next commit.

bhariharan:
Ready for review now.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, salesforce.com, inc.
+ * Copyright (c) 2014-2105, salesforce.com, inc.
```
khawkins: The future is now!

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -337,7 +335,44 @@ private void syncUp(SyncState sync, SyncUpdateCallback callback) throws Exceptio
...
+    private boolean isNewerThanServer(String objectType, String objectId,
+    		long lastModifiedDate) throws JSONException, IOException {
+    	boolean isNewer = false;
+    	long serverLastModified = UNCHANGED;
+    	final SOQLBuilder builder = SOQLBuilder.getInstanceWithFields(Constants.LAST_MODIFIED_DATE);
+        builder.from(objectType);
+        builder.where(Constants.ID + " = '" + objectId + "'");
+        final String query = builder.build();
+        RestResponse lastModResponse = null;
+        lastModResponse = sendSyncWithSmartSyncUserAgent(RestRequest.getRequestForQuery(apiVersion, query));
```
khawkins: This will always occur within the context of an asynchronous transaction, yes?  Otherwise, a sync request here could be problematic....

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -346,15 +381,38 @@ else if (record.getBoolean(LOCALLY_CREATED))
...
+         * on the server AFTER the client's last sync down. If the merge mode
+         * passed in tells us to leave the record alone under these
+         * circumstances, we will do nothing and return here.
+         */
+        if (mergeMode == MergeMode.LEAVE_IF_CHANGED &&
+        		(action == Action.update || action == Action.delete) &&
+        		!isNewerThanServer(objectType, objectId, lastModifiedDate)) {
+
+        	// Nothing to do for this record
+        	return true;
```
khawkins: Should we log something here?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
khawkins: Not before I exhaust my command of SOQL. :)

khawkins:
Otherwise, looks fine to me.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: Yes - this is called from ```runSync```, which spawns a new ```Thread``` from the available ```ThreadPool```.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: Sure - I'll add a log line.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: We could do group of records with a selectLastModifiedDate where id in (...) but there is a limit to the length of a soql query, so one might have to do several round trips if many records were locally modified.

wmathurin:
Looks good.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: @wmathurin Does that work across different entity types, or do we have to batch them per entity type?

bhariharan:
Added native tests. Debugging hybrid tests as we speak - should have a pull request for that in the ```shared``` repo soon.

PR #823, opened 2015-02-06
Small fix

wmathurin:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -224,7 +224,7 @@ public SyncState reSync(long syncId, SyncUpdateCallback callback) throws JSONExc
         if (sync == null) {
             throw new SmartSyncException("Cannot run reSync:" + syncId + ": no sync found");
         }
-        if (sync.getType() != SyncState.Type.syncDown || sync.getTarget().getQueryType() != SyncTarget.QueryType.soql) {
```
wmathurin: The following if takes care of that check

khawkins:
:+1:

bhariharan:
:+1:

PR #826, opened 2015-02-12
Hybrid NoteSync sample app for Android repo with custom target for soql with content fields

wmathurin:


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "external/shared"]
 	path = external/shared
-        url = https://github.com/forcedotcom/SalesforceMobileSDK-Shared.git
```
wmathurin: I'll fix that once the other pull request (on the shared repo) goes through

hybrid/SampleApps/NoteSync/src/com/salesforce/samples/notesync/ContentSoqlSyncTarget.java
```diff
@@ -0,0 +1,257 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Target for sync defined by a SOQL query
+ */
+public class ContentSoqlSyncTarget extends SoqlSyncTarget {
```
wmathurin: Here is the custom target that is being called doing sync down by NoteSync.html
It's a SOQL sync target that uses SOAP instead of REST and therefore gets content fields inlined (and base64 encoded).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -224,9 +224,6 @@ public SyncState reSync(long syncId, SyncUpdateCallback callback) throws JSONExc
         if (sync.getType() != SyncState.Type.syncDown) {
             throw new SmartSyncException("Cannot run reSync:" + syncId + ": wrong type:" + sync.getType());
         }
-        if (sync.getTarget().getQueryType() != SyncTarget.QueryType.soql) {
```
wmathurin: Re-sync should not be limited to one type of sync down (since custom sync target can support resync as well).

hybrid/SampleApps/NoteSync/src/com/salesforce/samples/notesync/ContentSoqlSyncTarget.java
```diff
@@ -0,0 +1,257 @@
...
+
+    private static final String REQUEST_TEMPLATE = "<?xml version=\"1.0\"?>" +
+            "<se:Envelope xmlns:se=\"http://schemas.xmlsoap.org/soap/envelope/\">" +
+            "<se:Header xmlns:sfns=\"urn:partner.soap.sforce.com\">" +
+            "<sfns:SessionHeader><sessionId>%s</sessionId></sfns:SessionHeader>" +
+            "</se:Header>" +
+            "<se:Body>%s</se:Body>" +
+            "</se:Envelope>";
+
+    private static final String QUERY_TEMPLATE = "<query xmlns=\"urn:partner.soap.sforce.com\" xmlns:ns1=\"sobject.partner.soap.sforce.com\">" +
```
bhariharan: ```SOAP```! So much fun!

hybrid/SampleApps/NoteSync/src/com/salesforce/samples/notesync/ContentSoqlSyncTarget.java
```diff
@@ -0,0 +1,257 @@
...
+    public JSONObject asJSON() throws JSONException {
+        JSONObject target = super.asJSON();
+        target.put(ANDROID_IMPL, getClass().getName());
+        return target;
+    }
+
+    @Override
+    public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws IOException, JSONException {
+        String queryToRun = maxTimeStamp > 0 ? SoqlSyncTarget.addFilterForReSync(getQuery(), maxTimeStamp) : getQuery();
+        syncManager.getRestClient().sendSync(RestRequest.getRequestForResources(syncManager.apiVersion)); // cheap call to refresh session
```
bhariharan: Are we refreshing the session ourselves here because of ```SOAP```?

hybrid/SampleApps/NoteSync/src/com/salesforce/samples/notesync/ContentSoqlSyncTarget.java
```diff
@@ -0,0 +1,257 @@
...
+     * @throws UnsupportedEncodingException
+     */
+    private RestRequest buildSoapRequest(String sessionId, String body) throws UnsupportedEncodingException {
+        Map<String, String> customHeaders = new HashMap<String, String>();
+        customHeaders.put("SOAPAction", "\"\"");
+
+        StringEntity entity = new StringEntity(String.format(REQUEST_TEMPLATE, sessionId, body), HTTP.UTF_8);
+        entity.setContentType("text/xml");
+
+        return new RestRequest(RestRequest.RestMethod.POST, "/services/Soap/u/32.0", entity, customHeaders);
```
bhariharan: Not a big deal, but could we read the ```32.0``` from ```ApiVersionStrings```. Makes it easier to update across the board.

bhariharan:
Looks pretty good otherwise.

hybrid/SampleApps/NoteSync/src/com/salesforce/samples/notesync/ContentSoqlSyncTarget.java
wmathurin: Soap expects the session id in the body of the post. So we are making a cheap call to make sure we have a valid session id first.

bhariharan: :+1:

PR #829, opened 2015-02-17
Fixing quirky loader

bhariharan:
Turns out that loader callbacks won't be triggered unless the load is triggered by ```LoaderManager```. For any other type of load, such as background sync, the ```Activity``` needs to be notified by a traditional eventing mechanism. This is primarily for the case where the ```Activity``` is foregrounded, but sync happens in the background.


khawkins:
Seems fine.

wmathurin:
We have to do that?

bhariharan:
@wmathurin Yes we do (for background sync). All the auto-magic stuff only works when its run through a ```LoaderManager```, which only applies to a class of the ```View``` hierarchy, namely ```Activity``` or ```Fragment```. If a ```Service``` is used like we do for background sync, the ```Activity``` won't refresh if it is in the foreground while the ```Service``` kicks in. If the ```Activity``` is backgrounded, no problem, since it'll load afresh next time it is foregrounded (in ```onResume()```).

PR #832, opened 2015-02-20
fix for null "integer" and "floating" SmartStore indexes crash

PiotrPiesiak:
SmartStore was crashing if values for integer/floating indexes were null


wmathurin:
Looks good. Thanks.

PR #844, opened 2015-03-17
Sync up target for android

wmathurin:
NB: diverging a bit from iOS implementation / will align iOS implementation in future pull request
TODO: hybrid side changes

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2014, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of salesforce.com, inc. nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission of salesforce.com, inc.
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.smartsync.util;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import com.salesforce.androidsdk.smartsync.manager.SyncManager;
+
+/**
+ * Target for sync down:
+ * - what records to download from server
+ * - how to download those records
+ */
+public abstract class SyncDownTarget implements SyncTarget {
+
+    // Constants
+	public static final String QUERY_TYPE = "type";
+    public static final String ANDROID_IMPL = "androidImpl";
+
+    // Fields
+	protected QueryType queryType;
+    protected int totalSize; // set during a fetch
+
+	/**
+	 * Build SyncDownTarget from json
+	 * @param target as json
+	 * @return
+	 * @throws JSONException 
+	 */
+	@SuppressWarnings("unchecked")
+	public static SyncDownTarget fromJSON(JSONObject target) throws JSONException {
+		if (target == null)
+			return null;
+
+		QueryType queryType = QueryType.valueOf(target.getString(QUERY_TYPE));
+
+        switch (queryType) {
+        case mru:     return MruSyncDownTarget.fromJSON(target);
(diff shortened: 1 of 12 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
```diff
@@ -0,0 +1,133 @@
+/*
```
wmathurin: Simply renamed SyncTarget to SyncDownTarget
But because I also introduced a interface SyncTarget (implemented by SyncDownTarget and SyncUpTarget), the diff doesn't see SyncDownTarget as a simple rename.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncTarget.java
```diff
@@ -26,110 +26,19 @@
  */
```
wmathurin: Old SyncTarget is now SyncDownTarget
SyncTarget is now a simple interface with asJson as method.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Save locally created record back to server
+     * @param syncManager
+     * @param objectType
+     * @param fields
+     * @return server record id or null if creation failed
+     * @throws JSONException
+     * @throws IOException
+     */
+    public String createOnServer(SyncManager syncManager, String objectType, Map<String, Object> fields) throws JSONException, IOException {
```
wmathurin: The logic to do a create/update/delete based on the dirty flag in the record remains in SyncManager. Every custom sync up target shouldn't have to implement it.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
```diff
@@ -0,0 +1,200 @@
...
+        return serverLastModified;
+    }
+
+    /**
+     * Return ids of records to sync up
+     * @param syncManager
+     * @param soupName
+     * @return
+     */
+    public Set<String> getIdsOfRecordsToSyncUp(SyncManager syncManager, String soupName) throws JSONException {
```
wmathurin: Custom target might never override that method.
But at least with that method, we actually define the target for the sync up i.e. set of records to send up.

bhariharan:
:+1:

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncTarget.java
```diff
@@ -26,110 +26,19 @@
...
-        return totalSize;
-    }
-
-    /**
-     * @return QueryType of this target
-     */
-    public QueryType getQueryType() {
-        return queryType;
-    }
+public interface SyncTarget {
```
khawkins: Seems like a good idea.  I thought the same for iOS, but didn't get there on the first pass.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public long fetchLastModifiedDate(SyncManager syncManager, String objectType, String objectId) throws JSONException, IOException {
+        boolean isNewer = false;
+        long serverLastModified = -1L;
+        final String query = SOQLBuilder.getInstanceWithFields(Constants.LAST_MODIFIED_DATE)
+                .from(objectType)
+                .where(Constants.ID + " = '" + objectId + "'")
+                .build();
+
+        RestResponse lastModResponse = syncManager.sendSyncWithSmartSyncUserAgent(RestRequest.getRequestForQuery(syncManager.apiVersion, query));
```
khawkins: Are we concerned about this happening synchronously?  Or is this in an area whose parent context is already asynchronous?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
```diff
@@ -0,0 +1,200 @@
...
+                final JSONArray records = responseJSON.optJSONArray(Constants.RECORDS);
+                if (records != null && records.length() > 0) {
+                    final JSONObject obj = records.optJSONObject(0);
+                    if (obj != null) {
+                        final String lastModStr = obj.optString(Constants.LAST_MODIFIED_DATE);
+                        if (!TextUtils.isEmpty(lastModStr)) {
+                            try {
+                                serverLastModified = Constants.TIMESTAMP_FORMAT.parse(lastModStr).getTime();
+                            } catch (ParseException e) {
+                                Log.e("SyncUpTarget:fetchLastModifiedDate", "Error during date parsing", e);
```
khawkins: Would we not bubble up the exception as a failed sync up?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
wmathurin: During sync, we use synchronous network calls - but the sync itself happens on a separate thread

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
khawkins: Yeah, seems better that the sync up target owns the records, like the sync down target does. :+1:

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
bhariharan: All the sync code should be called from a ```ContentProvider``` or an ```AsyncTask```. Technically, it is possible for an app to get here on the main thread, but there would be a bunch of other stuff that fails too, such as DB calls, etc. ```Android``` has gotten pretty strict about ```StrictModeViolations``` in recent times, so any app that attempts to use the main thread for even ```SmartStore``` access would get a bunch of red flags at development time itself.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
wmathurin: If we can't get a last modified date from the server, should we:
- return -1 (like it is now), causing the "sync" to think that the record hasn't changed and therefore can be written over regardless of merge mode
- return a large value (max long), to cause the "sync" to think that the record has changed and therefore shouldn't not be written over if the merge mode is not "overwrite"
- throw an error and stop the sync

bhariharan: Bubbling up the error is okay with me, but what would the recovery path be? App level logic to retry/abandon sync?

Should we fallback on ```OVERWRITE``` or ```LEAVE_IF_CHANGED```?

khawkins: I would think that the error would almost always be some sort of network error attempting to retrieve the remote modification date.  What do we do with network errors for standard sync up / sync down operations?

wmathurin:
We let the exception bubble up to the runSync method which catches it, logs it and update the sync status as failed. So I guess I could throw an exception in that method if we can't get the date back.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
khawkins: :+1:

khawkins:
Overall, looks good to me.

bhariharan:
:+1:

wmathurin:
All hybrid tests are passing

PR #845, opened 2015-03-17
First stab at adding global SmartStore

bhariharan:
TODO Items:

* Change the filter for deletion to exclude global DBs.
* Add new tests.
* Verify existing tests pass.

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SalesforceSDKManagerWithSmartStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SalesforceSDKManagerWithSmartStore.java
@@ -26,7 +26,9 @@
  */
 package com.salesforce.androidsdk.smartstore.app;
 
+import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 
 import net.sqlcipher.database.SQLiteDatabase;
 import net.sqlcipher.database.SQLiteOpenHelper;
@@ -36,7 +38,6 @@ import android.content.Context;
 import android.text.TextUtils;
 
 import com.salesforce.androidsdk.accounts.UserAccount;
-import com.salesforce.androidsdk.accounts.UserAccountManager;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.smartstore.store.DBOpenHelper;
 import com.salesforce.androidsdk.smartstore.store.SmartStore;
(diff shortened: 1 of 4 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SalesforceSDKManagerWithSmartStore.java
```diff
@@ -218,6 +223,25 @@ public synchronized void changePasscode(String oldPass, String newPass) {
     }
 
     /**
+     * Returns the database used by smart store in the global context.
+     *
+     * @param dbName The database name. This must be a valid file name without a
+     * 				 filename extension such as ".db". Pass 'null' for default.
+     * @return SmartStore instance.
+     */
+    public SmartStore getGlobalSmartStore(String dbName) {
```
bhariharan: Passing in ```null``` or empty string for ```dbName``` here will use the default name. We could expose another method that takes no parameters if we want to, but there are already 4 or 5 of these types of methods here.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SalesforceSDKManagerWithSmartStore.java
```diff
@@ -309,4 +347,55 @@ public boolean hasSmartStore(UserAccount account, String communityId) {
...
      	return DBOpenHelper.smartStoreExists(context, dbNamePrefix, account, communityId);
      }
+
+     /**
+      * Removes the global smart store.
+      *
+      * @param dbName Database name. This must be a valid file name without a
+      * 			  filename extension such as ".db". Pass 'null' for default.
+      */
+     public void removeGlobalSmartStore(String dbName) {
```
bhariharan: ```remove``` was never exposed here before, although it was possible through ```DbOpenHelper```.

bhariharan:
Added re-encryption of global DBs in [this](https://github.com/bhariharan/SalesforceMobileSDK-Android/commit/59b6449e2ef383fcafc1afff09dd43f74e329107) commit.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -95,20 +105,19 @@ public static synchronized DBOpenHelper getOpenHelper(Context ctx,
 	 * @param communityId Community ID.
 	 * @return DBOpenHelper instance.
 	 */
-	public static DBOpenHelper getOpenHelper(Context ctx, String dbNamePrefix, UserAccount account, String communityId) {
-		StringBuffer dbName = new StringBuffer(dbNamePrefix);
-		
-		// If we have account information, we will use it to create a database suffix for the user
+	public static DBOpenHelper getOpenHelper(Context ctx, String dbNamePrefix,
```
khawkins: Ah, so like you said, this functionality already existed.

khawkins:
Looks good to me, over all.

wmathurin:
Looks fine.

bhariharan:
Looks like we already had tests for global smart store in ```DbOpenHelperTest```. Nothing for me to do there :-)

bhariharan:
See [this](https://github.com/bhariharan/SalesforceMobileSDK-Android/commit/8dda2173f43c8d41137b6b9bf175223315fe2222) commit for the last round of changes.

* Added tests for removal of DBs.
* Fixed the file filter logic, so that global DBs are not removed upon logout.
* Misc. cleanup.

khawkins:
:+1:

PR #849, opened 2015-03-20
Tests for custom sync up

wmathurin:
Using a custom sync up target to test:
- custom sync up targets
- failed sync up with local create/update/delete (nb: soft failures not exceptions)

We could (and might) do more:
- hard failures (exceptions)
- failure when fetching date during a leave_if_changed sync
- the same (custom / failure scenarios) during sync down


bhariharan:
:+1:

PR #852, opened 2015-03-23
Fix for Eclipse static access warnings

bhariharan:


wmathurin:
Thanks.
Studio is not pointing those out (or maybe it is but not in a very noticeable way).

bhariharan:
One of these days I'm going to have to switch to ```Android Studio``` too, but I'm putting it off as long as I can ;-)

PR #854, opened 2015-03-26
Global store for sync manager / smart sync plugin

wmathurin:
Sync manager now only needs a smartstore to be instantiated (but still needs to have a rest client to run syncs)
New js tests hooked up and passing


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -108,62 +131,40 @@ public static synchronized SyncManager getInstance(UserAccount account, String c
...
-            instance = new SyncManager(account, communityId);
-            INSTANCES.put(uniqueId, instance);
-        }
-        return instance;
+
+        SmartStore smartStore = SmartSyncSDKManager.getInstance().getSmartStore(account, communityId);
+        SyncManager syncManager = getInstanceForStore(smartStore);
+
+        // Setting up rest client
+        RestClient restClient = SalesforceSDKManager.getInstance().getClientManager().peekRestClient(account);
```
khawkins: If I'm reading this right, if you get an instance for a user, you set the RestClient, but if you go straight to getting an instance for a store, you don't?  How does that play out across the different usage scenarios of `SyncManager`?

khawkins: Presumably, the developer would manually specify the RestClient for that manager, in that case?

khawkins:
Looks fine to me, over all.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: Alternatively, we could change the signature for the sync methods and add a rest client parameter there.
On the iOS side, SFRestAPI is a singleton and can be obtained from anywhere, and therefore the sync manager doesn't have to concern itself with it.

BTW another thing we need to figure out on the Android side is unauthenticated calls.

khawkins: Makes sense. :+1:

wmathurin:
Will revisit SyncManager instantiation in a future pull request

bhariharan:
:+1:

PR #856, opened 2015-03-27
Revisiting initiation of SyncManager

wmathurin:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -128,33 +127,40 @@ public static synchronized SyncManager getInstance(UserAccount account) {
...
+            smartStore = SmartSyncSDKManager.getInstance().getSmartStore(account, communityId);
+        }
 
-        // Setting up rest client
-        RestClient restClient = SalesforceSDKManager.getInstance().getClientManager().peekRestClient(account);
-        syncManager.setRestClient(restClient);
+        if (UserAccount.INTERNAL_COMMUNITY_ID.equals(communityId)) {
+            communityId = null;
+        }
+        String uniqueId = (account != null ? account.getUserId() : "") + ":"
```
bhariharan: We can take ```communityId``` out of this ```uniqueId``` computation, and with it, the line above that sets ```communityId = null```. The DB path accounts for ```userId, orgId, communityId```, and all we need to associate it with a user account is ```userId```. If a portal user if logged into a community, the ```userId``` will be different. If an internal user is logged into both org and community, the DB path will be different. Hence, the ```uniqueId``` should always be unique with ```userId + dbPath```.

wmathurin: Makes sense
Made the change here https://github.com/wmathurin/SalesforceMobileSDK-Android/commit/bd3089ab03f3862a6f2b3f15872f68a681ca8a9a

bhariharan:
:+1:

khawkins:
:+1:

PR #859, opened 2015-03-30
Unauthenticated user support for Android

wmathurin:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -363,6 +370,44 @@ public URI resolveUrl(String path) {
 		}
 	}
 
+    /**
+     * Use a unauthenticated client info when do not need authentication support (e.g.
+     * if you are talking to non-salesforce servers)
+     *
+     * NB: Your RestRequest's path will need to be a complete URL
+     */
+    public static class UnauthenticatedClientInfo extends ClientInfo {
```
wmathurin: If we were to pass a null ClientInfo instead, we would have to handle null in a number of places

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestRequest.java
```diff
@@ -128,7 +128,7 @@ public String getPath(Object... args) {
 	 * 
 	 * @param method				the HTTP method for the request (GET/POST/DELETE etc)
 	 * @param path					the URI path, this will automatically be resolved against the users current instance host.
-	 * @param httpEntity			the request body if there is one, can be null.
```
wmathurin: Just fixing a bunch of outdated java docs in this file

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -212,6 +215,15 @@ public void testGetResources() throws Exception {
         checkKeys(response.asJSONObject(), "sobjects", "search", "recent");
     }
 
+    /**
+     * Testing a get resources async call to the server - check response
+     * @throws Exception
+     */
+    public void testGetResourcesAsync() throws Exception {
```
wmathurin: Adding a couple of tests that use client.sendAsync since client.sendAsync and client.sendSync are quite different. Only the former uses Volley queues.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -380,11 +404,6 @@ public void testWrappedRestRequestUrl() throws Exception {
 		checkWrappedRestRequestUrl(FileRequests.uploadFile(new File("fakePath"), "MyFile", "Description", "image/png"), clientInfo.instanceUrl + "/services/data/" + ApiVersionStrings.VERSION_NUMBER + "/chatter/users/me/files");
     }
 
-    private void checkWrappedRestRequestUrl(RestRequest restRequest, String expectedUrl) throws Exception {
```
wmathurin: Moved the private helper methods below all the public methods

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -416,6 +429,87 @@ public void testWrappedRestRequestBody() throws Exception {
...
+        checkWrappedRestRequestBodyContentType(RestRequest.getRequestForUpdate(TestCredentials.API_VERSION, "account", "fakeId", fields), "application/json; charset=UTF-8");
+        checkWrappedRestRequestBodyContentType(FileRequests.uploadFile(new File("fakePath"), "MyFile", "Description", "image/png"), "multipart/form-data");
+    }
+
+    /**
+     * Testing doing a sync request against a non salesforce public api with a RestClient that uses an UnauthenticatedClientInfo
+     * @return
+     * @throws Exception
+     */
+    public void testRestClientUnauthenticatedlientInfo() throws Exception {
```
wmathurin: Test for unauthenticated user (going against a public api)

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -416,6 +429,87 @@ public void testWrappedRestRequestBody() throws Exception {
...
+    //
+
+    /**
+     * Send request using sendAsync method
+     * @param client
+     * @param request
+     * @return
+     * @throws InterruptedException
+     */
+    private RestResponse sendAsync(RestClient client, RestRequest request) throws InterruptedException {
```
wmathurin: Using a blocking queue to make the async call into a sync operation

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -128,6 +128,14 @@ public void getRestClient(Activity activityContext, RestClientCallback restClien
         }
     }
 
+    /**
+     * Method to create a unauthenticated RestClient.
+     * @return
+     */
+    public RestClient  peekUnauthenticatedRestClient() {
```
bhariharan: Do we need the equivalent asynchronous ```getUnauthenticatedRestClient()``` as well, that activities could use?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
bhariharan: So, only 1 unauthenticated ```RestClient``` instance can be procured per app?

wmathurin: Do we need more than one?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
wmathurin: Done with https://github.com/wmathurin/SalesforceMobileSDK-Android/commit/f223a9eadff0f196e1d38f0f418f6d6d5e9f9109

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
bhariharan: Not sure :-) If someone wanted 2 clients to 2 different URIs, perhaps. But then, they probably wouldn't use the Mobile SDK for that.

bhariharan:
Looks good! :+1:

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
wmathurin: You will get a different instance with every call but internally it will leverage the same volley RequestQueue

bhariharan: Right - that's what I meant. That should be fine though.

khawkins:
Seems pretty good to me.  :+1:

Issue #862, opened 2015-04-06
Hybrid app crash (ArrayIndexOutOfBoundsException) when OAuth scopes aren't defined to match bootconfig.json


daveespo:
If you don't include the proper scopes when setting up your Connected App, rather than getting "1804: the requested scope is not allowed" upon Login, the app crashes with:

````
04-06 14:21:56.398      322-322/com.salesforce.samples.vfconnector E/AndroidRuntime﹕ FATAL EXCEPTION: main
    Process: com.salesforce.samples.vfconnector, PID: 322
    java.lang.ArrayIndexOutOfBoundsException: length=0; index=-1
            at com.salesforce.androidsdk.app.SalesforceSDKManager.removeAccount(SalesforceSDKManager.java:848)
            at com.salesforce.androidsdk.app.SalesforceSDKManager.logout(SalesforceSDKManager.java:809)
            at com.salesforce.androidsdk.app.SalesforceSDKManager.logout(SalesforceSDKManager.java:770)
            at com.salesforce.androidsdk.app.SalesforceSDKManager.logout(SalesforceSDKManager.java:746)
````

AccountManager returns an empty array if there are no accounts, never null so that should be the test (#835 is this same bug, just in different words)

The offending code was introduced in 633df05

bhariharan:
@daveespo Thanks for reporting this issue - I'll take a look.

daveespo:
PR attached ;-)

bhariharan:
Could you send a pull request to the ```unstable``` branch?

daveespo:
Yep, sorry about that .. didn't notice the request for 'unstable' until it was too late

Hopefully I got it right this time ;-)

bhariharan:
:+1:

PR #863, opened 2015-04-06
Tolerate the lack of existing Accounts on the device when app logs in for the first time

daveespo:
Fix for #862


kchitalia:
One suggestion, can we have a check for null and length > 0?

daveespo:
The Android SDK docs make it pretty clear that null is not a valid return value:

> @return An array of {@link Account}, one per matching account.  Empty (never null) if no accounts of the specified type have been added.

kchitalia:
Sorry, my bad.

bhariharan:
Thanks for the contribution @daveespo!

PR #867, opened 2015-04-08
SmartStoreInspector can show user smartstore or default global smartstore

wmathurin:
Now making use of isGlobalStore passed to showInspector plugin


wmathurin:
Not as flexible as https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/988
The two options are: default global smartstore or user smartore
On iOS, you get to pass the smartstore to the view controller so it could be any smartstore

wmathurin:
With https://github.com/wmathurin/SalesforceMobileSDK-Android/commit/1a481dd940b4638cf27714c2034c7c9f3643dcd3, it is as flexible as iOS

bhariharan:
:+1:

PR #885, opened 2015-05-14
Getting rid of httpcore dependency

bhariharan:
It turned out to be a bit more intrusive than I was hoping it would be.


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -548,37 +583,131 @@ public HttpResponse performRequest(int method, URI url, HttpEntity httpEntity, M
...
+			}
+
+			// Done.
+			return response;
+		}
+
+		/**
+		 * Swaps the existing access token for a new one.
+		 */
+		private void refreshAccessToken() {
```
bhariharan: Moved this into its own method, so that it could be reused.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/files/FileRequests.java
```diff
@@ -238,40 +250,4 @@ private static HttpEntity makeFileShare(String fileId, String entityId, String s
...
-     *            The name/title of this file.
-     * @param description
-     *            A description of the file.
-     * @param mimeType
-     *            The mime-type of the file, if known.
-     * @return A RestRequest that can perform this upload.
-     * 
-     * @throws UnsupportedEncodingException
-     */
-    public static RestRequest uploadFile(File theFile, String name, String description,
```
bhariharan: I had to move this out of here, since there's no way to simply return an MPE request. It's all low level stuff using ```HttpUrlConnection```, so it didn't make sense to leave it here.

libs/test/SalesforceSDKTest/AndroidManifest.xml
```diff
@@ -27,4 +27,6 @@
 
     <uses-sdk android:minSdkVersion="17"
         android:targetSdkVersion="21" />
+
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```
bhariharan: Need this to create a file for the upload test (we didn't have a file upload test).

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -439,7 +438,6 @@ public void testWrappedRestRequestBodyContentType() throws Exception {
         fields.put("name", "NewAccount");
         checkWrappedRestRequestBodyContentType(RestRequest.getRequestForCreate(TestCredentials.API_VERSION, "account", fields), "application/json; charset=UTF-8");
         checkWrappedRestRequestBodyContentType(RestRequest.getRequestForUpdate(TestCredentials.API_VERSION, "account", "fakeId", fields), "application/json; charset=UTF-8");
-        checkWrappedRestRequestBodyContentType(FileRequests.uploadFile(new File("fakePath"), "MyFile", "Description", "image/png"), "multipart/form-data");
```
bhariharan: This API is gone now.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -472,6 +470,34 @@ public void testRestClientUnauthenticatedlientInfoAsync() throws Exception {
         checkKeys(jsonResponse.getJSONObject("artists"), "href", "items", "limit", "next", "offset", "previous", "total");
     }
 
+    /**
+     * Tests if the file upload API is working per design.
+     *
+     * @throws Exception
+     */
+    public void testFileUpload() throws Exception {
```
bhariharan: Test for file upload. Not quite done - deleting the remote file is pending.

wmathurin:
Looks fine.

PR #897, opened 2015-06-15
Re sync only disallowed to run when targeted sync is still actually running

wmathurin:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -289,6 +291,20 @@ private void updateSync(SyncState sync, SyncState.Status status, int progress, S
...
     		sync.save(smartStore);
+
+            switch (status) {
+
+                case NEW:
+                    break;
+                case RUNNING:
+                    runningSyncIds.add(sync.getId());
+                    break;
+                case DONE:
```
bhariharan: Meant to ask this earlier, but doesn't it have to be removed from the map when it gets to the ```DONE``` status?

wmathurin: For DONE and FAILED, it gets removed (it's falling through the case).

bhariharan: Oh right. I misread a nonexistent ```break;``` :-)

bhariharan:
:+1:

PR #902, opened 2015-06-18
Fix for MDM login host parsing

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -199,6 +199,8 @@ public void reset() {
 		} catch (Exception e) {
 			Log.w("LoginServerManager.getLoginServersFromRuntimeConfig",
 					"Exception thrown while attempting to read array, attempting to read string value instead");
+		}
+		if (mdmLoginServers == null) {
```
bhariharan: Turns out that if the hosts are not in an ```array```, an ```Exception``` is not thrown anymore :-( So, we need to attempt to parse a ```string``` outside the ```catch``` block for it to work.

wmathurin:
:+1:

PR #903, opened 2015-06-22
Use custom ID field if it is set

bhariharan:
Looks like we were hardcoding the ID field on ```Android```, but using the custom one on ```iOS```.


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -447,10 +447,10 @@ private void syncDown(SyncState sync, SyncUpdateCallback callback) throws Except
         int totalSize = target.getTotalSize();
         sync.setTotalSize(totalSize);
         updateSync(sync, SyncState.Status.RUNNING, 0, callback);
-
+        final String idField = sync.getTarget().getIdFieldName();
```
bhariharan: Falls back on ```Constants.ID``` if there's no custom field set.

wmathurin:
:+1:

PR #907, opened 2015-06-27
Advanced user tracking

wmathurin:
Adding device id to user agent


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -151,7 +153,8 @@ public static SalesforceSDKManager getInstance() {
      */
     protected SalesforceSDKManager(Context context, KeyInterface keyImpl, 
     		Class<? extends Activity> mainActivity, Class<? extends Activity> loginActivity) {
-    	this.context = context;
+        this.uid = Settings.Secure.getString(context.getContentResolver(), Settings.Secure.ANDROID_ID);
```
bhariharan: Do we need any extra ```<permission>``` for this?

bhariharan:
Looks good to me. :+1:

PR #909, opened 2015-06-29
SmartStore deadlock fix

wmathurin:
If you call SmartStore's beginTransaction outside a synchronized(store.getDatabase()), you will lock the actual db (because you are in a transaction) without locking the java db object.
Another thread does a smartstore operation at that point, it will lock on the java db object but will be stuck with the db operation because of your transaction. If you attempt to do a smartstore operation then, you will be stuck waiting for that second thread (DEADLOCK!!)

Added a comment to beginTransaction, and added synchronized() {..} in callers of the SDK.
(Maybe this API should go entirely)

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/phonegap/SmartStorePlugin.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/phonegap/SmartStorePlugin.java
@@ -140,10 +140,10 @@ public class SmartStorePlugin extends ForcePlugin {
 			@Override
 			public void run() {
 
-		    	// All smart store action need to be serialized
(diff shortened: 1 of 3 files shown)


bhariharan:
How is this different from DB level locking done by ```SQLite```? I vaguely remember switching to class level locking, because we had some static calls and references that could break in the multi user scenario (if you try to access 2 users' databases at the same time). Are we okay on that front?

wmathurin:
The setLockingEnabled is to keep two java threads from interacting with the database (it's in the java code only). The locking during a transaction is handled at the lowest level by sqlite and cannot be disabled (https://www.sqlite.org/lang_transaction.html). More advanced database would only lock target rows (or in some cases tables).

bhariharan:
:+1:

Issue #917, opened 2015-07-06
null integer and floating values cause crash


toddhalfpenny:
SDK: v3.3 

When running an upsert with null integer or floating values a crash is seen. It looked like this was fixed in [PR 832](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/832) but it does not seem to have worked as expected.

The following can be used to re-produce the issue (from JS console) of a new hybrid app.

```
var smartstore = cordova.require("com.salesforce.plugin.smartstore");

smartstore.registerSoup('testSoup', [{path:"MyString",type:"string"},{path:"MyInt",type:"integer"}], function(res){console.log(res);}, function(err){console.error(err);});

var upsertArr = [];

// All fields present
// All OK, as expected
upsertArr = [{"MyString" : "1", "MyInt" : 1}];
smartstore.upsertSoupEntries('testSoup', upsertArr, function(res){console.log(res);}, function(err){console.error(err);});

// No int field supplied
// All OK, as expected
upsertArr = [{"MyString" : "2"}];
smartstore.upsertSoupEntries('testSoup', upsertArr, function(res){console.log(res);}, function(err){console.error(err);});

// null string value supplied
// All OK, as expected
upsertArr = [{"MyString" : null, "MyInt" : 4}];
smartstore.upsertSoupEntries('testSoup', upsertArr, function(res){console.log(res);}, function(err){console.error(err);});

// null int value supplied
// FAILS  :(
upsertArr = [{"MyString" : "3", "MyInt" : null}];
smartstore.upsertSoupEntries('testSoup', upsertArr, function(res){console.log(res);}, function(err){console.error(err);});

// this is the error seen 
com.salesforce.smartstore:pgUpsertSoupEntries failed com.salesforce.util.exec.js:42
(anonymous function) com.salesforce.util.exec.js:42
cordova.callbackFromNative cordova.js:295
processMessage cordova.js:1054
androidExec.processMessages cordova.js:1091
pollOnce cordova.js:956
pollOnceFromOnlineEvent
...
org.json.JSONObject$1 cannot be cast to java.lang.Number

```

I am using the following in *src/com/androidsdk/smartstore/store/smartstore.java* and it seems to work as expected.

I was going to put in a PR but was unsure of how/where the tests should be written etc. If you can give me guidance I'd be happy to give this a go.

```
...

	/**
     * @param soupElt
     * @param contentValues
     * @param indexSpec
     */
[...]

PR #918, opened 2015-07-07
Bugfix/null integer + floating values cause crash

toddhalfpenny:
This PR fixes issue [917](https://github.com/forcedotcom/SalesforceMobileSDK-Android/issues/917).

There may be a better way of doing this null JSONObject checking, but this at least seems to work.

The manual tests used can be found in the issue ticket.


bhariharan:
Thanks @toddhalfpenny!

PR #923, opened 2015-07-18
Fix for the elusive 'No authentication challenges found' bug in HttpUrlConnection

bhariharan:


bhariharan:
```HttpUrlConnection``` throws an ```IOException``` when a certain header field is not present, but reads the input stream, so the 2nd attempt works fine. Without this fix, the user will be stuck in a state where he/she is unable to get a new access token in the ```401``` case.

PR #924, opened 2015-07-23
Bug fix for inability to add custom server to MDM configured app

bhariharan:


bhariharan:
The issue was that we weren't persisting runtime configs, so even if a custom server was added, it wouldn't shown, since we weren't reading from ```SharedPrefs```, we were reading from ```ManagedPrefs``` instead.

wmathurin:
Why use a separate pref file?

bhariharan:
I started out using the same file, but it gets tricky, due to the dependencies on indexes for selected server. Besides, in an ```Android for Work``` device, there are 2 APKs (personal and enterprise), so there should be only 1 file used at a time. This was the simpler implementation.

PR #927, opened 2015-07-28
Updated license file

wmathurin:
I put in the most recent licenses for sqlcipher and cordova. What else should we put in there?
@akhileshgupta Do we have an updated license for salesforce itself (first part of the file)?


bhariharan:
:+1: 

@wmathurin We haven't received a new ```Salesforce``` license AFAIK.

PR #928, opened 2015-07-31
Add file for travis CI

kchitalia:
Right now we are only running the build for the projects. Will add tests to run on travis CI as well in next iteration.


wmathurin:
Looks fine

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+   components:
+      - build-tools-21.1.1
+      - android-21
+env:
+   global:
+      # install timeout in minutes
+      - ADB_INSTALL_TIMEOUT=8
+
+script:
+   - ./gradlew assembleRelease
```
bhariharan: @kchitalia I think we should do ```assembleDebug``` here, so that console logging in enabled.

kchitalia:
Sure, I'll change it in the PR I send for the tests.

PR #930, opened 2015-08-03
Be sure to grab "nextRecordsUrl" from every request when sync-ing.

randalln:
Otherwise it just loops forever.


wmathurin:
Thanks. Do you mind resubmitting your pull request against the unstable branch?

PR #933, opened 2015-08-03
Shrinking npm package a bit by not including some folders (e.g. the s…

wmathurin:
…hared/doc folder which has the mobile sdk dev guide)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,10 +1,17 @@
+external/shared/doc/
+external/shared/gen/
+external/shared/test/
+external/shared/tools/
 external/shared/samples/accounteditor/
 external/shared/samples/contactexplorer/
 external/shared/samples/fileexplorer/
 external/shared/samples/simplesync/
+external/shared/samples/simplesyncreact/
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -1,10 +1,17 @@
+external/shared/doc/
+external/shared/gen/
+external/shared/test/
+external/shared/tools/
 external/shared/samples/accounteditor/
 external/shared/samples/contactexplorer/
 external/shared/samples/fileexplorer/
 external/shared/samples/simplesync/
+external/shared/samples/simplesyncreact/
```
bhariharan: Is ```simplesyncreact``` committed on ```unstable```? I didn't know.

bhariharan:
:+1:

.npmignore
wmathurin: It is. The big saving is the doc directory (which has the 500 pages PDF), it causes the tgz to go down from 25 to 21 MB. On the iOS side it's only 1.3 MB now.

bhariharan: :+1:

PR #935, opened 2015-08-04
MRU bug fix

bhariharan:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/MetadataManager.java
```diff
@@ -1092,12 +1092,7 @@ private String getReturnFieldsForObjectType(String objectTypeName) {
             }
             String whereClause;
             if (objContainsLastViewedDate) {
-
-            	/*
-            	 * TODO: This should be replaced with 'using SCOPE MRU'
-            	 * in 'v32.0'.
-            	 */
-                queryBuilder.from(String.format("%s using MRU", objectTypeName));
+                queryBuilder.from(String.format("%s using SCOPE MRU", objectTypeName));
```
bhariharan: I guess no one's using MRU.....

wmathurin:
:+1:

PR #940, opened 2015-08-21
Fixes for org split

kchitalia:


wmathurin:
Looks fine.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -548,16 +555,18 @@ public String getNewAuthToken() {
                         }
 
                         // Broadcasts an intent that the access token has been revoked.
-                        final Intent revokeIntent = new Intent(ACCESS_TOKEN_REVOKE_INTENT);
-                        revokeIntent.setPackage(SalesforceSDKManager.getInstance().getAppContext().getPackageName());
-                        SalesforceSDKManager.getInstance().getAppContext().sendBroadcast(revokeIntent);
+                        broadcastIntent = new Intent(ACCESS_TOKEN_REVOKE_INTENT);
+                    } else if (newInstanceUrl != null && !newInstanceUrl.equalsIgnoreCase(lastNewInstanceUrl)) {
+                        // Broadcasts an intent that the instance server has changed (implicitly token refreshed too)
+                        broadcastIntent = new Intent(INSTANCE_URL_UPDATE_INTENT);
```
bhariharan: In this ```else``` block, ```ACCESS_TOKEN_REFRESH_INTENT``` needs to be broadcast as well, since this occurs only when refresh occurs?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -597,38 +601,87 @@ public HttpResponse performRequest(int method, URI url, HttpEntity httpEntity,
...
 		 * Swaps the existing access token for a new one.
 		 */
-		private void refreshAccessToken() {
-			if (authTokenProvider != null) {
+		private void refreshAccessToken(HttpResponse response) throws IOException {
+            // If we haven't retried already and we have an accessTokenProvider
+            // Then let's try to get a new authToken
+            if (authTokenProvider != null) {
+                // remember to consume this response so the connection can get re-used
+                HttpEntity entity = response.getEntity();
```
bhariharan: Where's this coming from? I don't think we need ```consume()``` anymore, since we don't use the ```Apache HTTP client``` anymore.

bhariharan: Never mind. Looks like it was always there.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
kchitalia: I remember now why I did this. When the instance url gets refreshed, I would assume the token gets refreshed too. Is that assumption correct? If yes, then maybe I should change the name of this intent.

bhariharan: @kchitalia Yes - that assumption is correct, because the only way we would know about ```instanceUrl``` changing is in the refresh flow.

bhariharan: I'm okay with leaving this as it is, since listeners of this broadcast will probably except an access token update.

kchitalia: Sounds good. So no changes required :)

bhariharan:
:+1:

PR #945, opened 2015-08-26
Fix for MRU test flappers

bhariharan:


bhariharan:
```SalesforceSDK``` tests fixed [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/c80e309160286575489d9dc86211a9f8e1b31575).

bhariharan:
The ```Environment.getExternalStorageDirectory()``` isn't really a fix. The other way works too. It's based on the emulator config - it has to be configured with an SD card. I changed it to future-proof it.

PR #946, opened 2015-08-27
Add tests to Travis CI

kchitalia:


.travis.yml
```diff
@@ -15,5 +15,14 @@ env:
       # install timeout in minutes
       - ADB_INSTALL_TIMEOUT=8
 
+# Create and start the emulator
+before_script:
+   - echo no | android create avd --force -n test -t android-21  --abi armeabi-v7a -c 128M
```
bhariharan: I think we should change the ```abi``` to ```x86```. ```arm``` is too slow. Is ```-c``` for SD card?

kchitalia: Yep -c is for SD card. Will change to x86

wmathurin:
Looks fine. Should we run the tests against different emulators (different OS versions - may be min and max that we support)? 
Should do the same on iOS (build/test against min and max OS versions we support)?

.travis.yml
kchitalia: Apparently there are some issues with the x86 emulator:

https://github.com/travis-ci/travis-ci/issues/1419

But based on some suggestions on that thread, I could use the -PdisablePreDex to speed up the build.

bhariharan:
Matrix builds are ideal, but we faced tonnes of issues with the variants in the past with ```Jenkins```. I hope ```Travis``` works better :-)

bhariharan:
:+1:

PR #947, opened 2015-09-01
Sample apps cleanup (removal)

wmathurin:


bhariharan:
You also need to edit the top level ```settings.gradle``` file and remove the deleted projects from there. And then remove some entries from ```.npmignore```. Otherwise, looks good!

PR #948, opened 2015-09-02
Sample apps cleanup: readme updates + symlink updates 

wmathurin:


bhariharan:
:+1:

kchitalia:
:+1:

bhariharan:
@wmathurin Should we remove the ```shared``` repo resources too, for the deleted sample apps?

wmathurin:
We could.
There is a readme file in the shared repo explaining how do build an iOS/Android app from the samples.
Of course, because they are not in the main repo, there is a chance they regress without us noticing.
In any case, we need to keep UserList and VFConnector since they are used when creating hybrid local and hybrid remote apps with forceios/forcedroid.

bhariharan:
We can leave them alone for now. Doesn't hurt.

PR #950, opened 2015-09-16
SimpleSync is now SmartSyncExplorerHybrid

wmathurin:
Also found and fixed a bug with fts when empty records are created (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/92bc95a7e750d79d7c09b1d3468a2a2dc8a4298f).
Will add a test for that use case soon.


hybrid/SampleApps/SmartSyncExplorerHybrid/res/values/strings.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="account_type">com.salesforce.samples.simplesync.login</string>
```
bhariharan: The ```account_type``` and ```app_package``` should also be renamed. ```app_package``` is used for PN.

bhariharan:
Looks good otherwise.

PR #957, opened 2015-09-29
Added content length and chunking hints to POST connections

clibc:
Large POST requests (i.e. file uploads) load all entity content into memory. Android now uses okhttp internally and this is default behaviour. This is a problem if more data is being sent over the network than can fit into an application's heap.

okhttp expects that multipart form uploads give the connection a hint on content length or if content is streaming, set a chunk size. Setting either of these should prevent OOM errors from occurring. This
fix isn't available for SDK users on API levels < 19


bhariharan:
@clibc Could you turn off your code formatter and re-submit with only the relevant code changes? Thanks!

clibc:
@bhariharan done.

I can also submit a new, clean PR that doesn't have the revert commits.

PR #958, opened 2015-09-30
Compiling for API 23 (needed to react native for android)

wmathurin:
Using org.apache.http.legacy (because it's no longer supported on API 23)

Using gradle android plugin 1.3.1 instead of 1.0.0 (to be able to do useLibrary org.apache.http.legacy)

Moved SampleApps to NativeSampleApps and HybridSampleApps because of the error (only with gradle android plugin 1.3):
Your project contains 2 or more modules with the same identification com.salesforce.androidsdk:SampleApps
at ":hybrid:SampleApps" and ":native:SampleApps".
 You must use different identification (either name or group) for each modules.


hybrid/HybridSampleApps/NoteSync/build.gradle
```diff
@@ -5,8 +5,9 @@ dependencies {
 }
 
 android {
-  compileSdkVersion 21
-  buildToolsVersion "21.1.1"
+  compileSdkVersion 23
+  buildToolsVersion '23.0.1'
+  useLibrary  'org.apache.http.legacy'
```
bhariharan: Is this required here? Won't it be automatically included if it's in the ```build.gradle``` for ```SalesforceSDK``` (chained dependency)?

native/NativeSampleApps/SmartSyncExplorer/build.gradle
```diff
@@ -5,8 +5,8 @@ dependencies {
 }
 
 android {
-  compileSdkVersion 21
-  buildToolsVersion "21.1.1"
+  compileSdkVersion 23
+  buildToolsVersion '23.0.1'
```
bhariharan: If this works without ```useLibrary  'org.apache.http.legacy'```, I think the others would too.

bhariharan:
Looks fine otherwise.

hybrid/HybridSampleApps/NoteSync/build.gradle
wmathurin: ContentSoqlSyncDownTarget in the NoteSync project makes use of org.apache.http.entity.StringEntity, otherwise you are right, it wouldn't need the useLibrary.

bhariharan: Gotcha. I thought every project built on ```SalesforceSDK``` would automatically get this, since it's a dependency on the library's gradle file.

PR #960, opened 2015-10-02
Cordova 4.0

bhariharan:


external/cordova
```diff
@@ -1 +1 @@
-Subproject commit ef4383561e127521449aeb8da90806f6ea6f699d
+Subproject commit a67bddbd821cebe7adb7e284592a09df45c9cc37
```
bhariharan: What the diff doesn't show is all that new magic in this submodule with new ```Cordova``` ;-)

gradle.properties
```diff
@@ -0,0 +1,3 @@
+cdvCompileSdkVersion=android-21
```
bhariharan: These properties are required by ```Cordova``` library project, to add as a ```Gradle``` dependency. Unfortunately, it doesn't work with ```Marshmallow``` quite yet - ```targetApi``` I mean.

libs/SalesforceSDK/build.gradle
```diff
@@ -2,6 +2,8 @@
 apply plugin: 'android-library'
 
 dependencies {
+  compile project(':external:cordova:framework')
```
bhariharan: True ```Gradle``` library project dependency.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/sfhybrid/SalesforceDroidGapActivity.java
```diff
@@ -131,31 +131,17 @@ protected ClientManager buildClientManager() {
...
-    		webSettings.setCacheMode(WebSettings.LOAD_DEFAULT);
-    		EventsObservable.get().notifyEvent(EventType.GapWebViewCreateComplete, appView);
-      	}
+    	EventsObservable.get().notifyEvent(EventType.GapWebViewCreateComplete, appView);
 	}
 
 	@Override
-    protected CordovaWebViewClient makeWebViewClient(CordovaWebView webView) {
-        return new SalesforceIceCreamWebViewClient(this, webView);
+    protected CordovaWebViewEngine makeWebViewEngine() {
```
bhariharan: This is where most of the interfaces had to be changed and where the meat of the upgrade is.

libs/SalesforceSDK/build.gradle
wmathurin: :+1:

wmathurin:
Looks good. Why the "do not merge", what's missing?

bhariharan:
@wmathurin I don't want to merge yet, because whitelisting is not implemented yet - hence, network requests are broken in hybrid apps. Also, there's a rogue seg fault every now and then in the ```WebView``` that I'd like to investigate a bit more before merging. Third reason (this can be done in a separate pull request) is to remove ```ant``` files and ensure ```cordova build``` works with ```Gradle```.

bhariharan:
@wmathurin To add the ```Whitelist``` plugin, I need to generate new hybrid apps from the ```CordovaPlugin``` repo, which requires a bunch of changes anyway because all our scripts use ```ant``` and that needs to change to ```Gradle```. I'm going to merge this in now and send another pull request when that part is ready. For now, ```SmartSyncExplorerHybrid``` works fine, since it uses the native ```REST``` library. The ones that use ```forcetk``` are a bit broken at this point. The plugins themselves work though - ```SmartStore``` and ```SmartSync```.

PR #961, opened 2015-10-05
Update SoqlSyncDownTarget.java

noornabee:
When quering more than 2000 records, the continueFetch method keeps extracting the same url this cause an infinite loop


bhariharan:
@noornabee Thanks for your pull request, but we have already made this fix in our ```unstable``` branch [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/931). We won't be patching this on ```master``` at this time.

PR #965, opened 2015-10-05
First Round of forcedroid Updates

bhariharan:


node/forcedroid.js
```diff
@@ -300,7 +291,6 @@ function createNativeApp(config, showNextSteps) {
     // copy <Android Package>/libs/SalesforceSDK -> <App Folder>/forcedroid/libs/SalesforceSDK
     var salesforceSDKRelativePath = path.join('libs', 'SalesforceSDK');
     copyFromSDK(packageSdkRootDir, config.targetdir, salesforceSDKRelativePath);
```
wmathurin: Are we still going to do that or at we going to have a dependency in the build.gradle file?

node/forcedroid.js
```diff
@@ -319,45 +308,26 @@ function createNativeApp(config, showNextSteps) {
     if (config.usesmartstore) {
         var smartStoreRelativePath = path.join('libs', 'SmartStore');
         copyFromSDK(packageSdkRootDir, config.targetdir, smartStoreRelativePath);
```
wmathurin: Same question as above

node/forcedroid.js
```diff
@@ -319,45 +308,26 @@ function createNativeApp(config, showNextSteps) {
...
-        shelljs.exec(androidExePath + ' update project -p ' + path.join(config.targetdir, path.basename(packageSdkRootDir), smartStoreRelativePath));
         var smartSyncRelativePath = path.join('libs', 'SmartSync');
         copyFromSDK(packageSdkRootDir, config.targetdir, smartSyncRelativePath);
```
wmathurin: Same question as above

node/forcedroid.js
bhariharan: Turns out you can't specify a repo path for Gradle dependency - it has to be on MavenCentral, so until we publish, we need a physical copy to link to. Also, the generated project needs a ```settings.gradle``` at the root level that specifies the projects, which I need to generate for a template app that we generate, and then I need to tweak the ```build.gradle``` files to reflect the correct relative path. It's easier than ```ant```, because there's only 1 line in 1 file to tweak (for native). We can get to a point where none of this necessary once we get on ```MavenCentral```.

bhariharan: But to your point, we won't need this copy step - we can link it in ```build.gradle```. I'll do that in the next pull request.

wmathurin: Maybe we could use https://jitpack.io/ ?

bhariharan: I looked into some solutions, including that one, but it's a bit intrusive. The ```Gradle``` team is working on a solution to this, and the next gen ```Android``` plugin is said to solve some of these problems too, so I figured we could wait a bit.

node/forcedroid.js
```diff
@@ -409,6 +374,24 @@ function copyFromSDK(packageSdkRootDir, targetDir, srcDirRelative) {
 }
 
 //
+// Creates a root level 'settings.gradle' file for the app
+//
+function createAppGradleFile(appFolderName, appName, usesSmartStore) {
```
bhariharan: Generating ```settings.gradle``` here.

wmathurin:
:+1:

PR #969, opened 2015-10-15
React native app generation with forcedroid

wmathurin:


reactnative/ReactNativeTemplateApp/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "ReactNativeTemplateApp",
+  "name": "ReactNativeTemplate",
```
bhariharan: Should this name match the declaration in ```MainActivity``` below, i.e., ```ReactNativeTemplateApp```? I'm not sure how the flow works here.

reactnative/ReactNativeTemplateApp/build.gradle
```diff
@@ -1,7 +1,7 @@
 apply plugin: 'android'
 
 dependencies {
-    compile project(':libs:SmartSync')
+    compile project(':libs:SalesforceSDK')
```
bhariharan: Shouldn't this be ```:libs:SmartSync```, because we set ```useSmartStore``` to ```true``` in ```forcedroid```?

reactnative/ReactNativeTemplateApp/package.json
wmathurin: The value in MainActivity has to match the registerComponent call in index.android.js (last line) and they do.
The name in the package.json really doesn't matter.

reactnative/ReactNativeTemplateApp/build.gradle
wmathurin: Your build.gradle tweaking code (which puts in the :forcedroid:libs:xxx dependency) only looks for libs:SalesforceSDK.

wmathurin: If you were to run the template app from the SDK repo, you would be okay because the template app only does rest api calls (no smartstore or smartsync).

wmathurin: All react native apps don't need to use smartstore/smartsync, so we could change forcedroid to ask the user what they want. Thoughts?

bhariharan: Oh right - now I remember. We generate everything with ```:libs:SalesforceSDK``` and replace with ```:libs:SmartSync``` if required. We could do the same thing as ```native``` apps by offering the option, or we could do what we do on ```hybrid```, which is including ```SmartSync``` by default. Given the current trend in questions on the community, it looks like people are using ```SmartSync``` quite a bit, so maybe we should bundle by default? I'm okay either way, since we have both paths already coded :-)

reactnative/ReactNativeTemplateApp/package.json
bhariharan: Sounds good.

bhariharan:
:+1:

reactnative/ReactNativeTemplateApp/build.gradle
wmathurin: I think we should include it by default. Otherwise we might need to support/document adding it after the fact.

PR #972, opened 2015-10-18
Fixing 'sed' and 'mv' to work on all platforms including Windows

bhariharan:


bhariharan:
```exec``` is now used only to run ```cordova``` CLI.

wmathurin:
:+1:

PR #980, opened 2015-10-29
Switch library JARs to be pulled from MavenCentral instead

bhariharan:


kchitalia:
:+1:

bhariharan:
In [this](https://github.com/bhariharan/SalesforceMobileSDK-Android/commit/e07ec231e39df719cad15ed30d3389018162c8c7) commit, I have removed the dependency on the ```Zutubi``` test runner, since we don't use the XML test report generation anymore, since we don't use ```Jenkins``` anymore.

PR #985, opened 2015-11-09
Reusing existing socket factory instance

bhariharan:


wmathurin:
:+1:

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SalesforceTLSSocketFactory.java
```diff
@@ -48,8 +48,16 @@
...
     private SSLSocketFactory ssLSocketFactory;
 
+    public static SalesforceTLSSocketFactory getInstance() throws KeyManagementException, NoSuchAlgorithmException {
+        if (INSTANCE == null) {
+            INSTANCE = new SalesforceTLSSocketFactory();
+        }
+        return INSTANCE;
+    }
+
     public SalesforceTLSSocketFactory() throws KeyManagementException, NoSuchAlgorithmException {
```
ashishmody: This should be made private.

bhariharan: @ashishmody I deliberately didn't make this ```private```. There might be some use cases where I want a new instance as opposed to the same.

PR #986, opened 2015-11-12
RestResponses can now be consumed as InputStreams

clibc:
I've added the asInputStream() method to RestResponse. In order to keep backwards-compatibility, the `consume()` method should be called in a `finally` block after reading the stream. I've also added `consumeQuietly()` as the IOException in `consume()` will never be thrown with the stream method. 

Here's a Java 6 example:
```java
try {
    final InputStream in = restResponse.asInputStream();
    // Read the stream (or don't)
} finally {
    restResponse.consumeQuietly();
}
```

and a Java 7 example:

```java
try (final InputStream in = restResponse.asInputStream()) {
    // Read the stream (or don't)
} finally {
    restResponse.consumeQuietly();
}
```

The existing RestResponse methods favor returning null instead of exceptions, so `asInputStream()` will return null if the content has already been streamed once. I can turn this into an IOException if it makes more sense. I think it's more intuitive:

```java
try (final InputStream in = restResponse.asInputStream()) {
    // No need to null-check "in"
} catch (IOException e) {
    // The stream couldn't be read for some reason
} finally {
    restResponse.consumeQuietly();
}
```

### DIFF
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
@@ -30,6 +30,7 @@ import android.os.Environment;
 import android.test.InstrumentationTestCase;
 
 import com.android.volley.Request;
+import com.google.common.io.CharStreams;
 import com.salesforce.androidsdk.TestCredentials;
 import com.salesforce.androidsdk.auth.HttpAccess;
 import com.salesforce.androidsdk.auth.OAuth2;
@@ -41,14 +42,18 @@ import com.salesforce.androidsdk.rest.RestRequest.RestMethod;
 
 import org.apache.http.HttpStatus;
 import org.json.JSONArray;
+import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
@@ -56,7 +61,6 @@ import java.util.Map;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
-
 /**
  * Tests for RestClient
  *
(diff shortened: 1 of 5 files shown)


clibc:
The tests appear to be failing because of this:

```
Could not find com.android.support:appcompat-v7:23.0.1.
```

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -552,11 +556,166 @@ public void testFileUpload() throws Exception {
...
 
+    /**
+     * Tests if a stream from {@link RestResponse#asInputStream()} is readable.
+     *
+     * @throws Exception
+     */
+    public void testResponseStreamIsReadable() throws Exception {
+        final RestResponse response = getStreamTestResponse();
+
+        try (InputStream in = response.asInputStream()) {
```
bhariharan: @clibc I don't know if we enforce ```Java 7``` as a minimum, so it might be best to go with the ```Java 6``` syntax here.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -552,11 +556,166 @@ public void testFileUpload() throws Exception {
...
+
+    /**
+     * Tests if a stream from {@link RestResponse#asInputStream()} is consumed (according to the REST client) by fully reading the stream.
+     *
+     * @throws Exception
+     */
+    public void testResponseStreamConsumedByReadingStream() throws Exception {
+        final RestResponse response = getStreamTestResponse();
+
+        try (InputStream in = response.asInputStream()) {
```
bhariharan: Same as above, and every other ```try``` block.

bhariharan:
@clibc Do you know which components require ```com.android.support:appcompat-v7:23.0.1```?

clibc:
@bhariharan it looks like Facebook react requires it. It should be in the Google Maven repository but I'm not sure how to add that to Travis.

bhariharan:
@clibc Yes I'm aware that ```React``` requires it, but the ```SalesforceSDK``` project shouldn't require it. Only the ```SalesforceReact``` project requires it. Since these tests are in the ```SalesforceSDKTest``` project, they should work fine without the compat library, since they're in no way tied to ```React```, correct?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -552,11 +556,171 @@ public void testFileUpload() throws Exception {
...
+    /**
+     * Tests if a stream from {@link RestResponse#asInputStream()} is consumed (according to the REST client) by fully reading the stream.
+     *
+     * @throws Exception
+     */
+    public void testResponseStreamConsumedByReadingStream() throws Exception {
+        final RestResponse response = getStreamTestResponse();
+
+        try {
+            InputStream in = response.asInputStream()
```
bhariharan: Missing semi-colon at the end. No compilation error? ;-)

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -552,11 +556,171 @@ public void testFileUpload() throws Exception {
...
+        final RestResponse response = getStreamTestResponse();
+
+        try {
+            InputStream in = response.asInputStream();
+            inputStreamToString(in);
+        } catch (IOException e) {
+            fail("The InputStream should be readable and an IOException should not have been thrown");
+        }
+
+        try (InputStream in = response.asInputStream()) {
```
bhariharan: One more instance of ```Java 7``` syntax to be replaced.

bhariharan:
@clibc You should be able to run ```SalesforceSDKTest``` standalone locally from the command line. See ```tools/sdk.sh```.

clibc:
@bhariharan Gradle configures all the projects even if you only want to run the tasks from a subproject. The missing dependency in SalesforceReact is causing that project's configuration to fail, the entire configuration phase to fail and then the entire build to fail.

Unfortunately this is the Gradle lifecycle and that dependency will have to be found before any tasks can be run in any subproject.

clibc:
The dependency is in my local build, so I can run the tests through IntelliJ

bhariharan:
@clibc Crap - you're right. We used to have that in there, I wonder when it dropped off. Could you add ```compile 'com.android.support:appcompat-v7:23.0.1'``` to ```build.gradle``` for the ```SalesforceReact``` module, and check if that fixes it? If it does, could you add that to this pull request? Thanks!

clibc:
Done. I think that the Android support repository might need to be added to Travis:
```
    - extra-google-m2repository
    - extra-android-m2repository
```

bhariharan:
Looks good. Thanks @clibc!

PR #987, opened 2015-11-12
Switched logging to INFO for connected Android test tasks

clibc:
When the connected Android tests are running, there's no log output and Travis thinks that the build has hung after 10 minutes. This change makes the test task (and only the test task) print out both passed and failed tests. I think that having a 10 minute watchdog is a good idea and it's better to make the  test task more verbose.

Unfortunately due to the way the Android Gradle plugin works, I have to add a snippet of Groovy to each build file instead of being able to add it once in an `allprojects` block.

### DIFF
--- a/libs/SalesforceHybrid/build.gradle
+++ b/libs/SalesforceHybrid/build.gradle
@@ -39,3 +39,9 @@ android {
     exclude 'META-INF/NOTICE'
   }
 }
+
+afterEvaluate {
+  // Find all connected Android tests and ensure they log all passed tests. This keeps the Travis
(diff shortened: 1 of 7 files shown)


clibc:
@kchitalia take a look at this too

bhariharan:
@clibc Could you add this to ```RestExplorer```, ```TemplateApp``` and ```SmartSyncExplorer``` too? Those projects have associated tests too. Thanks!

wmathurin:
:+1:

kchitalia:
This is awesome. :100:

PR #1005, opened 2015-12-13
Saving identity url into the bundle of Account options.

harsha973:
Saving identity url into the bundle of Account options.


bhariharan:
@harsha973 Could you send this pull request against our ```unstable``` branch? Thanks!

PR #1025, opened 2016-01-08
Added timeout to fix the following situation (related to issue #1026)

coderReview:
1- Open an Salesforce application (SmartSync explorer example for instance)
2- Let it finish loading and sync down data
3- Turn Airplane mode on
4- Edit a record and click save
5- Turn Airplane mode off
6- Synchronize up records

Result: Synchronization takes a very long time (>20min) or hangs indefinitely 
Expected: Synchronization should happen quickly

PS: Values could be configurable in XML file


coderReview:
Related to issue #1026

bhariharan:
:+1:

PR #1029, opened 2016-01-13
Make cert flow configurable

bhariharan:


bhariharan:
No code changes. Just moved it into a protected method.

kchitalia:
:+1:

PR #1036, opened 2016-01-21
Fixing source directory for SmartStore

bhariharan:
The problem with the absolute here (```external...```) is that it doesn't exist in the ```CordovaPlugin```. When ```SmartStore``` is used in the plugin, this is broken. It should be ```libs``` - the binaries in there have symlinks to ```external...``` anyway.


kchitalia:
:+1:

PR #1039, opened 2016-01-28
Simple cleanup

bhariharan:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
```diff
@@ -149,7 +149,7 @@ public QueryType getQueryType() {
     public long getLatestModificationTimeStamp(JSONArray records) throws JSONException {
         long maxTimeStamp = -1;
         for (int i = 0; i < records.length(); i++) {
-            String timeStampStr = JSONObjectHelper.optString(records.getJSONObject(i), getModificationDateFieldName());
+            String timeStampStr = JSONObjectHelper.optString(records.getJSONObject(i), modificationDateFieldName);
```
bhariharan: It is a ```protected``` member.

wmathurin: Shouldn't we use getters instead and not make the field protected?

bhariharan: I suppose we could. This field is likely to be set only once in the lifecycle of the target (while creation), so I figured the protected member would probably be set/overridden in the constructor for the subclass and that'd be that. The getter would be used by other classes that use the target as public API to get the field name. Would you prefer making it private again? I'm okay either way.

wmathurin:
It's usually considered better to use a getter (see discussion here: http://stackoverflow.com/questions/2279662/java-protected-fields-vs-public-getters) except maybe on Android (see: http://developer.android.com/training/articles/perf-tips.html#GettersSetters).

In this case, I don't think the possible performance degradation of the method call would matter, since we are doing network and database calls around it which are order of magnitude slower.

So I would vote for the getter + private fields.

bhariharan:
Interesting! I'll switch them to getter + private.

bhariharan:
@wmathurin Done [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/commit/9b9decbcd5a36204b4cb8ca856a805a4a29bc1a4).

PR #1040, opened 2016-01-28
Add ID field if it's missing from query spec

bhariharan:


bhariharan:
@wmathurin I'm considering doing something similar for ```LAST_MODIFIED_DATE``` if it's missing, since that's special too. What do you think?

wmathurin:
Should we add a test?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -65,8 +68,20 @@ public SoqlSyncDownTarget(String query) {
...
 	}
-	
+
+    private void addIdFieldIfRequired() {
+        if (!TextUtils.isEmpty(query)) {
+            final String idFieldName = getIdFieldName();
+
+            // Inserts the mandatory 'Id' field if it doesn't exist.
+            if (!query.contains(idFieldName)) {
+                query = query.replaceFirst("([sS][eE][lL][eE][cC][tT] )", "select " + idFieldName + ", ");
```
wmathurin: I guess it should work even with subqueries (https://developer.salesforce.com/docs/atlas.en-us.200.0.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select.htm) thanks to the replaceFirst.

bhariharan: Yes it should.

bhariharan:
Sure - I'll add a test. The way it's setup now, ```SmartSyncExplorer``` will fail if this stops working, because I removed ```Id``` from its index specs.

bhariharan:
Adding ```LAST_MODIFIED_DATE``` [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/commit/67982df2a475fe742bc4620ff73d7010d632724a).

wmathurin:
:+1:

bhariharan:
Added test [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/commit/8c8f86387d0ee486dd9cd4f2f294b804af118fa5).

PR #1043, opened 2016-02-05
Shelljs update is breaking forcedroid

wmathurin:
While testing with test_force.js I was very puzzled to have all android app generation fail (even though they were working last night). For a while, I thought I had lost something in my recent refactoring.

It turns out that shelljs released v0.6.0 a few hours ago and now forcedroid breaks.
This is a quick fix to make it work again - a better fix would be to revisit our use of shelljs and see how they need to be changed.

People that now install forcedroid and didn't previously have shelljs will run into the issue. So we shouldn't take too long releasing our next release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     ],
     "bin" : { "forcedroid" : "node/forcedroid.js" },
     "dependencies": {
-        "shelljs": ">=0.1.4"
+        "shelljs": "~0.5.3"
     },
     "scripts": {
         "prepublish": "external/shared/node/prepublish.js",


package.json
```diff
@@ -10,7 +10,7 @@
     ],
     "bin" : { "forcedroid" : "node/forcedroid.js" },
     "dependencies": {
-        "shelljs": ">=0.1.4"
+        "shelljs": "~0.5.3"
```
wmathurin: ~a.b.c means accept a.b.x where x >= c
according to http://stackoverflow.com/questions/22343224/difference-between-tilde-and-caret-in-package-json
So 0.6.0 won't be used

khawkins:
:+1:

kchitalia:
forceios is not affected by this, right?

wmathurin:
@kchitalia good point.
Looking at https://github.com/forcedotcom/SalesforceMobileSDK-Shared/pull/327#issuecomment-180222102, I see there is one app type (hybrid_local) where running forceios fails. I looked at it and it's very similar to the failures I was getting with android. By specify 0.5.3 for the shelljs version, the issue goes away, see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1378.

bhariharan:
:+1:

kchitalia:
:+1:

PR #1044, opened 2016-02-09
Fix for instanceUrl bug

bhariharan:


wmathurin:
:+1:

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -900,18 +900,6 @@ public String getAccountType() {
         return context.getString(getSalesforceR().stringAccountType());
     }
 
-    /**
-     * Indicates whether the app is running on a tablet.
-     *
-     * @return True if the application is running on a tablet.
-     */
-    public static boolean isTablet() {
```
bhariharan: This is an inaccurate API that has outlived it's purpose. We can't make the determination of phone (vs) tablet, with a 7" screen in the mix. Also, with the advent of fragments, this API means very little now. It's not being used anywhere, since we switched to ```ActionBar```.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
```diff
@@ -167,8 +166,6 @@ public Bundle getAccountRemovalAllowed(AccountAuthenticatorResponse response, Ac
 
 		/**
          * Uses the refresh token to get a new access token.
-         * Remember that the authenticator runs under its own separate process, so if you want to debug you
```
bhariharan: Not accurate anymore. Same process.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -628,7 +628,10 @@ public String getNewAuthToken() {
                     Log.w("AccMgrAuthTokenProvider:fetchNewAuthToken", "accountManager.getAuthToken returned null bundle");
                 } else {
                     newAuthToken = bundle.getString(AccountManager.KEY_AUTHTOKEN);
-                    newInstanceUrl = bundle.getString(AuthenticatorService.KEY_INSTANCE_URL);
+                    final String encryptedInstanceUrl = bundle.getString(AuthenticatorService.KEY_INSTANCE_URL);
```
bhariharan: We send the ```instanceUrl``` encrypted, but never bothered to decrypt it. I guess it was working because we never did any org split testing where the ```instanceUrl``` actually changes.

ktanna:
:+1:

PR #1045, opened 2016-02-10
Fix for NPE crash

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -650,8 +650,12 @@ private void refreshAccessToken(HttpResponse response) throws IOException {
 				if (newAuthToken != null) {
 					setAuthToken(newAuthToken);
 				}
+
                 // Check if the instanceUrl changed
                 String instanceUrl = authTokenProvider.getInstanceUrl();
+				if (instanceUrl == null) {
```
bhariharan: When refresh token is revoked, ```instanceUrl``` is guaranteed to be ```null``` here. We try to URI parse ```instanceUrl```, which breaks. We didn't notice it on our sample apps because sample apps use ```sendAsync```, which bubbles up the error, whereas ```sendSync``` doesn't.

wmathurin:
:+1:

kchitalia:
Does it make sense to bubble up the error when sendSync is used?

ktanna:
:+1:

bhariharan:
@kchitalia Yes - the alternative is a crash, which prevents normal flow of the app. ```sendSync``` throws ```IOException```, so this is the expected response for in-flight requests post-logout.

PR #1049, opened 2016-02-16
Fixing user agent computation for Android

bhariharan:


bhariharan:
This makes it ```HybridSmartSyncLocal``` like ```iOS```, as opposed to ```HybridLocalSmartSync```.

wmathurin:
:+1:

PR #1057, opened 2016-02-22
Adding response headers to Execution

bhariharan:
We were swallowing the response headers, without propagating them upwards.


bhariharan:
@a-joshi-sfdc Here you go. This will be in the next ```Mobile SDK``` release.

wmathurin:
:+1:

PR #1061, opened 2016-02-22
Fix for bug revealed in template app (query returning 400)

wmathurin:
Query params were lost for delete/get/head


bhariharan:
:+1:

wmathurin:
Re-submitting against master

PR #1070, opened 2016-02-26
Added context as a parameter in PushNotificationInterface

fungi-fungi:
In case if you want to start ```IntentService``` after receiving push notification you need Context which is not currently available in implementation of  ```PushNotificationInterface```. Since ```PushService``` is actually anther ```IntentService``` we can pass context as a parameter of ```PushNotificationInterface```.


bhariharan:
```Context``` is available through ```SalesforceSDKManager.getInstance().getAppContext()```, which can be used to start an ```IntentService```. It probably shouldn't be started from ```PushService``` anyway.

PR #1080, opened 2016-03-17
Encrypting Auth Token

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -381,8 +381,6 @@ public Bundle createNewAccount(String accountName, String username, String refre
             // WARNING! This assumes all user data is a String!
             accountManager.setUserData(acc, key, extras.getString(key));
         }
-
-        accountManager.setAuthToken(acc, AccountManager.KEY_AUTHTOKEN, authToken);
```
bhariharan: This is a useless call, because ```AccountManager``` automatically caches auth token in memory anyway.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -617,17 +613,20 @@ public String getNewAuthToken() {
                 gettingAuthToken = true;
             }
 
-            // Invalidate current auth token
-            clientManager.invalidateToken(lastNewAuthToken);
+            // Invalidate current auth token.
+            final String cachedAuthToken = clientManager.accountManager.peekAuthToken(acc, AccountManager.KEY_AUTHTOKEN);
```
bhariharan: This is important, because unless we invalidate the correct token, the subsequent call to ```getAuthToken``` won't go to our implementation - it will go directly to ```AccountManager``` cache instead.

bhariharan:
Tested regular API calls and refresh calls for expired tokens. The only remaining piece of work is to test upgrade path from unencrypted auth token to encrypted auth token. I might have to do a one-time wipe of existing auth token and force a refresh for older versions of Mobile SDK.

wmathurin:
Are users going to be logged out if the app is upgraded from a older SDK to 4.2?

bhariharan:
No. That's the upgrade step that I had mentioned. I'm working on it. Since
there are no changes to refresh token, logout shouldn't be required.

wmathurin:
That's true.

wmathurin:
Looks fine. All the tests passing?

bhariharan:
* Fixed test failure [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1080/commits/fe3e632934a0c1ff0900781350394e2e3d2de5fa).
* Upgrade steps work fine with no code changes. I verified by revoking the token, and the refresh worked fine too.
* Relevant tests are running fine.

PR #1086, opened 2016-03-22
Networking was broken on 4.4

wmathurin:
Bringing back our ssl socket factory

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SalesforceTLSSocketFactory.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SalesforceTLSSocketFactory.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2015-present, salesforce.com, inc.
+ * All rights reserved.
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SalesforceTLSSocketFactory.java
```diff
@@ -0,0 +1,111 @@
...
+ * Lollipop (5.0) and above.
+ *
+ * FIXME: Remove this class once minApi >= Lollipop.
+ *
+ * @author bhariharan
+ */
+public class SalesforceTLSSocketFactory extends SSLSocketFactory {
+
+    private static SalesforceTLSSocketFactory INSTANCE;
+    private SSLSocketFactory delegate; // XXX needs to be called delegate to work around okhttp3 bug: see https://github.com/square/okhttp/issues/2323#issuecomment-185925554
```
wmathurin: Had to rename the private variable to delegate 
See https://github.com/square/okhttp/issues/2323#issuecomment-185925554

bhariharan:
:+1:

PR #1089, opened 2016-03-24
Adds pass-through constructors so Hybrid apps can implement their own WebView and WebViewClient subclasses

daveespo:
The upgrade to MobileSDK 4 has made it impossible to implement `WebView` and `WebViewClient` subclasses for Hybrid apps

I can override `makeWebViewEngine` in my `Activity` ... and implement my own `WebViewEngine` subclass which derives from `SalesforceWebViewEngine`

However, inside my subclass, I'm unable to call the `SystemWebViewEngine` constructors to properly set the appView 

These pass-through constructors are necessary so I can call `super(webView,preferences)` in my subclass the same way `SalesforceWebViewEngine` does


bhariharan:
Seems reasonable. :+1:

daveespo:
Thanks!

PR #1090, opened 2016-03-24
Fixing regressions

wmathurin:


bhariharan:
Interesting! Is this a new requirement with ```OkHttp``` or always been there?

bhariharan:
:+1:

wmathurin:
The callback with okhttp is running on a network thread, so we have to be careful not to do any UI work.
A serious app should not do network calls directly from the UI code but our sample apps do ;-)

bhariharan:
Sounds good. :+1:

wmathurin:
Missed one, now fixed see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1090/commits/9dcbb2693e06198eae8ac1537f2b8174133458b3

wmathurin:
Repro steps:
- create hybrid remote application
- login
- through a browser, login, go to setup, look for session, from the user session information page, remove all sessions
- reload hybrid remote application

Before the fix: it would crash
After the fix: it load the page (after doing a token / cookies refresh)

bhariharan:
:+1:

PR #1091, opened 2016-03-24
Adding remote delete sync API

bhariharan:
I wanted to send the pull request early, to get the review process going.

Outstanding Items:
* Implementation for ```SOSL``` and ```MRU```.
* Misc. cleanup.
* Tests.


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,58 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
     	return sync;
     }
 
+    /**
+     * Deletes local copies of records that have been deleted on the server.
+     *
+     * @param syncId Sync ID.
+     * @throws JSONException
+     */
+    public void syncRemoteDeletes(long syncId) throws JSONException {
```
bhariharan: New API to sync remotely deleted records.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,58 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
...
+        final String soupName = sync.getSoupName();
+        final String idFieldName = sync.getTarget().getIdFieldName();
+
+        /*
+         * Fetches list of IDs present in local soup that have not been modified locally.
+         */
+        final Set<String> localIds = new HashSet<String>();
+        QuerySpec querySpec = QuerySpec.buildAllQuerySpec(soupName, idFieldName, QuerySpec.Order.ascending, 10);
+        int count = smartStore.countQuery(querySpec);
+        querySpec = QuerySpec.buildSmartQuerySpec("SELECT {" + soupName + ":" + idFieldName + "} FROM {" + soupName + "} WHERE {" + soupName + ":" + LOCAL + "}='false'", count);
```
bhariharan: Fetches all IDs of records not modified locally.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,58 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
...
+                    localIds.add(idJson.optString(0));
+                }
+            }
+        }
+
+        /*
+         * Fetches list of IDs still present on the server from the list of local IDs
+         * and removes the list of IDs that are still present on the server.
+         */
+        final Set<String> remoteIds = ((SyncDownTarget) sync.getTarget()).getListOfRemoteIds(this, localIds);
```
bhariharan: At this point, we know it has to be a sync down operation, so we can safely assume so here to get remote IDs.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,58 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
...
+        if (remoteIds != null) {
+            localIds.removeAll(remoteIds);
+        }
+
+        // Deletes extra IDs from SmartStore.
+        if (localIds.size() > 0) {
+            // TODO: Remove entries from SmartStore.
```
bhariharan: TODO: Actual deletion from ```SmartStore```.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,58 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
...
+            }
+        }
+
+        /*
+         * Fetches list of IDs still present on the server from the list of local IDs
+         * and removes the list of IDs that are still present on the server.
+         */
+        final Set<String> remoteIds = ((SyncDownTarget) sync.getTarget()).getListOfRemoteIds(this, localIds);
+        if (remoteIds != null) {
+            localIds.removeAll(remoteIds);
```
bhariharan: Removes entries available locally that no longer exist on the server.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/MruSyncDownTarget.java
```diff
@@ -110,6 +110,12 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
         return null;
     }
 
+    @Override
+    public Set<String> getListOfRemoteIds(SyncManager syncManager, Set<String> localIds) {
```
bhariharan: Abstract method, so that custom sync down targets can choose how to deal with this.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -130,6 +138,80 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
         return records;
     }
 
+    @Override
+    public Set<String> getListOfRemoteIds(SyncManager syncManager, Set<String> localIds) {
```
bhariharan: ```SOQL``` implementation.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -130,6 +138,80 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
...
+            return null;
+        }
+        final String idFieldName = getIdFieldName();
+        final Set<String> remoteIds = new HashSet<String>();
+
+        /*
+         * Compute how many SOQL queries are required to get the IDs.
+         * SOQL has a limit of 2000 records per query, so we need to break it up.
+         */
+        final List<Set<String>> uberQueryList = new ArrayList<Set<String>>();
```
bhariharan: ```SOQL``` has a limit of ```2000``` records, so we need to break the set into subsets and send them as separate ```SOQL``` queries.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -130,6 +138,80 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
...
+                uberQueryList.add(smallSet = new HashSet<String>());
+            }
+            smallSet.add(value);
+        }
+
+        // Iterates through the uber list of sets and creates multiple SOQL queries.
+        for (final Set<String> queryList : uberQueryList) {
+
+            // Constructs a SOQL query to get IDs.
+            final StringBuilder soql = new StringBuilder("SELECT ");
```
bhariharan: Constructing ```SOQL``` with ```IN``` clause.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: Should we have a more generic names, since records that no longer the query (for whatever reason) should also be purged?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -130,6 +138,80 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
...
+            soql.append(idFieldName);
+            soql.append(" FROM ");
+
+            // Reads SObject name from the SOQL query itself.
+            final String[] result = query.toLowerCase().split("from");
+            result[1] = result[1].trim();
```
bhariharan: Getting entity name from the original sync down query.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -130,6 +138,80 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
...
+                final RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
+                if (response != null && response.isSuccess()) {
+                    final JSONObject responseJson = response.asJSONObject();
+                    if (responseJson != null) {
+                        final JSONArray records = responseJson.getJSONArray(Constants.RECORDS);
+                        if (records != null) {
+                            for (int i = 0; i < records.length(); i++) {
+                                final JSONObject idJson = records.optJSONObject(i);
+                                if (idJson != null) {
+                                    remoteIds.add(idJson.optString(idFieldName));
```
bhariharan: Reading ```ID``` from response.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoslSyncDownTarget.java
```diff
@@ -91,11 +91,16 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
         return null;
     }
 
+    @Override
+    public Set<String> getListOfRemoteIds(SyncManager syncManager, Set<String> localIds) {
+        // TODO: Do a SOQL query for each object type with 'IN' clause from localIds.
```
bhariharan: TODO: ```SOSL``` implementation. This is not going to be fun - need to iterate and build separate ```SOQL``` queries for different entities.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
```diff
@@ -126,6 +127,15 @@ public JSONObject asJSON() throws JSONException {
     public abstract JSONArray continueFetch(SyncManager syncManager) throws IOException, JSONException;
 
     /**
+     * Fetches list of IDs still present on the server from the list of local IDs.
+     *
+     * @param syncManager SyncManager instance.
+     * @param localIds Local IDs from SmartStore.
+     * @return List of IDs still present on the server.
+     */
+    public abstract Set<String> getListOfRemoteIds(SyncManager syncManager, Set<String> localIds);
```
bhariharan: Abstract method, so that custom sync down targets can choose how to deal with this.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: Instead of running a count query, couldn't you do what we do in SyncManager:getDirtyRecordIds where we keep getting the next page until we get a non-full page.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: Coming up with a good API name seems to be the hardest problem. ;-) How about ```syncNonExistentRecords```?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
wmathurin: We don't want to deal just with deleted records, we also want to purge out any records that no longer matches the SOQL query for the sync down.

So we should:
run the SOQL sync down with a modified SELECT to return only ids
Or
run the SOQL sync down with a modified SELECT to return only ids and an additional "and id in (local ids)" clause.
The second one is tricky because there are limits on SOQL query size.

bhariharan: Makes sense. I'll switch it to run with a modified ```SELECT Id``` query - should be fairly straightforward, except for the paging bit.

wmathurin: The paging bit is already there in startFetch/continueFetch.

Maybe you should create another SOQLSyncDownTarget with the modified soql query and call startFetch/continueFetch on it to avoid duplicating code. In that case, instead of having a method to get remote ids, you would have a method on SOQLSyncDownTarget to get back the ids-only SyncDownTarget?

Or alternatively, you could refactor startFetch/continueFetch to have an extra parameter: queryToRun. You would have 

```java
startFetch(SyncManager syncManager, long maxTimeStamp) {
  queryToRun = ...
  startFetch(syncManager, maxTimeStamp, queryToRun)
}

startFetch(SyncManager syncManager, long maxTimeStamp, queryToRun) {
  // pretty much the current implementation
}
```
In that approach, the getRemoteIds would call startFetch with the ids-only query.

Personally, I kind of like the first approach better, the looping through startFetch/continueFetch would be done in the sync manager.

bhariharan:
@wmathurin I did some refactoring based on your recommendations.
* Renamed new API to ```reSyncRemoteUpdates``` (couldn't think of anything better).
* Reused existing ```SOQL``` query.
* Reused ```startFetch``` and ```continueFetch```, by adding an overloaded version of ```startFetch``` that takes a query parameter.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,59 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
     	return sync;
     }
 
+    /**
+     * Removes local copies of records that have been deleted on the server
+     * or do not match the query results on the server anymore.
+     *
+     * @param syncId Sync ID.
+     */
+    public void reSyncRemoteUpdates(long syncId) throws JSONException {
```
bhariharan: Renamed API.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/MruSyncDownTarget.java
```diff
@@ -86,22 +87,25 @@ public JSONObject asJSON() throws JSONException {
...
+        return startFetch(syncManager, maxTimeStamp, soql);
+    }
 
-        // Get recent items attributes from server
-        request = RestRequest.getRequestForQuery(syncManager.apiVersion, soql);
-        response = syncManager.sendSyncWithSmartSyncUserAgent(request);
-        JSONObject responseJson = response.asJSONObject();
-        JSONArray records = responseJson.getJSONArray(Constants.RECORDS);
+    @Override
+    public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String queryRun) throws IOException, JSONException {
```
bhariharan: New overloaded version of ```startFetch```.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -100,16 +104,21 @@ public JSONObject asJSON() throws JSONException {
 
     @Override
     public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws IOException, JSONException {
-        String queryToRun = maxTimeStamp > 0 ? SoqlSyncDownTarget.addFilterForReSync(query, maxTimeStamp) : query;
+        return startFetch(syncManager, maxTimeStamp, query);
+    }
+
+    @Override
+    public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String queryRun) throws IOException, JSONException {
```
bhariharan: New overloaded version of ```startFetch```.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
bhariharan: The implementation here is really simple now. It simply breaks down the ```SOQL``` query to get only ```Id```, and reuses ```startFetch``` and ```continueFetch``` to build records.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoslSyncDownTarget.java
```diff
@@ -76,13 +77,17 @@ public JSONObject asJSON() throws JSONException {
 
     @Override
     public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws IOException, JSONException {
-        RestRequest request = RestRequest.getRequestForSearch(syncManager.apiVersion, query);
+        return startFetch(syncManager, maxTimeStamp, query);
+    }
+
+    @Override
+    public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String queryRun) throws IOException, JSONException {
```
bhariharan: New overloaded version of ```startFetch```.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
```diff
@@ -118,6 +120,16 @@ public JSONObject asJSON() throws JSONException {
...
 
     /**
+     * Starts fetching records conforming to target query.
+     *
+     * @param syncManager SyncManager instance.
+     * @param maxTimeStamp Max timestamp.
+     * @param queryRun Query to run.
+     * @throws IOException, JSONException
+     */
+    public abstract JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String queryRun) throws IOException, JSONException;
```
bhariharan: New overloaded version of ```startFetch```.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
```diff
@@ -175,4 +195,23 @@ public long getLatestModificationTimeStamp(JSONArray records) throws JSONExcepti
...
         custom
     }
+
+    /**
+     * Helper method to parse IDs from a network response to a SOQL query.
+     *
+     * @param records SObject records.
+     * @return Set of IDs.
+     */
+    protected Set<String> parseIdsFromResponse(JSONArray records) {
```
bhariharan: Simple util method used by all sync down targets.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: How about reSyncCleanGhosts ?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
rwhitley: Or `syncMissingRecords`?

wmathurin: It will only remove local records that shouldn't be there. It's not
bringing down updates or records. Maybe it should be called:
cleanReSyncGhosts?

bhariharan:
Fixed API name [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1091/commits/6013ac71bce67e83afd5d862ff8e1cb3fde5e780).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
```diff
@@ -119,17 +128,49 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
...
+            return null;
+        }
+        final String idFieldName = getIdFieldName();
+        final Set<String> remoteIds = new HashSet<String>();
+
+        // Alters the SOQL query to get only IDs.
+        final StringBuilder soql = new StringBuilder("SELECT ");
+        soql.append(idFieldName);
+        soql.append(" FROM");
+        final String[] fromClause = query.toLowerCase().split("from");
```
wmathurin: Do you know if object type / field names need to be properly capitalized? If they do, then we should locate "from" in the lower case query, and split the original query at that location.

bhariharan: ```SOQL``` doesn't appear to be case-sensitive AFAIK. I tried the query in our sample app both ways on the ```Contact``` object, and all lowercase appears to work fine too.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: Maybe we should use TreeSet's for speed.

wmathurin:
Looks fine. I would call it cleanReSyncGhosts instead of reSyncCleanGhosts, since it really does a clean.
Also we should add new tests.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -287,6 +286,65 @@ public SyncState syncUp(SyncUpTarget target, SyncOptions options, String soupNam
...
+         * Fetches list of IDs still present on the server from the list of local IDs
+         * and removes the list of IDs that are still present on the server.
+         */
+        final Set<String> remoteIds = ((SyncDownTarget) sync.getTarget()).getListOfRemoteIds(this, localIds);
+        if (remoteIds != null) {
+            localIds.removeAll(remoteIds);
+        }
+
+        // Deletes extra IDs from SmartStore.
+        if (localIds.size() > 0) {
```
bhariharan: Added delete implementation here.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/MruSyncDownTarget.java
bhariharan: Added ```MRU``` implementation here. Verified it works with sample app.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SoqlSyncDownTarget.java
bhariharan: Verified SOQL end-to-end using sample app.

bhariharan:
Outstanding Items:

* Tests.
* Verifying ```SOSL``` implementation with sample app (```SOQL``` and ```MRU``` verified).

bhariharan:
Added tests for ```SOQL```, ```SOSL``` and ```MRU```, and verified behavior on all 3. This should be good to go now.

wmathurin:
Looks good.

PR #1095, opened 2016-03-29
Oauth bridge has authenticate method

wmathurin:
Application no longer does a login by default, instead the javascript needs to call authenticate if it needs to be authenticated.

First I had a shouldAuthenticate method in SalesforceReactActivity, but it was creating problems when the login automatically started in the middle of the react page running. 

This implementation allows for react native driven splash screen (see comment below).

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceOAuthPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceOAuthPlugin.java
@@ -28,7 +28,6 @@ package com.salesforce.androidsdk.phonegap.plugin;
 
 import android.util.Log;
 
-import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.phonegap.ui.SalesforceDroidGapActivity;
 import com.salesforce.androidsdk.phonegap.ui.SalesforceWebViewClientHelper;
 
@@ -36,12 +35,13 @@ import org.apache.cordova.CallbackContext;
 import org.apache.cordova.PluginResult;
 import org.json.JSONArray;
 import org.json.JSONException;
-import org.json.JSONObject;
 
 /**
  * PhoneGap plugin for Salesforce OAuth.
  */
 public class SalesforceOAuthPlugin extends ForcePlugin {
+
+    private static String TAG = "SalesforceOAuthPlugin";
 
     /**
      * Supported plugin actions that the client can take.
@@ -75,7 +75,7 @@ public class SalesforceOAuthPlugin extends ForcePlugin {
                     default: throw new Exception("No handler for action " + action);
                 }
             } catch (Exception e) {
-                Log.w("SalesforceOAuthPlugin.execute", e.getMessage(), e);
+                Log.w(TAG, "execute:" +  e.getMessage(), e);
                 callbackContext.error(e.getMessage());
             }
             }
@@ -90,7 +90,7 @@ public class SalesforceOAuthPlugin extends ForcePlugin {
      * @throws JSONException
      */
     protected void authenticate(final CallbackContext callbackContext) throws JSONException {
-        Log.i("SalesforceOAuthPlugin.authenticate", "authenticate called");
+        Log.i(TAG, "authenticate called");
         ((SalesforceDroidGapActivity) cordova.getActivity()).authenticate(callbackContext);
 
         // Done.
@@ -105,15 +105,8 @@ public class SalesforceOAuthPlugin extends ForcePlugin {
      * @throws JSONException
      */
     protected void getAuthCredentials(CallbackContext callbackContext) throws JSONException {
-        Log.i("SalesforceOAuthPlugin.getAuthCredentials", "getAuthCredentials called");
-    	JSONObject credentials = ((SalesforceDroidGapActivity) cordova.getActivity()).getJSONCredentials();
-        if (credentials == null) {
-            Log.w("SalesforceOAuthPlugin.getAuthCredentials", "getAuthCredentials failed - never authenticated");
-            callbackContext.error("Never authenticated");
-        } else {
-            Log.i("SalesforceOAuthPlugin.getAuthCredentials", "getAuthCredentials successful");
-            callbackContext.success(credentials);
-        }    
+        Log.i(TAG, "getAuthCredentials called");
+        ((SalesforceDroidGapActivity) cordova.getActivity()).getAuthCredentials(callbackContext);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceOAuthPlugin.java
```diff
@@ -105,15 +105,8 @@ protected void authenticate(final CallbackContext callbackContext) throws JSONEx
...
-        Log.i("SalesforceOAuthPlugin.getAuthCredentials", "getAuthCredentials called");
-    	JSONObject credentials = ((SalesforceDroidGapActivity) cordova.getActivity()).getJSONCredentials();
-        if (credentials == null) {
-            Log.w("SalesforceOAuthPlugin.getAuthCredentials", "getAuthCredentials failed - never authenticated");
-            callbackContext.error("Never authenticated");
-        } else {
-            Log.i("SalesforceOAuthPlugin.getAuthCredentials", "getAuthCredentials successful");
-            callbackContext.success(credentials);
-        }    
+        Log.i(TAG, "getAuthCredentials called");
```
wmathurin: Refactored to keep the cordova plugin and the react bridge as similar as possible.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -534,30 +542,6 @@ private void addSidCookieForDomain(CookieManager cookieMgr, String domain, Strin
        cookieMgr.setCookie(domain, cookieStr);
    }
 
-   /**
```
wmathurin: Moved to RestClient - now being called from two places: react and cordova

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/SalesforceNetReactBridge.java
```diff
@@ -150,15 +151,7 @@ private RestRequest prepareRestRequest(ReadableMap args) throws UnsupportedEncod
...
-        if (restClient == null) {
-            UserAccount account = SalesforceSDKManager.getInstance().getUserAccountManager().getCurrentUser();
-            if (account == null) {
-                restClient = SalesforceSDKManager.getInstance().getClientManager().peekUnauthenticatedRestClient();
-            } else {
-                restClient = SalesforceSDKManager.getInstance().getClientManager().peekRestClient(account);
-            }
-        }
-        return restClient;
+        return ((SalesforceReactActivity) getCurrentActivity()).getRestClient();
```
wmathurin: Using the rest client owned by the react activity

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java
```diff
@@ -102,50 +96,80 @@ public void onPause() {
...
+                     SalesforceReactActivity.this.client = client;
+
+
+	                /*
+                     * Do a cheap REST call to refresh the access token if needed.
+                     * If the login took place a while back (e.g. the already logged
+                     * in application was restarted), then the returned session ID
+                     * (access token) might be stale.
+                     */
+                     client.sendAsync(RestRequest.getRequestForResources(ApiVersionStrings.getVersionNumber(SalesforceReactActivity.this)), new RestClient.AsyncRequestCallback() {
```
wmathurin: We do the same for cordova.
If you use the session token directly, it better be fresh.
That code could also be moved to the getAuthCredentials method.

reactnative/ReactNativeTemplateApp/js/index.android.js
```diff
@@ -37,22 +37,40 @@ var {
     Navigator
 } = React;
 var forceClient = require('./react.force.net.js');
+var oauth = require('./react.force.oauth.js');
 
 var App = React.createClass({
+    getInitialState: function() {
```
wmathurin: NEW: Your app needs to call authenticate before doing anything that requires authentication

wmathurin:
One sample app updated here: https://github.com/wmathurin/MyUserPicReactNative/pull/2

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -312,40 +305,55 @@ public void authenticatedRestClient(RestClient client) {
...
-                        	if (callbackContext != null) {
-                            	callbackContext.error(exception.getMessage());
-                        	}
-                        }
-                    });
-	            }
+						@Override
+						public void onError(Exception exception) {
+							if (callbackContext != null) {
+								callbackContext.error(exception.getMessage());
```
bhariharan: Weird spacing between ```callbackContext``` and ```.error...```.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/SalesforceNetReactBridge.java
bhariharan: Does this tie the bridge to ```SalesforceReactActivity```, or was that always the case anyway?

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java
```diff
@@ -26,39 +26,46 @@
...
     protected void onCreate(Bundle savedInstanceState) {
+        Log.i(TAG, "onCreate called");
         super.onCreate(savedInstanceState);
 
+        // Get clientManager
+        clientManager = buildClientManager();
+
         // Gets an instance of the passcode manager.
         passcodeManager = SalesforceSDKManager.getInstance().getPasscodeManager();
-        userSwitchReceiver = new ActivityUserSwitchReceiver();
```
bhariharan: What happened to the user switcher receiver?

wmathurin: There is a story for providing a sf account manager bridge for react. So right now, there is no way of doing it in a react native application. Until then, I don't think we need the user switcher in SalesforceReactActivity.

bhariharan: Gotcha. Maybe we should add a ```TODO:``` somewhere in this class, so we don't forget to add it back.

bhariharan:
Looks good to me otherwise.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java
wmathurin: Done here https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1095/commits/9e7d69000cb272c94f883d966054fb8e6453254a

PR #1098, opened 2016-04-01
Supporting automatic login at startup (in addition to deferred login)

wmathurin:


libs/SalesforceReact/res/values/strings.xml
```diff
@@ -2,4 +2,5 @@
 <resources>
     <string name="app_name">SalesforceReact</string>
     <string name="app_package">com.salesforce.androidsdk.reactnative</string>
+    <string name="sf__should_authenticate_but_is_offline">Should authenticate but is offline - cannot proceed</string>
```
bhariharan: This should be in ```sf_strings.xml```, with the other standard error message strings we have and other ```sf__``` prefixed strings.

bhariharan:
Looks fine otherwise.

wmathurin:
@bhariharan Done here https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1098/commits/6540188931b0140d4a17b829e48812c2d4774369

PR #1109, opened 2016-04-27
Fixing 404 happening in tests

wmathurin:
It looks like missing the whitelist plugin and having an invalid refresh token caused the 404.
Adding the whitelist plugin, and still having an invalid refresh token gives a 401 instead.
Removing the whitelist plugin and having a valid refresh token works fine.


bhariharan:
👍 

Gonna merge this in, so that I can rebase my open pull request and verify test runs before merging it.

PR #1110, opened 2016-04-27
Adding 'cleanReSyncGhosts' ReactNative API

bhariharan:


libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/ReactBridgeHelper.java
```diff
@@ -68,6 +68,9 @@ public static void invokeSuccess(Callback successCallback, int value) {
         successCallback.invoke("" + value);
     }
 
+    public static void invokeSuccess(Callback successCallback) {
+        successCallback.invoke();
```
wmathurin: What's the benefit of calling:
```java
ReactBridgeHelper.invokeSuccess(successCallback)
```
Instead of
```java
successCallback.invoke()
```

bhariharan: No benefit. Was following the pattern with copy-paste. ;-) I'll switch it out.

bhariharan:
@wmathurin Reverted that.

wmathurin:
👍

PR #1114, opened 2016-05-04
Moving to SQLCipher 3.4

wmathurin:
Because the SQLCipher distributed on maven central has FTS/JSON1 enabled we no longer need to build the .so ourselves and ship them with the Mobile SDK


wmathurin:
Ran all native and hybrid tests successfully

bhariharan:
Nice! So, no fork anymore. 👍

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/AbstractSmartStoreTest.java
```diff
@@ -86,6 +86,8 @@ public void testCompileOptions() {
 
 		assertTrue("ENABLE_FTS4 flag not found in compile options", compileOptions.contains("ENABLE_FTS4"));
 		assertTrue("ENABLE_FTS3_PARENTHESIS flag not found in compile options", compileOptions.contains("ENABLE_FTS3_PARENTHESIS"));
+		assertTrue("ENABLE_FTS5 flag not found in compile options", compileOptions.contains("ENABLE_FTS5"));
+		assertTrue("ENABLE_JSON1 flag not found in compile options", compileOptions.contains("ENABLE_JSON1"));
 	}
```
wmathurin: Here is the test that checks that the sqlcipher in use was compiled with the right flags

bhariharan:
👍

kchitalia:
Nice!!!

PR #1117, opened 2016-05-08
Removing unnecessary PRNG fix

bhariharan:
This fix is only applicable to ```API 17```, and not to ```< API 16``` or ```> API 18```. Since our minimum is now ```API 19```, we don't need this anymore.


kchitalia:
:+1:

PR #1120, opened 2016-05-10
fix the test timeout issue using intel emulator accelerator 

helenren:


kchitalia:
Looks like the build failed. I would suggest try this out on your fork first, that way we don't pollute the public fork :)

.travis.yml
```diff
@@ -11,6 +11,7 @@ android:
       - build-tools-23.0.1
       - android-23
       - extra-google-m2repository
+      - sys-img-x86-android-10
```
bhariharan: We don't support ```< android-19``` anyway, so this won't work.

PR #1122, opened 2016-05-11
JSON1 index spec for Android

wmathurin:


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBHelper.java
```diff
@@ -234,9 +252,40 @@ public Cursor countQuery(SQLiteDatabase db, String table, String whereClause, St
 	 */
 	public Cursor limitRawQuery(SQLiteDatabase db, String sql, String limit, String... whereArgs) {
 		String limitSql = String.format(LIMIT_SELECT, sql, limit);
+		if (captureExplainQueryPlan) {
```
wmathurin: Added the ability to turn on/off explain query plan capture.
Used by tests (to validate that the right db indexes are being used when running certain queries) but could also be used when debugging perf issues.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBHelper.java
```diff
@@ -460,6 +509,23 @@ public String getSoupTableName(SQLiteDatabase db, String soupName) {
        return soupTableName;
    }
 
+	/**
+	 * If turned on, explain query plan is run before executing a query and stored in lastExplainQueryPlan
+	 * and also get logged
+	 * @param captureExplainQueryPlan true to turn capture on and false to turn off
+     */
+	public void setCaptureExplainQueryPlan(boolean captureExplainQueryPlan) {
```
wmathurin: Those two methods are also exposed on SmartStore.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartSqlHelper.java
```diff
@@ -120,16 +120,24 @@ else if (path.equals(SmartStore.SOUP_LAST_MODIFIED_DATE)) {
 				// {soupName:path}
 				else {
 					String columnName = getColumnNameForPathForSmartSql(db, soupName, path, position);
-					matcher.appendReplacement(sql, columnName);
+					matcher.appendReplacement(sql, columnName.replace("$", "\\$") /* treat any $ as litteral */);
```
wmathurin: When doing select {soup:path} on a json1 indexed path, you will get sql that looks like:
select json_extract(soup, '$.x.y.z')
We don't want appendReplacement to think $ is referring to a previous match

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -294,10 +311,17 @@ protected void registerSoupUsingTableName(String soupName, IndexSpec[] indexSpec
 
         int i = 0;
         for (IndexSpec indexSpec : indexSpecs) {
-            // for create table
+            // Column name or expression the db index is on
             String columnName = soupTableName + "_" + i;
-            String columnType = indexSpec.type.getColumnType();
-            createTableStmt.append(", ").append(columnName).append(" ").append(columnType);
+            if (TypeGroup.value_indexed_with_json_extract.isMember(indexSpec.type)) {
+                columnName = "json_extract(" + SOUP_COL + ", '$." + indexSpec.path + "')";
```
wmathurin: That's the key of it all, instead of the column name being an actual column, it's an expression.
So when creating the index / selecting on / selecting by that field, we want to use the expression instead of a column name.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -667,42 +694,25 @@ private JSONArray getDataFromRow(Cursor cursor) throws JSONException {
...
-    		    			row.put(d);
-    		    			// Note: we could end up returning an integer for a string column if you have a string value that contains just an integer
-    	    			}
-    		    		// It must be holding a string then
-    	    			catch (NumberFormatException ne) {
-    		    			row.put(raw);
-    	    			}
-    	    		}
-    			}
+            int valueType = cursor.getType(i);
```
wmathurin: This is a change we could have made a long while back - getType was not available before API 11. See TODO on line 681 in the old code.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -1230,7 +1245,33 @@ public String getColumnType() {
             return columnType;
         }
     }
-    
+
+    /**
+      * Enum for type groups
+      */
+    public enum TypeGroup {
```
wmathurin: In case we add more types in the future, this groupings should limit code change above.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/EncryptedSmartStoreTest.java
```diff
@@ -1,39 +0,0 @@
-/*
```
wmathurin: There is no real value testing with and without encryption. 
At the very beginning, non-encrypted db was done with plain SQLite and encrypted with SQLCipher but it is no longer the case. So getting rid of EncryptedSmartStoreTest and PlainSmartStoreTest and renaming AbstractSmartStoreTest to SmartStoreTest.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreAlterTest.java
```diff
@@ -148,47 +147,56 @@ public void testAlterSoupTypeChangeStringToInteger() throws JSONException {
...
-        JSONObject soupElt2 = new JSONObject();
-        soupElt2.put(CITY, PARIS);
-        soupElt2.put(COUNTRY, FRANCE);
-
-        long elt1Id = idOf(store.create(TEST_SOUP, soupElt1));
-        long elt2Id = idOf(store.create(TEST_SOUP, soupElt2));
-
-        // Checking db
-        checkDb(new long[]{elt1Id, elt2Id}, indexSpecs[0].type, indexSpecs[1].type);
+        tryAlterSoupTypeChange(SmartStore.Type.string, SmartStore.Type.full_text);
```
wmathurin: Trying out all the permutations to/from string/json1 and full_text

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreLoadTest.java
```diff
@@ -91,237 +83,180 @@ protected void tearDown() throws Exception {
...
-        for (int k=1; k<MAX_NUMBER_ENTRIES; k*=2) {        
-        	upsertManyEntries(k);
-        }
+
+    //
+    // Tests
+    //
+
+    public void testUpsertQuery1StringIndex1field20characters() throws JSONException {
+        tryUpsertQuery(Type.string, NUMBER_ENTRIES, 1, 20, 1);
```
wmathurin: Rewrote most of the tests here. Now you can specify number of indexes / type of index / number of rows / number of fields per soup element / number of characters per field. And timings for upserts and various queries (exact/like/all) will be logged. We should probably run various permutations on actual devices and collect the number.
Locally, I observed json1 and string to be comparable at query time, but json1 to be faster for inserts (as expected).

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreLoadTest.java
```diff
@@ -91,237 +83,180 @@ protected void tearDown() throws Exception {
...
+        queryEntries(QuerySpec.buildExactQuerySpec(TEST_SOUP, "k_0", "missing", null, null, 1));
+    }
+
+
+    private void queryEntries(QuerySpec querySpec) throws JSONException {
+        List<Long> times = new ArrayList<Long>();
+        int countMatches = 0;
+        boolean hasMore = true;
+        for (int pageIndex = 0; hasMore; pageIndex++) {
+            long start = System.nanoTime();
```
wmathurin: We used to use currentTimeMillis, now using nanoTime instead

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTestCase.java
```diff
@@ -86,7 +93,72 @@ protected boolean hasTable(String tableName) {
...
 		}
 	}
-	
+
+	/**
+	 * Helper method to check columns of table
+	 * @param tableName
+	 * @param expectedColumnNames
+     */
+	protected void checkColumns(String tableName, List<String> expectedColumnNames) {
```
wmathurin: A bunch of new helper methods:
- to check the columns in the database
- to check the database indexes
- to check the explain query plan of the last query run

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTest.java
```diff
@@ -44,12 +44,13 @@
 import org.json.JSONObject;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 
 /**
- * Abstract super class for plain and encrypted smart store tests
+ * Main test suite for SmartStore
```
wmathurin: All the relevant tests come now in two flavors: with regular indexes (string/floating etc) and with the new json1 index

wmathurin:
NB: A lot of new tests in SmartStoreTest / SmartStoreAlterTest and SmartStoreLoadTest

bhariharan:
Looks good to me. 👍

PR #1124, opened 2016-05-12
Extend the Travis CI Wait time to make more reliable to completed the test

helenren:
It also fix the build failure that caused by the wrong name for native sample app.


bhariharan:
👍

PR #1126, opened 2016-05-12
Fixt test failures

helenren:
For following tests:

testGetSDKInfo

samples.templateapp

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
@@ -203,7 +203,7 @@ public class PasscodeActivity extends Activity implements OnEditorActionListener
     public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
         Log.i("onEditorAction", "view=" + v + " actionId=" + actionId + " event=" + event);
         // Processing the editor action only on key up to avoid sending events like pass code manager unlock twice.
-        if (event != null && event.getAction() == KeyEvent.ACTION_UP) {
+        if ((event != null && event.getAction() == KeyEvent.ACTION_UP) || event == null) {
             String pc = entry.getText().toString();
             if (pc.length() > 0 && pc.length() < getMinPasscodeLength()) {
                 error.setText(getMinLengthInstructions(getMinPasscodeLength()));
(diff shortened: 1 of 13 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
```diff
@@ -203,7 +203,7 @@ public void enableLogout(boolean b) {
     public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
         Log.i("onEditorAction", "view=" + v + " actionId=" + actionId + " event=" + event);
         // Processing the editor action only on key up to avoid sending events like pass code manager unlock twice.
-        if (event != null && event.getAction() == KeyEvent.ACTION_UP) {
```
bhariharan: ```event``` should never be ```null``` here, and if it is, it should fall in the ```else``` block like it does now.

libs/test/SalesforceHybridTest/src/com/salesforce/androidsdk/phonegap/SDKInfoPluginTest.java
```diff
@@ -76,9 +79,17 @@ public void testGetSDKInfo() throws NameNotFoundException, JSONException {
 		assertTrue("Wrong bootconfig oauthScopes", sdkInfoOAuthScopes.contains("api"));
 		assertEquals("Wrong bootconfig oauthRedirectURI", bootconfig.getOauthRedirectURI(), sdkInfoBootConfig.getString("oauthRedirectURI"));
 		assertEquals("Wrong bootconfig remoteAccessConsumerKey", bootconfig.getRemoteAccessConsumerKey(), sdkInfoBootConfig.getString("remoteAccessConsumerKey"));
-		assertEquals("Wrong bootconfig androidPushNotificationClientId", bootconfig.getPushNotificationClientId(), sdkInfoBootConfig.getString("androidPushNotificationClientId"));
+        try {
+            sdkInfoBootConfig.getString("androidPushNotificationClientId");
```
bhariharan: Instead of doing this, let's add a dummy value for ```androidPushNotificationClientId``` for the test project's ```bootconfig.xml```, and revert this change. The test should pass then.

helenren: yes, if i add a dummy value there, the test can pass, while for that empty id check in the product code, want to use this verification to make sure it works as expected.

bhariharan: This test is verifying that the value is read if it exists. So adding a dummy push notification ID should be fine.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
helenren: so after calling setText, call doEditorAction with EditorInfo.IME_ACTION_GO, the event should be set? While with the debug, after  doEditorAction, the even is null, seems like not enter Enter manually after type a passcode there.

bhariharan: Hitting ```ENTER``` manually triggers 2 events - ```ACTION_DOWN``` and then ```ACTION_UP```. We need to simulate ```ACTION_UP``` in the test. Let's revert the changes to this file.

bhariharan:
👍

PR #1128, opened 2016-05-17
Small improvement to test

wmathurin:


libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreLoadTest.java
```diff
@@ -243,12 +243,12 @@ private void tryAlterSoup(Type indexType) throws JSONException {
         // Without indexing for new index specs
         alterSoup("Adding one index / no re-indexing", false, new IndexSpec[]{new IndexSpec("k_0", indexType), new IndexSpec("k_1", indexType)});
         alterSoup("Adding one index / dropping one index / no re-indexing", false, new IndexSpec[] {new IndexSpec("k_0", indexType), new IndexSpec("k_2", indexType)});
-        alterSoup("Dropping two indexes / no re-indexing", false, new IndexSpec[] {new IndexSpec("k_3", indexType)});
+        alterSoup("Dropping one index / no re-indexing", false, new IndexSpec[] {new IndexSpec("k_0", indexType)});
 
         // With indexing for new index specs
```
wmathurin: That way we are back to having just one index on k_0
So we can really compare the execution times with and without re-indexing

bhariharan:
👍

PR #1133, opened 2016-05-19
fix fail to submit passcode test issue

helenren:
Since IME_ACTION_GO action does not trigger a key DOWN and UP key event, so we need to trigger those specific key events by test.


wmathurin:
👍

PR #1137, opened 2016-05-23
Bugfix in test setup

bhariharan:


bhariharan:
@helenren Test suite is running now. All tests are passing except the files tests. I'm guessing that's because of the recent changes in file API paths. Should be simple to fix those assertions.

helenren:
👍

PR #1138, opened 2016-05-24
enable the build and test for Android version 19 to 23

helenren:
also fix the file request tests


bhariharan:
@helenren Would you mind opening a new pull request with clean commit history? Looks like there are 20 different merge commits here. Thanks!

PR #1141, opened 2016-05-24
Replace shelljs.sed with our own function to replace text in a file

kchitalia:


node/forcedroid.js
```diff
@@ -462,7 +462,17 @@ function fixSdkGradleFiles(config) {
...
-    shelljs.sed('-i', "compile project(':libs:", "compile project(':forcedroid:libs:", path.join(appFolderName, "forcedroid", "libs", lib, "build.gradle"));    
+    replaceInFile(path.join(appFolderName, "forcedroid", "libs", lib, "build.gradle"), "compile project(':libs:", "compile project(':forcedroid:libs:"); 
+}
+
+function replaceInFile(fileName, textInFile, replacementText) {
+    var contents = fs.readFileSync(fileName, 'utf8');
+    var lines = contents.split(/\r*\n/);
+    var result = lines.map(function (line) {
+      return line.replace(textInFile, replacementText);
+    }).join('\n');
```
bhariharan: Why're we adding a newline character?

kchitalia: Because we split the file contents on \r\n. If we don't join on \n the file contents will be on one line. The file becomes unreadable.

wmathurin: We are probably going to need that function in forceios too, should we move it to SalesforceMobileSDK-Shared/node/util.js (or another js in the same folder)?

kchitalia: That sounds like a good idea. I'll close this PR. Will send a new PR for the shared repo and one more here with updated submodule pointer to shared repo and code changes in forcedroid.js I'll move this function to node/util.js

PR #1143, opened 2016-05-25
Replace all occurences of sed with our own function

kchitalia:


node/forcedroid.js
```diff
@@ -280,15 +280,14 @@ function createNativeOrReactNativeApp(config) {
     if (config.usesmartstore) {
         console.log('Adding SmartStore/SmartSync support.');
         console.log('Extending SmartSyncSDKManager instead of SalesforceSDKManager.');
-        shelljs.sed('-i', /SalesforceSDKManager/g, 'SmartSyncSDKManager', appClassPath);
-        shelljs.sed('-i',
-            /com\.salesforce\.androidsdk\.app\.SmartSyncSDKManager/g,
+        miscUtils.replaceTextInFile(appClassPath, 'SalesforceSDKManager', 'SmartSyncSDKManager');
+        miscUtils.replaceTextInFile(appClassPath,
```
bhariharan: Does the function replace all occurrences? ```/g``` was for global replace.

wmathurin:
Have you run test_force.js (https://github.com/forcedotcom/SalesforceMobileSDK-Shared/blob/master/node/test_force.js) to make sure all the app generation works.
Also we should change test_force.js to no longer install shelljs 0.5.3.

node/forcedroid.js
kchitalia: It will replace all occurrences of the text in the file. The function splits the file into lines and applies the replacement text to all occurrences of the text to be replaced in that file.

bhariharan: Cool!

bhariharan:
👍

kchitalia:
@wmathurin I was able to successfully generate the native and hybrid apps using forcedroid on my local to test this. I will change test_force.js and run it to make sure we are covered.

wmathurin:
👍

kchitalia:
@wmathurin @bhariharan All feedback has been incorporated.

bhariharan:
👍

PR #1144, opened 2016-05-25
update the call to the deprecated method sendJavaScript 

helenren:
Also disable the run for api 23 for the issue on travis ci. Will add it back once the issue is fixed or find out the workaround.


wmathurin:
👍

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/util/test/JSTestCase.java
```diff
@@ -143,4 +145,19 @@ protected void runTest(String testName)  {
         assertNotNull("No test result", result);
         assertTrue(result.testName + " " + result.message, result.success);
     }
+
+	@TargetApi(Build.VERSION_CODES.KITKAT)
```
bhariharan: Remove ```@TargetApi``` annotation here. Our minimum is ```KitKat```, so we don't need this annotation.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/util/test/JSTestCase.java
```diff
@@ -143,4 +145,19 @@ protected void runTest(String testName)  {
...
+	private void sendJavascript(final CordovaWebView webView, final String javascript) {
+
+		webView.getView().post(new Runnable() {
+			@Override
+			public void run() {
+				if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
bhariharan: Remove this check too.

PR #1145, opened 2016-05-25
Remove from soup by query

wmathurin:
With native/hybrid/react native support
Also added other smartstore tests recently added on iOS


external/shared
```diff
@@ -1 +1 @@
-Subproject commit 14e5606d366e4b9711d4085b18ac0fa2e5a65ce0
+Subproject commit 03cc709564f6052f3bb0ff44c0350494f1bfab2f
```
bhariharan: Might want to revert the change to ```shared``` commit here. This one is older than the latest, which is already checked in to ```unstable```.

bhariharan:
Looks good otherwise.

PR #1148, opened 2016-05-27
fix couple things

helenren:
1. handle the 2nd lock screen on device before running test
2. keep the order as the put order for makeFileShare instead of the random order. The order for version 19 is different to the order for version 21 and 22 when using hashmap instead of linkedhashmap.


bhariharan:
👍

PR #1153, opened 2016-06-01
DDL operations are transactional for SQLite

wmathurin:
Using transactions where we were not (register table, alter soup)


bhariharan:
👍

PR #1154, opened 2016-06-03
Update syncManager and cacheManager to use new delete API

helenren:


bhariharan:
👍

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -335,13 +335,11 @@ public void cleanResyncGhosts(long syncId) throws JSONException {
 
         // Deletes extra IDs from SmartStore.
         if (localIds.size() > 0) {
-            final Long[] soupEntryIds = new Long[localIds.size()];
-            int index = 0;
             for (final String localId : localIds) {
-                soupEntryIds[index] = smartStore.lookupSoupEntryId(soupName, idFieldName, localId);
```
wmathurin: We used to do n lookups and 1 delete, and now we do n deletes.
We could bring it down to a single delete by using a smart query for the removeByQuery:
```java
// NB: I have not tested the code so typos possible
String smartSql = String.format("SELECT {%s:%s} FROM {%s} WHERE {%s:%s} IN (%s}", soupName, SmartStore.SOUP_ENTRY_ID, soupName, soupName, idFieldName, "'" + TextUtils.join("', '", localIds) + "'");
smartStore.deleteByQuery(soupName, querySpec);
```

wmathurin:
👍

PR #1155, opened 2016-06-03
Full-text indexes now using FTS5

wmathurin:
NB: In the future we should provide a way for developers to use advanced FTS features like selecting the tokenizer (https://www.sqlite.org/fts5.html#section_4_3) or sorting using bm25 (https://www.sqlite.org/fts5.html#section_5_1_1).

Added a way to select FTS4 or FTS5 -- not meant for app but used by tests to make sure no behavior changed.

For a soup created before 4.2, one simply needs to do an alterSoup passing in the original index specs, and the FTS virtual table will be recreated using FTS5 (there is a test for that).

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/AlterSoupLongOperation.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/AlterSoupLongOperation.java
@@ -495,9 +495,9 @@ public class AlterSoupLongOperation extends LongOperation {
 			List<String> oldColumnsFts = new ArrayList<String>();
 			List<String> newColumnsFts = new ArrayList<String>();
 
-			// Adding docid column
+			// Adding rowid column
 			oldColumnsFts.add(SmartStore.ID_COL);
-			newColumnsFts.add(SmartStore.DOCID_COL);
+			newColumnsFts.add(SmartStore.ROWID_COL);
 
 			// Adding indexed path columns that we are keeping
 			for (String keptPath : keptPaths) {
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/QuerySpec.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/QuerySpec.java
@@ -36,6 +36,8 @@ import org.json.JSONObject;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Simple class to represent a query spec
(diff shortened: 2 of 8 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/AlterSoupLongOperation.java
```diff
@@ -495,9 +495,9 @@ private void copyOldData() {
 			List<String> oldColumnsFts = new ArrayList<String>();
 			List<String> newColumnsFts = new ArrayList<String>();
 
-			// Adding docid column
```
wmathurin: With FTS5, they don't alias the id column as docid anymore.
The good news is that using rowid works against fts4 and fts5 tables.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/QuerySpec.java
```diff
@@ -346,12 +348,9 @@ private String computeFromClause() {
     private String computeWhereClause() {
         if (path == null && queryType != QueryType.match /* null path allowed for fts match query */) return "";
 
-        String field;
+        String field = null;
 
-        if (queryType == QueryType.match && path == null) {
```
wmathurin: fts5 doesn't allow WHERE column MATCH 'value' - only allows WHERE table MATCH 'column:value'

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -1212,6 +1215,22 @@ private String buildInStatement(String col, String inPredicate) {
...
+	public FtsExtension getFtsExtension() {
+		return ftsExtension;
+	}
+
+	/**
+	 * Sets the ftsX to be used when creating the virtual table to support full_text queries
+	 * NB: only used in tests
+	 * @param ftsExtension
+     */
+	public void setFtsExtension(FtsExtension ftsExtension) {
```
wmathurin: Store wide - meant for tests only

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreAlterTest.java
```diff
@@ -230,6 +230,53 @@ public void testAlterSoupWithFullTextIndexesToGetIndexesOnCreatedAndLastModified
...
 
+    /**
+     * Create soup with fts4 virtual table
+     * Call alterSoup passing in same index specs
+     * Make sure virtual table is recreated with fts5
+     * That way soup created before 4.2 (using fts4 virutal table) can be migrated to fts5 by calling alterSoup
+     * @throws JSONException
+     */
+    public void testAlterSoupwithFullTextIndexesFromFts4ToFts5() throws JSONException {
+        IndexSpec[] indexSpecs = new IndexSpec[] {new IndexSpec(CITY, SmartStore.Type.full_text), new IndexSpec(COUNTRY, SmartStore.Type.full_text)};
```
wmathurin: This test demonstrates how one can use alterSoup an existing soup from using fts4 to using fts5

bhariharan:
Looks good! 👍

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
kchitalia: Can this be moved to test classes as its only for tests?

wmathurin: How would you do that, it's a setter?

kchitalia: Sorry, my bad.

PR #1157, opened 2016-06-06
Add test_force.js to travis file

kchitalia:
I will add hybrid apps later. Right now when the test script creates a hybrid app, it fails on ./install.sh as its trying to do a xcdoebuild. Will have to see how we can build only the android sources and skip the iOS ones.


bhariharan:
👍

wmathurin:
👍

PR #1159, opened 2016-06-09
Fix for JSONException at SmartSyncManager:isNewerThanServer

wmathurin:
- we shouldn't go to the server for locally created records
- but if the app code unchecks the locally_created flag (e.g. after doing an local update or delete), the smart sync code should not fail the sync up
- added tests

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -439,11 +439,11 @@ public class SyncManager {
         boolean locallyUpdated = record.getBoolean(LOCALLY_UPDATED);
 
         Action action = null;
-        if (record.getBoolean(LOCALLY_DELETED))
+        if (locallyDeleted)
             action = Action.delete;
-        else if (record.getBoolean(LOCALLY_CREATED))
+        else if (locallyCreated)
             action = Action.create;
-        else if (record.getBoolean(LOCALLY_UPDATED))
+        else if (locallyUpdated)
             action = Action.update;
 
         if (action == null) {
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -425,7 +425,7 @@ public class SyncManager {
 
             return (serverLastModifiedDate <= lastModifiedDate);
         } catch (Exception e) {
-            Log.e("SmartSyncManager:isNewerThanServer", "Couldn't figure out last modified date", e);
+            Log.e("SmartSyncMgr:isNewerThanServer", "Couldn't figure out last modified date", e);
             throw new SmartSyncException(e);
         }
     }
@@ -434,6 +434,10 @@ public class SyncManager {
                                     JSONObject record, MergeMode mergeMode) throws JSONException, IOException {
 
         // Do we need to do a create, update or delete
+        boolean locallyDeleted = record.getBoolean(LOCALLY_DELETED);
+        boolean locallyCreated = record.getBoolean(LOCALLY_CREATED);
+        boolean locallyUpdated = record.getBoolean(LOCALLY_UPDATED);
+
         Action action = null;
         if (record.getBoolean(LOCALLY_DELETED))
             action = Action.delete;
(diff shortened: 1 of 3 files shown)


libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTest.java
```diff
@@ -282,6 +293,31 @@ public void testSyncUpWithLocallyDeletedRecords() throws Exception {
...
 
+    /**
+     * Create accounts locally, delete them locally, sync up with merge mode LEAVE_IF_CHANGED, check smartstore
+     *
+     * Ideally an application that deletes locally created records should simply remove them from the smartstore
+     * But if records are kept in the smartstore and are flagged as created and deleted (or just deleted), then
+     * sync up should not throw any error and the records should end up being removed from the smartstore
+     *
+     */
+    public void testSyncUpWithLocallyCreatedAndDeletedRecords() throws Exception {
```
wmathurin: If you revert the SyncUpTarget.java change above, this test will fail with the JSONException reported in the bug.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -460,7 +464,7 @@ else if (record.getBoolean(LOCALLY_UPDATED))
          * circumstances, we will do nothing and return here.
          */
         if (mergeMode == MergeMode.LEAVE_IF_CHANGED &&
-        		(action == Action.update || action == Action.delete) &&
+        		!locallyCreated &&
```
bhariharan: Is ```!locallyCreated``` the same as either of the other two actions? Could all 3 be ```false```?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -491,7 +495,9 @@ else if (record.getBoolean(LOCALLY_UPDATED))
                 }
                 break;
             case delete:
-                statusCode = target.deleteOnServer(this, objectType, objectId);
+                statusCode = (locallyCreated
```
bhariharan: Nice catch!

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
```diff
@@ -144,18 +144,22 @@ public int updateOnServer(SyncManager syncManager, String objectType, String obj
...
-    public String fetchLastModifiedDate(SyncManager syncManager, String objectType, String objectId) throws JSONException, IOException {
-        final String query = SOQLBuilder.getInstanceWithFields(getModificationDateFieldName())
-                .from(objectType)
-                .where(getIdFieldName() + " = '" + objectId + "'")
-                .build();
-
-        RestResponse lastModResponse = syncManager.sendSyncWithSmartSyncUserAgent(RestRequest.getRequestForQuery(syncManager.apiVersion, query));
-        JSONArray records = lastModResponse.asJSONObject().optJSONArray(Constants.RECORDS);
-        return records != null && records.length() > 0 ? records.optJSONObject(0).optString(getModificationDateFieldName()) : null;
+    public String fetchLastModifiedDate(SyncManager syncManager, String objectType, String objectId) {
```
bhariharan: No code changes here? Just wrapping in ```try-catch```?

bhariharan:
Looks good.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: if it is marked as created and deleted then action will be deleted, but we really shouldn't go to the server.
Ideally the app should just delete the record if it was created locally instead of waiting until sync up to do it.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncUpTarget.java
wmathurin: Try-catch
On error we return null which the caller expects - if we throw an exception, the caller will give up on the sync up

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: Gotcha.

bhariharan:
👍

PR #1161, opened 2016-06-13
Fix a build break

helenren:
There is a change for create Account name, that we have a generic helper to create record name based on the passed in record type.  While the recent changes checked in not align to that change.

Also enable hybrid tests


bhariharan:
Looks good. 👍

wmathurin:
👍

bhariharan:
👍

PR #1164, opened 2016-06-16
Move to latest react native (0.27.2)

wmathurin:


build.gradle
```diff
@@ -12,6 +12,11 @@ allprojects {
   group = 'com.salesforce.androidsdk'
   
   repositories {
+    maven {
+      // All of React Native (JS, Android binaries) is installed from npm
```
wmathurin: Starting in 0.21, react native is no longer on maven central.
Instead we need to get the npm package and point to it from the build.gradle file.

install.sh
```diff
@@ -5,3 +5,6 @@
 git submodule init
 git submodule sync
 git submodule update
+
+# for react-native
+npm install react-native@0.27.2
```
wmathurin: install.sh now needs to bring in the npm package for react native otherwise SalesforceReact won't compile

install.vbs
```diff
@@ -70,7 +70,16 @@ End If
 ' Copy the hard files to where their symlinks would be.
 Call CopySymlinkFiles()
 
-WScript.Echo vbCrLf & "Successfully initialized submodules and fixed symlinks."
```
wmathurin: That's the code for windows developer (but it hasn't been tested - I don't have a windows box).

reactnative/ReactNativeTemplateApp/js/index.android.js
```diff
@@ -26,16 +26,17 @@
 
 'use strict';
 
-var React = require('react-native');
-var {
+import React from 'react';
+import {
```
wmathurin: Same changes we did for index.ios.js

reactnative/ReactNativeTemplateApp/package.json
```diff
@@ -6,6 +6,6 @@
     "start": "node_modules/react-native/packager/packager.sh --root js/"
   },
   "dependencies": {
-    "react-native": "^0.20.0"
+    "react-native": "^0.27.2"
```
wmathurin: We'll have to keep that version number lined up with the one we npm install in install.sh, otherwise the SDK repo and generated app will work against different versions of react native.

bhariharan:
👍

PR #1166, opened 2016-06-16
Latest shared + small fix

wmathurin:


install.sh
```diff
@@ -7,4 +7,4 @@ git submodule sync
 git submodule update
 
 # for react-native
-npm install react-native@0.27.2
+npm install react-native@0.27.2 react@15.1.0 --silent
```
bhariharan: Do we need ```--silent```? We may not know if something goes wrong with no console output.

bhariharan:
👍

install.sh
wmathurin: It doesn't silence errors, just reduces the ouput.

PR #1168, opened 2016-06-23
 Passcode can no longer be guessed as many times as desired

wmathurin:
* Keeping failed passcode attempts in pref file
* Not showing "forgot passcode" in passcode create/confirm screens


bhariharan:
👍

PR #1170, opened 2016-06-27
Changes to support 64 bits devices with react native

wmathurin:
Based on workaround offered here: https://corbt.com/posts/2015/09/18/mixing-32-and-64bit-dependencies-in-android.html

### DIFF
--- a/reactnative/ReactNativeTemplateApp/build.gradle
+++ b/reactnative/ReactNativeTemplateApp/build.gradle
@@ -28,11 +28,19 @@ android {
         }
     }
 
-  packagingOptions {
-    exclude 'META-INF/LICENSE'
-    exclude 'META-INF/LICENSE.txt'
-    exclude 'META-INF/DEPENDENCIES'
-    exclude 'META-INF/NOTICE'
-  }
+    // To avoid could find DSO to load: libreactnativejni.so
+    // See https://corbt.com/posts/2015/09/18/mixing-32-and-64bit-dependencies-in-android.html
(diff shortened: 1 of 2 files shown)


reactnative/ReactNativeTemplateApp/build.gradle
```diff
@@ -28,11 +28,19 @@ android {
...
 
-  packagingOptions {
-    exclude 'META-INF/LICENSE'
-    exclude 'META-INF/LICENSE.txt'
-    exclude 'META-INF/DEPENDENCIES'
-    exclude 'META-INF/NOTICE'
-  }
+    // To avoid could find DSO to load: libreactnativejni.so
+    // See https://corbt.com/posts/2015/09/18/mixing-32-and-64bit-dependencies-in-android.html
+    defaultConfig {
```
bhariharan: Should this go in ```build.gradle``` for ```SalesforceReact```, so all apps get it?

wmathurin: We could, but that way developer can take it out if for whatever reason they don't want it.

bhariharan: Sounds good.

bhariharan:
👍

reactnative/ReactNativeTemplateApp/build.gradle
bhariharan: More I think about it, it makes sense at the app level. It's not something a library would do.

PR #1171, opened 2016-06-28
install the needed react-native for recent change

helenren:


node/forcedroid.js
```diff
@@ -38,7 +38,8 @@ var path = require('path');
 var outputColors = require('../external/shared/node/outputColors');
 var commandLineUtils = require('../external/shared/node/commandLineUtils');
 var shelljs = require('shelljs') ;
-var fs = require('fs'); 
+var fs = require('fs');
```
bhariharan: We chose ```fs``` because it comes preinstalled with ```node```. Is that the case for ```fs-extra``` too? I fear that we'll run into the same issues as ```shelljs``` by introducing an extra dependency.

helenren: so, you prefer to use fs only, even the code will not look that clean, right?

bhariharan: @helenren Yes, I would prefer to use ```fs``` only, for now. In ```5.0```, I have a plan to get rid of the post install steps and some of these copy steps.

bhariharan:
Looks good. 👍

node/forcedroid.js
```diff
@@ -392,17 +393,35 @@ function getTemplateSourceFilePaths(config) {
...
+            }
+        }
+        else {
+            console.log(outputColors.cyan + 'INFO:' + outputColors.reset + ' ' + srcDirRelative + ' already exists.  Skipping copy.');
+        }
+    }
+    else {
+        destDir = path.join(targetDir, path.basename(packageSdkRootDir), path.dirname(srcDirRelative));
+        if (!fse.existsSync(path.join(destDir, path.basename(srcDirRelative)))) {
+            shelljs.mkdir('-p', destDir);
```
wmathurin: Why are we copying with fse in one case and shelljs in the other?

helenren: shelljs of 0.7.0 version has issue for folder copy, while fse copy file hit permission for copying file, that's why we cannot use shelljs or fse for both file and folder copy

.travis.yml
```diff
@@ -36,6 +36,7 @@ before_install:
     - nvm install 0.12.0
     - nvm use 0.12.0
     - npm install shelljs@0.7.0
+    - npm install react-native@0.27.2 react@15.1.0 --silent
```
wmathurin: Looks good.

PR #1175, opened 2016-07-01
disable the test for template app for the weird failure that only happened on travis ci and for version 22

helenren:


wmathurin:
👍

PR #1176, opened 2016-07-01
Magic link

huminzhi:


bhariharan:
Still has merge conflicts. Let's do an offline review.

PR #1180, opened 2016-07-07
Support magic link login

helenren:


bhariharan:
@helenren Closing this for now, so we don't accidentally merge this before ```4.2```. Let's reopen next week after the release.

PR #1189, opened 2016-07-11
Upgrade to SQLCipher 3.5.2 for Android N Support

bhariharan:


wmathurin:
3.5.2 came out a few days ago. Should we use that instead?

bhariharan:
@wmathurin Yeah - makes sense. I'll bump it up and test with it.

bhariharan:
I tested running ```SmartSyncExplorer``` with the new version of ```SQLCipher```. I also tested running the upgrade scenario. Everything seems to work great. I think we're good to go on merging.

kchitalia:
Hybrid SmartSyncExplorer looks good too?

bhariharan:
Yup, looks good to me.

PR #1194, opened 2016-07-14
Fixed tests, tab spaces, and updated shared submodule ptr

eric-romero:

PR #1195, opened 2016-07-15
External storage post merge work (incomplete)

wmathurin:
Code to prevent soup from having json1 index and external storage (and new test for it)
1) New tests for externally stored soup with full text
2) New test for delete by query against externally stored soup
3) New tests for alter soup with storage change and column type changes
NB: There doesn't seem to be an API to alter soup with soup spec on Android - so the tests are incomplete (have TODOs) and fail

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/Encryptor.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/Encryptor.java
@@ -120,7 +120,7 @@ public class Encryptor {
      * @return decrypted data
      */
     public static String decrypt(String data, String key) {
-        if (key == null || data == null) {
+        if (key == null || key.isEmpty() || data == null) {
             return data;
         }
         try {
@@ -145,7 +145,7 @@ public class Encryptor {
      * @return base64, aes256 encrypted data
      */
     public static String encrypt(String data, String key) {
-        if (key == null || data == null) {
+        if (key == null || key.isEmpty() || data == null) {
             return data;
         }
         try {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/Encryptor.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/Encryptor.java
@@ -43,6 +43,7 @@ import javax.crypto.spec.SecretKeySpec;
 import android.app.Service;
 import android.app.admin.DevicePolicyManager;
 import android.content.Context;
+import android.text.TextUtils;
 import android.util.Base64;
 import android.util.Log;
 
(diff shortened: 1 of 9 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/Encryptor.java
```diff
@@ -120,7 +120,7 @@ public static boolean isFileSystemEncrypted() {
      * @return decrypted data
      */
     public static String decrypt(String data, String key) {
-        if (key == null || data == null) {
+        if (key == null || key.isEmpty() || data == null) {
```
bhariharan: Why not just use ```TextUtils.isEmpty(key)```? Why're we doing this anyway?

wmathurin: Some tests don't encrypt smartstore (they pass empty string). That used to work fine, but the external storage code blows up when trying to encrypt or decrypt files if the key is "".

I think we should support no encryption. During development, it makes sense to disable decryption to be able to introspect the database with regular sqlite viewer.

bhariharan: Sounds good. Let's replace it with ```TextUtils.isEmpty(key)```.

bhariharan:
Looks good otherwise.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/Encryptor.java
wmathurin: Done here https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1195/commits/05eb1882950f114fb1308b6f0651254bedd95e37

bhariharan:
👍

wmathurin:
@bhariharan  I'm going to merge that.

There is only one thing missing to complete the post-merge work for Android: we are missing an alter soup method that takes a soup spec (like we have on iOS), therefore allowing a soup to be changed to/from using external storage.

@eric-romero is working on it (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1194#issuecomment-233014889).

I added tests for that already (TDD!) - they all fail obviously. 
They are in SmartStoreAlterTest.java.
```
testAlterSoupTypeChangeStringToJSON1ExternalToInternal
testAlterSoupTypeChangeJSON1ToStringInternalToExternal
testAlterSoupTypeChangeStringToFullTextExternalToInternal
testAlterSoupTypeChangeFullTextToStringInternalToExternal
testAlterSoupStringInternalToExternalToInternal
testAlterSoupFullTextInternalToExternalToInternal
```
They all call: 
```
tryAlterSoupChange(SmartStore.Type fromType, SmartStore.Type toType, boolean fromStorageInternal,boolean toStorageInternal, boolean toStorageInternal2)
```
which calls the existing alterSoup method passing the new index specs, it should instead call the new alterSoup method passing in a new soup spec along with the new index specs. 

Once the alterSoup method is added, we can modify the above test helper method (it has todo's in the right places) and then the six tests above should start passing.

The iOS implementation is complete and can also serve as a model (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/unstable/libs/SmartStore/SmartStoreTests/SFSmartStoreAlterTests.m#L358).

bhariharan:
Thanks @wmathurin! I'll look at those ```TODO:s``` and wrap this up.

bhariharan:
The outstanding items here are done in [this](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1197) pull request. All ```290``` tests passed successfully.

PR #1196, opened 2016-07-15
Populating application's dataDir in constructor of DBOpenHelper

eric-romero:
Also added a null check for removeExternalBlobsDirectory if in case it remains null


eric-romero:
This should fix issue for ktanna

bhariharan:
👍

PR #1197, opened 2016-07-18
Support for changing storage type for alterSoup

eric-romero:
Allowed AlterSoupLongOperation to support changing soup spec. This enables a store to be
converted from internal storage to external and vice versa. Storing the new/old soup specs
in the long operations table allows resuming of this kind of operation. If these entries are not found
in the db (in the case of client upgrade), they will be assumed to be internal (see SoupSpec.fromJSON)


bhariharan:
Looks good. 👍

kchitalia:
LGTM

PR #1202, opened 2016-07-22
fix the passcode ui test failure and extend the travis ci timeout for libs test

helenren:


.travis.yml
```diff
@@ -51,7 +51,7 @@ script:
    #run template app generation
    - travis_wait 30 set -o pipefail && node external/shared/node/test_force.js --os=android --test=native
 
-   - travis_wait 30 ./gradlew createDebugCoverageReport
+   - travis_wait 50 ./gradlew createDebugCoverageReport
```
helenren: since we added some new tests, especially the load test, the test need longer time to complete.

native/test/TemplateAppTest/src/com/salesforce/samples/templateapp/PasscodeActivityTest.java
```diff
@@ -443,9 +443,6 @@ private void doEditorAction(final int textViewId, final int actionCode) {
                     v.onEditorAction(actionCode);
                 }
             });
-            waitSome();
```
helenren: recently, IME_ACTION_GO action can trigger key_down and key_up event successfully, so we don't need to send enter key separately as before, otherwise will trigger it twice and cause to enter empty passcode, which cause test failed.

bhariharan:
👍

kchitalia:
I am assuming the IME_ACTION_GO behavior is same for different versions of Android.

helenren:
Yes it is, at least based on the try on different versions. @kchitalia Thanks.

kchitalia:
👍

PR #1203, opened 2016-07-26
add test org cleanup class 

helenren:
update the yml file to call it before all tests to make sure each test pass starts with a good status.


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright (c) 2011-2015, salesforce.com, inc.
```
bhariharan: Change to ```2016```.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
...
+import com.salesforce.androidsdk.rest.RestRequest.RestMethod;
+import com.salesforce.androidsdk.rest.RestResponse;
+
+import java.net.HttpURLConnection;
+import java.net.URI;
+/**
+ * Do org data cleanup
+ *
+ */
+public class ClearupOrgData extends InstrumentationTestCase {
```
bhariharan: Could you put this under the ```util``` subpackage and rename this class to ```TestDataCleanupHelper``` or something like that?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
...
+    private RestClient restClient;
+    private String authToken;
+    public static final String TEST_FIRST_NAME = "firstName";
+    public static final String TEST_LAST_NAME = "lastName";
+    public static final String TEST_DISPLAY_NAME = "displayName";
+    public static final String TEST_EMAIL = "test@email.com";
+    public static final String TEST_PHOTO_URL = "http://some.photo.url";
+    public static final String TEST_THUMBNAIL_URL = "http://some.thumbnail.url";
+    private final int MaxTransactCount = 100;
+    private final String EXPECTED_DELETE = "ExpectedDeletes";
```
bhariharan: ```static```.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
...
+     * Testing a customer apex api call.
+     * @throws Exception
+     */
+    public void testcleanOrgData() throws Exception {
+        int stop = 0;
+        int expectedCount = 0;
+        String requestFormat = "/services/apexrest/Cleanup?start=%d&stop=%d";
+        do {
+            int start = stop;
+            stop = stop + MaxTransactCount;
```
bhariharan: Use lowercase ```m``` in ```maxTransactCount```.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
...
+
+            RestResponse response = restClient.sendSync(new RestRequest(RestMethod.DELETE, String.format(requestFormat, start, stop), null));
+            checkResponse(response, HttpURLConnection.HTTP_OK, false);
+            expectedCount = response.asJSONObject().getInt(EXPECTED_DELETE);
+        } while (expectedCount >= MaxTransactCount);
```
bhariharan: Can we batch them up and make a single ```DELETE``` call, instead of looping and making multiple API calls?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
...
+     * Helper method to validate responses
+     * @param response
+     * @param expectedStatusCode
+     */
+    private void checkResponse(RestResponse response, int expectedStatusCode, boolean isJsonArray) {
+        // Check status code
+        assertEquals(expectedStatusCode  + " response expected", expectedStatusCode, response.getStatusCode());
+
+        // Try to parse as json
+        try {
```
bhariharan: Why're we parsing the response here? We don't do anything with it.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
```diff
@@ -0,0 +1,124 @@
...
+            if (isJsonArray) {
+                response.asJSONArray();
+            }
+            else {
+                response.asJSONObject();
+            }
+        }
+        catch (Exception e) {
+            fail("Failed to parse response body");
+            e.printStackTrace();
```
bhariharan: Remove this line.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
helenren: Which one do you mean, the util inside Salesforcesdktest or com.salesorce.androidsdk.util? Thanks

bhariharan: Inside ```SalesforceSDKTest```.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
helenren: The apex class needs to know the number of records needs to be deleted next time from the last api call, via batch call, how to figure out how many records need to be deleted? Could you show me more details or Is there any example that I can take a look?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/TestDataCleanupHelper.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) 2011-2016, salesforce.com, inc.
```
bhariharan: I meant change to ```2016-present```. It was just introduced.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/TestDataCleanupHelper.java
```diff
@@ -0,0 +1,96 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk;
```
bhariharan: Package should be ```com.salesforce.androidsdk.util```. Look at ```LogUtilTest``` for an example.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ClearupOrgData.java
helenren: hmm, the one over there is for testing logutil  specifically, that's why I am hesitate to put it there. I will add a helpers folder and put it there.

bhariharan: @helenren No - that's a generic ```util``` folder. It can be put there.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/helpers/TestDataCleanupHelper.java
```diff
@@ -0,0 +1,96 @@
...
+ * Do org data cleanup
+ *
+ */
+public class TestDataCleanupHelper extends InstrumentationTestCase {
+
+    private ClientInfo clientInfo;
+    private HttpAccess httpAccess;
+    private RestClient restClient;
+    private String authToken;
+    public static final String TEST_FIRST_NAME = "firstName";
```
bhariharan: These can be ```private``` too.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/helpers/TestDataCleanupHelper.java
```diff
@@ -0,0 +1,96 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.helpers;
```
bhariharan: Let's move this to the `util` subpackage.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/util/TestDataCleanupHelper.java
```diff
@@ -0,0 +1,96 @@
...
+    private ClientInfo clientInfo;
+    private HttpAccess httpAccess;
+    private RestClient restClient;
+    private String authToken;
+    public static final String TEST_FIRST_NAME = "firstName";
+    public static final String TEST_LAST_NAME = "lastName";
+    public static final String TEST_DISPLAY_NAME = "displayName";
+    public static final String TEST_EMAIL = "test@email.com";
+    public static final String TEST_PHOTO_URL = "http://some.photo.url";
+    public static final String TEST_THUMBNAIL_URL = "http://some.thumbnail.url";
```
bhariharan: All of these can be ```private```, right?

helenren: yes, while would prefer to leave them public since plan to extend this method as a test base class, which will be the part of the cleanup story we plan to do. So..

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/util/TestDataCleanupHelper.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Cleanup all records that created by test account.
+     *
+     * @throws Exception
+     */
+    public void testCleanupOrgData() throws Exception {
+        int stop = 0;
+        int expectedCount = 0;
+        String requestFormat = "/services/apexrest/Cleanup?start=%d&stop=%d";
```
bhariharan: @huminzhi @helenren If we're triggering ```DELETE``` via API from the client, why would we use ```apex```, which has no visibility to a consumer of Mobile SDK, as opposed to using ```DELETE REST API``` directly here?

helenren: This cleanup class is not designed for sdk consumer, just for our internal test purpose, to make sure each test pass starting with a good/known status. @bhariharan.

bhariharan: I understand. Is there any reason we can't use ```DELETE REST API``` to achieve the same result?

helenren: We need this cleanup for all supported platforms. So, we want to share some work as much as possible.

helenren: And could you please point out the drawback on using apex? Don't have much experience with it on my side. Thanks

bhariharan: My main concern is that this `apex` class is a black box to external developers and it exists only on our org. An external developer running our tests against his/her own org won't have access to this and the tests will fail. I view this `apex` class as a tool for us to ensure our data is clean. We should certainly use it, but not from the unit tests themselves. We should use it as a script that we run periodically to clean the data, maybe from `Travis` or any other `CI` that we decide to use. I just had a discussion with @huminzhi about this offline too, and I think we agree that it shouldn't be driven from the tests themselves. We can discuss more in detail tomorrow.

huminzhi: sorry I missed the whole conversation here, I agree with @bhariharan , let's not use the apex here since it's a blackbox to our customers... 
@helenren can we just use restAPI to achieve what we want here? Say, we use API to get a list of accounts/contacts? owned by the user, then delete them through rest API also?

bhariharan:
Closing this based on offline discussion with @huminzhi and @helenren.

PR #1204, opened 2016-07-27
fix jwt flow bugs

huminzhi:


huminzhi:
@kchitalia @helenren @bhariharan This is to fix some issue discovered

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -289,6 +289,9 @@ public LoginOptions getLoginOptions(String jwt, String url) {
                 loginOptions = new LoginOptions(url, getPasscodeHash(), config.getOauthRedirectURI(),
                         config.getRemoteAccessConsumerKey(), config.getOauthScopes(), null, jwt);
             }
+        } else {
```
huminzhi: so this is to fix a scenario where the app is first launched normally, then background and foreground through the link, we are not updating loginOption

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -406,10 +406,13 @@ protected TokenEndpointResponse performRequest(LoginOptions options) {
         @Override
         protected void onPostExecute(TokenEndpointResponse tr) {
             if (tr != null && tr.authToken != null) {
-                loginOptions.jwt = tr.authToken;
+                loginOptions.setJwt(tr.authToken);
                 doLoadPage(true);
             }
-            loginOptions.jwt = null;
+            else {
```
huminzhi: This is to fix a scenario where if the jwt expired, the app will stuck at a blank screen

bhariharan:
👍

PR #1206, opened 2016-07-28
forcedroid default to current folder as output folder

kchitalia:


node/forcedroid.js
```diff
@@ -106,6 +106,11 @@ function createApp(config) {
         process.exit(8);
     }
 
+    //If no target directory specified, default to current directory
+    if(config.targetdir == null || config.targetdir.length == 0) {
```
bhariharan: If `config.targetdir` is not set, let's automatically set it to `appName`. All the other changes in this file can be reverted. I feel like that would be a much cleaner approach.

node/forcedroid.js
```diff
@@ -151,7 +156,7 @@ function getAndroidSDKToolPath() {
 // Helper to create hybrid application
 //
 function createHybridApp(config) {
-    config.projectDir = path.join(config.targetdir, config.appname);
+    config.projectDir = config.targetdir;
```
kchitalia: Had to do this, otherwise we end up one level too deep.

bhariharan:
👍

PR #1207, opened 2016-07-29
Fix react native app crash during launch time

helenren:
The helper method is declared as public and react method wrongly.


bhariharan:
Nice! 👍

PR #1210, opened 2016-08-02
Adding screen overlay permission support

bhariharan:


bhariharan:
Don't merge this yet. I'm still testing and making some tweaks.

wmathurin:
Doesn't the super class do it already?
See https://github.com/facebook/react-native/blob/v0.27.2/ReactAndroid/src/main/java/com/facebook/react/ReactActivity.java#L133

bhariharan:
@wmathurin Interesting. I didn't notice that we subclass `ReactActivity`. I thought we were subclassing `Activity` like the other `SalesforceActivity` classes. Somehow, it doesn't work on `Marshmallow`. I'll dig a bit more.

bhariharan:
The crash happens because of the super class. `ReactActivity` launches the screen overlay permission screen in `onCreate()` and then tries to draw layout over the screen. But it doesn't have permission to do so, because it's now in the `Settings` app, which causes a permissions crash. Ideally, it shouldn't try to draw over the `Settings` app before getting permission. I don't know if we can fix this on our end, without fixing the super class.

PR #1213, opened 2016-08-03
add callback for jwt fail flow

huminzhi:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -394,13 +394,8 @@ protected void onAuthFlowComplete(TokenEndpointResponse tr) {
     private class SwapJWTForAccessTokenTask extends BaseFinishAuthFlowTask<LoginOptions> {
 
         @Override
-        protected TokenEndpointResponse performRequest(LoginOptions options) {
-            try {
```
huminzhi: no need to wrap the exception here, there is an existing exception hander..

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -410,7 +405,8 @@ protected void onPostExecute(TokenEndpointResponse tr) {
                 doLoadPage(true);
             }
             else {
-                doLoadPage(false);
+                onAuthFlowError("jwt_oauth_error", "fail to swap jwt");
+                callback.finish();
```
huminzhi: it will make a toast of the error, instead of fail silently..

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -410,7 +405,8 @@ protected void onPostExecute(TokenEndpointResponse tr) {
                 doLoadPage(true);
             }
             else {
-                doLoadPage(false);
+                onAuthFlowError("jwt_oauth_error", "fail to swap jwt");
```
bhariharan: @huminzhi So, with this change, `doLoadPage` never gets called with `false`?

huminzhi:
@bhariharan @kchitalia @helenren Please review

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
huminzhi: I just tried, it will load the page still, and have a toast showing the error.
seems all handled in OnAuthFlowError(....)

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
bhariharan: The `OAuth2.swapJWTForTokens` call throws a bunch of exceptions though. We need to catch them here, but instead of just logging them, we should call `onAuthFlowError` here.

huminzhi: Thanks @bhariharan , will update accordingly

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -394,11 +394,12 @@ protected void onAuthFlowComplete(TokenEndpointResponse tr) {
     private class SwapJWTForAccessTokenTask extends BaseFinishAuthFlowTask<LoginOptions> {
 
         @Override
-        protected TokenEndpointResponse performRequest(LoginOptions options) {
+        protected TokenEndpointResponse performRequest(LoginOptions options) throws Exception {
```
bhariharan: We should remove `throws Exception` here, since we're catching it anyway and triggering `onAuthFlowError`.

huminzhi: Thank you @bhariharan ,  I just removed it.

huminzhi:
Don't merge yet, still something on going..

PR #1216, opened 2016-08-08
Handling JWT errors

bhariharan:


bhariharan:
Ready to merge now. I've added generic `Intent` broadcasts for `OAuth` errors too, so consuming apps can do more than just the default `Toast` message we display, without having to override the class.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -240,6 +243,8 @@ public void run() {
             }, t.getDuration());
             t.show();
         }
+        final Intent intent = new Intent(AUTHENTICATION_FAILED_INTENT);
```
csoon03: I was able to receive the intent! Just wanted to confirm before I report back to UX and PM, there's no way to differentiate the errors?

csoon03: Also, I was looking at UserAccountManager.USER_SWITCH_INTENT_ACTION but this seems to only be sent if the user is switching to another user that has logged in previously, are there no intents sent for users who have successfully completed the login flow?

huminzhi: So you will need to define and check the error description to differentiate..

libs/SalesforceSDK/res/values/sf__strings.xml
```diff
@@ -8,6 +8,7 @@
 	<string name="sf__invalid_server_url">Please type in a valid URL starting with https://</string>
 	<string name="sf__access_token_revoked">Your access token has been revoked by the administrator. You will now be logged out.</string>
     <string name="sf__managed_app_error">Authentication only allowed from managed device.</string>
+	<string name="sf__jwt_authentication_error">JWT authentication error. Please try again.</string>
```
huminzhi: you can redefine the error description here..

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
csoon03: but there's nothing passed to the Intent that I can check against?

bhariharan: @csoon03 I can pass in `error` and `errorDesc` if you like. `error` is a `String` that represents the error like `access_denied`, for instance, and `errorDesc` is a human readable description of that. Typically, these fields are set for errors when the request fails that are returned from the server. If nothing is returned from the server, these fields might be empty. Does that work for you?

libs/SalesforceSDK/res/values/sf__strings.xml
bhariharan: Actually, you wouldn't modify this file. You would override the value for this key in your app's `strings.xml`.

csoon03: Gotcha!

bhariharan:
I worked with @csoon03 offline and tested the flow end-to-end. Seems to be working well. This should be good to go.

PR #1217, opened 2016-08-08
Adding support for alter with soup spec in hybrid and react native apps

wmathurin:


bhariharan:
👍

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartStorePlugin.java
```diff
@@ -59,25 +59,11 @@
 public class SmartStorePlugin extends ForcePlugin {
 
 	// Keys in json from/to javascript
-	public static final String BEGIN_KEY = "beginKey";
```
huminzhi: Just curious @wmathurin if it's ok to remove all those keys in terms of upgrading..the change should not impact upgrading right?

wmathurin: These constants moved to QuerySpec.java (
https://forcedotcom.github.io/SalesforceMobileSDK-Android/com/salesforce/androidsdk/smartstore/store/QuerySpec.html)
 many releases back.

bhariharan:
👍

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartStorePlugin.java
huminzhi: Thank you!

wmathurin:
Actually this PR was not finished.
I'll send another one soon with the remaining work.

PR #1224, opened 2016-08-16
loose exception type

huminzhi:
@bhariharan @wmathurin FYI, please review

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
@@ -567,7 +567,7 @@ public class DBOpenHelper extends SQLiteOpenHelper {
 		try {
 			result = new JSONObject(loadSoupBlobAsString(soupTableName, soupEntryId, passcode));
 
-		} catch (JSONException ex) {
+		} catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -567,7 +567,7 @@ public JSONObject loadSoupBlob(String soupTableName, long soupEntryId, String pa
 		try {
 			result = new JSONObject(loadSoupBlobAsString(soupTableName, soupEntryId, passcode));
 
-		} catch (JSONException ex) {
```
huminzhi: Tests are failing because the exception type is NullPointException here if the soupBlob is deleted (so instead of a mal-format json, it's a null)

bhariharan: @huminzhi I don't think we should catch an NPE.

huminzhi: Talked with @bhariharan , it should be ok in this case.

bhariharan:
Cleaner fix [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1225). Closing this.

PR #1226, opened 2016-08-17
Prevent sync if a logout is in progress

bhariharan:


native/NativeSampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/sync/ContactSyncAdapter.java
```diff
@@ -58,8 +59,13 @@ public ContactSyncAdapter(Context context, boolean autoInitialize,
 	@Override
 	public void onPerformSync(Account account, Bundle extras, String authority,
```
bhariharan: This is async and triggered by the OS (`Android's` background sync paradigm). Hence, we shouldn't attempt to sync when a logout is in progress.

wmathurin:
👍

Issue #1237, opened 2016-08-24
SalesforceSDKManager.getInstance().hasNetwork() is always true


stvnhg:
1. Version of Mobile SDK Used: 4.1
2. Issue found in Native App or Hybrid App: Hybrid
3. OS Version: Android 6.0
4. Device: Moto G

**Problem**
When users are offline, not logged in and they open the app for the first time they are not getting redirected to the error page. 

When you look at line 210 in SalesforceDroidGapActivity.java "onResumeNotLoggedIn()". When i debug my application and I turn off WiFi and Mobile Data I always get true for the expression "SalesforceSDKManager.getInstance().hasNetwork()" on line 216. 

Where is this variable set? I can only find the onReceive method in HttpAccess.java but it is being executed after the onResumeNotLoggedIn()... Also, the last if statement should also check the detailedState and not only null value because it isn't null when your connection is down so that if would never be reached.

` // Tries mobile connection.` 
` final NetworkInfo mobileInfo = conMgr.getNetworkInfo(ConnectivityManager.TYPE_MOBILE);` 
` if (mobileInfo == null || mobileInfo.getDetailedState() == NetworkInfo.DetailedState.DISCONNECTED) {` 
            `             setHasNetwork(false);` 
            `             return;` 
        ` }`

We have similar issue in iOS version. Do I need to create the same issue on the other repo or can you tell me how to fix it here?

Thanks!

huminzhi:
Thank you for your info @stvnhg , I confirm it's an issue in Android and will fix it in next release.
You mentioned you see similar issue in iOS, will you please file another issue there since it's totally different lib and path? Thanks a lot!

bhariharan:
Fixed [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1238).

stvnhg:
Thank you for that but issue is not solved, please re-open. What about the fact when my connection is down and I enter the "onResumeNotLoggedIn()" method in "SalesforceDroidGapActivity.java" on line 210 when I open the app when I'm not logged in? "SalesforceSDKManager.getInstance().hasNetwork()" on line 216 still returns true and tries to authenticate. I want it to enter the else clause and do the "loadErrorPage()" on line 224. Thank you!

stvnhg:
@bhariharan @huminzhi Any update on this? Thank you.

stvnhg:
@bhariharan @huminzhi ???

[...]

PR #1238, opened 2016-08-24
fix issue where not setting network to false as should be

huminzhi:
For issue 
https://github.com/forcedotcom/SalesforceMobileSDK-Android/issues/1237

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
@@ -176,10 +176,12 @@ public class HttpAccess extends BroadcastReceiver {
 
         // Tries mobile connection.
         final NetworkInfo mobileInfo = conMgr.getNetworkInfo(ConnectivityManager.TYPE_MOBILE);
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
```diff
@@ -176,7 +176,7 @@ public void onReceive(Context context, Intent intent) {
 
         // Tries mobile connection.
         final NetworkInfo mobileInfo = conMgr.getNetworkInfo(ConnectivityManager.TYPE_MOBILE);
-        if (mobileInfo == null) {
+        if (mobileInfo == null || mobileInfo.getDetailedState() == NetworkInfo.DetailedState.DISCONNECTED) {
```
bhariharan: We should probably use the same pattern as above, and change this block as follows:

```
    if (mobileInfo != null && mobileInfo.isConnected()) {
        setHasNetwork(true);
        return;
    }
    setHasNetwork(false);
    return;
}
```

huminzhi: Good point. updated..

bhariharan:
👍

PR #1239, opened 2016-09-03
fix the build and test failure on travis ci

helenren:
The test_force part still fail for the pull request build since the tool pick the code from unstable branch only instead the unstable+the change. Locally, the failure is gone with this fix. And the build triggered by push on unstable should reflect the fix correctly.


wmathurin:
👍

huminzhi:
👍

PR #1243, opened 2016-09-12
Fixing network status on Android

bhariharan:
Funnily enough, we monitor change in network status, but never propagate those changes anywhere. So, if an app starts offline, we're going to act like network is available.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2015, salesforce.com, inc.
+ * Copyright (c) 2011-present, salesforce.com, inc.
  * All rights reserved.
  * Redistribution and use of this software in source and binary forms, with or
  * without modification, are permitted provided that the following conditions
@@ -26,10 +26,7 @@
  */
 package com.salesforce.androidsdk.auth;
 
-import android.content.BroadcastReceiver;
 import android.content.Context;
-import android.content.Intent;
-import android.content.IntentFilter;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Build;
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
```diff
@@ -149,50 +138,21 @@ public synchronized OkHttpClient getOkHttpClient() {
...
-     * @param intent Intent.
-     */
-    @Override
-    public void onReceive(Context context, Intent intent) {
-        if (conMgr == null) {
-            return;
-        }
-
-        // Checks if an active network is available.
-        final NetworkInfo activeInfo = conMgr.getActiveNetworkInfo();
```
bhariharan: We don't need the conditionals below or checking per type. Current type is all we need to establish if we're connected.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
```diff
@@ -149,50 +138,21 @@ public synchronized OkHttpClient getOkHttpClient() {
     }
 
     /**
-     * Detects network changes and sets the network connectivity status.
-     *
-     * @param context The context of the request.
-     * @param intent Intent.
-     */
-    @Override
-    public void onReceive(Context context, Intent intent) {
```
bhariharan: We don't need to monitor changes in network connectivity - we don't use them anywhere.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
```diff
@@ -149,50 +138,21 @@ public synchronized OkHttpClient getOkHttpClient() {
...
-
-    /**
      * Returns the status of network connectivity.
      *
      * @return True - if network connectivity is available, False - otherwise.
      */
     public synchronized boolean hasNetwork() {
-        return hasNetwork;
+        boolean isConnected = true;
+        if (conMgr != null) {
```
bhariharan: All we need to do is check the current status of network connectivity here.

wmathurin:
👍

PR #1245, opened 2016-09-13
update the shared submodule pointer to use the latest one

helenren:


bhariharan:
@helenren Looks like this needs to be rebased. I'll take care of updating it. I'm sending a PR soon.

PR #1246, opened 2016-09-14
Smartsync refresh

wmathurin:
Some apps might get records and store them (cache them) directly into smartstore instead of using SmartSync. Those apps might still want to refresh cached records when coming back from the background etc. 

For that purpose, there is a new sync down target type called "refresh" (in addition to the existing ones "mru", "soql", "sosl" and "custom"
It expects a soup name, an object type and the list of fields to refresh.
It first gets the ids from the records in the soup and then use SOQL to get the current values from the server.

1) Multiple round trips to the server could be necessary because of the 10k characters limit on soql query.
2) If you do a reSync for refresh-sync-down, it will only fetched updated records as expected but the timestamp used to compare against is computed from the stored records instead of being based the ones calculated the last time the sync down was run.

Tests were added for the new sync down / for reSync of refresh-sync-down / for clean ghosts of refresh-sync-down and also for the multiple round tips cases.


libs/SalesforceSDK/src/com/salesforce/androidsdk/util/JSONObjectHelper.java
```diff
@@ -82,5 +85,42 @@ public static String optString(JSONObject obj, String key) throws JSONException
 	public static String optString(JSONObject obj, String key, String fallback) throws JSONException {
 		return obj.isNull(key) ? fallback : obj.getString(key);
 	}
-
```
wmathurin: Those methods used to be in SoqlSyncDownTarget

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/CacheManager.java
```diff
@@ -749,22 +749,27 @@ private boolean doesMasterSoupContainSoup(String soupName) {
...
-        try {
-            int count = smartStore.countQuery(querySpec);
-            querySpec = QuerySpec.buildSmartQuerySpec(smartSql, count);
-			results = smartStore.query(querySpec, 0);
-		} catch (JSONException e) {
-            Log.e(TAG, "JSONException occurred while attempting to read cached data", e);
-		} catch (SmartStoreException e) {
-            Log.e(TAG, "SmartStoreException occurred while attempting to read cached data", e);
+
+        if (smartStore.hasSoup(SOUP_OF_SOUPS)) {
```
wmathurin: An exception was showing up in the log after calling hardReset on the CacheManager (which we do in the tear down of smartsync tests) when it had never been used.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -629,7 +629,14 @@ public RestResponse sendSyncWithSmartSyncUserAgent(RestRequest restRequest) thro
         return restClient.sendSync(restRequest, new HttpAccess.UserAgentInterceptor(SalesforceSDKManager.getInstance().getUserAgent(SMART_SYNC)));
     }
 
-	/**
+    /**
```
wmathurin: The new sync down target needs to read from smartstore to decide what to download from the server.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncDownTarget.java
```diff
@@ -120,16 +123,6 @@ public JSONObject asJSON() throws JSONException {
...
 
     /**
-     * Starts fetching records conforming to target query.
-     *
-     * @param syncManager SyncManager instance.
-     * @param maxTimeStamp Max timestamp.
-     * @param queryRun Query to run.
-     * @throws IOException, JSONException
-     */
-    public abstract JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String queryRun) throws IOException, JSONException;
```
wmathurin: Only certain sync down target have a queryRun

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTest.java
```diff
@@ -1255,7 +1401,7 @@ private String makeInClause(Collection<String> values) {
     /**
      Soql sync down target that pauses for a second at the beginning of the fetch
      */
-    static class SlowSoqlSyncDownTarget extends SoqlSyncDownTarget {
+    public static class SlowSoqlSyncDownTarget extends SoqlSyncDownTarget {
```
wmathurin: Unrelated bug fix - To avoid runtime error causing a test failure

bhariharan:
Looks good to me. 👍

PR #1248, opened 2016-09-15
fix the test failure that to use the clientinfo from oAuthRefreshInterceptor

helenren:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -233,7 +233,7 @@ public String getRefreshToken() {
 	 * @return The client info.
 	 */
 	public ClientInfo getClientInfo() {
-		return clientInfo;
+		return oAuthRefreshInterceptor.clientInfo;
```
helenren: the clientInfo is updated for the 2nd request using the access token from authTokenProvider once gotten a 401 return at the first time

bhariharan: Maybe we should remove `clientInfo` instance variable from this class and use `oAuthRefreshInterceptor.clientInfo` everywhere? I don't see any reason to keep the instance variable around. It makes it harder to keep the two copies in sync.

helenren: agree. Will update the code.

wmathurin:
👍

bhariharan:
👍

PR #1250, opened 2016-09-20
Android N (Nougat) Support

bhariharan:


bhariharan:
This bumps up the `targetApiVersion` to `24`, so apps can leverage `Nougat` APIs.

bhariharan:
Sample apps are working fine - both native and hybrid.

PR #1252, opened 2016-09-24
Latest shared + fix for bug revealed by refresh sync hybrid test

wmathurin:


wmathurin:
There are a few smartsync tests that flap (they pass when run alone but fail when run in the suite).
I'd like to fix them before merging.

PR #1263, opened 2016-10-21
Adding publisher service and persistence for on/off

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AILTNPublisherService.java
```diff
@@ -0,0 +1,85 @@
...
+            if (ACTION_PUBLISH.equals(action)) {
+                handleActionPublish();
+            }
+        }
+    }
+
+    /**
+     * Handles the publish action in the provided background thread.
+     */
+    private void handleActionPublish() {
```
bhariharan: Simple `IntentService` that we can launch on demand to publish stored logs.

wmathurin:
👍

PR #1267, opened 2016-10-25
Updating publish frequency should update scheduler

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -73,6 +74,7 @@
 
     private static Map<String, SalesforceAnalyticsManager> INSTANCES;
     private static boolean PUBLISH_HANDLER_ACTIVE;
+    private static ScheduledFuture SCHEDULER;
```
wmathurin: Style-wise, shouldn't only static final (i.e. constants) be all upper case?

bhariharan: Typically, yes, you're right. Static would be something like `sScheduler` as opposed to instance variable being `mScheduler`, but we have been using uppercase all over our projects, so I kept it consistent with that.

bhariharan:
@wmathurin Renamed [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1267/commits/70bdfc90596e7cd97ee2a924bc6104675b2c984c).

PR #1270, opened 2016-10-29
Android N 7.1 Support

bhariharan:


native/TemplateApp/AndroidManifest.xml
```diff
@@ -23,7 +23,7 @@
 	</application>
 
     <uses-sdk android:minSdkVersion="19"
-        android:targetSdkVersion="24" />
```
bhariharan: @wmathurin You'll need to pick up these changes in your template app repo PR. The target API version is now `25` in `AndroidManifest.xml`, `build.gradle` for the template apps and the template test apps.

wmathurin: Fortunately I have not started with the android templates :-)

PR #1279, opened 2016-11-04
Add SmartSync Feature to report

ivanbogdanov:


helenren:
build

helenren:
build

helenren:
build

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/app/SmartSyncSDKManager.java
```diff
@@ -115,6 +118,7 @@ public static void initNative(Context context, KeyInterface keyImpl,
      */
     public static SmartSyncSDKManager getInstance() {
```
wmathurin: Any app compiled with smartstore/smartsync will use SmartSyncSDKManager.
Instead you should set that flag from SyncManager.java getInstance

bhariharan: +1

ivanbogdanov:
cleaned up SmartSyncSDKManager

PR #1283, opened 2016-11-08
Adding token refresh event

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -685,6 +693,33 @@ public long getLastRefreshTime() {
         public String getInstanceUrl() { return lastNewInstanceUrl; }
     }
 
+    private static void logAnalyticsEvent(String name) {
```
wmathurin: All the logAnalyticsEvent look alike - I wonder if we could have a helper method in SalesforceAnalyticsManager that would take in name, context and custom attributes?

bhariharan:
@wmathurin Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1283/commits/23e50339101d793dca944a5d7c46bbfc07aef432). I added a helper class for it.

wmathurin:
👍

PR #1290, opened 2016-11-11
Adding logout event

bhariharan:


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -398,7 +398,7 @@ public void onError(Exception exception) {
 
                 // Only logout if we are NOT offline
                 if (!(exception instanceof NoNetworkException)) {
-                    SalesforceSDKManager.getInstance().logout(SalesforceDroidGapActivity.this);
+                    logout(null);
```
bhariharan: Unrelated bugfix.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -782,6 +784,7 @@ public void logout(Activity frontActivity, final boolean showLoginPage) {
      * @param showLoginPage If true, displays the login page after removing the account.
      */
     public void logout(Account account, Activity frontActivity, final boolean showLoginPage) {
+        EventBuilderHelper.createAndStoreEvent("userLogout", null, TAG, null);
```
bhariharan: Yes - that's why the call is before `logout`. The missing piece here (coming in another PR) is a fire-and-forget `publish` call just before `logout`.

Issue #1273, opened 2016-11-01
Clean Resync Ghosts get Object Name After From


mzaklama:
Please fill out the following details:

1. Version of Mobile SDK Used: 4.3.1
2. Issue found in Native App or Hybrid App: Native
3. OS Version: Android 5.0
4. Device: Samsung
5. Steps to reproduce:
SoqlSyncDownTarget.java Line 154
If you have a field containing FROM, the split does not get the correct Soup/Object.
eg: I have a field called From_Customer__c and an Object called Stock_Transfers__c; the soql StringBuilder after line 154 is equal to "SELECT Id FROM_Customer__c," instead of "SELECT Id FROM Stock_Transfers".
I had to change this line to be soql.append(fromClause[fromClause.length - 1]);

mzaklama:
My fix will not work if the Object itself contains "FROM"

mzaklama:
A better fix is to change lines 152 and 153 to
soql.append(" FROM ");
final String[] fromClause = query.split("([ ][fF][rR][oO][mM])[ ]");

This way it will insure that we get the FROM that is not included in a Field name or an Object Name

bhariharan:
@mzaklama Thanks for reporting this. Mind sending us a pull request against `unstable`?

bhariharan:
Fixed [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1296).

PR #1296, opened 2016-11-14
Bug fix in regex

bhariharan:


bhariharan:
Fix for this [issue](https://github.com/forcedotcom/SalesforceMobileSDK-Android/issues/1273).

bhariharan:
We already have a test for `cleanResyncGhosts`. We need to add a custom object with a custom field with `from` in it to explicitly test this.

wmathurin:
I see. Or if the soql transforming code was in its own helper method, we could test it in isolation of the rest. Anyway, it's up to you.

bhariharan:
It's specific to the `SOQL` target right now. Extracting it out for this test might need a little work, so I'll leave it as is for now.

PR #1297, opened 2016-11-16
Removing deprecated APIs

bhariharan:


bhariharan:
Any feedback? We don't use these APIs anymore. `5.0` seems to be the right train to remove APIs.

PR #1301, opened 2016-11-17
fix SmartSyncExplorerReactNative logout issue

ivanbogdanov:


ivanbogdanov:
@wmathurin

wmathurin:
The problem probably resides in SalesforceReacActivity and / or with the app itself.

bhariharan:
Agree with @wmathurin. One of the biggest differences in `React` app and `Cordova` app is that `SalesforceReactActivity` extends `ReactActivity` and not our `Activity`, whereas `SalesforceDroidGapActivity` is ours, so we setup the `getRestClient` properly. We need to ensure the `registerSoup` call happens only after `getRestClient` returns with a valid `RestClient`.

ivanbogdanov:
I think it reveals larger problem: most bridge methods shouldn't be called if a user is not logged in.

wmathurin:
The oauth bridge logout method calls logout on the SalesforceReactActivity which calls logout on SalesforceSDKManager which should finish the SalesforceReactActivity. After login, it should be re-created. Somehow the react code starts too soon. I'm pretty sure that was working correctly with older version of React native. I wonder if they keep the javascript context around.

ivanbogdanov:
Developers can call these bridge methods outside of  SalesforceReactActivity just from ReactActivity right?

wmathurin:
They need to use SalesforceReactActivity (just like cordova developers that use our SDK must use SalesforceDroidGapActivity).

ivanbogdanov:
I have just double-checked: registerSoup is called first time BEFORE login.

bhariharan:
@ivanbogdanov Looks like we need to fix that in the app, so that it gets called only AFTER login.

ivanbogdanov:
Do we care about scenarios of app functionality without / before login?

bhariharan:
App could do anything that doesn't require login before login. For example, global `SmartStore` would probably work.

ivanbogdanov:
Is global `SmartStore` call allowed for an authenticated user?

wmathurin:
There is no problem after first logging in. The problem only arise on
logout. I think the react context is live too soon (or was never tear down).

bhariharan:
@ivanbogdanov Let's close this PR and discuss app level solutions offline.

PR #1302, opened 2016-11-18
Fixing bug with account state in plugins

bhariharan:
Based on what we discussed @ivanbogdanov and @wmathurin.

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartSyncPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartSyncPlugin.java
@@ -28,6 +28,7 @@ package com.salesforce.androidsdk.phonegap.plugin;
 
 import android.util.Log;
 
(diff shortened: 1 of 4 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartSyncPlugin.java
```diff
@@ -288,7 +295,7 @@ private SyncManager getSyncManager(boolean isGlobal,String storeName) {
      * @param arg0 first argument passed in plugin call
      * @return
      */
-    private static SmartStore getSmartStore(JSONObject arg0) {
+    private static SmartStore getSmartStore(JSONObject arg0) throws Exception {
```
wmathurin: Could we only have the next two methods in SmartStorePlugin.java and have SmartSyncPlugin.java call them there - like we do in the React bridges?

bhariharan:
Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1302/commits/ed6942c650d0487635e36d3caddfd539bd24eb1a).

PR #1306, opened 2016-11-22
Load react native App only when logged in

ivanbogdanov:
This is probably the most economical way to fix it without major rewrite


ivanbogdanov:
@wmathurin @bhariharan

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivityDelegate.java
```diff
@@ -0,0 +1,33 @@
+package com.salesforce.androidsdk.reactnative.ui;
```
bhariharan: Add copyright here. You can copy it from the other file.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivityDelegate.java
```diff
@@ -0,0 +1,33 @@
...
+        super(activity, mainComponentName);
+    }
+
+    public SalesforceReactActivityDelegate(FragmentActivity fragmentActivity, @Nullable String mainComponentName) {
+        super(fragmentActivity, mainComponentName);
+    }
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+//        super.onCreate(savedInstanceState);
```
bhariharan: Remove commented out line.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java
```diff
@@ -100,6 +103,9 @@ public void onDestroy() {
 
     @Override
     public void onResume() {
+        if(client != null && reactActivityDelegate != null){
```
bhariharan: This call should come after `super.onResume()`. Also, I'm not sure if this will work for first time load, because `client` will be `null` the first time, since the `peekRestClient()` call is below and that's what initiates login flow.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java
```diff
@@ -100,6 +103,9 @@ public void onDestroy() {
...
+            reactActivityDelegate.onReadyCreate();
```
bhariharan: In general, from a lifecycle perspective, I don't know if it's a good idea to call `onCreate()` from within `onResume()`. `onResume()` gets triggered every time this view loads, whereas `onCreate()` is meant for heavy lifting that's typically done once in the lifecycle of that view controller. If the parent activity does something different in `onCreate()` in a future version of `ReactNative`, this could be bad for us. Also, we lose the ability to load from saved `Bundle`. @wmathurin What do you think?

wmathurin: if they override shouldAuthenticate() to return false, then it means they expect to run the react native code without being authenticated (they might choose to authenticate later by using the oauth bridge). In that case, we should do the onReadyCreate() without having a client.

bhariharan:
Looking at the code, I would instead override `loadApp()` in the delegate and do something a little different. Right now, it looks like this:

```
protected void loadApp(String appKey) {
    if (mReactRootView != null) {
      throw new IllegalStateException("Cannot loadApp while app is already running.");
    }
    mReactRootView = createRootView();
    mReactRootView.startReactApplication(
      getReactNativeHost().getReactInstanceManager(),
      appKey,
      getLaunchOptions());
    getPlainActivity().setContentView(mReactRootView);
  }
```

I would move the `startReactApplication...` call to its own method in the subclass and not call that, until `onResume()`. That way, we don't muck around with other stuff in the lifecycle of `Android`, but we defer `loadApp`.

ivanbogdanov:
Fixed. Let me know

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivityDelegate.java
```diff
@@ -0,0 +1,68 @@
...
+
+
+    public void onReadyLoadApp() {
+        super.loadApp(appComponentName);
+        restClient = salesforceReactActivity.getRestClient();
+    }
+
+    @Override
+    protected void loadApp(String appKey) {
+        if(salesforceReactActivity != null && !salesforceReactActivity.shouldAuthenticate()){
```
wmathurin: This conditional (or a similar one) is in multiple place. I don't see why you need the restClient here.

What about doing the following:

- Have a shouldReactBeRunning() method in SalesforceReactActivity that internally checks shouldAuthenticate() -- those details do not need to be known to the delegate class.

- Have a isRunning member in the SalesforceReactActivityDelegate class (initially false).

- Have loadApp() do:
```java
if (!isRunning && salesforceReactActivity != null && salesforceReactActivity.shouldReactBeRunning()) {
  super.loadApp(appComponentName);
  isRunning = true;
}
```

- Have onResume do:
```java
loadApp(); // NB loadApp does nothing if it is already loaded
```

ivanbogdanov: I am checking for restClient to cover the following situation:
* the activity shouldReactBeRunning returns `true`
* the user is not yet logged in
* the the simplest way to check it is to see if `restClient` is not null

wmathurin: shouldReactBeRunning should not return true if the login has not completed - it should only return true if authentication is not required or it is required and has completed.
I'd rather not have the delegate know to much about the internals of the SalesforceReactActivity.

wmathurin:
It looks better but what happens when you first login.
SalesforceReactActivity's resume calls super.resume which calls onResume on the delegate - and since the client has not yet been peeked by SalesforceReactActivity, then it will not load the react app. Does one have to leave and go back for it to load?

ivanbogdanov:
Fixed. Let me know.

wmathurin:
Still unsure how it works when you first login
shouldReactBeRunning will only be true after https://github.com/ivanbogdanov/SalesforceMobileSDK-Android/blob/8f9b12e58203fd4c70668b05b7ff9e07065d0540/libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java#L111

But the delegate resume will already have run.

What do you observe when you run the app?

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivityDelegate.java
```diff
@@ -0,0 +1,65 @@
...
+
+
+    @Override
+    protected void loadApp(String appKey) {
+        loadReactAppOnceIfReady(appKey);
+    }
+
+    @Override
+    protected void onResume() {
+        loadReactAppOnceIfReady(appComponentName);
```
bhariharan: Instead of overriding `onResume` here, maybe you could call `loadReactAppOnceReady` from `onResume` of `SalesforceReactActivity` once `client` is set if authentication is required (after the `peekRestClient` or `getRestClient` call). I think that should work for first time load too.

ivanbogdanov: Everything works on first time login. I can move the call to `SalesforceReactActivity` but it will be more stylistic change. Let me know

bhariharan: This is fine by me if it works on first time login.

wmathurin: I don't understand how it works, but I guess I'll step through the code.

ivanbogdanov:
@bhariharan

bhariharan:
👍

wmathurin:
It's not working properly:
- launch react native template app (or app generated from it)
- login
-> you get a blank/grey screen
if you hit home button and resume, then you will see the list of users as expected
if you kill the app and relaunch, then you get the blank/grey screen again
Again, if you hit home button and resume, then the list of users come back

PR #1308, opened 2016-11-29
Fix react-native app activation logic after login

ivanbogdanov:


ivanbogdanov:
@bhariharan @wmathurin

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/ui/SalesforceReactActivity.java
```diff
@@ -230,12 +230,10 @@ public RestClient getRestClient() {
     }
 
     protected void setRestClient(RestClient restClient) {
-        if(restClient!= null && client != restClient){
-            if(reactActivityDelegate != null){
-                reactActivityDelegate.loadReactAppOnceIfReady(getMainComponentName());
-            }
-        }
         client = restClient;
+        if(client != null && reactActivityDelegate != null){
```
bhariharan: @ivanbogdanov Does this fix the first time load gray screen issue that @wmathurin noticed?

ivanbogdanov: Yes: It was `client` should be set before `shouldReactBeRunning` call

PR #1310, opened 2016-11-30
Remove Guava as a compile dependency.

mattcreaser:
I've kept it available for unit tests to minimize the changes.


huminzhi:
ok to test (For CI)

libs/SmartStore/build.gradle
```diff
@@ -4,6 +4,7 @@ apply plugin: 'com.android.library'
 dependencies {
   compile project(':libs:SalesforceSDK')
   compile 'net.zetetic:android-database-sqlcipher:3.5.2'
+  androidTestCompile 'com.google.guava:guava:18.0'
```
bhariharan: @mattcreaser Why does the `SmartStore` test project need `Guava`?

wmathurin:
👍

PR #1313, opened 2016-12-02
Null handling fix

wmathurin:
Fix for bug reported https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/issues/239

Also added a method to completely disable EventBuilderHelper so that test can be run individually (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/46e1fee96a9485c6a511a078c7fadbffc0fa64c9)

### DIFF
--- a/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTestCase.java
+++ b/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTestCase.java
@@ -34,6 +34,7 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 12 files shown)


libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTestCase.java
```diff
@@ -60,6 +61,7 @@
 	@Override
 	public void setUp() throws Exception {
 		super.setUp();
+		EventBuilderHelper.enableDisable(false);
```
bhariharan: We should probably add this to the base test class of the other libraries too.

wmathurin:
Note: the bug only affected String index not JSON1 index. The problem was in the project function, and it's only used with String indexes.

PR #1317, opened 2016-12-05
Add check for user account existence

bhariharan:


bhariharan:
@helenren This should fix it.

PR #1321, opened 2016-12-07
Switching to Base64.NO_WRAP for Bas64.encodeToString

smcnulty-sfdc:
Also moving removeNewLine method to PasscodeManager, making it private and a little more efficient.


bhariharan:
build

PR #1325, opened 2016-12-12
separate build results for each of API and enable code coverage for h…

helenren:
…ybrid tests


helenren:
As I remembered we have enabled the code coverage for hybrid tests, somehow is not there. Just add it back.

wmathurin:
👍

PR #1338, opened 2016-12-20
add suport fastlane

helenren:


helenren:
@huminzhi

fastlane/Fastfile
```diff
@@ -0,0 +1,72 @@
...
+lane :All do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SalesforceSDK:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest :native:NativeSampleApps:RestExplorer:connectedAndroidTest",
+flags: "--continue --no-daemon --profile --stacktrace"
+)
+end
+
+lane :PR do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest",
```
bhariharan: @helenren Why is the PR lane missing `SalesforceSDK`?

fastlane/Fastfile
```diff
@@ -0,0 +1,72 @@
+system("cd .. && npm install shelljs@0.7.0")
+system("cd .. && ./install.sh")
+
+lane :All do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SalesforceSDK:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest :native:NativeSampleApps:RestExplorer:connectedAndroidTest",
```
bhariharan: Add `:libs:SalesforceReact:assembleDebug` to this lane.

fastlane/Fastfile
bhariharan: Add `:libs:SalesforceReact:assembleDebug` to this lane.

fastlane/Fastfile
```diff
@@ -0,0 +1,72 @@
...
+lane :PR do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest",
+flags: "--continue --no-daemon --profile --stacktrace"
+    )
+end
+
+lane :noSDK do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest :native:NativeSampleApps:RestExplorer:connectedAndroidTest",
```
bhariharan: Add `:libs:SalesforceReact:assembleDebug` to this lane.

fastlane/Fastfile
```diff
@@ -0,0 +1,72 @@
...
+flags: "--continue --no-daemon --profile --stacktrace"
+)
+end
+
+lane :noSmartStore do
+gradle(
```
bhariharan: Add `:libs:SalesforceReact:assembleDebug` to this lane.

fastlane/Fastfile
```diff
@@ -0,0 +1,72 @@
...
+task:":libs:SalesforceHybrid:connectedAndroidTest",
+flags: "--continue --no-daemon --profile --stacktrace"
+)
+end
+
+lane :NativeSampleApps do
+gradle(
+task:":native:NativeSampleApps:RestExplorer:connectedAndroidTest",
+flags: "--continue --no-daemon --profile --stacktrace"
+)
```
bhariharan: Add a lane for `:SalesforceReact` that does `:libs:SalesforceReact:assembleDebug`.

fastlane/Fastfile
helenren: Since our PR is running on android-23 version that is most reliable version we recognized so far, for which the emulator has an known issue that cannot setup proxy for emulator to use. We only can do that for android-19. Without the proxy server setup, test testRestClientUnauthenticatedlientInfo will consistently failed.

fastlane/Fastfile
helenren: added.

fastlane/Fastfile
helenren: added

fastlane/Fastfile
```diff
@@ -0,0 +1,86 @@
...
+lane :All do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SalesforceSDK:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest :libs:SalesforceReact:assembleDebug :native:NativeSampleApps:RestExplorer:connectedAndroidTest",
+flags: "--continue --no-daemon --profile --stacktrace"
+)
+end
+
+lane :PR do
+gradle(
+task:":libs:SalesforceAnalytics:connectedAndroidTest :libs:SmartStore:connectedAndroidTest :libs:SmartSync:connectedAndroidTest :libs:SalesforceHybrid:connectedAndroidTest :libs:SalesforceReact:assembleDebug",
```
bhariharan: @helenren It's okay if the one test fails. I think we should still add `SalesforceSDK` in here - at least we'll know if more tests fail.

helenren: My concern is that every PR build will fail, and the people will easily to think the failure is a known one always, unless to check the results every single time. we will do that? And after each change pushed into dev branch, there will be a full test pass run automatically triggered on all supported versions, including sdk test on 19. In this case, we still don't lose the control. It is not a idea solution, while balance the effort and the coverage, I would prefer not include sdk test in our PR, until we figure a way to solve that emulator issue.

bhariharan: @helenren Okay, let's leave it as is for now then.

PR #1340, opened 2017-01-06
NPE: LoginActivity.initAnlayticsManager()

narakim:
NullPointerException: SalesforceAnalyticsManager.getInstance(Account).updateLoggingPrefs() called when not logged in.

- LoginActivity doesn’t always finish with an account. User can back out instead, but SalesforceAnalyticsManager.getInstance() can return null when there isn’t any account.

Steps:
1) Start Login Activity
2) Press back without logging in


wmathurin:
Looks good. Thanks.

urthakkar:
Thanks @narakim

PR #1343, opened 2017-01-11
SyncUp can now have separate fieldlist for create and update

wmathurin:
Android equivalent of https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1887

Added two optional members in SyncOptions: createFieldlist and updateFieldlist.
During sync up:
* created records use createFieldlist if available (and fieldlist otherwise).
* updated records use updateFieldlist if available (and fieldlist otherwise).

Code change very simple.

Tests changes
Tests used to deal with just one field (name) now they deal with multiple fields (name and description).
New tests were added for createUpdatelist and updateFieldlist:
* testSyncUpWithUpdateFieldList
* testSyncUpWithCreateFieldList
* testSyncUpWithCreateAndUpdateFieldList

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncOptions.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncOptions.java
@@ -43,10 +43,14 @@ public class SyncOptions {
 
     public static final String MERGEMODE = "mergeMode";
 	public static final String FIELDLIST = "fieldlist";
+	public static final String CREATE_FIELDLIST = "createFieldlist";
+	public static final String UPDATE_FIELDLIST = "updateFieldlist";
 
     private MergeMode mergeMode;
 	private List<String> fieldlist;
-	
+	private List<String> createFieldlist;
+	private List<String> updateFieldlist;
+
 	/**
 	 * Build SyncOptions from json
 	 * @param options as json
@@ -60,7 +64,10 @@ public class SyncOptions {
         String mergeModeStr = JSONObjectHelper.optString(options, MERGEMODE);
         MergeMode mergeMode = mergeModeStr == null ? null : MergeMode.valueOf(mergeModeStr);
 		List<String> fieldlist = toList(options.optJSONArray(FIELDLIST));
-		return new SyncOptions(fieldlist, mergeMode);
+		List<String> createFieldlist = toList(options.optJSONArray(CREATE_FIELDLIST));
+		List<String> updateFieldlist = toList(options.optJSONArray(UPDATE_FIELDLIST));
+
+		return new SyncOptions(fieldlist, createFieldlist, updateFieldlist, mergeMode);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


wmathurin:
Can I get a review ;-) ?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncOptions.java
```diff
@@ -106,13 +126,23 @@ public JSONObject asJSON() throws JSONException {
 		JSONObject options = new JSONObject();
         if (mergeMode != null) options.put(MERGEMODE, mergeMode.name());
 		if (fieldlist != null) options.put(FIELDLIST, new JSONArray(fieldlist));
+		if (createFieldlist != null) options.put(CREATE_FIELDLIST, new JSONArray(createFieldlist));
+		if (updateFieldlist != null) options.put(UPDATE_FIELDLIST, new JSONArray(updateFieldlist));
 		return options;
 	}
```
trooper2013: What would happen if fieldList AND createFieldlist/updateFieldlist both have been set. Do we have any special handling for which takes precedence?

trooper2013: Seems like the create & update list does in other parts of code. Then maybe an if..else?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncOptions.java
```diff
@@ -77,24 +84,37 @@ public static SyncOptions optionsForSyncUp(List<String> fieldlist) {
...
     }
 
+	/**
+	 * @param fieldlist
+	 * @param createFieldlist
+	 * @param updateFieldlist
+	 * @param mergeMode
+	 * @return
+	 */
+	public static SyncOptions optionsForSyncUp(List<String> fieldlist, List<String> createFieldlist, List<String> updateFieldlist, MergeMode mergeMode) {
```
trooper2013: Do we need a public API with both fieldList & create/update Fieldlist or are these mutually exclusive.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncOptions.java
wmathurin: The code making use of it is here: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1343/files#diff-4201649df1e9c61d455833a013ad7b4bR498
During sync up, there might be creates / updatse and deletes.
But a given record is either created or updated or deleted.
During create, the create list is used if available otherwise fieldlist is used.
During update, the update list is used if available otherwise fieldlist is used.

wmathurin: By the way, having both is the whole point of this PR.

wmathurin:
Can you do a review -> approve ?
Thanks.

khawkins:
Looks pretty in line with the iOS side, to me. 👍

PR #1344, opened 2017-01-11
Passing Additional Parameters to Auth endpoints

trooper2013:
Added ability to pass Additional Parameters using LoginOptions. Changed OAuth2  and ClientManager to ingest these parameters into auth/token endpoint & account manager.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
@@ -76,6 +76,7 @@ public class AuthenticatorService extends Service {
     public static final String KEY_DISPLAY_NAME = "display_name";
     public static final String KEY_PHOTO_URL = "photoUrl";
     public static final String KEY_THUMBNAIL_URL = "thumbnailUrl";
+    public static final String KEY_ADDL_PARAMS = "addlParams";
 
     private Authenticator getAuthenticator() {
         if (authenticator == null)
@@ -223,6 +224,15 @@ public class AuthenticatorService extends Service {
                     }
                 }
             }
+
+            final String encAddlParams = mgr.getUserData(account, AuthenticatorService.KEY_ADDL_PARAMS);
+            Map<String,String> addlParamsMap = null;
+            if(encAddlParams != null) {
+                String addlParams = SalesforceSDKManager.decryptWithPasscode(encAddlParams,
+                        SalesforceSDKManager.getInstance().getPasscodeHash());
+                addlParamsMap = OAuth2.parameterStringToMap(addlParams);
+            }
+
             final String encCommunityId = mgr.getUserData(account, AuthenticatorService.KEY_COMMUNITY_ID);
             String communityId = null;
             if (encCommunityId != null) {
@@ -237,7 +247,7 @@ public class AuthenticatorService extends Service {
             }
             final Bundle resBundle = new Bundle();
             try {
-                final TokenEndpointResponse tr = OAuth2.refreshAuthToken(HttpAccess.DEFAULT, new URI(loginServer), clientId, refreshToken, clientSecret);
+                final TokenEndpointResponse tr = OAuth2.refreshAuthToken(HttpAccess.DEFAULT, new URI(loginServer), clientId, refreshToken, clientSecret,addlParamsMap);
 
                 // Handle the case where the org has been migrated to a new instance, or has turned on my domains.
                 if (!instServer.equalsIgnoreCase(tr.instanceUrl)) {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
@@ -76,7 +76,7 @@ public class AuthenticatorService extends Service {
     public static final String KEY_DISPLAY_NAME = "display_name";
     public static final String KEY_PHOTO_URL = "photoUrl";
     public static final String KEY_THUMBNAIL_URL = "thumbnailUrl";
-    public static final String KEY_ADDL_PARAMS = "addlParams";
+   // public static final String KEY_ADDL_PARAMS = "addlParams";
 
     private Authenticator getAuthenticator() {
         if (authenticator == null)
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -790,6 +801,19 @@ public LoginOptions(String loginUrl, String passcodeHash, String oauthCallbackUr
...
         }
 
+        public LoginOptions(String loginUrl, String passcodeHash, String oauthCallbackUrl,
+                            String oauthClientId, String[] oauthScopes, String clientSecret, String jwt,
+                            HashMap<String,String> additionalParameters) {
+            this(loginUrl, passcodeHash, oauthCallbackUrl, oauthClientId, oauthScopes,clientSecret,jwt);
+            this.setAdditionalParameters(additionalParameters);
+        }
+
+        public void setAdditionalParameters(HashMap<String,String> additionalParameters) {
```
skyler-b: Should this take a Map as an argument (not a HashMap)?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -790,6 +801,19 @@ public LoginOptions(String loginUrl, String passcodeHash, String oauthCallbackUr
...
+            this.additionalParameters = additionalParameters;
+            bundle.putSerializable(AuthenticatorService.KEY_ADDL_PARAMS,additionalParameters);
```
skyler-b: If this bundle addition is required, should we not expose 'additionalParameters' publicly, so that we make sure all writes to it come through this method?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
trooper2013: Needs to be  a serializable.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
trooper2013: @skyler-b are you looking for a getter in loginoptions? Or a way  set it directly through the SDK Manager?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
skyler-b: Then should additionalParameters be declared as a HashMap?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
skyler-b: The first time I was testing this out, I just set the field directly, but it wasn't working right, then I found the setter which did this bundle step. If it's required to go through this path, I was just suggesting we lower the visibility of the field.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
trooper2013: Agree. Good catch.Will change it to HashMap.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
trooper2013: Tried to be consistent with the definition of this already designed class. Clearly this problem exists with all the class members. Will look into the overall design and see if I can clean this class up. I don't see why the bundle cannot be created on demand(asBundle).

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
```diff
@@ -237,7 +247,7 @@ public Bundle getAuthToken(
             }
             final Bundle resBundle = new Bundle();
             try {
-                final TokenEndpointResponse tr = OAuth2.refreshAuthToken(HttpAccess.DEFAULT, new URI(loginServer), clientId, refreshToken, clientSecret);
+                final TokenEndpointResponse tr = OAuth2.refreshAuthToken(HttpAccess.DEFAULT, new URI(loginServer), clientId, refreshToken, clientSecret,addlParamsMap);
```
wmathurin: Are the additional parameters needed for refreshing the token or just when initiating login? If we only need them for login, then I don't think we need to change the AuthenticatorService.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -389,6 +392,12 @@ public Bundle createNewAccount(String accountName, String username, String refre
             }
         }
         Account acc = new Account(accountName, getAccountType());
+
+        String encodedAdlParams = OAuth2.parameterMapToString(loginOptions.additionalParameters);
```
wmathurin: What we store in the account is stuff we got from the server, why would we store the additional parameters if they are only used when calling getAuthorizationUrl?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
trooper2013: Its required for both. This is what i got from the W-3608610.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
trooper2013: Will remove them from the account.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
trooper2013: @skyler-b Please take a look at Login options. I would have preferred to make all members of the class private, but the change made now is  a good compromise. I have deferred the creation of the bundle to on-demand. Directly manipulating the members or using setters yields the same result.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -149,38 +150,123 @@
...
-            return getAuthorizationUrl(loginServer, clientId, callbackUrl, scopes, clientSecret, displayType);
+            return getAuthorizationUrl(loginServer, clientId, callbackUrl, scopes, clientSecret, displayType, addlParams);
         }
         final StringBuilder sb = new StringBuilder(instanceURL);
         sb.append(FRONTDOOR);
         sb.append(SID).append(EQUAL).append(accessToken);
         sb.append(AND).append(RETURL).append(EQUAL).append(Uri.encode(getAuthorizationUrl(loginServer,clientId,callbackUrl,
                 scopes, clientSecret, displayType).toString()));
+
+        if(addlParams != null && addlParams.size() > 0) {
```
wmathurin: The if (addParams != null) {...} is duplicated in two methods - maybe we could have it extracted to its own helper method?

wmathurin: Looks good otherwise.

PR #1349, opened 2017-02-09
Adding Support for jCenter Maven Publishing

bhariharan:


build.gradle
```diff
@@ -5,18 +5,22 @@ buildscript {
...
   dependencies {
     classpath 'com.android.tools.build:gradle:2.2.2'
+    classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
   }
 }
 
-allprojects {
-  group = 'com.salesforce.androidsdk'
+plugins {
+  id "com.jfrog.bintray" version "1.7.3"
```
bhariharan: This plugin and the one above help us with publisher hooks in `Gradle` scripts.

libs/SalesforceAnalytics/build.gradle
```diff
@@ -52,3 +53,68 @@ afterEvaluate {
   // from timing out since executing tests are printed to the console.
   tasks.matching { it.name ==~ /^connected.*AndroidTest$/ }*.logging*.setLevel(LogLevel.INFO)
 }
+
+install {
```
bhariharan: Packaging and uploading multiple libraries/artifacts from a single repo is very tricky. Hence, even though a lot of this seems repeated, we need them in individual `Gradle` files. You'll notice `artifactId` is different, and there are some other differences too.

libs/SalesforceAnalytics/build.gradle
```diff
@@ -52,3 +53,68 @@ afterEvaluate {
...
+                    url 'https://github.com/forcedotcom/SalesforceMobileSDK-Android'
+
+                }
+            }
+        }
+    }
+}
+
+task sourcesJar(type: Jar) {
+    from android.sourceSets.main.java.srcDirs
```
bhariharan: Sources are required for it to be accepted in `jCenter`.

libs/SalesforceAnalytics/build.gradle
```diff
@@ -52,3 +53,68 @@ afterEvaluate {
...
+artifacts {
+    archives sourcesJar
+}
+
+bintray {
+    user = System.getenv('BINTRAY_USER')
+    key = System.getenv('BINTRAY_KEY')
+    configurations = ['archives']
+    pkg {
+        repo = 'salesforcemobilesdk'
```
bhariharan: I created our own repo on `Bintray`.

libs/SalesforceAnalytics/build.gradle
```diff
@@ -52,3 +53,68 @@ afterEvaluate {
...
+        name = 'salesforce-analytics'
+        userOrg = 'forcedotcom'
```
bhariharan: I created our own org on `Bintray` too, instead of just my user.

libs/SalesforceAnalytics/build.gradle
```diff
@@ -52,3 +53,68 @@ afterEvaluate {
...
+        websiteUrl = 'https://github.com/forcedotcom/SalesforceMobileSDK-Android'
+        issueTrackerUrl = 'https://github.com/forcedotcom/SalesforceMobileSDK-Android/issues'
+        publicDownloadNumbers = true
+        githubRepo = 'forcedotcom/SalesforceMobileSDK-Android'
+        githubReleaseNotesFile = 'README.md'
+        licenses = ['Apache-2.0']
+        labels = ['android', 'salesforce', 'mobilesdk']
+        version {
+            name = '5.0.0'
+            released  = new Date()
```
bhariharan: Automatically insert publish time.

libs/SalesforceAnalytics/build.gradle
```diff
@@ -52,3 +53,68 @@ afterEvaluate {
...
+    from android.sourceSets.main.java.srcDirs
+    classifier = 'sources'
+}
+
+artifacts {
+    archives sourcesJar
+}
+
+bintray {
+    user = System.getenv('BINTRAY_USER')
```
bhariharan: This is setup such that it can be pulled from the env variables when you run the script. I have it setup on my `.bash_profile` for instance.

bhariharan:
I tested it by publishing the artifacts. It works great! Some observations:

- It works out the chain of dependency by itself magically. So, if you include `SmartSync`, it automatically pulls down the 3 lower libraries that it depends on.
- `compile 'com.salesforce.mobilesdk:SmartSync:5.0.0'` is basically all you need to add to your app's `build.gradle`. Since we publish a `JAR` of sources too (as required by `jCenter`), debugging and viewing sources is seamless.
- `SalesforceHybrid` and `SalesforceReact` don't work, because `Cordova` and `React` aren't published on `jCenter`. It complains of missing libraries. Since that doesn't work, we won't be publishing `SalesforceHybrid` and `SalesforceReact`. Bummer, but I hope to file issues with `Cordova` and `React` to get this going at least in the future.

I'll demo it to the team tomorrow @wmathurin @trooper2013

wmathurin:
Since SalesforceHybrid can't be published, does it mean that we can't change the way hybrid apps work and that we are stuck with the library projects embedded in the plugin?

bhariharan:
@wmathurin I have tried to hook it up yet, but my gut feeling tells me we're stuck for now (for `SalesforceHybrid`), but I am planning on sending a pull request to `Cordova` with a `Gradle` task to upload to `MavenCentral`. If they accept my PR and publish, we can switch the plugin over completely.

bhariharan:
test

bhariharan:
build

bhariharan:
Build is failing only because `Fastlane` isn't available on `master`. I have tested locally.

PR #1358, opened 2017-02-24
New sync down target for parent-children

wmathurin:
The new sync down target ParentChildrenSyncDownTarget is initialized with information about parent / children (fields to get / soup to use / relationship fields in the soup etc).
* Unlike SoqlSyncDownTarget (its superclass), it doesn't take a full SOQL query but instead only expects the portion of it to use in the where clause. It generates the SOQL query to get parents with their children, and save them to separate soups.
* Leave-if-changed merge mode is supported. If a parent or any of its children is locally modified, then the whole parent-children set will be left alone.
* Re-sync is also supported although, it will only download parent-children sets where the parent changed (because SOQL cannot express the more complex query).
* Clean ghosts will clean parent and children soups and should only be used if the records in those soups come exclusively through a parent-children sync down.


hybrid/HybridSampleApps/NoteSync/src/com/salesforce/samples/notesync/ContentSoqlSyncDownTarget.java
```diff
@@ -35,7 +35,7 @@
 import com.salesforce.androidsdk.rest.RestResponse;
 import com.salesforce.androidsdk.smartsync.manager.SyncManager;
 import com.salesforce.androidsdk.smartsync.util.Constants;
-import com.salesforce.androidsdk.smartsync.util.SoqlSyncDownTarget;
```
wmathurin: The util package was getting crowded so I moved all the targets to a new package.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/JSONObjectHelper.java
```diff
@@ -139,4 +139,21 @@ public static String optString(JSONObject obj, String key, String fallback) thro
...
 
+	/**
+	 * Given an array of JSONObject, return list made of values for key
+	 * @param jsonObjects
+	 * @param key
+	 * @param <T>
+	 * @return
+	 * @throws JSONException
+	 */
+	@SuppressWarnings("unchecked")
```
wmathurin: That method used to be somewhere else

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -296,39 +297,17 @@ public void cleanResyncGhosts(long syncId) throws JSONException {
...
-         * and removes the list of IDs that are still present on the server.
-         */
-        final Set<String> remoteIds = target.getListOfRemoteIds(this, localIds);
-        if (remoteIds != null) {
-            localIds.removeAll(remoteIds);
-        }
-
-        // Deletes extra IDs from SmartStore.
-        int localIdSize = localIds.size();
+        // Ask target to clean up ghosts
```
wmathurin: Delegating to the target (required for advanced targets like ParentChildrenSyncDownTarget).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/MruSyncDownTarget.java
```diff
@@ -118,7 +118,7 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
     }
 
     @Override
-    public Set<String> getListOfRemoteIds(SyncManager syncManager, Set<String> localIds) {
```
wmathurin: That method does not need to be public.

wmathurin: Not catching the exception - it should bubble up as far as possible

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java
```diff
@@ -98,21 +105,20 @@ private void addSpecialFieldsIfRequired() {
...
-    private JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String queryRun) throws IOException, JSONException {
-        String queryToRun = maxTimeStamp > 0 ? SoqlSyncDownTarget.addFilterForReSync(queryRun, getModificationDateFieldName(), maxTimeStamp) : queryRun;
-        RestRequest request = RestRequest.getRequestForQuery(syncManager.apiVersion, queryToRun);
+    protected JSONArray startFetch(SyncManager syncManager, String query) throws IOException, JSONException {
+        RestRequest request = RestRequest.getRequestForQuery(syncManager.apiVersion, query);
         RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
-        JSONObject responseJson = response.asJSONObject();
-        JSONArray records = responseJson.getJSONArray(Constants.RECORDS);
+        JSONObject responseJson = getResponseJson(response);
+        JSONArray records = getRecordsFromResponseJson(responseJson);
```
wmathurin: Extraction of records is moved to a separate method.
In the case of ParentChildrenSyncDownTarget, we are getting parent and children records together and we needed some control over the parsing without having to override the entire method,

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java
```diff
@@ -122,55 +128,68 @@ private JSONArray startFetch(SyncManager syncManager, long maxTimeStamp, String
         return records;
     }
 
+    protected JSONObject getResponseJson(RestResponse response) throws IOException {
+        JSONObject responseJson;
```
wmathurin: In case of bad soql for instance, we get a json object with the error.
That would get lost. Now the error message from the server will end up in the exception.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java
```diff
@@ -183,7 +202,15 @@ public static String addFilterForReSync(String query, String modificationFieldDa
...
-        return query;
+    public String getQuery() {
+        return getQuery(0);
+    }
+
+    /**
+     * @return soql query for this target
+     * @param maxTimeStamp
+     */
+	public String getQuery(long maxTimeStamp) {
```
wmathurin: The subclass is building the SOQL query and so does need to rely on some nasty regexp replace to inject the time stamp predicate

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncDownTarget.java
```diff
@@ -91,6 +94,12 @@ public SyncDownTarget() {
         super();
     }
 
+
+    public SyncDownTarget(String idFieldName, String modificationDateFieldName) {
```
wmathurin: Previously id field name and modification date field name could be specified when building from json but not with other constructors

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncDownTarget.java
```diff
@@ -128,14 +137,41 @@ public JSONObject asJSON() throws JSONException {
...
 
+
+    /**
+     * Delete from local store records that a full sync down would no longer download
+     * @param syncManager
+     * @param soupName
+     * @return
+     * @throws JSONException, IOException
+     */
+    public int cleanGhosts(SyncManager syncManager, String soupName) throws JSONException, IOException {
```
wmathurin: New method called from SyncManager's cleanResyncGhosts

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/ChildrenInfo.java
```diff
@@ -0,0 +1,83 @@
...
+
+import com.salesforce.androidsdk.util.JSONObjectHelper;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Simple object to capture details of children in parent-child relationship
+ */
+public class ChildrenInfo extends ParentInfo {
```
wmathurin: Class used to capture details about children for parent-children sync down

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/ParentInfo.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.salesforce.androidsdk.util.JSONObjectHelper;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Simple object to capture details of parent in parent-child relationship
+ */
+public class ParentInfo {
```
wmathurin: Class used to capture details about parent for parent-children sync down

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTargetTest.java
```diff
@@ -0,0 +1,78 @@
...
+import com.salesforce.androidsdk.smartsync.util.Constants;
+import com.salesforce.androidsdk.smartsync.util.SOQLBuilder;
+
+import java.util.Date;
+
+/**
+ * Test class for SoqlSyncDownTarget
+ *
+ */
+public class SoqlSyncDownTargetTest extends SyncManagerTestCase {
```
wmathurin: Those tests are not new - they used to be in SyncManagerTest - but because they call methods that are now protected, I had to move them to the smartsync.target package in a new suite.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncTest.java
```diff
@@ -0,0 +1,909 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.SortedSet;
+
+/**
+ * Test class for ParentChildrenSyncDownTarget
+ *
+ */
+public class ParentChildrenSyncTest extends SyncManagerTestCase {
```
wmathurin: All the tests for parent-children sync down.
Some low level tests (checking the SOQL generated etc).
Tests doing regular sync down, sync down with leave-if-changed, re-sync, clean ghosts.

wmathurin:
Could I get a review? Thanks.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncDownTarget.java
```diff
@@ -0,0 +1,412 @@
...
+                soupName, SmartStore.SOUP_ENTRY_ID,
+                getDirtyRecordIdsSql(soupName, SmartStore.SOUP_ENTRY_ID)
+        );
+    }
+
+    @Override
+    public void saveRecordsToLocalStore(SyncManager syncManager, String soupName, JSONArray records) throws JSONException {
+        // NB: method is called during sync down so for this target records contain parent and children
+        SmartStore smartStore = syncManager.getSmartStore();
+        synchronized(smartStore.getDatabase()) {
```
bhariharan: Were we always synchronizing on the DB? Should we synchronize on `SmartStore` itself?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/RefreshSyncDownTarget.java
```diff
@@ -211,26 +213,20 @@ private JSONArray fetchFromServer(SyncManager syncManager, List<String> ids, Lis
...
-            Log.e(TAG, "IOException thrown while fetching records", e);
-        } catch (JSONException e) {
-            Log.e(TAG, "JSONException thrown while fetching records", e);
+        List<String> localIdsList = new ArrayList<>(localIds);
+        int sliceSize = getCountIdsPerSoql();
+        int countSlices = (int) Math.ceil((double) localIds.size() / sliceSize);
+        for (int slice = 0; slice < countSlices; slice++) {
+            List<String> idsToFetch = localIdsList.subList(slice * sliceSize, Math.min(localIdsList.size(), (slice + 1) * sliceSize));
+            JSONArray records = fetchFromServer(syncManager, idsToFetch, Arrays.asList(getIdFieldName()), 0 /* get all */);
+            remoteIds.addAll(parseIdsFromResponse(records));
```
bhariharan: Is the consuming code handling the exceptions since they're being thrown now?

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncTest.java
```diff
@@ -0,0 +1,909 @@
+package com.salesforce.androidsdk.smartsync.target;
```
bhariharan: Minor, but maybe move this line below the copyright like the other files.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTargetTest.java
```diff
@@ -0,0 +1,78 @@
+package com.salesforce.androidsdk.smartsync.target;
```
bhariharan: Minor, but maybe move this line below the copyright like the other files.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncDownTarget.java
wmathurin: We were. We are doing it in other places to.
In there is that comment https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java#L259

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/RefreshSyncDownTarget.java
wmathurin: cleanResyncGhosts was always throwing a bunch of different exceptions - so I expect the app to be handling them

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTargetTest.java
wmathurin: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1358/commits/e3affe5f45c9175256398cfdaab00ffb37820146)

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncTest.java
wmathurin: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1358/commits/e3affe5f45c9175256398cfdaab00ffb37820146)

PR #1359, opened 2017-02-28
Switching to Cordova 6.1.2

bhariharan:
This PR:
- switches to using `Cordova 6.1.2` instead of `Cordova 6.1.0`.
- Fetches `Cordova` from `jCenter` instead of consuming it as a submodule.


bhariharan:
build

bhariharan:
I had forgotten to include `cordova.js` and update the symlinks for the hybrid apps and the test suite in the workspace. Should be good to go now.

bhariharan:
build

PR #1361, opened 2017-03-01
Better fix for class path of Maven plugins

bhariharan:


bhariharan:
@wmathurin This should fix the `hybrid` packaging issue. The only clean fix is to ensure each library project's `build.gradle` has the plugins it uses in the classpath, as opposed to relying on the top level `build.gradle`, which won't be there in generated projects.

build.gradle
```diff
@@ -5,14 +5,9 @@ buildscript {
 
   dependencies {
     classpath 'com.android.tools.build:gradle:2.2.3'
```
bhariharan: I kept the basic `Android` plugin at the top level, because the sample apps in the repo use it.

bhariharan:
build

PR #1369, opened 2017-03-14
Updated build.gradle to specify 0.35.0 *or greater*

rgoldiez:
Small update so that newer versions of react native can be used without having to modify build.gradle for each version.

### DIFF
--- a/libs/SalesforceReact/build.gradle
+++ b/libs/SalesforceReact/build.gradle
@@ -5,7 +5,7 @@ apply plugin: 'com.jfrog.bintray'
 dependencies {
   compile project(':libs:SmartSync')
   compile 'com.android.support:appcompat-v7:23.0.1'
-  compile 'com.facebook.react:react-native:0.35.0'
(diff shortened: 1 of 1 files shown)


bhariharan:
@rgoldiez But we want to tie it to the latest version of `ReactNative` that we've tested on and can support as opposed to automatically pulling in the latest.

rgoldiez:
🙁 I understand... was just looking for an easier way to run with more current versions of React Native, without having to fork or modify the repo.  On the iOS side, I simply `npm install --save react-native@[version]` and then do a `pod update` and I'm good.  I was trying to make the Android side just as easy.

bhariharan:
@rgoldiez I'm no `Gradle` expert, but maybe we could pull the version from an overridable constant that the app could override in its `build.gradle`. I'll look into this in a future release, but close this for now. Does that sound reasonable?

rgoldiez:
Sounds good!

PR #1370, opened 2017-03-14
Bridge fixed to accepted objects in body

rgoldiez:
Updated the sendRequest method and associated helper methods to accept objects in bodies instead of simply strings.  Required for posting Feed Elements and to other standard (and custom) endpoints.


wmathurin:
Awesome, thanks.

PR #1372, opened 2017-03-15
Adding the ability to override AILTN app name

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -164,6 +165,26 @@ public static boolean hasInstance() {
     }
 
     /**
+     * Sets the app name to be used by the analytics framework.
+     *
+     * @param appName App name.
+     */
+    public static void setAiltnAppName(String appName) {
```
bhariharan: The getter, setter and the variable itself have to be static, because the analytics framework initializes this very early in the lifecycle and does it only once per app.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTest.java
```diff
@@ -0,0 +1,194 @@
...
+                new KeyImpl(), MainActivity.class);
+        SalesforceSDKTestManager.setAiltnAppName(TEST_APP_NAME);
+        SalesforceSDKTestManager.getInstance().setIsTestRun(true);
+        compareAiltnAppNames(TEST_APP_NAME);
+    }
+
+    /**
+     * Test for default analytics app name.
+     */
+    public void testDefaultAiltnAppName() {
```
bhariharan: Test without setting - for default value.

PR #1373, opened 2017-03-16
Bumping up Gradle and build tools versions to latest recommended versions

bhariharan:


bhariharan:
CI failure is because the `License Agreement` for the new build tools haven't been accepted. Build works fine locally. Merging.

PR #1380, opened 2017-03-21
Fixing content-length bug with GZIP on OkHttp

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AILTNPublisher.java
```diff
@@ -129,6 +138,28 @@ public boolean publishLogLines(JSONArray logLines) {
...
+        return new RequestBody() {
+
+            @Override
+            public MediaType contentType() {
+                return requestBody.contentType();
+            }
+
+            @Override
+            public long contentLength() {
+                return buffer.size();
```
bhariharan: We need to buffer the whole thing first to get the exact size to set `Content-Length` header. Some servers are sensitive to this and return `411` if it's missing, such as our community endpoints. It works fine against our org endpoints though.

bhariharan:
@wmathurin Nope - `iOS` was working with communities and `Android` wasn't. That's what exposed this bug. `iOS` has `Content-Length` set already.

PR #1382, opened 2017-03-23
Fixing passcode change flow

skyler-b:


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -397,9 +397,7 @@ public boolean isLocked() {
      * @param ctx
      */
     public void lock(Context ctx) {
-        locked = true;
```
skyler-b: the first problem i found in the passcode change flow is that these two lines were happening in 'lock()' but not in 'showLockActivity' which is what the passcode change flow calls. I've just moved them to happen in that method.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -511,22 +518,23 @@ public void setMinPasscodeLength(int minPasscodeLength) {
...
 
     public boolean shouldLock() {
         return timeoutMs > 0 && now() >= (lastActivity + timeoutMs);
     }
 
     public void showLockActivity(Context ctx, boolean changePasscodeFlow) {
-        if (ctx == null) {
-        	return;
-        }
+        locked = true;
```
skyler-b: after this move, the passcode change dialog actually started showing up

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -511,22 +518,23 @@ public void setMinPasscodeLength(int minPasscodeLength) {
              * triggered later from OAuthWebviewHelper.
              */
             if (hasStoredPasscode(SalesforceSDKManager.getInstance().getAppContext())) {
-        		showLockActivity(SalesforceSDKManager.getInstance().getAppContext(),
```
skyler-b: added ability to pass in the context, as well as return a boolean, will explain this in the next file.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -538,7 +538,7 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
                 final PasscodeManager passcodeManager = mgr.getPasscodeManager();
                 passcodeManager.storeMobilePolicyForOrg(account, id.screenLockTimeout * 1000 * 60, id.pinLength);
                 passcodeManager.setTimeoutMs(id.screenLockTimeout * 1000 * 60);
-                passcodeManager.setMinPasscodeLength(id.pinLength);
+                boolean changeRequired = passcodeManager.setMinPasscodeLength((Activity) getContext(), id.pinLength);
```
skyler-b: We need to pass in the Activity context here, so that LoginActivity is used, and so that LoginActivity can get onActivityResult from the PasscodeActivity. Without this, the application context is used, and we don't get a callback when the passcode is done.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -551,9 +551,9 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
                  */
                 if (!passcodeManager.hasStoredPasscode(mgr.getAppContext())) {
```
skyler-b: the reason the boolean was added is because this line, because technically, you do have a stored passcode, so it should not enter this 'if', we anyways have already set the lock screen

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -551,9 +551,9 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
...
                     // This will bring up the create passcode screen - we will create the account in onResume
-                    mgr.getPasscodeManager().setEnabled(true);
-                    mgr.getPasscodeManager().lockIfNeeded((Activity) getContext(), true);
```
skyler-b: lock if needed would not run anyways because there wasn't actually a timeout.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -551,9 +551,9 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
...
-                } else {
+                    passcodeManager.setEnabled(true);
+                    passcodeManager.lockIfNeeded((Activity) getContext(), true);
+                } else if (!changeRequired) { // If a passcode change is required, the lock screen will have already been set in setMinPasscodeLength
```
skyler-b: previously, our code was falling into this 'else' which would add our account and kill the login activity (callback.finish()) before the passcode had been entered, and we again fail to get the activity result.

skyler-b:
@bhariharan @chrlie

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -511,22 +518,23 @@ public void setMinPasscodeLength(int minPasscodeLength) {
...
-        				true);
+                setEnabled(true);
+                showLockActivity(ctx, true);
+                return true;
```
bhariharan: @skyler-b Wouldn't this lead to `minPasscodeLength` not being updated? Previously, after the lock activity is shown, control would return here and we would execute `storeMobilePolicy` below. Now, since the code returns here, the min length won't be updated, right?

skyler-b: great catch, you're right, storeMobilePolicy is not being executed in this case, I'll fix that. 

`minPasscodeLength` was still being updated (looks like it's updated twice), maybe I can clean that up too.

bhariharan: Yeah - this stuff is quite convoluted. That's why I have avoided changing code here since I wrote it years ago ;-)

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -511,22 +519,23 @@ public void setMinPasscodeLength(int minPasscodeLength) {
              * triggered later from OAuthWebviewHelper.
              */
             if (hasStoredPasscode(SalesforceSDKManager.getInstance().getAppContext())) {
```
bhariharan: @skyler-b This should be `ctx` too, right?

bhariharan: `hasStoredPasscode(ctx)`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -511,22 +519,23 @@ public void setMinPasscodeLength(int minPasscodeLength) {
...
-        		showLockActivity(SalesforceSDKManager.getInstance().getAppContext(),
-        				true);
+                setEnabled(true);
```
bhariharan: Why do we need `setEnabled(true)` here? If it enters this code, passcode is stored and hence it is already enabled right?

bhariharan:
@skyler-b Few more comments.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
skyler-b: This one and `storeMobilePolicy` did not really _need_ the Activity context, so I had just left it with the app context. I'll move it over, no problem.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
skyler-b: You're right, this is not needed.

bhariharan:
I'll merge this once the CI run is complete.

PR #1383, opened 2017-03-23
Comment fixes and using batch in cleanup

wmathurin:


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -1037,12 +1034,22 @@ private void cleanup() {
         try {
             RestResponse response = restClient.sendSync(RestRequest.getRequestForSearch(TestCredentials.API_VERSION, "find {" + ENTITY_NAME_PREFIX + "}"));
             JSONArray jsonResults = response.asJSONObject().getJSONArray("searchRecords");
+            List<RestRequest> requests = new ArrayList<>();
             for (int i = 0; i < jsonResults.length(); i++) {
                 JSONObject jsonResult = jsonResults.getJSONObject(i);
                 String objectType = jsonResult.getJSONObject("attributes").getString("type");
                 String id = jsonResult.getString("Id");
-                restClient.sendSync(RestRequest.getRequestForDelete(TestCredentials.API_VERSION, objectType, id));
+                RestRequest deleteRequest = RestRequest.getRequestForDelete(TestCredentials.API_VERSION, objectType, id);
```
wmathurin: That's what I'm doing on the iOS side so I wanted to keep both implementations in sync

PR #1385, opened 2017-03-27
Added an optional Bundle of extras to switchToUser

mahendrn:
This Bundle of extras could be used to pass additional
information during user switch. It will be part of the
broadcast intent sent after the user switch is completed.


mahendrn:
SmartSyncJSTest#testSyncDownToGlobalStoreNamed is failing in the CI run, but passes for me locally.

PR #1387, opened 2017-04-03
Persisting last server selection by user

bhariharan:


bhariharan:
Yes - I'll add tests.

bhariharan:
@wmathurin Actually we've already got tests for `get` and `set` server selection [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/LoginServerManagerTest.java#L114). All I'm doing is adding persistence for those selections even when the app is swapped out of memory. I can't think of a good way to simulate that scenario in a unit test. I did test manually though.

bhariharan:
Unrelated single test failure in the hybrid tests (flapper).

PR #1388, opened 2017-04-04
Upgrading to ReactNative 0.43.0

bhariharan:


bhariharan:
`SalesforceReact` compiles fine.

install.sh
```diff
@@ -7,4 +7,4 @@ git submodule sync
 git submodule update
 
 # for react-native
-npm install react-native@0.35.0 react@15.3.1 --silent
```
wmathurin: Looking at https://github.com/facebook/react-native/blob/v0.43.0/package.json#L129,
I think it should be react@16.0.0-alpha.6

install.vbs
```diff
@@ -72,7 +72,7 @@ Call CopySymlinkFiles()
 
 ' Getting npm dependencies (for react-native)
 WScript.Echo "Getting npm dependencies"
-intReturnVal = objShell.Run("npm install react-native@0.35.0 react@15.3.1 --silent", 1, True)
```
wmathurin: Looking at https://github.com/facebook/react-native/blob/v0.43.0/package.json#L129,
I think it should be react@16.0.0-alpha.6

bhariharan:
@wmathurin Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1388/commits/7e34294b1bc7d78d2c9de94f8815e341c8c9b234).

wmathurin:
Could you try running the sample app with 0.43?

Simply edit package.json locally (point to correct react and react native and to your SalesforceMobileSDK-Android on line https://github.com/forcedotcom/SmartSyncExplorerReactNative/blob/master/package.json#L14).

If it builds and runs then we are golden.

bhariharan:
@wmathurin We should be good to go here. `TemplateApp` and `SmartSyncExplorerReactNative` are working fine with these changes.

PR #1389, opened 2017-04-07
Latest shared

wmathurin:
+ small cleanup

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -68,7 +68,6 @@ import java.util.concurrent.Executors;
 public class SyncManager {
 
     // Constants
-    public static final int PAGE_SIZE = 2000;
     private static final int UNCHANGED = -1;
     private static final String TAG = "SyncManager";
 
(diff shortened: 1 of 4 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -68,7 +68,6 @@
 public class SyncManager {
 
     // Constants
-    public static final int PAGE_SIZE = 2000;
```
wmathurin: Not used

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncTarget.java
```diff
@@ -196,7 +196,6 @@ public void saveRecordsToLocalStore(SyncManager syncManager, String soupName, JS
                 smartStore.beginTransaction();
                 for (int i = 0; i < records.length(); i++) {
                     JSONObject record = new JSONObject(records.getJSONObject(i).toString());
-                    cleanRecord(record);
```
wmathurin: The next method called (cleanAndSaveInSmartStore) does a cleanRecord

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTestCase.java
```diff
@@ -148,6 +148,40 @@ protected String createLocalId() {
     }
```
wmathurin: Method moved in file

PR #1390, opened 2017-04-12
Moved deleting InstanceID to background thread

schilakala:


schilakala:
@bhariharan Fix for the push notification on unregister when logging out from passcode screen

PR #1392, opened 2017-04-13
Better revoke handling

wmathurin:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -708,15 +708,16 @@ private void refreshAccessToken() throws IOException {
...
+				if (newAuthToken == null || authTokenProvider.getInstanceUrl() == null) {
+                    throw new RefreshTokenRevokedException("Could not refresh token");
                 }
 
+                // Use new token
+                setAuthToken(newAuthToken);
+
                 // Check if the instanceUrl changed
                 String instanceUrl = authTokenProvider.getInstanceUrl();
-                if (instanceUrl == null) {
```
wmathurin: After a revoke, authTokenProvider.getInstanceUrl() would be null, so the old code would throw a IOException "Instance URL is null".
Now it throws a RefreshTokenRevokedException (which is a subclass of IOException - so backward compatible).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -257,6 +257,10 @@ public void run() {
                             break;
                     }
                     updateSync(sync, SyncState.Status.DONE, 100, callback);
+                } catch (RestClient.RefreshTokenRevokedException re) {
```
wmathurin: The catch (Exception e) block that follows does an updateSync, which will get a SmartStore instance (while logging out is taking place) - and then terrible things will happen - a database gets created for the outgoing user that won't be openable by the returning user causing the app to crash at logout.

This is a somewhat superficial fix - SmartStore should not return an instance for an outgoing user.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -708,15 +708,16 @@ private void refreshAccessToken() throws IOException {
             // Then let's try to get a new authToken
             if (authTokenProvider != null) {
                 final String newAuthToken = authTokenProvider.getNewAuthToken();
-                if (newAuthToken != null) {
-                    setAuthToken(newAuthToken);
+
+				if (newAuthToken == null || authTokenProvider.getInstanceUrl() == null) {
+                    throw new RefreshTokenRevokedException("Could not refresh token");
```
bhariharan: We need to let internal teams know to catch this exception now instead of `IOException`.

native/NativeSampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/ui/MainActivity.java
```diff
@@ -132,6 +132,15 @@ protected void onCreate(Bundle savedInstanceState) {
 	}
 
 	@Override
+	protected void logoutCompleteActions() {
+		super.logoutCompleteActions();
```
wmathurin: If there is a revoke, the client manager will call logout without passing a front activity and without requesting login to be shown. So the login screen will only be shown the next time the user changes activities.

wmathurin: Maybe that code belongs in the SDK Salesforce*Activity classes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
wmathurin: It's a subclass of IOException so they don't need to change their code.

native/NativeSampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/ui/MainActivity.java
bhariharan: Yeah, I think that's a good idea. Should we put this in `SalesforceDroidGapActivity` as well then?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
bhariharan: Yes - I just noticed :-)

native/NativeSampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/ui/MainActivity.java
wmathurin: @bhariharan do you think that code should move to the super classes (the SDK Salesforce*Activity.java)?

bhariharan: Yes - I think that's a good idea to move it to the super classes. We also need to check if we should add this to `SalesforceDroidGapActivity` - it inherits from `CordovaActivity` instead of ours.

wmathurin: Maybe we don't really need to pass the activity to logout and have the SalesforceSDKManager finish activities.

wmathurin: Let me merge that PR with the recreate in the sample app code.
I'll do more testing before making the change in the SDK activity classes to see if it doesn't have unforeseen side effects in hybrid / react native apps.

bhariharan: Yeah - maybe make that change first thing in `5.2` instead of now.

PR #1394, opened 2017-04-14
Fixing links in Windows

wmathurin:


wmathurin:
When I removed repoUtils.js from shared, I broke build_windows_file_list.js so the links file has been off since 5.0. I guess few developers actually clone the repo (on Windows at least).

PR #1415, opened 2017-05-19
Adding user agent marker for related records

bhariharan:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncTargetHelper.java
```diff
@@ -60,10 +62,10 @@
 
     public static void saveRecordTreesToLocalStore(SyncManager syncManager, SyncTarget target, ParentInfo parentInfo, ChildrenInfo childrenInfo, JSONArray recordTrees) throws JSONException {
         SmartStore smartStore = syncManager.getSmartStore();
+        SmartSyncSDKManager.getInstance().registerUsedAppFeature(FEATURE_RELATED_RECORDS);
```
bhariharan: @wmathurin Is this the best place to add this? I figured this would cover the `SyncUp` and `SyncDown` targets.

wmathurin: It's only used during sync down.
I recommend doing the registerUsedAppFeature call from the constructors of parent children sync down and sync up targets.

bhariharan:
@wmathurin Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1415/commits/0dbd2e125fdf250af58bc77b823c134c52b6ca11).

PR #1416, opened 2017-05-23
Fixing incorrect code 

wmathurin:
NB: runs correctly if children and parent are using standard ID field => for that reason it went unnoticed


bhariharan:
Should we add tests for custom `id` field and custom `lastModifiedDate` field?

wmathurin:
There are tests for custom id and mod date field, but not exercising that particular method.

PR #1418, opened 2017-05-26
Treat 15 and 18 char community IDs the same

skyler-b:
Avoids duplicate database creation

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
@@ -59,7 +59,7 @@ public class UserAccount {
 	public static final String CLIENT_ID = "clientId";
 	public static final String COMMUNITY_ID = "communityId";
 	public static final String COMMUNITY_URL = "communityUrl";
-	public static final String INTERNAL_COMMUNITY_ID = "000000000000000000";
+	public static final String INTERNAL_COMMUNITY_ID = "000000000000000";
 	public static final String INTERNAL_COMMUNITY_PATH = "internal";
     public static final String EMAIL = "email";
     public static final String FIRST_NAME = "first_name";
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -452,7 +452,7 @@ public String getUserLevelStoragePath() {
 	 */
 	public String getCommunityLevelStoragePath() {
 		String leafDir = INTERNAL_COMMUNITY_PATH;
-		if (!TextUtils.isEmpty(communityId) && !communityId.equals(INTERNAL_COMMUNITY_ID)) {
+		if (!TextUtils.isEmpty(communityId) && !communityId.startsWith(INTERNAL_COMMUNITY_ID)) {
```
bhariharan: @skyler-b What if the `communityId` starts with 15 chars of 0s and the last 3 are non-zero chars? I know it's highly unlikely, but our API has always expected and accepted only 18 char IDs. Maybe we should doc it better?

skyler-b: Actually our app started normalizing 15 char community IDs to have AAA as the last 3 characters. That's how I found this problem, so not too unlikely 😉 

I thought this would be a safer approach since (I think) we try to make 15 or 18 char IDs interoperable everywhere. I believe through any server API, passing in 15 vs 18 char is required to work the same.

bhariharan: It's meant to be interoperable on the server because that's internal code. 15-char ID is an internal only representation of IDs. External facing entity IDs are required to be 18-char only, so IDs coming back from a public API endpoint, such as the token endpoint should be 18-char only.

PR #1421, opened 2017-05-30
Adding persistence of logging preferences

bhariharan:
This stores the log levels and file size per component in `SharedPreferences`, so that these preferences are used even if the app is swapped out of memory or the device is restarted.

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/FileLogger.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/FileLogger.java
@@ -27,6 +27,7 @@
 package com.salesforce.androidsdk.analytics.logger;
 
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.text.TextUtils;
 import android.util.Log;
 
@@ -47,9 +48,12 @@ public class FileLogger {
     private static final String LOG_SUFFIX = "_log";
     private static final String UTF8 = "UTF-8";
     private static final String ASCII = "US-ASCII";
+    private static final String FILE_LOGGER_PREFS = "sf_file_logger_prefs";
     private static final String TAG = "FileLogger";
     private static final int MAX_SIZE = 10000;
 
+    private Context context;
+    private String componentName;
     private QueueFile file;
     private int maxSize;
 
@@ -61,9 +65,11 @@ public class FileLogger {
      * @throws IOException If file initialization was unsuccessful.
      */
     public FileLogger(Context context, String componentName) throws IOException {
+        this.context = context;
+        this.componentName = componentName;
+        readFileLoggerPrefs();
         final File filename = new File(context.getFilesDir(), componentName + LOG_SUFFIX);
         file = new QueueFile(filename);
-        maxSize = MAX_SIZE;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogger.java
```diff
@@ -269,25 +272,53 @@ public void log(Level level, String tag, String message, Throwable e) {
...
-                        logLine = String.format(LOG_LINE_FORMAT, level, tag, message);
+                        logLine = String.format(LOG_LINE_FORMAT, curTime, level, tag, message);
                     }
                     fileLogger.addLogLine(logLine);
                 }
             }
         });
     }
+
+    private String getTimeFromUTC() {
```
bhariharan: Also added logging of timestamp.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/FileLogger.java
```diff
@@ -102,11 +108,11 @@ public int getMaxSize() {
      *
      * @param size Maximum number of log lines.
      */
-    public synchronized void setMaxSize(int size) {
+    public void setMaxSize(int size) {
         if (size < 0) {
             size = 0;
         }
-        maxSize = size;
+        storeFileLoggerPrefs(size);
```
wmathurin: Shouldn't maxSize also be set?
If I construct a FileLogger then call setMaxSize, woudn't the chosen maxSize only be used the next time I construct a FileLogger?

bhariharan: Nope - it should be used this time. It's set at the end of `storeFileLoggerPrefs` [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1421/files#diff-a6cbed71575822a567cf352b98d0d547R269).

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogger.java
```diff
@@ -49,8 +52,10 @@
 public class SalesforceLogger {
 
     private static final String TAG = "SalesforceLogger";
-    private static final String LOG_LINE_FORMAT = "LEVEL: %s, TAG: %s, MESSAGE: %s";
-    private static final String LOG_LINE_FORMAT_WITH_EXCEPTION = "LEVEL: %s, TAG: %s, MESSAGE: %s, EXCEPTION: %s";
+    private static final String LOG_LINE_FORMAT = "TIME: %s, LEVEL: %s, TAG: %s, MESSAGE: %s";
+    private static final String LOG_LINE_FORMAT_WITH_EXCEPTION = "TIME: %s, LEVEL: %s, TAG: %s, MESSAGE: %s, EXCEPTION: %s";
+    private static final String US_DATE_FORMAT = "yyyy.MM.dd G 'at' HH:mm:ss z";
```
wmathurin: Could we add the milliseconds (HH:mm:ss.SSS)?
Also should we just mirror the logcat date/time format e.g.
05-30 15:18:15.839 ?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/FileLogger.java
wmathurin: 👍

bhariharan:
Switched date format [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1421/commits/acc532690f07c6d94b508c65b8a2398ede136cca) and switched to 1 thread [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1421/commits/0a93a5bdf4a9831ebbcef44ad74de0c34df330d7).

PR #1422, opened 2017-05-31
Replacing log lines in SalesforceSDK library with new logger framework

bhariharan:


bhariharan:
No real code changes. I've replaced all `Log` calls with `SalesforceLogger` calls.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -791,7 +798,7 @@ private synchronized void postPushUnregister(BroadcastReceiver pushReceiver,
             try {
                 context.unregisterReceiver(pushReceiver);
             } catch (Exception e) {
-            	Log.e("SalesforceSDKManager", "Exception occurred while un-registering.", e);
+                logger.log(SalesforceLogger.Level.ERROR, TAG, "Exception occurred while unregistering", e);
```
wmathurin: Could you 
```java
import static SalesforceLogger.Level.*;
```
So you could just write
```java
logger.log(ERROR, TAG, ...)
```

Or should we introduce warn / info / error / debug methods on logger?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -115,7 +109,9 @@ protected synchronized void upgradeAccMgr() {
             	upgradeTo2Dot2();
             }
         } catch (NumberFormatException e) {
-        	Log.e("UpgradeManager:upgradeAccMgr", "Failed to parse installed version.");
+            SalesforceLogger.getLogger(SalesforceSDKManager.SF_SDK_COMPONENT_NAME,
```
wmathurin: Can we make that shorter?
Maybe SalesforceSDKManager could have a static method getLogger(String componentName) and the logger could have a e method (for errors), and you could import static SF_SDK_COMPONENT_NAME, so it would look like
```java
SalesforceSDKManager.getLogger(SF_SDK_COMPONENT_NAME).e(TAG, "Failed ...")
```

wmathurin: Or could we go further and have a SalesforceSDKManager aware helper class with static methods for info/warning/error, so the call would look like
```java
SDKLogger.e(SF_SDK_COMPONENT_NAME, TAG, ...)
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
wmathurin: So the call would then be 
```java
logger.e(TAG, ...)
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
wmathurin: I'm not sure about the class name but it should be as short as possible and live in the SalesforceSDK lib.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
bhariharan: I'll add methods on the logger for each level (like `Log.e`).

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
bhariharan: Will do.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
bhariharan: Added explicit methods [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1422/commits/99c3a06377337ec981a8be480c3352e3e3cdbdce).

bhariharan:
@wmathurin Added a util [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1422/commits/01b04bc6b5e4704aad1f20f7440ff17f5db4c4e8) for `SalesforceSDK`. Log lines now look like `SalesforceSDKLogger.e(tag, message, e);`. Should be good to go now.

PR #1424, opened 2017-06-02
Switching logs in SmartSync to use new framework

bhariharan:


bhariharan:
I think those two loggers should be consolidated. I'll take a stab at it.

PR #1427, opened 2017-06-03
Consolidating SmartSync logger utils

bhariharan:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SmartSyncLogger.java
```diff
@@ -61,6 +70,20 @@ public static void e(String tag, String message, Throwable e) {
...
 
     /**
+     * Logs an error log line.
+     *
+     * @param tag Log tag.
+     * @param origin Origin class.
+     * @param msg Log message.
+     * @param obj Object to be logged.
+     */
+    public static void e(String tag, Object origin, String msg, Object obj) {
```
wmathurin: Do we need tag and origin?

SyncManagerLogger had a fixed tag ("SyncManagerLogger") and origin was used to differentiate log lines from different classes.

Also does SalesforceLogger print the component name anywhere in the log line?

bhariharan: True. I can get rid of `origin` here.

And no, `SalesforceLogger` does not print component name in the console. The component name matters only for the file backed storage - each component has its own file. Also, depending on the log levels for different components, log lines printed will vary. However, the component names themselves wouldn't be a part of the log line. I don't know if there's much value in spewing those out to the console.

bhariharan:
Removed `origin` [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1427/commits/22e627dac9676afecf000d4e31cfa0c75aafe862).

PR #1428, opened 2017-06-05
Replacing calls in SalesforceAnalytics library with new logger

bhariharan:


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/util/SalesforceAnalyticsLogger.java
```diff
@@ -0,0 +1,222 @@
...
+     * @param message Log message.
+     * @param e Exception to be logged.
+     */
+    public static void v(Context context, String tag, String message, Throwable e) {
+        log(context, SalesforceLogger.Level.VERBOSE, tag, message, e);
+    }
+
+    private static void log(Context context, SalesforceLogger.Level level, String tag, String message) {
+        if (context != null) {
+            SalesforceLogger.getLogger(COMPONENT_NAME, context).log(level, tag, message);
```
wmathurin: Question:
Could this [line](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/util/SalesforceSDKLogger.java#L149) have some unwanted side effect in some cases e.g. in tests maybe - creating a SalesforceSDKManager sooner than wanted?

bhariharan: Nope - it shouldn't. I did consider that carefully when adding those calls.
- Our library will throw a `RuntimeException` if any code is executed before `SalesforceSDKManager.init...` has been called. This is by design, because most parts of our library operate under the assumption that an instance of `SalesforceSDKManager` (and thereby an `appContext`) will be available.
- None of the tests use this logging framework for that very reason (to avoid such a situation).

PR #1434, opened 2017-06-12
Fixing bug where ActionBar could be null

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -234,7 +234,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
 	@Override
 	public void loadingLoginPage(String loginUrl) {
 		final ActionBar ab = getActionBar();
-		ab.setTitle(loginUrl);
+		if (ab != null) {
```
bhariharan: `getActionBar()` is `nullable`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -846,9 +846,9 @@ public void run() {
                 }
             });
 		} catch (KeyChainException e) {
-			e.printStackTrace();
+            SalesforceSDKLogger.e(TAG, "Exception thrown while retrieving X.509 certificate", e);
```
bhariharan: Fixing logging.

PR #1436, opened 2017-06-21
ManagerTestCase was not compiling anymore

wmathurin:
Added public static method to set log level

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SmartSyncLogger.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SmartSyncLogger.java
@@ -213,6 +213,16 @@ public class SmartSyncLogger {
         v(tag, msg + ": " + objStr);
     }
 
+    /**
+     * Sets the log level to be used.
+     *
+     * @param level Log level.
+     */
+    public static void setLogLevel(SalesforceLogger.Level level) {
+        getLogger().setLogLevel(level);
+    }
+
+
     private static SalesforceLogger getLogger() {
         return SalesforceLogger.getLogger(COMPONENT_NAME,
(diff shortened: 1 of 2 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SmartSyncLogger.java
```diff
@@ -213,6 +213,16 @@ public static void v(String tag, String msg, Object obj) {
         v(tag, msg + ": " + objStr);
     }
 
+    /**
+     * Sets the log level to be used.
+     *
+     * @param level Log level.
+     */
+    public static void setLogLevel(SalesforceLogger.Level level) {
```
wmathurin: We should offer a similar method (or exposed getLogger() publicly) in all the *Logger classes otherwise there is no way for developers to control log levels per component.

bhariharan: Yup, will do. They could directly call SalesforceLogger with the right component name to achieve the same though, since it's a quasi-singleton.

wmathurin: That's true. But then we should make the COMPONENT_NAME in all those classes public.

bhariharan: True. I like your approach of adding a method to set log level better. It's more obvious and easier to document. I'll send a PR.

bhariharan: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1438).

PR #1439, opened 2017-06-26
Adding support for branded login path

bhariharan:


bhariharan:
@trooper2013 FYI

bhariharan:
Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1439/commits/8ec5642ceab8d584ee964beac18113df02de6271).

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -582,12 +583,31 @@ public synchronized String getEncryptionKeyForPasscode(String actualPass) {
...
+     */
+    public String getBrandedLoginPath() {
+    	return brandedLoginPath;
+    }
+
+    /**
+     * Sets the branded login path. In the following example, "<brand>" should be set here.
+     * https://community.force.com/services/oauth2/authorize/<brand>?response_type=code&...
+     *
+     * @return Branded login path.
```
wmathurin: Setter doesn't return anythinf

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -582,12 +583,31 @@ public synchronized String getEncryptionKeyForPasscode(String actualPass) {
     }
 
     /**
+     * Returns the branded login path.
+     *
+     * @return Branded login path, or empty string if not configured.
+     */
+    public String getBrandedLoginPath() {
```
wmathurin: Should we just call it brand?
The name is a bit misleading. This method doesn't return the full path of the login URL.

khawkins:
I'm trying to envision how the branding information is actually going to be bootstrapped in a real-world case.  Presumably, the developer makes an out-of-band call to...somewhere, to establish the branding information, sets that, and proceeds with authentication?

To me, the devils are in those "somewhere" details.  Isn't it likely to be org-specific information?  At which point, wouldn't we want branding information to reside at the user-level, rather than the app level?

khawkins:
But then I guess this branding will necessarily come in before a user has even been established.  So maybe at the login host level?  It seems like it's not at the app level.

bhariharan:
`SalesforceSDKManager` is the established way to provide global one-time optional customizations in an `Android Mobile SDK` app, typically. We have a similar pattern established for other points of customization such as `appName`, `displayType`, etc. The first call when an app starts up has to be `SalesforceSDKManager.init...`, followed by these optional configs. For something a little more mainstream we would probably use something like `bootconfig` or somewhere in `xml` that can be overridden, but in this case, I feel like this is the right place for it because it's not tied to login host itself, it's decoupled.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
bhariharan: I can rename it to `loginBrand`. I don't want it to be confused with community/org level branding - that's something else we don't really support today.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
bhariharan: Copy-paste error ;-)

khawkins:
I'm not concerned with where it lives as much as I'm concerned with its scope.  Display Type and App Name are app-level configurations.  Arguably, branding information will not be.  Seems entirely likely to be information provided at an Org level, in the form of an entry in a /.well-known configuration on a My Domains URL.

In that scenario, it's not going to make sense to have a "single" brand setting at the app level, because it's simply the wrong vehicle to represent a configuration that can have many values.

bhariharan:
We have other org-specific customizations in `SalesforceSDKManager` like `additionalOauthKeys`. This could use a refactor. I agree with you conceptually on encapsulating these elsewhere. I'll make a note to do this in `6.0`.

bhariharan:
@wmathurin Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1439/commits/89d9b12c20df004f8b04a1b4b24a97442365eb8f).

wmathurin:
I wonder why the server team could not have the brand at the beginning at the path so it could be captured in the login host URL. We should capture the brand wherever we capture the login hosts (host picker screen, preference files, MDM parameter etc). But of course that's a more involved piece of code. Should we document the new API as being temporary?

bhariharan:
We asked them to implement it as a param (we have support for plugging in params today), like `?brand=<brand>`. They gave us an explanation of why they couldn't do that (which I didn't understand and/or don't remember). AFAIK, this is a new feature that only 1 customer wants - `LV`. Let's see if it gets any adoption over native login flow. There are other ways to brand a community login page which are easier and more preferred.

PR #1442, opened 2017-06-28
Logger crash bug fix

bhariharan:
There could be cases where the app extends `SalesforceSDKManager` but uses some hybrid components. If we automatically typecast to `SmartSyncSDKManager`, this would cause a crash. Besides, `appContext` is in the lowest level of the hierarchy anyway.


bhariharan:
@skyler-b Could you try it with this patch?

PR #1444, opened 2017-06-29
Added check for null okHttpClient before publishing log lines

eric-romero:
There are some cases when publishLogLines may be called before RestClient has
an okHttpClient. In these cases, lets return early and publish logs later when its ready.


bhariharan:
As I mentioned in the other PR, we should investigate why we have a `RestClient` instance without a valid backing `OkHttpClient`. Looking at the code, this should never happen unless there's an invalid cached instance of `RestClient` or the app is explicitly setting `OkHttpClient` to null, both of which need app level fixes.

PR #1445, opened 2017-06-29
Update Sync Status to fail when refresh token is invalid.

narakim:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -255,7 +255,8 @@ public void run() {
                     updateSync(sync, SyncState.Status.DONE, 100, callback);
                 } catch (RestClient.RefreshTokenRevokedException re) {
                     SmartSyncLogger.e(TAG, "Exception thrown in runSync", re);
-                    // Do not do anything - let the logout go through!
+                    // Failed due to invalid refresh token. Update status to failed.
+                    updateSync(sync, SyncState.Status.FAILED, UNCHANGED, callback);
```
wmathurin: We didn't do an updateSync on purpose because it will try and write to the database, but the revoke token will cause a logout which will cause the database to be wiped out.
Maybe we can have updateSync take an extra argument to decide whether to just notify listeners or notify listeners and update status in db. We would use the only notify listeners in this case?

bhariharan: We send out an intent when the refresh token is revoked. The app could have a listener for this intent and cancel any in-flight sync. @narakim 

Here's the action on the intent:
```
public static final String ACCESS_TOKEN_REVOKE_INTENT = "access_token_revoked";
```

Look in `ClientManager` for more details.

narakim: Ah I see, let me look into cancelling syncs. Thank you @wmathurin @bhariharan

PR #1458, opened 2017-07-14
Capturing start and end time in sync state

wmathurin:
Reporting run time in ailtn event

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/EventBuilderHelper.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/EventBuilderHelper.java
@@ -44,6 +44,10 @@ import java.util.concurrent.Executors;
  */
 public class EventBuilderHelper {
 
+    // Event startTime and endTime can be specified by passing values for attributes START_TIME and END_TIME
+    public static final String START_TIME = "startTime";
+    public static final String END_TIME = "endTime";
+
     private static final String TAG = "EventBuilderHelper";
     private static boolean enabled = true;
 
@@ -122,7 +126,9 @@ public class EventBuilderHelper {
         final InstrumentationEventBuilder builder = InstrumentationEventBuilder.getInstance(manager.getAnalyticsManager(),
                                                                                             SalesforceSDKManager.getInstance().getAppContext());
         builder.name(name);
-        builder.startTime(System.currentTimeMillis());
+        builder.startTime(attributes.optLong(START_TIME));
+        if (attributes.has(END_TIME)) builder.endTime(attributes.optLong(END_TIME));
+
         final JSONObject page = new JSONObject();
         try {
             page.put("context", className);
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -354,6 +354,9 @@ public class SyncManager {
                         }
                         attributes.put("syncId", sync.getId());
                         attributes.put("syncTarget", sync.getTarget().getClass().getName());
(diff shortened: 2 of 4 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -354,6 +354,9 @@ private void updateSync(SyncState sync, SyncState.Status status, int progress, S
                         }
                         attributes.put("syncId", sync.getId());
                         attributes.put("syncTarget", sync.getTarget().getClass().getName());
+                        if (sync.getEndTime() > 0 && sync.getStartTime() > 0) {
+                            attributes.put("runTime", sync.getEndTime() - sync.getStartTime());
```
bhariharan: We shouldn't put this in `attributes`. The `attributes` part of the payload is for free-form JSON that doesn't have well-defined fields in the schema or for additional information. In this case, there's a `duration` field in the schema that is automatically extracted in `Splunk`, making it easy to look up. The library has 2 fields - [startTime](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/model/InstrumentationEvent.java#L66) and [endTime](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/model/InstrumentationEvent.java#L67). We automatically compute `duration` and put it in the payload [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/transform/AILTNTransform.java#L114), if `endTime` is available, since `endTime` is an optional field. For this case, I suggest we tweak the implementation of [this](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/EventBuilderHelper.java#L107) helper method to extract `startTime` and `endTime` from the `attributes` payload and put them in their respective fields of the event if they exist. So this line would also change to put `startTime` and `endTime` in `attributes`.

wmathurin: On android, startTime / duration are long and represent milliseconds.
On iOS, they are integers and represent seconds.
We need to make them consistent and hold the best precision possible.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTestCase.java
```diff
@@ -262,6 +262,12 @@ protected void checkStatus(SyncState sync, SyncState.Type expectedType, long exp
         if (expectedTotalSize != TOTAL_SIZE_UNKNOWN) {
             assertEquals("Wrong total size", expectedTotalSize, sync.getTotalSize());
         }
+        if (sync.getStatus() != SyncState.Status.NEW) {
```
wmathurin: That's the test part.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/EventBuilderHelper.java
```diff
@@ -122,7 +126,9 @@ private static void createAndStore(String name, UserAccount userAccount, String
         final InstrumentationEventBuilder builder = InstrumentationEventBuilder.getInstance(manager.getAnalyticsManager(),
                                                                                             SalesforceSDKManager.getInstance().getAppContext());
         builder.name(name);
-        builder.startTime(System.currentTimeMillis());
+        builder.startTime(attributes.optLong(START_TIME));
```
wmathurin: buildEvent() will use currentTimeMillis if startTime is 0, so no need to provide it here.

PR #1461, opened 2017-07-19
Restart app / show login if refresh token revoked

wmathurin:
This is the default behavior on iOS but was not the default behavior on Android.
If this is not the desired behavior, a dev simply needs to override logoutCompleteActions.

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
@@ -551,6 +551,11 @@ public class SalesforceDroidGapActivity extends CordovaActivity {
      * Performs actions on logout complete.
      */
     protected void logoutCompleteActions() {
+        // NB: without this code a refresh token revoke does not restart app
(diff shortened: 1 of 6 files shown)


wmathurin:
The other option is to keep it the way it was (and have them be different on iOS and Android) and document the way to achieve the restart on revoke.

bhariharan:
I don't think we should move this to the super classes in the SDK and here's why:

This method is triggered when the intent `LOGOUT_COMPLETE_INTENT_ACTION` is broadcasted from `SalesforceSDKManager`. The purpose of this intent is for apps to perform any extra tasks in their app in addition to what the SDK does. The behavior within the SDK when a logout occurs is slightly different between `iOS` and `Android`. On `Android`, we decide what to show next based on the number of users remaining in `SalesforceSDKManager` (look at the logic in `startSwitcherActivityIfRequired()`). This method is overridable by an app to dictate its own custom behavior. Also, whether to show the login screen or not is also customizable by passing in a `boolean` parameter for that. By recreating the activity in the base classes, we're circumventing all those customization points in the SDK. I think we should leave it in the sample app's activity and leave the SDK's activities the way they are since this impacts way more than just refresh token revocation - it impacts multi user scenarios, etc.

wmathurin:
That makes sense.

PR #1462, opened 2017-07-19
Bumping up minApi to API 21 (Lollipop)

bhariharan:


bhariharan:
`API 21` and above use `TLS 1.2` by default.

PR #1479, opened 2017-08-20
Android O Support

bhariharan:


bhariharan:
Haven't tested the entire matrix yet. That part is in progress.

bhariharan:
@wmathurin Tested `SmartSyncExplorer` and `SmartSyncExplorerHybrid` on an older emulator and an `Android O` emulator (`API 26`). Both apps work fine on both. We're good to go!

PR #1481, opened 2017-08-23
Adding in-memory cache for encryption keys

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -79,17 +84,23 @@ public static synchronized String getUniqueId(String name) {
      * @return Encryption key.
      */
     public static synchronized String getEncryptionKey(String name) {
+        final String keyString = getUniqueId(name);
```
wmathurin: Why not use name as the key in CACHED_ENCRYPTION_KEYS?

bhariharan: Sure - we could. I can change that.

PR #1483, opened 2017-08-24
Caching auth token right away

wmathurin:
Also fixing issues with RestClient caches
- no point building a fresh oauth refresh interceptor if a cached ok http client ends up being used
- cleaning cache on logout

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -369,6 +369,11 @@ public class ClientManager {
         }
         Account acc = new Account(accountName, getAccountType());
         accountManager.addAccountExplicitly(acc, SalesforceSDKManager.encrypt(refreshToken), new Bundle());
+
+        // Caching auth token otherwise the first call to accountManager.getAuthToken will go to the AuthenticatorService which will do a refresh
+        // That is problematic when the refresh token is set to expire immediately
+        accountManager.setAuthToken(acc, AccountManager.KEY_AUTHTOKEN, SalesforceSDKManager.encrypt(authToken));
+
         // There is a bug in AccountManager::addAccountExplicitly() that sometimes causes user data to not be
         // saved when the user data is passed in through that method. The work-around is to call setUserData()
         // for all the user data manually after passing in empty user data into addAccountExplicitly().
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -369,6 +369,11 @@ public Bundle createNewAccount(String accountName, String username, String refre
         }
         Account acc = new Account(accountName, getAccountType());
         accountManager.addAccountExplicitly(acc, SalesforceSDKManager.encrypt(refreshToken), new Bundle());
+
+        // Caching auth token otherwise the first call to accountManager.getAuthToken will go to the AuthenticatorService which will do a refresh
+        // That is problematic when the refresh token is set to expire immediately
+        accountManager.setAuthToken(acc, AccountManager.KEY_AUTHTOKEN, SalesforceSDKManager.encrypt(authToken));
```
wmathurin: The comment says it all.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -132,48 +136,71 @@
      * @param authTokenProvider
 	 */
 	public RestClient(ClientInfo clientInfo, String authToken, HttpAccess httpAccessor, AuthTokenProvider authTokenProvider) {
-		this(httpAccessor, new OAuthRefreshInterceptor(clientInfo, authToken, authTokenProvider));
-	}
-
-	public RestClient(HttpAccess httpAccessor, OAuthRefreshInterceptor httpInterceptor) {
+		this.clientInfo = clientInfo;
```
wmathurin: The freshly created oauth refresh interceptor ends up not being used if we find a cached ok http client.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -132,48 +136,71 @@
...
 	}
 
 	/**
-	 * Clear cache of org-id/user-id to OkHttpClient.Builder
+	 * Remove cached OkHttpClient.Builder, OkHttpClient and OAuthRefreshInterceptor for the given org id, user id
 	 */
-	public static void clearOkClientBuildersCache() {
-		OK_CLIENT_BUILDERS = null;
+	public synchronized static void clearCaches(String orgId, String userId) {
+		String cacheKey = computeCacheKey(orgId, userId);
```
wmathurin: If we don't clean the caches, and a user logs out and then logs back in, he will use the old okhttpclient (with the old oauth refresh interceptor).
As a result the first request will fail since the auth header will use the old auth token.
In the case where refresh expires immediately, you end up being send back to login right away.

PR #1485, opened 2017-08-28
Fixing auth related crash in Android for Work container on Android O

bhariharan:
In `Android 8.0 (API 26)`, they have changed the default `AccountManager` permissions accessible to an app. An app must now call `newChooseAccountIntent()`, which will bring up an account picker screen for the user to make a choice, or directly use the custom `Authenticator` for the app's account type. Since we have a custom `Authenticator`, we should use that to access account data instead of calling `accountManager.getAuthToken()`. Directly calling `getAuthToken()` will result in a `SecurityException` being thrown and the app will crash. This PR fixes refactors our access of account data to work on `Android O` in the `Android for Work` environment and some other cleanup. More info here:
- [Android 8.0 Changes](https://developer.android.com/about/versions/oreo/android-8.0-changes.html#aaad).
- [Android 8.0 API Changes](https://developer.android.com/about/versions/oreo/android-8.0.html#naa).

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -30,7 +30,7 @@ import android.accounts.Account;
 import android.accounts.AccountManager;
 import android.accounts.AccountManagerCallback;
 import android.accounts.AccountManagerFuture;
-import android.accounts.AccountsException;
+import android.accounts.NetworkErrorException;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
@@ -42,10 +42,10 @@ import com.salesforce.androidsdk.analytics.EventBuilderHelper;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.auth.AuthenticatorService;
 import com.salesforce.androidsdk.auth.HttpAccess;
+import com.salesforce.androidsdk.auth.OAuth2;
 import com.salesforce.androidsdk.rest.RestClient.ClientInfo;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
-import java.io.IOException;
 import java.io.Serializable;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -513,12 +509,12 @@ public String getNewAuthToken() {
             }
 
             // Invalidate current auth token.
-            final String cachedAuthToken = clientManager.accountManager.peekAuthToken(acc, AccountManager.KEY_AUTHTOKEN);
+            final String cachedAuthToken = clientManager.peekRestClient(acc).getAuthToken();
```
bhariharan: Get data instead of directly calling `peekAuthToken` on `AccountManager`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -513,12 +509,12 @@ public String getNewAuthToken() {
...
             clientManager.invalidateToken(cachedAuthToken);
             String newAuthToken = null;
             String newInstanceUrl = null;
             try {
-                final Bundle bundle = clientManager.accountManager.getAuthToken(acc, AccountManager.KEY_AUTHTOKEN, null, false, null, null).getResult();
+                final Bundle bundle = refreshStaleToken(acc);
```
bhariharan: Refresh the token internally instead of calling `getAuthToken()` on `AccountManager`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -582,6 +578,88 @@ public long getLastRefreshTime() {
 
         @Override
         public String getInstanceUrl() { return lastNewInstanceUrl; }
+
+        private Bundle refreshStaleToken(Account account) throws NetworkErrorException {
```
bhariharan: A simple refresh module -> uses `OAuth2.java` to refresh the token and constructs a `Bundle` with the new access token and instance URL after updating data in `AccountManager`. Encryption of data is done to keep it interoperable with `AuthenticatorService`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
wmathurin: Do we need the refresh code [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/blob/authenticator/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java#L185) ?
Can we avoid the duplication?

bhariharan: Yes, we need that too, for external modules that might use our authenticator to auth using Salesforce. I could extract it into a util class but I didn't want to expose it in a public method. I'll try to organize this better when I do the IDP flow on Android.

PR #1486, opened 2017-08-29
Logout improvements

wmathurin:


wmathurin:
See commits for details

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
```diff
@@ -133,26 +132,14 @@ public static SmartStoreSDKManager getInstance() {
...
         } else {
             DBOpenHelper.deleteAllUserDatabases(getAppContext());
         }
 
-        /*
-         * Checks how many accounts are left that are authenticated. If only one
-         * account is left, this is the account that is being removed. In this
-         * case, we can safely reset all DBs.
-         */
-        final List<UserAccount> users = getUserAccountManager().getAuthenticatedUsers();
```
bhariharan: Don't we still need this? There's something special about the last user in a multi-user scenario. I don't remember the details. Or maybe we cleaned it up along the way.

wmathurin: 1) If there is only one account left, wouldn't that be the one passed to the cleanUp(UserAccount) method?
2) If somehow null is passed, but there is one account left, then I need to handle in the cleanUp(Activity, Account) method. cleanUp(UserAccount) could be running in a separate process as a result of the CLEANUP_INTENT_ACTION broadcast, and the account manager might already have been updated by then.
Do you think cleanup(null) could happen?

bhariharan: Maybe this was put in to handle logout from `Settings` at some point, because `Settings` didn't have a handle on the account that was being removed. Looking at the code, I don't see a way `null` would be passed in for account either. We should be good.

PR #1487, opened 2017-08-29
Removing meaningless code

wmathurin:
Once upon a time, we used to pad with the number of pad bytes. https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/d65974a70df81eed6d88109af96f36f937f55fee

Then we stopped doing that. https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/7becd771898d87843dc444dd758d93449199c824
But only the encrypt() method was updated, not the decrypt method.

We were left with that meaningless and actually dangerous code that can throw ArrayIndexOutOfBoundException.
NB: Such exception happened sometimes and a superficial change was made (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/8d9ef6bb45d3129f5a312688bbeee700fef28334).


wmathurin:
No migration required. Padding has long been gone (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/7becd771898d87843dc444dd758d93449199c824).

The padding related code in decrypt was rarely entered because the padding value would have to be 0 and 16. When it was entered, it was probably messing up the result value randomly.

bhariharan:
👍

PR #1490, opened 2017-09-07
Fixing tests

wmathurin:
While testing dev53, I discovered tests failures that also exist in dev.
This PR is to fix them (then I'll merge dev into dev53).


wmathurin:
NB: do not merge yet, there are more test failures I'm working on

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/files/FileRequestsTest.java
```diff
@@ -185,18 +183,16 @@ public void testFileShares() {
      * @throws Exception
      */
     public void testFileUpload() throws Exception {
-        final String filename  = "MyFile.txt";
```
wmathurin: We were getting a permission denied on Android O.
We don't actually need to create a file on external storage, so using File.createTempFile instead.

PR #1507, opened 2017-10-06
Static config for setting up smartstore

wmathurin:
Two new methods on SmartStoreSDKManager:
* `setupGlobalStoreFromDefaultConfig()`
Call at any time to register soups defined in `/res/raw/globalstore.json` into the default global smartstore.

* `setupUserStoreFromDefaultConfig()`
Call once a user has logged in to register soups defined in `/res/raw/userstore.json` into the user's default smartstore.

Notes:
* The JSON format is based on what we use in our Cordova plugin and React Native module.
* It just does a bunch of `registerSoup` calls, it will not do an `alterSoup` if the soup already exists.
* The only feedback is through the logs - that's intentional, those config files are static (can't be changed at runtime) so the developer should get them right while developing the app.
* You can't define a soup that defines external storage through the config file - that's intentional, we want to phase that feature out at some point.

The sample application SmartSyncExplorer now defines its soup in a config file.
I also added a new test suite called StoreConfigTest.

Example of JSON config file
```json
{
  "soups": [
    {
      "soupName": "soup1",
      "indexes": [
        { "path": "stringField1", "type": "string"},
        { "path": "integerField1", "type": "integer"},
        { "path": "floatingField1", "type": "floating"},
        { "path": "json1Field1", "type": "json1"},
        { "path": "ftsField1", "type": "full_text"}
      ]
    }
  ]
}
```


libs/SalesforceSDK/src/com/salesforce/androidsdk/util/test/JSONTestHelper.java
```diff
@@ -43,28 +43,34 @@
 	 * @throws JSONException
 	 */
 	public static void assertSameJSON(String message, Object expected, Object actual) throws JSONException {
```
wmathurin: Refactored that class so that when there is a mismatch, you see the full expected value and the full actual value.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
```diff
@@ -384,4 +389,31 @@ public void removeAllUserStores() {
...
+        setupStoreFromConfig(getSmartStore(), R.raw.userstore);
+    }
+
+    /**
+     * Setup given store using config found in given json resource file
+     *
+     * @param store
+     * @param resourceId
+     */
+    private void setupStoreFromConfig(SmartStore store, int resourceId) {
```
wmathurin: Should we make that third one public to allow configuring any stores from any resource file?

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartSqlExternalStorageTest.java
```diff
@@ -57,7 +57,7 @@
 	private static final String DEPARTMENTS_SOUP = "departments";
 
 	@Override
-	protected String getPasscode() {
+	protected String getEncryptionKey() {
```
wmathurin: Unrelated to the config work - but thought that we should rename that method (getPasscode was a bit misleading).

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTestCase.java
```diff
@@ -57,24 +57,25 @@
...
 	@Override
 	public void setUp() throws Exception {
 		super.setUp();
 		EventBuilderHelper.enableDisable(false);
 		targetContext = getInstrumentation().getTargetContext();
 		dbOpenHelper = DBOpenHelper.getOpenHelper(targetContext, null);
-		DBHelper.getInstance(dbOpenHelper.getWritableDatabase(getPasscode())).reset(targetContext, null);
-		store = new SmartStore(dbOpenHelper, getPasscode());
-		store.dropAllSoups();
+		dbHelper = DBHelper.getInstance(dbOpenHelper.getWritableDatabase(getEncryptionKey()));
```
wmathurin: We are doing the cleanup in tearDown so as to leave a clean slate to the next tests.

native/NativeSampleApps/SmartSyncExplorer/res/raw/userstore.json
```diff
@@ -0,0 +1,16 @@
+{
```
wmathurin: That's the smartstore config for SmartSyncExplorer.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
```diff
@@ -384,4 +389,31 @@ public void removeAllUserStores() {
                     UserAccountManager.getInstance().getCurrentUser().getCommunityId());
         }
     }
+
+    /**
+     * Setup global store using config found in res/raw/globalstore.json
+     */
+    public void setupGlobalStoreFromDefaultConfig() {
+        SmartStoreLogger.d(TAG, "Setting up global store using config sound in res/raw/globalstore.json");
```
bhariharan: Minor typo: I'm guessing you meant _found in_?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
```diff
@@ -384,4 +389,31 @@ public void removeAllUserStores() {
...
+        setupStoreFromConfig(getGlobalSmartStore(), R.raw.globalstore);
+    }
+
+    /**
+     * Setup user store using config found in res/raw/userstore.json
+     */
+    public void setupUserStoreFromDefaultConfig() {
+        SmartStoreLogger.d(TAG, "Setting up user store using config sound in res/raw/userstore.json");
```
bhariharan: Minor typo: I'm guessing you meant _found in_?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
bhariharan: I don't have a strong preference either way, but maybe we could expose this later? Developers might accidentally plug in other resources that aren't `JSON`.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartSqlExternalStorageTest.java
bhariharan: The encryption key is now automatically generated and applied internally by `SalesforceSDK`. This seems like a legacy artifact. Should we get rid of this interface (`getPasscode()`) to better represent the current design?

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/StoreConfigTest.java
```diff
@@ -0,0 +1,192 @@
...
+        super.setUp();
+
+        SmartStoreSDKTestManager.init(getInstrumentation().getTargetContext(), store);
+        sdkManager = SmartStoreSDKTestManager.getInstance();
+        globalStore = sdkManager.getGlobalSmartStore();
+        userStore = sdkManager.getSmartStore();
+    }
+
+    @Override
+    protected String getEncryptionKey() {
```
bhariharan: As mentioned above, should we remove this interface and its implementation and use `sdkManager.getEncryptionKey()` directly in all tests instead? Doesn't look like we care about the actual value of the encryption key in tests, so why bother setting it everywhere?

native/NativeSampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/sync/ContactSyncAdapter.java
```diff
@@ -63,12 +64,14 @@ public ContactSyncAdapter(Context context, boolean autoInitialize,
...
         final boolean syncDownOnly = extras.getBoolean(SYNC_DOWN_ONLY, false);
-        final SalesforceSDKManager sdkManager = SalesforceSDKManager.getInstance();
+        final SmartStoreSDKManager sdkManager = SmartStoreSDKManager.getInstance();
         final UserAccountManager accManager = sdkManager.getUserAccountManager();
         if (sdkManager.isLoggingOut() || accManager.getAuthenticatedUsers() == null) {
             return;
         }
         if (account != null) {
+            // Setup schema if needed
+            sdkManager.setupUserStoreFromDefaultConfig();
```
bhariharan: Shouldn't the setup happen from the loader? Seems like an anti-pattern to trigger this from the sync adapter.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/StoreConfigTest.java
wmathurin: SmartStoreTestCase builds a store directly without using a SDK manager.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreFullTextSearchTest.java
```diff
@@ -69,7 +68,7 @@
     private long eileenEvaId;
 
     @Override
-    protected String getPasscode() {
+    protected String getEncryptionKey() {
```
bhariharan: Should we replace these empty strings to some random non-empty encryption key to better represent reality?

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/StoreConfigTest.java
bhariharan: Gotcha.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreFullTextSearchTest.java
wmathurin: We could. Maybe a different PR for making smartstore tests more realistic?

bhariharan: Sounds good.

bhariharan:
👍

PR #1508, opened 2017-10-06
Expanding functionality for hybrid URLs and deferred authentication

khawkins:
Based on the 2x2 matrix we described:

| Deferred Login | Has Relative URL                       | Has Absolute URL                |
| -------------- | -------------------------------------- | ------------------------------- |
| No             | Require login and build abs URL        | Error since this is a misconfig |
| Yes            | Error since we can't build the abs URL | Load the abs URL w/o auth       |

This does sanity checking on the values, then updates the workflows mostly to support absolute URLs without authentication, which weren't allowed before.


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -181,8 +182,16 @@ private void restartIfUserSwitched() {
      */
     private void onResumeNotLoggedIn() {
 
+        BootConfigValidationResult validationResult = BootConfig.validateBootConfig(bootconfig);
```
khawkins: I coded this to ultimately allow validation of any boot config.  But since the only thing we're validating at the moment is hybrid configuration, I didn't plug this validation step in anywhere else.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -98,6 +98,49 @@ public static BootConfig getBootConfig(Context ctx) {
...
 		return INSTANCE;
 	}
+
+	/**
+	 * Validates a boot config's inputs against basic sanity tests.
+	 * @param config The BootConfig instance to validate.
+	 * @return A BootConfigValidationResult describing the success or failure of validation, with
+	 * an the error message associated with a failed validation.
+	 */
+	public static BootConfigValidationResult validateBootConfig(BootConfig config) {
```
khawkins: The logic was getting a little crazy, trying to juxtapose valid URL values against local/remote configs and authentication, in tandem.  This sanity checks all of the conditions that aren't subject to workflow.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -98,6 +98,49 @@ public static BootConfig getBootConfig(Context ctx) {
...
+				return new BootConfigValidationResult(false, "No start page configured.");
+			}
+
+			// Sanity check local URLs against absolute URL values.
+			if (config.isLocal() && config.isStartPageAbsoluteUrl()) {
+				return new BootConfigValidationResult(false, "Local start page should not be absolute URL.");
+			}
+
+			// Start page makeup for remote apps beyond this point is subject to conditional configuration.
+			if (!config.isLocal()) {
```
wmathurin: Would it be easier to read like this:
```java
if (!config.isLocal() && config.shouldAuthenticate() && config.isStartPageAbsoluteUrl()) {
   return new BootConfigValidationResult(false, "Cannot authenticate using an absolute URL for start page.");
}
if (!config.isLocal() && !config.shouldAuthenticate() && !config.isStartPageAbsoluteUrl()) {
    return new BootConfigValidationResult(false, "Cannot configure relative URL start page with deferred authentication.");
}
```

khawkins: I could do that, sure.

khawkins:
@wmathurin @bhariharan Thinking about this after sleeping on it, I had a couple of questions.

1. In Android/Java, would it be more consistent to use a throw/catch paradigm for surfacing boot config failures in `BootConfig.validateBootConfig()`, instead of the custom `BootConfigValidationResult` object returning state?
2. I made mention of not having any particular validation on the native side, but should we change that?  I was thinking about that through the iOS lens, where `SalesforceSDKManager` actually validates the standard information with some nominal sanity checks.  Should this be part of that process, a little higher up in the execution order?

wmathurin:
1. I agree it should blow up if the config is wrong.
The config is static (can't be changed at runtime), so the dev should get it right.
2. Makes sense to me as well.

bhariharan:
1. Agreed - probably throw a `BootConfigException` with details on the error or something.
2. What sort of validation are we talking about moving to `SalesforceSDKManager`? I'm not sure I understand what's missing on the native side.

khawkins:
@bhariharan We have sanity tests on iOS to make sure that things like Consumer Key and Redirect URI actually have values.  I don't see that on the Android side, unless I'm missing it.  On the iOS side, there's an inheritance model between the basics of this app config, and the additional app config settings that are specific to hybrid.  If we had an equivalent model, we could simply validate the config at app startup, rather than somewhere down the Activity path.

bhariharan:
@khawkins This is already handled in `Android` by `BootConfig` throwing a `JSONException` if any of those values don't exist. This would cause a failure right away in `SalesforceSDKManager` as soon as `init` is called. I don't see the need for explicit checks there.

khawkins:
Ah okay, I missed that validation.  I'm working on an approach to align those on the iOS side, since that side of the house is a little disconnected in the way that the validation is owned.  On Android, I'll continue with the approach I've started, and clean up the error handling a bit.  Stay tuned!

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -453,10 +461,21 @@ public void loadLocalStartPage() {
...
+     * @param loadThroughFrontDoor Whether or not to load through front-door.
+     */
+    private void loadRemoteStartPage(boolean loadThroughFrontDoor) {
         assert !bootconfig.isLocal();
         String startPage = bootconfig.getStartPage();
+        String url = startPage;
+        if (loadThroughFrontDoor) {
+            url = getFrontDoorUrl(url, false);
+        }
         SalesforceHybridLogger.i(TAG, "loadRemoteStartPage called - loading: " + startPage);
```
bhariharan: Minor: change this log line to `url` to correctly reflect the URL being loaded.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -98,6 +98,39 @@ public static BootConfig getBootConfig(Context ctx) {
 		}
 		return INSTANCE;
 	}
+
+	/**
+	 * Validates a boot config's inputs against basic sanity tests.
+	 * @param config The BootConfig instance to validate.
+	 * @throws BootConfigException If the boot config is invalid.
+	 */
+	public static void validateBootConfig(BootConfig config) {
```
bhariharan: Should this be renamed to `validateHybridBootConfig`, since it's validating only hybrid config?

khawkins: I left it generic, because it could validate native-based configuration in the future.  I.e. if you call this for a native config at this point, it will still "validate"—there's just no validation.  But we are only calling it in the hybrid case at this point, so I could see it that way too.

khawkins:
Cool.  Should I merge this, or wait until we finalize our new rules for bootstrap configuration and get that implemented in this PR first?

bhariharan:
Let's merge this and do another PR for that.

PR #1509, opened 2017-10-11
Large scale OAuth cleanup

bhariharan:
Before I send the IDP and SP pull requests out, I wanted to clean up our legacy `OAuth` implementation and start off with a clean base, so I've sent that portion out separately as a pull request.

**Since this PR is large and probably not reviewable, here's a summary of changes I made:**
- Consolidated a bunch of overridden methods that we had accumulated over the years as we added more parameters for features. Most of them were unnecessary and confusing for consumers.
- Removed `client_secret` in any shape, size or form from our codebase, since this is a legacy `OAuth` flow that doesn't apply anymore.
- Removed `clientId` from the `UserAccount` object. This field was never getting populated anyway because we were (incorrectly) assigning the value of `client_secret` to this field.
- All the other changes are simply rippling effects of the API changes.

**Testing:**
- All libraries and sample apps compile successfully.
- Ran the following sample apps and tested various scenarios including revoking tokens, refresh flow, multi-user login, switching users and log out. Works fine.
    - [x] `SmartSyncExplorer`.
    - [x] `RestAPIExplorer`.
    - [x] `SmartSyncExplorerHybrid`.
- Running tests:
    - [x] `SmartSync` tests - 100% passing.
    - [x] `SmartStore` tests - 100% passing.
    - [x] `SalesforceHybrid` tests - 100% passing.
    - [x] `SalesforceSDK` tests - 100% passing.
- Emulators used:
    - [x] `API 25`.
    - [x] `API 23`.

Should be good to go!

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
@@ -56,7 +56,6 @@ public class UserAccount {
 	public static final String USER_ID = "userId";
 	public static final String USERNAME = "username";
 	public static final String ACCOUNT_NAME = "accountName";
-	public static final String CLIENT_ID = "clientId";
 	public static final String COMMUNITY_ID = "communityId";
 	public static final String COMMUNITY_URL = "communityUrl";
 	public static final String INTERNAL_COMMUNITY_ID = "000000000000000AAA";
@@ -82,7 +81,6 @@ public class UserAccount {
 	private String userId;
 	private String username;
 	private String accountName;
-	private String clientId;
 	private String communityId;
 	private String communityUrl;
     private String firstName;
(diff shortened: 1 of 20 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -146,8 +116,8 @@ public UserAccount(String authToken, String refreshToken,
 	public UserAccount(String authToken, String refreshToken,
```
wmathurin: Should we put in a builder for 6.0? There are a lot of args in that constructor.

bhariharan: Great idea! I'd prefer to do that in another PR later though.

wmathurin:
Sounds good.

PR #1516, opened 2017-10-12
Using Builder pattern to build UserAccount

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -113,7 +113,7 @@
 	 * @param thumbnailUrl Thumbnail URL.
 	 * @param additionalOauthValues Additional OAuth values.
 	 */
-	public UserAccount(String authToken, String refreshToken,
+	UserAccount(String authToken, String refreshToken,
```
bhariharan: Package level access to ensure consumers use the `Builder` going forward.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
```diff
@@ -0,0 +1,278 @@
...
+package com.salesforce.androidsdk.accounts;
+
+import java.util.Map;
+
+/**
+ * This class helps build a UserAccount object.
+ *
+ * @author bhariharan
+ */
+public class UserAccountBuilder {
```
bhariharan: A simple `Builder` pattern class for `UserAccount`.

PR #1521, opened 2017-10-19
Adding foundation pieces for IDP flow

bhariharan:
This PR adds the foundation components for the IDP portion. I'm now able to trigger the IDP flow for an existing user from an SP and go back all the way to the SP with a `code` from the server. The IDP auth piece would typically be triggered from the SP like this:
```
final Intent intent = new Intent(this, IDPWebViewActivity.class);
intent.putExtra(IDPWebViewActivity.SP_CONFIG_BUNDLE_KEY, spConfig.toBundle());
intent.putExtra(IDPWebViewActivity.USER_ACCOUNT_BUNDLE_KEY, userAccount.toBundle());
startActivityForResult(intent, LAUNCH_CODE);
```
The result would be handled by the SP in a callback like this:
```
@Override
public void onActivityResult(int requestCode, int resultCode, Intent data) {
    if (requestCode == 777) {
        if (resultCode == RESULT_CANCELED) {
            final String error = data.getStringExtra(IDPWebViewActivity.ERROR_KEY);
        } else if (resultCode == RESULT_OK) {
            final String code = data.getStringExtra(IDPWebViewActivity.CODE_KEY);
        }
    }
}
```
Next PR will contain the SP pieces and the actual flow between various components. It will also have tweaks to the existing user switcher screen and more tests.


libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -79,6 +79,12 @@
             android:excludeFromRecents="true"
             android:theme="@style/SalesforceSDK.ActionBarTheme" />
 
+        <!-- IDP webview activity -->
+        <activity android:name="com.salesforce.androidsdk.auth.idp.IDPWebViewActivity"
```
bhariharan: New activity for the `WebView` portion of the IDP flow. This is also the central orchestrator of this flow.

libs/SalesforceSDK/res/layout/sf__idp_webview.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
bhariharan: A simple layout that only contains a `WebView` right now. I wanted to keep it in XML so that apps can override the screen in parts or entirely.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -217,7 +218,65 @@ public static URI getAuthorizationUrl(URI loginServer, String clientId, String c
...
+     * @param accessToken IDP's access token.
+     * @param loginUrl SP's login URL.
+     * @param displayType IDP's display type.
+     * @param clientId SP's client ID.
+     * @param callbackUrl SP's callback URL.
+     * @param scopes SP's scopes.
+     * @param codeChallenge SP's code challenge.
+     * @return A 'frontdoor' URL that the IDP can load in a WebView.
+     */
+    public static URI getIDPFrontdoorUrl(String instanceUrl, String accessToken, String loginUrl,
```
bhariharan: Builds the IDP frontdoor request. Similar to what we have for JWT today.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -217,7 +218,65 @@ public static URI getAuthorizationUrl(URI loginServer, String clientId, String c
...
+     *
+     * @param loginUrl SP's login URL.
+     * @param displayType IDP's display type.
+     * @param clientId SP's client ID.
+     * @param callbackUrl SP's callback URL.
+     * @param scopes SP's scopes.
+     * @param codeChallenge SP's code challenge.
+     * @return An approval URL that the IDP can use to construct its 'frontdoor' URL.
+     */
+    public static URI getIDPApprovalUrl(String loginUrl, String displayType,
```
bhariharan: Builds the IDP auth request. Similar to what we have for JWT today.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPRequestHandler.java
```diff
@@ -0,0 +1,173 @@
...
+                accessToken = restClient.getAuthToken();
+            }
+        }
+        return accessToken;
+    }
+
+    private void basicValidation() throws IDPRequestHandlerException {
+
+        // Ensures that we have all the values we need in order to perform an IDP request.
+        if (spConfig == null || TextUtils.isEmpty(spConfig.getOauthClientId()) ||
```
bhariharan: Basic validation to ensure we have all the pieces we need.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPRequestHandler.java
```diff
@@ -0,0 +1,173 @@
...
+     * @return Valid access token.
+     * @throws IDPRequestHandlerException
+     */
+    public String getValidAccessToken() throws IDPRequestHandlerException {
+
+        /*
+         * Ensures the access token associated with this user account is valid and refreshes the
+         * token if it's not valid. Any request to 'frontdoor' must be made with a valid token.
+         */
+        final Context context = SalesforceSDKManager.getInstance().getAppContext();
```
bhariharan: Utility to make a simple REST call to refresh the token (if required). This must be called before attempting to load frontdoor.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPRequestHandler.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Kicks off the 'frontdoor' request in the supplied WebView instance.
+     *
+     * @param accessToken Valid access token.
+     * @param webView WebView instance.
+     * @throws IDPRequestHandlerException
+     */
+    public void makeFrontDoorRequest(String accessToken, WebView webView) throws IDPRequestHandlerException {
+        final Context context = SalesforceSDKManager.getInstance().getAppContext();
```
bhariharan: Loads the frontdoor request in the supplied `WebView`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+    private SPConfig spConfig;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        // Fetches the required extras.
+        final Intent intent = getIntent();
+        final Bundle extras = intent.getExtras();
+        if (extras != null) {
```
bhariharan: Easy way to pass in `SPConfig` and `UserAccount`. Ideally, this would come from the account switcher screen based on what account was chosen (will be in the next PR).

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+                WindowManager.LayoutParams.FLAG_SECURE);
+        setContentView(R.layout.sf__idp_webview);
+        final WebView webView = findViewById(R.id.sf__webview);
+        final WebSettings webSettings = webView.getSettings();
+        webSettings.setUseWideViewPort(true);
+        webSettings.setLayoutAlgorithm(WebSettings.LayoutAlgorithm.NORMAL);
+        webSettings.setJavaScriptEnabled(true);
+        webView.setWebViewClient(new IDPWebViewClient());
+        try {
+            final IDPRequestHandler idpRequestHandler = new IDPRequestHandler(spConfig, userAccount);
```
bhariharan: Basic validation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * This class is used to monitor redirects within the WebView to determine
+     * when the login flow is complete, parses the response and passes it back.
+     */
+    protected class IDPWebViewClient extends WebViewClient {
+
+        @Override
+        public boolean shouldOverrideUrlLoading(WebView view, String url) {
+            boolean isDone = url.replace("///", "/").toLowerCase(Locale.US).startsWith(spConfig.getOauthCallbackUrl().
```
bhariharan: Inspects redirects to determine if the flow is done.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+                    replace("///", "/").toLowerCase(Locale.US));
+            if (isDone) {
+                final Uri callbackUri = Uri.parse(url);
+                final Map<String, String> params = UriFragmentParser.parse(callbackUri);
+
+                // Determines if the authentication flow succeeded or failed.
+                if (params != null) {
+                    final String code = params.get(CODE_KEY);
```
bhariharan: Gets `code` back from the server and passes it along, or passes along an error.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+                }
+            }
+            return isDone;
+        }
+    }
+
+    private void handleError(String error) {
+        final Intent intent = new Intent();
+        intent.putExtra(ERROR_KEY, error);
+        setResult(RESULT_CANCELED, intent);
```
bhariharan: Sets the error status and finishes itself.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+        finish();
+    }
+
+    private void handleSuccess(String code) {
+        final Intent intent = new Intent();
+        intent.putExtra(CODE_KEY, code);
+        setResult(RESULT_OK, intent);
```
bhariharan: Sets `code` in the result and finishes itself.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+        public RefreshAuthTokenTask(IDPRequestHandler idpRequestHandler, WebView webView) {
+            this.idpRequestHandler = idpRequestHandler;
+            this.webView = webView;
+        }
+
+        @Override
+        protected String doInBackground(Void... nothings) {
+            String accessToken = null;
+            try {
+                accessToken = idpRequestHandler.getValidAccessToken();
```
bhariharan: Network request must be made in the background.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPWebViewActivity.java
```diff
@@ -0,0 +1,171 @@
...
+            } catch (Exception e) {
+                SalesforceSDKLogger.e(TAG, "Refreshing token failed", e);
+            }
+            return accessToken;
+        }
+
+        @Override
+        protected void onPostExecute(String accessToken) {
+            try {
+                if (accessToken != null) {
```
bhariharan: Once the network request is done, we load frontdoor in the `WebView` -> this is done on the UI thread.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPConfig.java
```diff
@@ -0,0 +1,171 @@
...
+import android.os.Bundle;
+
+import com.salesforce.androidsdk.auth.OAuth2;
+
+/**
+ * This class performs represents an SP app's config.
+ *
+ * @author bhariharan
+ */
+public class SPConfig {
```
bhariharan: A class to represent the config passed in from the SP.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPConfig.java
```diff
@@ -0,0 +1,171 @@
...
+        this.userHint = userHint;
+        computedScopeParameter = OAuth2.computeScopeParameter(oauthScopes);
+    }
+
+    /**
+     * Parameterized constructor for internal use only.
+     *
+     * @param config Bundle containing config.
+     */
+    SPConfig(Bundle config) {
```
bhariharan: Convenience methods to make and extract from a `Bundle` to easily pass this object between apps and/or activities.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
```diff
@@ -0,0 +1,35 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.auth.idp;
+
+/**
+ * This class performs requests made by the SP app based on the IDP app's response.
+ *
+ * @author bhariharan
+ */
+public class SPRequestHandler {
```
bhariharan: Will be fleshed out in the next PR.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/IDPRequestHandlerTest.java
```diff
@@ -0,0 +1,212 @@
...
+import com.salesforce.androidsdk.rest.ClientManager;
+
+import junit.framework.Assert;
+
+/**
+ * Tests for IDPRequestHandler.
+ *
+ * @author bhariharan
+ */
+public class IDPRequestHandlerTest extends InstrumentationTestCase {
```
bhariharan: Tests for `IDPRequestHandler`.

native/NativeSampleApps/RestExplorer/res/values/bootconfig.xml
```diff
@@ -5,6 +5,7 @@
     <string name="oauthRedirectURI">testsfdc:///mobilesdk/detect/oauth/done</string>
     <string-array name="oauthScopes">
         <item>api</item>
+        <item>web</item>
```
bhariharan: Need `web` scope for frontdoor and to act as an IDP.

wmathurin:
When you said in the description for the PR  
> I'm not able to trigger the IDP flow

Did you mean:
> I'm **now** able to trigger the IDP flow

libs/SalesforceSDK/AndroidManifest.xml
wmathurin: Should we name it `IDPLoginActivity` instead to describe what it does rather than what it uses (and it's possible that in the future it won't use a web view anymore)?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPRequestHandler.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Makes a simple REST call to and returns a valid access token or null if refresh failed.
+     * This method should NOT be called from the main thread since it makes a network request.
+     *
+     * @return Valid access token.
+     * @throws IDPRequestHandlerException
+     */
+    public String getValidAccessToken() throws IDPRequestHandlerException {
```
wmathurin: We are doing the same thing in `SalesforceDroidGapActivity.java`, may be should introduce a helper method on RestClient to do that - which we could call ensureValidAccessToken.

native/NativeSampleApps/RestExplorer/res/values/bootconfig.xml
wmathurin: Are we going to put a check in the code to make sure the web scope is available or are we just let it fail in the web view?

libs/SalesforceSDK/AndroidManifest.xml
bhariharan: I deliberately named it this way. There will be another `IDPLoginActivity` that will be a subclass of `LoginActivity` in the next PR. That will be the screen that does login for a nonexistent user without switching users.

native/NativeSampleApps/RestExplorer/res/values/bootconfig.xml
bhariharan: In the next PR, I will add more error handling. I plan on adding the error callback within `WebViewClient`, so that if an error callback occurs, I can propagate the error back.

bhariharan:
Yes - I'm *now* able to trigger IDP flow :-)

libs/SalesforceSDK/AndroidManifest.xml
wmathurin: Couldn't we use one activity for both?
If we need two separate  activities, could we still pick a name that describes what it does?

bhariharan: Well....maybe I could. I'll only know when I implement the flow part :-)

I hesitate to call it `IDPLoginActivity` because that sounds like it's the IDP getting logged in. This generates a code for the SP though. How about `IDPCodeGeneratorActivity`?

wmathurin: 👍

PR #1524, opened 2017-10-25
Adding support for unauthenticatedStartPage hybrid configuration

khawkins:
Per our updated requirements for deferred authentication in hybrid remote apps:

- New property introduced to bootconfig: `unauthenticatedStartPage`.
    - Required for deferred authentication in hybrid remote apps.
    - Ignored in other configurations (local, remote apps with authentication configured.  Framework will log a warning message in these cases).
    - Must be an absolute URL, if present.
    - Use of `unauthenticatedStartPage` vs. `startPage` in a remote app with deferred authentication will be based on the internal authentication state of a user.
        - If a user has not authenticated, the unauthenticated page will be loaded.
        - If the user has authenticated, the original start page will be used.
- Existing `startPage` property represents the "authenticated" path for remote apps, and all paths for local apps.
    - This property is always required.
    - The URL value must be relative.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
@@ -147,14 +147,12 @@ public class BootConfig {
 			}
 
 			// Lack of unauthenticatedStartPage with remote deferred authentication is an error.
-			if (!config.isLocal() && !config.shouldAuthenticate()
-					&& (config.getUnauthenticatedStartPage() == null || config.getUnauthenticatedStartPage().length() == 0)) {
+			if (!config.isLocal() && !config.shouldAuthenticate() && TextUtils.isEmpty(config.getUnauthenticatedStartPage())) {
 				throw new BootConfigException(UNAUTHENTICATED_START_PAGE + " required for remote app with deferred authentication.");
 			}
 
 			// unauthenticatedStartPage, if present, must be an absolute URL.
-			if (config.getUnauthenticatedStartPage() != null
-					&& config.getUnauthenticatedStartPage().length() > 0
+			if (!TextUtils.isEmpty(config.getUnauthenticatedStartPage())
 					&& !BootConfig.isAbsoluteUrl(config.getUnauthenticatedStartPage())) {
 				throw new BootConfigException(UNAUTHENTICATED_START_PAGE + " should be absolute URL.");
 			}
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
@@ -41,6 +41,7 @@ import android.text.TextUtils;
 import com.salesforce.androidsdk.R;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey;
+import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
 /**
  * Class encapsulating the application configuration (consumer key, oauth scopes, refresh behavior).
@@ -48,6 +49,8 @@ import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey;
  * @author wmathurin
  */
 public class BootConfig {
+
+	private static final String TAG = "BootConfig";
 
 	// We expect a assets/www/bootconfig.json file to be provided by hybrid apps.
 	private static final String HYBRID_BOOTCONFIG_PATH = "www" +
@@ -63,10 +66,13 @@ public class BootConfig {
 	private static final String SHOULD_AUTHENTICATE = "shouldAuthenticate";
 	private static final String ATTEMPT_OFFLINE_LOAD = "attemptOfflineLoad";
 	private static final String PUSH_NOTIFICATION_CLIENT_ID = "androidPushNotificationClientId";
+	private static final String UNAUTHENTICATED_START_PAGE = "unauthenticatedStartPage";
 
 	// Default for optional configs.
 	private static final boolean DEFAULT_SHOULD_AUTHENTICATE = true;
 	private static final boolean DEFAULT_ATTEMPT_OFFLINE_LOAD = true;
+
+	private boolean configIsHybrid;
 
 	private String remoteAccessConsumerKey;
 	private String oauthRedirectURI;
@@ -77,8 +83,23 @@ public class BootConfig {
 	private boolean shouldAuthenticate;
 	private boolean attemptOfflineLoad;
 	private String pushNotificationClientId;
+	private String unauthenticatedStartPage;
 
 	private static BootConfig INSTANCE = null;
+
+	BootConfig(JSONObject config, boolean isHybrid) {
+		this(isHybrid);
+		parseBootConfig(config);
+	}
+
+	BootConfig(Context ctx, boolean isHybrid) {
+		this(isHybrid);
+		readFromXML(ctx);
+	}
+
+	private BootConfig(boolean isHybrid) {
+		configIsHybrid = isHybrid;
+	}
 
 	/**
      * Method to (build and) get the singleton instance.
@@ -88,11 +109,12 @@ public class BootConfig {
 	 */
 	public static BootConfig getBootConfig(Context ctx) {
 		if (INSTANCE == null) {
-			INSTANCE = new BootConfig();
-			if (SalesforceSDKManager.getInstance().isHybrid()) {
-				INSTANCE.readFromJSON(ctx);
+			boolean isHybrid = SalesforceSDKManager.getInstance().isHybrid();
+			if (isHybrid) {
+				JSONObject bootConfigObj = readFromJSON(ctx);
+				INSTANCE = new BootConfig(bootConfigObj, isHybrid);
 			} else {
-				INSTANCE.readFromXML(ctx);
+				INSTANCE = new BootConfig(ctx, isHybrid);
 			}
 			INSTANCE.readFromRuntimeConfig(ctx);
 		}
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -77,9 +83,24 @@
 	private boolean shouldAuthenticate;
 	private boolean attemptOfflineLoad;
 	private String pushNotificationClientId;
+	private String unauthenticatedStartPage;
 
 	private static BootConfig INSTANCE = null;
 
+	BootConfig(JSONObject config, boolean isHybrid) {
```
khawkins: I added some package-visible constructors, for unit test support.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -63,11 +66,14 @@
 	private static final String SHOULD_AUTHENTICATE = "shouldAuthenticate";
 	private static final String ATTEMPT_OFFLINE_LOAD = "attemptOfflineLoad";
 	private static final String PUSH_NOTIFICATION_CLIENT_ID = "androidPushNotificationClientId";
+	private static final String UNAUTHENTICATED_START_PAGE = "unauthenticatedStartPage";
 
 	// Default for optional configs.
 	private static final boolean DEFAULT_SHOULD_AUTHENTICATE = true;
 	private static final boolean DEFAULT_ATTEMPT_OFFLINE_LOAD = true;
 
+	private boolean configIsHybrid;
```
khawkins: Added this property for unit test support.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -109,74 +131,89 @@ public static void validateBootConfig(BootConfig config) {
...
-    	String mdmRemoteAccessConsumeKey = runtimeConfig.getString(ConfigKey.ManagedAppOAuthID);
-    	String mdmOauthRedirectURI = runtimeConfig.getString(ConfigKey.ManagedAppCallbackURL);
-    	if (!TextUtils.isEmpty(mdmRemoteAccessConsumeKey)) {
-            remoteAccessConsumerKey = mdmRemoteAccessConsumeKey;
-        }
-    	if (!TextUtils.isEmpty(mdmOauthRedirectURI)) {
-            oauthRedirectURI = mdmOauthRedirectURI;
-        }
+
+	/**
```
khawkins: No code changes here.  Just fixing wonky spacing.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -109,74 +131,89 @@ public static void validateBootConfig(BootConfig config) {
...
-        return config;
+		config.put(IS_LOCAL, isLocal);
+		config.put(START_PAGE, startPage);
+		config.put(ERROR_PAGE, errorPage);
+		if (!TextUtils.isEmpty(pushNotificationClientId)) {
+			config.put(PUSH_NOTIFICATION_CLIENT_ID, pushNotificationClientId);
+		}
+		config.put(SHOULD_AUTHENTICATE, shouldAuthenticate);
+		config.put(ATTEMPT_OFFLINE_LOAD, attemptOfflineLoad);
+		config.put(UNAUTHENTICATED_START_PAGE, unauthenticatedStartPage);
```
khawkins: This is the only line added.  The rest is fixing wonky spacing.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
bhariharan: Do we need this? Can we use `SalesforceSDKManager.getInstance().isHybrid()` instead to make this determination? That flag would automatically be set by the corresponding init call used in that test (`initHybrid()` or `initNative()`).

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
bhariharan: Same question as above. Do we need this?

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
khawkins: It's for testing.  Without it, what can be an isolated unit test in the com.salesforce.androidsdk package, instead requires loading the tests in the context of a SalesforceHybridSDKManager instance, which in turn requires setting up an app context to load the manager.  We could do that, but it seems like overkill for simple unit test functionality that doesn't otherwise depend on any of that setup.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
khawkins: Do you mean the hybrid flag specifically, or the constructors generally?  Without the constructors, BootConfig instantiation is hard-wired into reading from www/bootconfig.json. Given that the BootConfig only exposes read-only values from that process, it's an onerous approach to formulate isolated tests that tweak values in that configuration to test their validity.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
bhariharan: Makes sense.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
bhariharan: Yes - I was referring to the hybrid flag specifically. For example, this constructor makes sense only when `isHybrid` is `true` and the next one makes sense only when it's `false`.

bhariharan: Could we have a package level static method that takes in a filename instead and avoid these constructors? We could revert the changes made to the existing static method and simply call the package level static method from there with the default filename. In the tests, you could pass in a bunch of fake boot configs.

bhariharan: Or a static method that takes `JSONObject`.

khawkins: > Could we have a package level static method that takes in a filename instead and avoid these constructors?

And then the thinking for the tests is canned bootconfigs in the text context, each representing a different test case?  That could be a good approach, so test consumers could see the configuration in its expected form, so to speak.  Let me know if that's what you're getting at, and I can rework it.

bhariharan: Yes - exactly!

khawkins:
@bhariharan I made the changes we talked about in 4ffba821d0c3f85aac1a988ff9c45b417ffac9e5.  There's now a static package-level factory method to create hybrid boot configs based on context and file path.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/config/BootConfigTest.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+public class BootConfigTest extends InstrumentationTestCase {
+
+    private static final String BOOTCONFIG_ASSETS_PATH_PREFIX = "www" + System.getProperty("file.separator");
+    private Context testContext;
+
+    @Override
+    public void setUp() throws Exception {
+        testContext = getInstrumentation().getContext();
```
bhariharan: Don't you need a `super.setUp()` call here?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/config/BootConfigTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public void tearDown() throws Exception {
+        testContext = null;
```
bhariharan: Don't you need a `super.tearDown()` call here at the end?

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -109,74 +130,92 @@ public static void validateBootConfig(BootConfig config) {
...
+			// unauthenticatedStartPage doesn't make sense in a remote setup with authentication.  Warn accordingly.
+			if (!config.isLocal() && config.shouldAuthenticate() && config.getUnauthenticatedStartPage() != null) {
+				SalesforceSDKLogger.w(TAG, UNAUTHENTICATED_START_PAGE + " set for remote app with authentication, but it will never be used.");
+			}
 
-			if (!config.isLocal() && config.shouldAuthenticate() && config.isStartPageAbsoluteUrl()) {
-				throw new BootConfigException("Cannot authenticate using an absolute URL for start page.");
+			// Lack of unauthenticatedStartPage with remote deferred authentication is an error.
+			if (!config.isLocal() && !config.shouldAuthenticate()
+					&& (config.getUnauthenticatedStartPage() == null || config.getUnauthenticatedStartPage().length() == 0)) {
```
bhariharan: Minor: You could use `TextUtils.isEmpty(config.getUnauthenticatedStartPage())` here instead if you like - it checks for `null` or empty string.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/BootConfig.java
```diff
@@ -109,74 +130,92 @@ public static void validateBootConfig(BootConfig config) {
...
+				throw new BootConfigException(UNAUTHENTICATED_START_PAGE + " required for remote app with deferred authentication.");
 			}
 
-			if (!config.isLocal() && !config.shouldAuthenticate() && !config.isStartPageAbsoluteUrl()) {
-				throw new BootConfigException("Cannot configure relative URL start page with deferred authentication.");
+			// unauthenticatedStartPage, if present, must be an absolute URL.
+			if (config.getUnauthenticatedStartPage() != null
```
bhariharan: Minor: You could use `!TextUtils.isEmpty(config.getUnauthenticatedStartPage())` here instead if you like - it checks for `null` or empty string.

khawkins:
Thanks @bhariharan , I've made those changes.

PR #1536, opened 2017-11-02
Adding account and completing flow on the SP side

bhariharan:
This PR closes the loop on the SP side once a `TokenEndpointResponse` is received. It takes this response, hits the identity endpoint, sets PIN policies if they exist, adds the account and switches to it (same as the standard flow). I managed to reuse the existing flow since it was well broken up.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
@@ -36,6 +36,7 @@ import com.salesforce.androidsdk.auth.HttpAccess;
 import com.salesforce.androidsdk.auth.OAuth2;
 import com.salesforce.androidsdk.config.BootConfig;
 import com.salesforce.androidsdk.security.SalesforceKeyGenerator;
+import com.salesforce.androidsdk.ui.LoginActivity;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
```diff
@@ -139,16 +143,16 @@ public TokenEndpointTask(String code, String loginUrl) {
...
         }
 
         @Override
         protected void onPostExecute(OAuth2.TokenEndpointResponse tokenResponse) {
-            /*
-             * TODO: Digest the token response and make a request to the identity URL, then run
-             * post login steps from OAuthWebviewHelper.
-             */
+            if (authCallback != null && tokenResponse != null) {
+                authCallback.receivedTokenResponse(tokenResponse);
```
bhariharan: This sends the response back to `LoginActivity` where we came from.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -398,4 +401,30 @@ public void onReceive(Context context, Intent intent) {
...
+     */
+    public class SPAuthCallback {
+
+        /**
+         * Called when the flow was successful and token response is received.
+         *
+         * @param tokenResponse Token response.
+         */
+        public void receivedTokenResponse(OAuth2.TokenEndpointResponse tokenResponse) {
+            webviewHelper.onAuthFlowComplete(tokenResponse);
```
bhariharan: Once we get the `TokenEndpointResponse`, we simply plug it in to the existing flow and the rest is taken care of automatically since it was well broken up.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -398,4 +401,30 @@ public void onReceive(Context context, Intent intent) {
...
+        }
+
+        /**
+         * Called when the flow was not successful.
+         *
+         * @param errorMessage Error message.
+         */
+        public void receivedErrorResponse(String errorMessage) {
+            Toast.makeText(LoginActivity.this, errorMessage, Toast.LENGTH_LONG).show();
```
bhariharan: Show a `Toast` for errors -> same handling as other errors with login like SSL errors today.

PR #1538, opened 2017-11-07
Adding support for 'user_hint' for IDP initiated flow

bhariharan:
This should close the loop on the IDP feature.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAccountPickerActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAccountPickerActivity.java
@@ -101,19 +101,22 @@ public class IDPAccountPickerActivity extends AccountSwitcherActivity {
         final List<UserAccount> accounts = getAccounts();
         if (accounts == null) {
             usersExist = false;
-            final String userHint = spConfig.getUserHint();
-            if (!TextUtils.isEmpty(userHint)) {
-                final String[] userParts = userHint.split(COLON);
+        }
+        final String userHint = spConfig.getUserHint();
+        if (!TextUtils.isEmpty(userHint)) {
+            final String[] userParts = userHint.split(COLON);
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAccountPickerActivity.java
```diff
@@ -90,9 +91,46 @@ public void onDestroy() {
...
+        /*
+         * If no users exist, launches the new user login flow directly. If we could build a user
+         * account from the 'user_hint' value passed in, launches SP login flow for that account.
+         */
+        if (!usersExist) {
             accountSelected(null);
+        } else {
+            if (selectedAccount != null) {
+                accountSelected(selectedAccount);
+            }
```
wmathurin: Looks like we don't really need usersExists flag. getUserFromOrgAndUserId handles not having any accounts, so couldn't we simply do:

```java
super.onresume();

UserAccount selectedAccount = null;

final String userHint = spConfig.getUserHint();
if (!TextUtils.isEmpty(userHint)) {
    final String[] userParts = userHint.split(COLON);

    /*
     * The value for 'user_hint' should be of the format 'orgId:userId' and should
     * use the 18-character versions of 'orgId' and 'userId'.
     */
    if (userParts.length == 2) {
        final String orgId = userParts[0];
        final String userId = userParts[1];
        selectedAccount = SalesforceSDKManager.getInstance().
            getUserAccountManager().getUserFromOrgAndUserId(orgId, userId);
    }
}

/*
 * If we could build a user account from the 'user_hint' value passed in, launches SP login flow for that account.
 * Otherwise, we launch the new user login flow directly (because selectedAccount will be null).
 */
accountSelected(selectedAccount);
```

bhariharan: There are cases where there is no account selection (more than 1 account), but you're right - it can be simplified. Will do.

PR #1539, opened 2017-11-08
Few corner case bug fixes for IDP flow

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAccountPickerActivity.java
```diff
@@ -153,7 +153,7 @@ protected void accountSelected(UserAccount account) {
          */
         if (account != null) {
             proceedWithIDPAuthFlow(account);
-        } else {
+        } else if (!SalesforceSDKManager.getInstance().isIDPAppLoginFlowActive()) {
```
bhariharan: This one ensures we don't kick off a new login flow when one is active currently. This can occur sometimes because once `LoginActivity` goes away, control comes back on `onResume()` at which point we don't want to kick off another login flow.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAccountPickerActivity.java
```diff
@@ -168,6 +168,7 @@ protected void finishActivity() {
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (requestCode == SPRequestHandler.IDP_REQUEST_CODE) {
             SalesforceSDKLogger.d(TAG, "Activity result obtained - IDP code exchange complete");
+            SalesforceSDKManager.getInstance().setIDPAppLoginFlowActive(false);
```
bhariharan: We want to set this to `false` only when the flow is truly done.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPCodeGeneratorActivity.java
```diff
@@ -121,6 +123,14 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
                 } else {
                     handleError("Code not returned from server");
                 }
+            } else if (url.contains(EC_301) || url.contains(EC_302)) {
```
bhariharan: This one's a bit ugly, but I see no other way until we get an API that's truly a first-class citizen. @trooper2013 You should probably add this error handling on the `iOS` side too.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPRequestHandler.java
```diff
@@ -71,9 +71,9 @@ public IDPRequestHandler(SPConfig spConfig, UserAccount userAccount) throws IDPR
         this.spConfig = spConfig;
         this.userAccount = userAccount;
         basicValidation();
-        loginUrl = spConfig.getLoginUrl();
+        loginUrl = userAccount.getLoginServer();
```
bhariharan: This one fixes a corner case bug when you swizzle out to the IDP and then decide to change login hosts.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
```diff
@@ -63,9 +63,20 @@
...
+    }
+
+    /**
+     * Parameterized constructor.
+     *
+     * @param loginUrl Login URL.
+     * @param userHint User hint. Must be of the format 'orgId:userId', both being 18-char IDs.
+     * @param authCallback Auth callback.
+     */
+    public SPRequestHandler(String loginUrl, String userHint, LoginActivity.SPAuthCallback authCallback) {
```
bhariharan: Added a convenience constructor to pass in `user_hint`. Need this for IDP initiated flow.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -400,7 +403,6 @@ public void finish(UserAccount userAccount) {
             intent.putExtra(IDPAccountPickerActivity.USER_ACCOUNT_KEY, userAccount.toBundle());
             sendBroadcast(intent);
         }
-        SalesforceSDKManager.getInstance().setIDPAppLoginFlowActive(false);
```
bhariharan: We want to set this to `false` only when the flow is truly done.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -146,6 +146,9 @@ protected void onDestroy() {
 	@Override
     protected void onNewIntent(Intent intent) {
         super.onNewIntent(intent);
+
+        // Reloads login page for every new intent to ensure the correct login server is selected.
+        webviewHelper.loadLoginPage();
```
bhariharan: Since this activity is a single task activity, `onCreate()` is triggered only once. Hence, we should update the login host here to the current value to ensure that we don't end up with a stale login host while coming in from the SP app. This can occur in some super corner cases without this fix.

PR #1540, opened 2017-11-09
Static config for setting up syncs

wmathurin:
Two new methods on SmartSyncSDKManager:
* `setupGlobalSyncsFromDefaultConfig()`
Call at any time to create syncs defined in `/res/raw/globalsyncs.json` into the default global smartstore.

* `setupUserSyncsFromDefaultConfig()`
Call once a user has logged in to create syncs defined in `/res/raw/usersyncs.json` into the user's default smartstore.

Notes:
* The JSON format is based on what we use in our Cordova plugin and React Native module.
* It just does a bunch of `SyncState.createSyncDown` and `SyncState.createSyncUp` calls, it will skip a sync configuration if there is already one by the same name.
* The only feedback is through the logs - that's intentional, those config files are static (can't be changed at runtime) so the developer should get them right while developing the app.

The sample application SmartSyncExplorer now defines its syncs in a config file.
I also added a new test suite called SyncsConfigTest.

Example of JSON config file
```json
{
  "syncs": [
    {
      "syncName": "syncDownContacts",
      "syncType": "syncDown",
      "soupName": "contacts",
      "target": {"type":"soql", "query":"SELECT FirstName, LastName, Title, MobilePhone, Email, Department, HomePhone FROM Contact LIMIT 10000"},
      "options": {"mergeMode":"OVERWRITE"}
    },
    {
      "syncName": "syncUpContacts",
      "syncType": "syncUp",
      "soupName": "contacts",
      "target": {"createFieldlist":["FirstName", "LastName", "Title", "MobilePhone", "Email", "Department", "HomePhone"]},
      "options": {"fieldlist":["Id", "FirstName", "LastName", "Title", "MobilePhone", "Email", "Department", "HomePhone"], "mergeMode":"LEAVE_IF_CHANGED"}
    }
  ]
}
```


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/ConfigHelper.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.InputStreamReader;
+import java.io.StringWriter;
+import java.io.Writer;
+
+/**
+ * Helper class for reading config files
+ *
+ */
+
+public class ConfigHelper {
```
wmathurin: Helper for StoreConfig and the new SyncsConfig

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/config/StoreConfig.java
```diff
@@ -99,6 +94,13 @@ public void registerSoups(SmartStore store) {
             try {
                 JSONObject soupConfig = soupsConfig.getJSONObject(i);
                 String soupName = soupConfig.getString(SOUP_NAME);
+
+                // Leaving soup alone if it already exists
```
wmathurin: Not changing the behavior here - the registerSoup that comes a few lines later would not do anything if the soup already exists. But that way, it's clearer to the developer looking at the logs that the soup was not recreated.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/config/SyncsConfig.java
```diff
@@ -0,0 +1,135 @@
...
+            return;
+
+        for (int i=0; i<syncsConfig.length(); i++) {
+            try {
+                JSONObject syncConfig = syncsConfig.getJSONObject(i);
+
+                String syncName = syncConfig.getString(SYNC_NAME);
+
+                // Leaving sync alone if it already exists
+                if (SyncState.hasSyncWithName(store, syncName)) {
```
wmathurin: If you change the static config, you need to logout / login.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/config/SyncsConfigTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public void tearDown() throws Exception {
+        deleteSyncs();
+        deleteGlobalSyncs();
+        super.tearDown();
+    }
+
+
+    public void testSetupGlobalSyncsFromDefaulltConfig() throws JSONException {
```
wmathurin: New tests.

native/NativeSampleApps/SmartSyncExplorer/src/com/salesforce/samples/smartsyncexplorer/ui/DetailActivity.java
```diff
@@ -153,29 +154,47 @@ public void onLoaderReset(Loader<ContactObject> loader) {
...
+			deleteConfirmationDialog.show(getFragmentManager(), "DeleteDialog");
+		}
 	}
 
 	/**
-	 * Performs the underlying deletion of a record.
+	 * Performs the underlying delete or undelete of a record.
 	 */
-	public void reallyDelete() {
+	public void deleteOrUndelete() {
```
wmathurin: See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1540/commits/8ba3c6ded16f0147f3d94d0225ca8141c9205d3d for details - now behaving like
hybrid and react native (and iOS) smart sync explorers: showing undelete for already deleted records, doing a db delete when deleting locally created records.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/ConfigHelper.java
bhariharan: Should we call this something a little more generic, like `ResourceReaderUtil`? Should this be in the `util` package?

wmathurin: Good idea. I’ll make the change.

PR #1542, opened 2017-11-10
Adding helpers for IDP initiated auth flow

bhariharan:
IDP initiated flow works end-to-end now.

Here's the code you need on the IDP side to trigger this flow:
```
final Intent intent = new Intent(IDPInititatedLoginReceiver.IDP_LOGIN_REQUEST_ACTION);
intent.putExtra(IDPInititatedLoginReceiver.USER_HINT_KEY, userHint);
sendBroadcast(intent);
```

This will be in the IDP template app.

### DIFF
--- a/libs/SalesforceSDK/AndroidManifest.xml
+++ b/libs/SalesforceSDK/AndroidManifest.xml
@@ -90,6 +90,15 @@
             android:theme="@style/SalesforceSDK.ActionBarTheme"
             android:exported="false" />
 
+        <!-- Receiver for IDP initiated login flow -->
+        <receiver android:name="com.salesforce.androidsdk.auth.idp.IDPInititatedLoginReceiver"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="com.salesforce.IDP_LOGIN_REQUEST" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
+        </receiver>
+
         <!-- Google Play Services Push Registration -->
         <!--
             Push notification services and receivers. The 'category' attribute
@@ -110,12 +119,14 @@
                 <action android:name="com.google.android.c2dm.intent.RECEIVE" />
             </intent-filter>
         </service>
+
         <service android:name="com.salesforce.androidsdk.push.SFDCInstanceIDListenerService"
             android:exported="false">
             <intent-filter>
                 <action android:name="com.google.android.gms.iid.InstanceID" />
             </intent-filter>
         </service>
+
         <service android:name="com.salesforce.androidsdk.push.SFDCRegistrationIntentService"
             android:exported="false">
         </service>
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPInititatedLoginReceiver.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPInititatedLoginReceiver.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2017-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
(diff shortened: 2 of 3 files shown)


libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -90,6 +90,15 @@
             android:theme="@style/SalesforceSDK.ActionBarTheme"
             android:exported="false" />
 
+        <!-- Receiver for IDP initiated login flow -->
+        <receiver android:name="com.salesforce.androidsdk.auth.idp.IDPInititatedLoginReceiver"
```
bhariharan: A simple `BroadcastReceiver` on the SP side to receive IDP auth requests.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPInititatedLoginReceiver.java
```diff
@@ -0,0 +1,102 @@
...
+    public static final String USER_HINT_KEY = "user_hint";
+    public static final String IDP_INIT_LOGIN_KEY = "idp_init_login";
+    private static final String COLON = ":";
+
+    private String userHint;
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        if (intent != null) {
+            if (IDP_LOGIN_REQUEST_ACTION.equals(intent.getAction())) {
```
bhariharan: Handles only the IDP auth action.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPInititatedLoginReceiver.java
```diff
@@ -0,0 +1,102 @@
...
+                final Bundle extras = intent.getExtras();
+                if (extras != null) {
+                    userHint = extras.getString(USER_HINT_KEY);
+                }
+
+                // Launches login flow if the user doesn't already exist on the SP app.
+                if (!doesUserExist()) {
```
bhariharan: Checks if the user passed in already exists on the SP app. If it does, there's nothing to do.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPInititatedLoginReceiver.java
```diff
@@ -0,0 +1,102 @@
...
+                    launchLoginActivity();
```
bhariharan: If not, launches login screen on the SP app.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -151,6 +153,18 @@ protected void onNewIntent(Intent intent) {
...
         // Reloads login page for every new intent to ensure the correct login server is selected.
         webviewHelper.loadLoginPage();
+
+        // Launches IDP login flow directly for IDP initiated login flow.
+        if (intent != null) {
+            final Bundle extras = intent.getExtras();
+            if (extras != null) {
+                userHint = extras.getString(IDPInititatedLoginReceiver.USER_HINT_KEY);
+                boolean isIdpInitFlow = extras.getBoolean(IDPInititatedLoginReceiver.IDP_INIT_LOGIN_KEY);
+                if (isIdpInitFlow) {
```
bhariharan: If this is an IDP initiated auth request, take `user_hint` and directly flow back to the IDP app.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPInititatedLoginReceiver.java
wmathurin: It feels like something is missing. The IDP should what we want the SP app to do once we have the desired user.
It would be like the returl on the server side (when you have to go through login in the middle of a flow but and want the flow to resume).

wmathurin: @trooper2013 what do you do on the iOS side?

bhariharan: Yeah, makes sense. I was going to enhance this in the next PR to take in an optional component/intent from the IDP that can be launched if present. Right now it ends back on the IDP since `LoginActivity` will be popped off the stack once logged in.

wmathurin: Sounds good.

PR #1545, opened 2017-11-11
PR CI Improvments

brandonpage:
- Run Tests Conditionally, based on modified files
- Fix Danger for all users

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,11 @@ jobs:
           command:  |
             LIBS_TO_TEST=$(ruby .circleci/gitChangedLibs.rb)
             echo "export LIBS_TO_TEST=${LIBS_TO_TEST}" >> "${BASH_ENV}"
-            echo -e "\n\nLibraries to Test-> ${LIBS_TO_TEST}"
+            if [ ! -z ${LIBS_TO_TEST} ]; then
+              echo -e "\n\nLibraries to Test-> ${LIBS_TO_TEST}"
+            else
+              echo -e "\n\nNo libraries to run"
+            fi
 
       - restore_cache:
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "build.gradle" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,8 +119,11 @@ jobs:
 
       - run:
           name: Combine lib Test Restuls
-          command: junit-merge libs/**/build/outputs/androidTest-results/connected/*.xml
           when: always
+          command: |
+            if [ ! -z ${LIBS_TO_TEST} ]; then
+              junit-merge libs/**/build/outputs/androidTest-results/connected/*.xml
+            fi
       - run:
           name:  Run Danger
           when: always
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,14 @@ jobs:
 
       - checkout
 
+      - run:
+          name: Determine Tests to Run
+          shell: /bin/bash
+          command:  |
+            LIBS_TO_TEST=$(ruby .circleci/gitChangedLibs.rb)
+            echo "export LIBS_TO_TEST=${LIBS_TO_TEST}" >> "${BASH_ENV}"
+            echo -e "\n\nLibraries to Test-> ${LIBS_TO_TEST}"
+
       - restore_cache:
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "build.gradle" }}
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -22,6 +22,14 @@ jobs:
 
       - checkout
 
+      - run:
+          name: Determine Tests to Run
+          shell: /bin/bash
+          command:  |
+            LIBS_TO_TEST=$(ruby .circleci/gitChangedLibs.rb)
+            echo "export LIBS_TO_TEST=${LIBS_TO_TEST}" >> "${BASH_ENV}"
+            echo -e "\n\nLibraries to Test-> ${LIBS_TO_TEST}"
```
brandonpage: Print libraries to be tested to quick reference while CI is running.

.circleci/config.yml
```diff
@@ -40,48 +48,88 @@ jobs:
...
           name: Combine lib Test Restuls
           command: junit-merge libs/**/build/outputs/androidTest-results/connected/*.xml
           when: always
       - run:
           name:  Run Danger
-          command: danger --dangerfile=.circleci/Dangerfile --danger_id=ci/circleci
           when: always
+          command: |
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              DANGER_GITHUB_API_TOKEN="c21349d8a97e1bf9cdd9""301fd949a83db862216b" danger --dangerfile=.circleci/Dangerfile --danger_id=ci/circleci
```
brandonpage: Unfortunately, this appears to be the only workaround to get danger to run consistently in CircleCI.  The token here only has public_repo access so if someone takes it I don't foresee them being able to do anything nefarious.

.circleci/gitChangedLibs.rb
```diff
@@ -0,0 +1,28 @@
...
+  path = prfile["filename"]
+  for lib in $libs
+    if path.include? lib
+      libs = libs.add(lib)
+    end
+  end
+end
+
+libs_ar = libs.to_a()
+if !libs_ar.empty?() && libs_ar.include?("SalesforceSDK")
```
brandonpage: If files are chanced in Core, run all tests.

.circleci/gitChangedLibs.rb
```diff
@@ -0,0 +1,28 @@
...
+  libs_ar = $libs.to_a()
+elsif libs_ar.include?("SmartStore") && !libs_ar.include?("SmartSync")
```
brandonpage: If changes in SmartStore, test SmartSync too.

.circleci/gitChangedLibs.rb
```diff
@@ -0,0 +1,28 @@
...
+  libs_ar.push("SmartSync")
+end
```
brandonpage: Any other dependencies I missed?

brandonpage:
If none of the libraries are changed should we run all tests?  Or none?  

Right now I have it running none because the only scenarios I could think of are: CI updates, Readme update, test app changes, configuration updates.

wmathurin:
SalesforceHybrid and SalesforceReact both depend on SmartSync.

brandonpage:
## Danger has errored
[!] Invalid `Dangerfile` file: No JUnit file was found at merged-test-results.xml. Updating the Danger gem might fix the issue. Your Danger version: 5.5.3, latest Danger version: 5.5.4

```
 #  from .circleci/Dangerfile:1
 #  -------------------------------------------
 >  junit.parse "merged-test-results.xml"
 #  junit.show_skipped_tests = true
 #  -------------------------------------------
```

<p align="right" data-meta="generated_by_ci/circleci">
</p>

.circleci/gitChangedLibs.rb
wmathurin: SalesforceSDK depends on SalesforceAnalytics (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceSDK/build.gradle#L17).
So line 22, we should test for SalesforceAnalytics.
We should have another elsif for SalesforceSDKCore that would push SmartStore, SmartSync, SalesforceHybridSDK, SalesforceReact.

wmathurin: We could simplify the code:

```ruby
libsTopoSorted = ["SalesforceAnalytics", "SalesforceSDK", "SmartStore", "SmartSync", "SalesforceHybridSDK", "SalesforceReact"] 
libsModified = libs.to_a()
libsToTest = libsTopoSorted.slice(libsModified.map{|l| libsTopoSorted.find_index(l)}.min() .. libsTopoSorted.length)
```
libsTopoSorted is topologically sorted - x before y means x does not depend on y.
It looks for the modified library that has the smallest index in libsTopoSorted.
Then tests that library and all the ones that follow in libsTopoSorted.

PR #1547, opened 2017-11-14
Fixing a couple of issues

bhariharan:
- `ec` may not be the first parameter.
- Removed flapping test due to platform search indexing time.

### DIFF
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
@@ -36,7 +36,6 @@ import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse;
 import com.salesforce.androidsdk.rest.RestClient.AuthTokenProvider;
 import com.salesforce.androidsdk.rest.RestClient.ClientInfo;
 import com.salesforce.androidsdk.rest.RestRequest.RestMethod;
-import com.salesforce.androidsdk.util.JSONObjectHelper;
 
 import org.json.JSONArray;
 import org.json.JSONException;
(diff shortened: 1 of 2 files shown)


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -576,24 +574,6 @@ public void testQuery() throws Exception {
...
 
     /**
-     * Testing a search call to the server.
-     * Create new account then look for it using sosl.
-     * @throws Exception
-     */
-    public void testSearch() throws Exception {
-        IdName idNameFirstAccount = createAccount();
-        IdName idNameSecondAccount = createAccount();
-        RestResponse response = restClient.sendSync(RestRequest.getRequestForSearch(TestCredentials.API_VERSION, "find {" + ENTITY_NAME_PREFIX + "}"));
```
wmathurin: Why are we killing that test?
If it flaps - we could wait a bit longer (the problem with search is that the update of the search indexes is asynchronous).

bhariharan: Yeah - adding time isn't helping much - it's still unpredictable. Based on my discussions with the search team, there's no guaranteed time taken for `SOLR indexing`.

wmathurin: Could we still run a search and make sure we don't get a server error - without actually checking the individual records returned? That way we would catch a regression in the soql rest api wrapper if they are ever changed.

bhariharan: Yeah, that's a good idea. As long as we don't check the number of records returned, we should be okay.

PR #1548, opened 2017-11-14
Adding search test back

bhariharan:
@wmathurin


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -588,7 +588,20 @@ public void testDoubleConsume() throws Exception {
         	fail("Calling consume should not have thrown an exception");
         }
     }
-    
+
+    /**
+     * Testing a search call to the server.
+     * Create new account then look for it using sosl.
```
wmathurin: Could we remove that line from the comment

bhariharan: Sure.

PR #1570, opened 2017-12-18
UserAccount null check in SyncManager.reset(UserAccount)

narakim:
Copied the similar null checks from other reset methods;
MetadataManager.reset(userAccount);
CacheManager.softReset(userAccount);

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -180,17 +180,19 @@ public class SyncManager {
      * @param account User account.
      */
     public static synchronized void reset(UserAccount account) {
-        Set<String> keysToRemove = new HashSet<>();
-        for (String key : INSTANCES.keySet()) {
-            if (key.startsWith(account.getUserId())) {
-                keysToRemove.add(key);
-                SyncManager syncManager = INSTANCES.get(key);
-                syncManager.threadPool.shutdownNow();
-            }
-        }
-        // NB: keySet returns a Set view of the keys contained in this map.
-        // The set is backed by the map, so changes to the map are reflected in the set, and vice-versa.
-        INSTANCES.keySet().removeAll(keysToRemove);
+	    if (account != null) {
+	        Set<String> keysToRemove = new HashSet<>();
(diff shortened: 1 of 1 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -180,17 +180,22 @@ public static synchronized void reset() {
      * @param account User account.
      */
     public static synchronized void reset(UserAccount account) {
-        Set<String> keysToRemove = new HashSet<>();
-        for (String key : INSTANCES.keySet()) {
-            if (key.startsWith(account.getUserId())) {
-                keysToRemove.add(key);
-                SyncManager syncManager = INSTANCES.get(key);
-                syncManager.threadPool.shutdownNow();
+        if (account == null) {
```
bhariharan: @narakim I don't think we should do this. This API specifically expects a valid `account` instance.  Why are we substituting it with `currentUser` if an `account` isn't passed in? I'm okay with the `null` check below though.

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Tests have failed, see below for more information.</td>
    </tr>
  </tbody>
</table>

### Tests: 

Name | Classname | Time|
--- | --- | ---|
testUpsertConcurrentEntries | com.salesforce.androidsdk.phonegap.SmartStoreLoadJSTest | 2.066|
testSyncSObjectDetectConflictUpdate | com.salesforce.androidsdk.phonegap.SmartSyncJSTest | 0.378|
testSyncUpLocallyDeleted | com.salesforce.androidsdk.phonegap.SmartSyncJSTest | 0.379|


<p align="right" data-meta="generated_by_ci/circleci">
</p>

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
narakim: Will do! Thank you Bharath!

PR #1591, opened 2018-01-11
Provided an option to generate a RSA-2048 keypair 

Songroid:
Option to get an encoded public key string.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -163,7 +163,7 @@ public class SalesforceKeyGenerator {
     /**
      * Generate a keypair and get the encoded public key string.
      *
-     * @param name Unique name associated with this unique ID.
+     * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
      * @return RSA public key string.
      */
     @Nullable
@@ -190,7 +190,7 @@ public class SalesforceKeyGenerator {
     /**
      * Get the private key from the generated key pair.
      *
-     * @param name Unique name associated with this unique ID.
+     * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
      * @return A private key for decryption.
      */
     @Nullable
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -31,33 +31,22 @@ import android.os.Build;
 import android.security.KeyPairGeneratorSpec;
 import android.security.keystore.KeyGenParameterSpec;
 import android.security.keystore.KeyProperties;
-import android.support.annotation.Nullable;
 import android.text.TextUtils;
 import android.util.Base64;
 
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
-import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.security.KeyFactory;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
-import java.security.KeyStore;
-import java.security.KeyStoreException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.security.NoSuchProviderException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.SecureRandom;
-import java.security.cert.CertificateException;
-import java.security.spec.InvalidKeySpecException;
-import java.security.spec.KeySpec;
-import java.security.spec.PKCS8EncodedKeySpec;
-import java.security.spec.X509EncodedKeySpec;
 import java.util.Calendar;
 import java.util.HashMap;
 import java.util.Locale;
@@ -166,7 +155,6 @@ public class SalesforceKeyGenerator {
      * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
      * @return RSA public key string.
      */
-    @Nullable
     public static synchronized String getRSAPublicString(String name, int length) {
         final SharedPreferences prefs = SalesforceSDKManager.getInstance().getAppContext().getSharedPreferences(ASYMM_SHARED_PREF_FILE, 0);
         final String id = prefs.getString(getSharedPrefKey(name), null);
@@ -193,7 +181,6 @@ public class SalesforceKeyGenerator {
      * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
      * @return A private key for decryption.
      */
-    @Nullable
     public static synchronized PrivateKey getRSAPrivateKey(String name, int length) {
         PrivateKey privateKey = null;
         KeyPair keyPair = getKeyPair(name, length);
@@ -272,6 +259,7 @@ public class SalesforceKeyGenerator {
                         .build());
                 kp = kpg.generateKeyPair();
             } else {
+                // TODO: Remove the 'else' block once minVersion > 23.
                 Calendar start = Calendar.getInstance();
                 Calendar end = Calendar.getInstance();
                 end.add(Calendar.YEAR, 30);
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -37,6 +37,7 @@ import android.util.Base64;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
+import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">📱 Needs testing on a real device if change is non-trivial.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_PR-Check">
</p>

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +139,33 @@ public static String getSHA256Hash(String privateKey) {
         return hashedString;
     }
 
+    /**
+     * Generate a RSA-2048 keypair and get the encoded public key string
+     *
+     * @return RSA-2048 public key string
+     */
+    @Nullable
+    public static synchronized String getRSA2048PublicString(String name) {
```
khawkins: Why does there need to be an API for a specific key length?  We wouldn't this just be `getRSAPublicString(name, keyLength)`?

khawkins: Where does the private key go?  How is the public key useful without its counterpart in existence somewhere?

bhariharan: I agree with @khawkins. Let's create `getRSAPublicString(name, keyLength)` instead.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +139,33 @@ public static String getSHA256Hash(String privateKey) {
...
+        final SharedPreferences prefs = SalesforceSDKManager.getInstance().getAppContext().getSharedPreferences(SHARED_PREF_FILE, 0);
```
bhariharan: I'd prefer not to reuse the existing shared pref file that houses symmetric encryption keys for this. Let's create a new one.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +139,33 @@ public static String getSHA256Hash(String privateKey) {
...
+        final String id = prefs.getString(getSharedPrefKey(name), null);
+
+        if (id != null) {
+            UNIQUE_IDS.put(name, id);
```
bhariharan: In line with the above comment, let's use a new `HashMap`, not reuse `UNIQUE_IDS`. Remember there could be collisions in `name` and the algorithm used is different.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +139,33 @@ public static String getSHA256Hash(String privateKey) {
...
+        } else {
+            String publicKeybase64 = null;
+            KeyPair keyPair = getRSA2048KeyPair();
+
+            if (keyPair != null) {
+                PublicKey publicKey = keyPair.getPublic();
+                publicKeybase64 = Base64.encodeToString(publicKey.getEncoded(), Base64.NO_WRAP);
+            }
+            prefs.edit().putString(getSharedPrefKey(name), publicKeybase64).commit();
+            UNIQUE_IDS.put(name, publicKeybase64);
```
bhariharan: I'm guessing you want to commit the `PrivateKey` to the shared pref file too.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -190,4 +221,17 @@ private static String getAddendum(String name) {
         final String suffix = TextUtils.isEmpty(name) ? "" : name;
         return String.format(Locale.US, ADDENDUM, suffix);
     }
+
+    @Nullable
```
bhariharan: Remove this annotation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
Songroid: Yeah I can make the change. Currently RSA-2048 is the only usage for the push notification encryption. The private key will be stored and handled locally by the system. I will add the method to get it and make it ready for decryption.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -190,4 +221,17 @@ private static String getAddendum(String name) {
...
+    private static KeyPair getRSA2048KeyPair() {
+        KeyPair kp = null;
```
bhariharan: I think this can be moved into the `else` block in the other method. Not a whole lot of value in splitting this piece out as a `private` method.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +160,49 @@ public static String getSHA256Hash(String privateKey) {
         return hashedString;
     }
 
+    /**
+     * Generate a keypair and get the encoded public key string.
+     *
+     * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
+     * @return RSA public key string.
+     */
+    @Nullable
```
bhariharan: Remove `@Nullable` annotation please. We don't use annotations from the `support` library in our SDK.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +160,49 @@ public static String getSHA256Hash(String privateKey) {
...
+        }
+    }
+
+    /**
+     * Get the private key from the generated key pair.
+     *
+     * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
+     * @return A private key for decryption.
+     */
+    @Nullable
```
bhariharan: Remove `@Nullable`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -190,4 +258,38 @@ private static String getAddendum(String name) {
...
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                KeyPairGenerator kpg = KeyPairGenerator.getInstance(
+                        KeyProperties.KEY_ALGORITHM_RSA, ANDROID_KEYSTORE);
+                kpg.initialize(new KeyGenParameterSpec.Builder(
+                        name,
+                        KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT)
+                        .setKeySize(length)
+                        .build());
+                kp = kpg.generateKeyPair();
+            } else {
```
bhariharan: Add this comment here:
```
TODO: Remove the 'else' block once minVersion > 23.
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +149,47 @@ public static String getSHA256Hash(String privateKey) {
...
+
+        if (id != null) {
+            RSA_IDS.put(name, id);
+            return id;
+        } else {
+            String publicKeyBase64 = null;
+            KeyPair keyPair = getKeyPair(name, length);
+            if (keyPair != null) {
+                PublicKey publicKey = keyPair.getPublic();
+                publicKeyBase64 = Base64.encodeToString(publicKey.getEncoded(), Base64.NO_WRAP);
```
bhariharan: Use `Base64.NO_WRAP | Base64.NO_PADDING | Base64.URL_SAFE` in the flags here, since we will be sending this to the server.

bhariharan:
Spoke with @Songroid offline. We're going to try to use `Keystore` instead of `SharedPreferences` for this.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +154,47 @@ public static String getSHA256Hash(String privateKey) {
...
+     * Get the private key from the generated key pair.
+     *
+     * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
+     * @return A private key for decryption.
+     */
+    public static synchronized PrivateKey getRSAPrivateKey(String name, int length) {
+        PrivateKey privateKey = null;
+        createRSAKeys(name, length);
+        try {
+            KeyStore.Entry entry = loadKeyStore().getEntry(name, null);
```
bhariharan: Shouldn't we try to call `getEntry(name, null)` first before we call `createRSAKeys`? If we have already created the keys they will exist in the `KeyStore`. However, if that call returns `null` or throws an `Exception`, then we should call `createRSAKeys`. If `createRSAKeys` returns `KeyPair` that would plug in here directly in this case.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -135,6 +154,47 @@ public static String getSHA256Hash(String privateKey) {
...
+     * Generate a keypair and get the encoded public key string.
+     *
+     * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
+     * @return RSA public key string.
+     */
+    public static synchronized String getRSAPublicString(String name, int length) {
+        String publicKeyBase64 = null;
+        createRSAKeys(name, length);
+        try {
+            publicKeyBase64 = Base64.encodeToString(loadKeyStore().getCertificate(name).getPublicKey().getEncoded(),
```
bhariharan: Same comment as below. I think this line should be called before `createRSAKeys` in the event that it has already been created.

PR #1592, opened 2018-01-11
Fixed the issue that the response can be consumed multiple times.

Taost:
- put 'consumed = true' into the finally block, so that it will be updated even when there is any exception in the try block. As a result, the same response won't be consumed again when consume() is called later on.
- updated consumeQuietly() to make sure it will swallow any exception.


Taost:
@bhariharan The 'ci/circleci: setup' test failed. Any idea? How do I re-trigger the test?

bhariharan:
@Taost I'll kick off a new build.

Taost:
@bhariharan It seems the tests are stuck.

PR #1603, opened 2018-01-18
Fix issue with decryption using RSA keypair

tsunami0ne:
The encryption padding needs to be set in order for the private key to be used to decrypt data.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -264,6 +264,7 @@ public class SalesforceKeyGenerator {
                             name,
                             KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT)
                             .setKeySize(length)
+                            .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1)
                             .build());
                     kpg.generateKeyPair();
                 } else {


tsunami0ne:
I wrote a simple unit test for this, but decided not to check it in because it required me to add code in a lot of different places outside of the unit test. This is what I did:

https://gist.github.com/tsunami0ne/5b987ff3cd2126cdd7573c0cca00cf96

The reason I added a "KEY_4" is because the keypairs aren't cleared from the Android keystore before every test, so if you had run the generateRSAKeysIfNecessary() code before, the old keypair would just be re-used and always fail this test.

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><a href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/e56f5f805c66425322f8a59cad55795d0188232a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java#L239">libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java#L239</a> - Consider using <code>apply()</code> instead; <code>commit</code> writes its data to persistent storage immediately, whereas <code>apply</code> will handle it in the background</td>
    </tr>
  </tbody>
</table>

</p>

Songroid:
Do we need the padding applied to the <M as well?

Songroid:
Okay. Eventually we have to test them anyways.

PR #1610, opened 2018-01-28
Adding support for runtime config of browser login through SSO auth config

bhariharan:
Reads from `.well-known/auth-configuration`.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -77,7 +77,6 @@ import com.salesforce.androidsdk.ui.DevInfoActivity;
 import com.salesforce.androidsdk.ui.LoginActivity;
 import com.salesforce.androidsdk.ui.PasscodeActivity;
 import com.salesforce.androidsdk.ui.SalesforceR;
-import com.salesforce.androidsdk.util.AuthConfigUtil;
 import com.salesforce.androidsdk.util.EventsObservable;
 import com.salesforce.androidsdk.util.EventsObservable.EventType;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
@@ -467,21 +466,6 @@ public class SalesforceSDKManager {
         if (!TextUtils.isEmpty(idpAppUrlScheme)) {
             INSTANCE.idpAppURIScheme = idpAppUrlScheme;
         }
-
-        // Kicks off a thread to fetch auth config for the current server.
-        (new Thread(new Runnable() {
-
-            @Override
-            public void run() {
-                final String loginServer = SalesforceSDKManager.getInstance().getLoginServerManager().getSelectedLoginServer().url;
-                final AuthConfigUtil.MyDomainAuthConfig authConfig = AuthConfigUtil.getMyDomainAuthConfig(loginServer);
-                boolean browserLoginEnabled = false;
-                if (authConfig != null) {
-                    browserLoginEnabled = authConfig.isBrowserLoginEnabled();
-                }
-                SalesforceSDKManager.getInstance().setBrowserLoginEnabled(browserLoginEnabled);
-            }
-        })).start();
     }
 
     /**
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -77,6 +77,7 @@ import com.salesforce.androidsdk.ui.DevInfoActivity;
 import com.salesforce.androidsdk.ui.LoginActivity;
 import com.salesforce.androidsdk.ui.PasscodeActivity;
 import com.salesforce.androidsdk.ui.SalesforceR;
+import com.salesforce.androidsdk.util.AuthConfigUtil;
 import com.salesforce.androidsdk.util.EventsObservable;
 import com.salesforce.androidsdk.util.EventsObservable.EventType;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
@@ -466,6 +467,21 @@ public class SalesforceSDKManager {
         if (!TextUtils.isEmpty(idpAppUrlScheme)) {
             INSTANCE.idpAppURIScheme = idpAppUrlScheme;
         }
+
+        // Kicks off a thread to fetch auth config for the current server.
+        (new Thread(new Runnable() {
+
+            @Override
+            public void run() {
+                final String loginServer = SalesforceSDKManager.getInstance().getLoginServerManager().getSelectedLoginServer().url;
+                final AuthConfigUtil.SSOAuthConfig authConfig = AuthConfigUtil.getSSOAuthConfig(loginServer);
+                boolean browserLoginEnabled = false;
+                if (authConfig != null) {
+                    browserLoginEnabled = authConfig.isBrowserLoginEnabled();
+                }
+                SalesforceSDKManager.getInstance().setBrowserLoginEnabled(browserLoginEnabled);
+            }
+        })).start();
     }
 
     /**
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -474,7 +474,7 @@ public class SalesforceSDKManager {
             @Override
             public void run() {
                 final String loginServer = SalesforceSDKManager.getInstance().getLoginServerManager().getSelectedLoginServer().url;
-                final AuthConfigUtil.SSOAuthConfig authConfig = AuthConfigUtil.getSSOAuthConfig(loginServer);
+                final AuthConfigUtil.MyDomainAuthConfig authConfig = AuthConfigUtil.getMyDomainAuthConfig(loginServer);
                 boolean browserLoginEnabled = false;
                 if (authConfig != null) {
                     browserLoginEnabled = authConfig.isBrowserLoginEnabled();
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
@@ -129,6 +129,9 @@ public class SPRequestHandler {
 
     private void handleError(String error) {
         SalesforceSDKLogger.e(TAG, "Error received from IDP app: " + error);
+        if (authCallback != null) {
+            authCallback.receivedErrorResponse(error);
+        }
     }
 
     private void handleSuccess(String code, String loginUrl) {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
@@ -62,6 +62,7 @@ import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey;
 import com.salesforce.androidsdk.rest.ClientManager.LoginOptions;
 import com.salesforce.androidsdk.security.PasscodeManager;
 import com.salesforce.androidsdk.ui.OAuthWebviewHelper.OAuthWebviewHelperEvents;
+import com.salesforce.androidsdk.util.AuthConfigTask;
 import com.salesforce.androidsdk.util.EventsObservable;
 import com.salesforce.androidsdk.util.EventsObservable.EventType;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
@@ -79,7 +80,7 @@ import java.util.Map;
  * The bulk of the work for this is actually managed by OAuthWebviewHelper class.
  */
 public class LoginActivity extends AccountAuthenticatorActivity
-		implements OAuthWebviewHelperEvents {
+		implements OAuthWebviewHelperEvents, AuthConfigTask.AuthConfigCallbackInterface {
 
     public static final int PICK_SERVER_REQUEST_CODE = 10;
     private static final String TAG = "LoginActivity";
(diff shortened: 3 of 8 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -466,6 +467,21 @@ public static void initInternal(Context context) {
         if (!TextUtils.isEmpty(idpAppUrlScheme)) {
             INSTANCE.idpAppURIScheme = idpAppUrlScheme;
         }
+
+        // Kicks off a thread to fetch auth config for the current server.
+        (new Thread(new Runnable() {
+
+            @Override
+            public void run() {
+                final String loginServer = SalesforceSDKManager.getInstance().getLoginServerManager().getSelectedLoginServer().url;
```
bhariharan: This runs at the end of `SalesforceSDKManager.init()` to update settings for the current server when the app starts up.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java
```diff
@@ -129,6 +129,9 @@ private SPConfig buildSPConfig(String loginUrl, String userHint) {
 
     private void handleError(String error) {
         SalesforceSDKLogger.e(TAG, "Error received from IDP app: " + error);
+        if (authCallback != null) {
```
bhariharan: Minor unrelated bug in the IDP flow where the error wasn't getting displayed as a `Toast`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
```diff
@@ -169,10 +171,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
      * @param v View.
      */
     public void setPositiveReturnValue(View v) {
-        setResult(Activity.RESULT_OK, null);
-        final Intent changeServerIntent = new Intent(CHANGE_SERVER_INTENT);
-        sendBroadcast(changeServerIntent);
-        finish();
+        (new AuthConfigTask()).execute();
```
bhariharan: This is kicked off when the `Apply` button is clicked on the `Change Server` screen. This indicates a server switch and we need to check if browser-based login is enabled for the new server.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
```diff
@@ -244,4 +243,27 @@ public void rebuildDisplay() {
             }
         }
     }
+
+    private class AuthConfigTask extends AsyncTask<Void, Void, Void> {
+
+        @Override
+        protected Void doInBackground(Void... nothings) {
+            final String loginServer = SalesforceSDKManager.getInstance().getLoginServerManager().getSelectedLoginServer().url;
```
bhariharan: Simple `AsyncTask` that runs the query in a background thread and posts the result on the main thread when done and dismissed itself.

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">📱 Needs testing on a real device if change is non-trivial.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_PR-Check">
</p>

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/AuthConfigUtil.java
```diff
@@ -0,0 +1,124 @@
...
+
+import okhttp3.Request;
+import okhttp3.Response;
+
+/**
+ * This class has utility methods that help with authentication related functionality.
+ *
+ * @author bhariharan
+ */
+public class AuthConfigUtil {
```
bhariharan: This class can be used for other auth-related util methods in the future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/AuthConfigUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final String TAG = "AuthConfigUtil";
+
+    /**
+     * Returns the auth config associated with an SSO login endpoint. This call
+     * should be made from a background thread since it makes a network request.
+     *
+     * @param loginUrl Login URL.
+     * @return Auth config.
+     */
+    public static SSOAuthConfig getSSOAuthConfig(String loginUrl) {
```
bhariharan: Gets SSO auth config from `well-known` endpoint.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/AuthConfigUtil.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        return authConfig;
+    }
+
+    /**
+     * This class represents SSO auth config.
+     *
+     * @author bhariharan
+     */
+    public static class SSOAuthConfig {
```
bhariharan: Encapsulation for SSO auth config.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/util/AuthConfigUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Tests for AuthConfigUtil.
+ *
+ * @author bhariharan
+ */
+@RunWith(AndroidJUnit4.class)
+@SmallTest
+public class AuthConfigUtilTest {
+
+    private static final String SSO_ENDPOINT = "https://mobsdktest.my.salesforce.com";
```
bhariharan: Test class. We have an SSO endpoint setup with `well-known` config.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
```diff
@@ -244,4 +243,27 @@ public void rebuildDisplay() {
             }
         }
     }
+
+    private class AuthConfigTask extends AsyncTask<Void, Void, Void> {
```
brandonpage: <table data-meta="generated_by_SalesforceSDK">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/13a92a93ed6f93479f8ed1b774cb2f4d10c7e044/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java#L247"></span>This AsyncTask class should be static or leaks might occur (com.salesforce.androidsdk.ui.ServerPickerActivity.AuthConfigTask)</td>
    </tr>
  </tbody>
</table>

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><a href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/13a92a93ed6f93479f8ed1b774cb2f4d10c7e044/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java#L129">libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java#L129</a> - Missing permissions required by ConnectivityManager.getActiveNetworkInfo: android.permission.ACCESS_NETWORK_STATE</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><a href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/13a92a93ed6f93479f8ed1b774cb2f4d10c7e044/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java#L147">libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java#L147</a> - Do not place Android context classes in static fields (static reference to <code>SalesforceSDKManager</code> which has field <code>context</code> pointing to <code>Context</code>); this is a memory leak (and also breaks Instant Run)</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><a href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/13a92a93ed6f93479f8ed1b774cb2f4d10c7e044/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java#L252">libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java#L252</a> - Using <code>getString</code> to get device identifiers is not recommended.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><a href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/13a92a93ed6f93479f8ed1b774cb2f4d10c7e044/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java#L141">libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/SPRequestHandler.java#L141</a> - This AsyncTask class should be static or leaks might occur (com.salesforce.androidsdk.auth.idp.SPRequestHandler.TokenEndpointTask)</td>
    </tr>
  </tbody>
</table>

</p>

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
wmathurin: Shouldn't we wait for that call to complete?
Could we use the same code as AuthConfigTask?

bhariharan: There's no good way to accomplish that in this case since there's no action after `init()` completes. Using an `AsyncTask` in this scenario would be the same as a `Runnable`. We recommend that the `init()` call be made from the `Application` class of the consuming application to initialize the SDK components, after which the main screen starts up. This should give ample time for the `Runnable` to finish execution. The only deterministic way I can think of is to maybe do this on initialization of `LoginActivity` instead, but that might be too frequent.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -93,19 +95,30 @@
...
 
-        // Setup content view
+        // Fetches auth config if required.
+        authConfigFetched = new AtomicBoolean(false);
+        if (!authConfigFetched.get()) {
+            (new AuthConfigTask(this)).execute();
+        }
+
+        // Waits for auth config to be fetched before continuing to load the activity.
+        while (!authConfigFetched.get()) {
```
wmathurin: To avoid the `while()` could you use a BlockingQueue (like we did in [TestRunnerPlugin](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/TestRunnerPlugin.java#L117) and [JSTestCase](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/util/test/JSTestCase.java#L89))

```java
private BlockingQueue<Boolean> authConfigReady = new ArrayBlockingQueue<Boolean>(1);

// Instead of the while
authConfigReady.take()

// In onAuthConfigFetched
authConfigReady.add(Boolean.TRUE);
```

bhariharan:
Have to make a couple more tweaks. Stay tuned.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -93,19 +96,31 @@
...
         final LoginOptions loginOptions = LoginOptions.fromBundle(getIntent().getExtras());
 
-        // Protect against screenshots
+        // Protect against screenshots.
         getWindow().setFlags(WindowManager.LayoutParams.FLAG_SECURE,
                 WindowManager.LayoutParams.FLAG_SECURE);
 
-        // Setup content view
+        // Fetches auth config if required.
+        authConfigReady = new ArrayBlockingQueue<>(1);
```
wmathurin: Shouldn't you move the test for production or sandbox here and not even create a AuthConfigTask if there is not auth config expected?

bhariharan: We need that check in the `AsyncTask` for the `Change Server` use case (and any other use situations where we might want to fetch that value in the future). I figured it was better to centralize that code there. It should return pretty quickly in that case since there's no network call.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -93,19 +96,31 @@
...
+        (new AuthConfigTask(this)).execute();
+
+        // Waits for auth config to be fetched before continuing to load the activity.
+        try {
+            authConfigReady.take();
```
bhariharan: @wmathurin There's a bit of a problem with this line here. Since it blocks the main thread, it's interfering with the `AsyncTask's MessageQueue` and preventing it from calling `onPostExecute()`. I'm not sure there's a good way to block the main thread and allow `AsyncTask` to post on the main thread without causing a deadlock. The call in `ServerPickerActivity` works beautifully though since it's using the same `AsyncTask` the way it should be - without blocking the main thread, but waiting instead for the callback.

wmathurin: Actually, we don't need the BlockingQueue.
You can just do `asyncTask.execute().get()` to wait for the async task to complete.
See: https://stackoverflow.com/a/14046475

bhariharan: I tried that already. It didn't work for me. Line of execution continued to the next line. Maybe it's because of `Void` return argument. I'll fiddle around with it a bit.

PR #1647, opened 2018-03-12
Fixed the incorrect format of the alias

Songroid:
The previous name is from an AES-256 key, which may contain "=". When an alias is fed to the KeyPairGeneratorSpec, it will throw IllegalArgumentException since it tries to parse the following Distinguished Names:

e.g. X.500 Distinguished Names: CN=cName, OU=orgUnit, O=org, L=city, S=state, C=countryCode
The fix will get rid of any non alphanumeric chars from the fed alias strings.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -158,9 +158,10 @@ public class SalesforceKeyGenerator {
      */
     public static synchronized PublicKey getRSAPublicKey(String name, int length) {
         PublicKey publicKey = null;
-        createRSAKeysIfNecessary(name, length);
+        String wrappedName = getNonAlphanumericString(name);
+        createRSAKeysIfNecessary(wrappedName, length);
         try {
-            publicKey = loadKeyStore().getCertificate(name).getPublicKey();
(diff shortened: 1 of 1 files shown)


bhariharan:
@Songroid It's probably better to fix this at the generation stage and pass in the correct key instead of manipulating it once it's in. Could you send a PR to my fork [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/blob/pn_encryption/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java#L297) to fix the string there instead?

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><a href="https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/00540e7a54f12c1556c85408c4323798bb7b1fad/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java#L250">libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java#L250</a> - Consider using <code>apply()</code> instead; <code>commit</code> writes its data to persistent storage immediately, whereas <code>apply</code> will handle it in the background</td>
    </tr>
  </tbody>
</table>

</p>

bhariharan:
Discussed with @Songroid offline. We'll be making these changes elsewhere in the caller (as my comment mentions above).

PR #1655, opened 2018-03-21
Deferred login should be given a chance to handle logout through receiver

smcnulty-sfdc:
Allowing the front activity to remain when we are a hybrid app and using deferred authentication.  Otherwise user runs into the experience where the app closes immediately after login instead of resuming the unauthenticated experience that is provided by a deferred login application.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -818,7 +818,9 @@ public class SalesforceSDKManager {
 
         // Finishes front activity if specified, and if this is the last account.
         if (frontActivity != null && (users == null || users.size() <= 1)) {
-            frontActivity.finish();
+            // only finish if it isn't hybrid or if hybrid, requires authentication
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -818,7 +818,9 @@ private void cleanUp(Activity frontActivity, Account account) {
 
         // Finishes front activity if specified, and if this is the last account.
         if (frontActivity != null && (users == null || users.size() <= 1)) {
-            frontActivity.finish();
+            // only finish if it isn't hybrid or if hybrid, requires authentication
+            if (!isHybrid() || BootConfig.getBootConfig(frontActivity).shouldAuthenticate())
```
bhariharan: @smcnulty-sfdc We do want to finish the hybrid activity in our hybrid apps though. Could the caller not pass in `frontActivity` instead? Or start it up again post-logout?

bhariharan: Can we move this to `SalesforceDroidGapActivity` instead to pass in a `null` value for `frontActivity` when `logout` is called in the deferred auth context?

PR #1657, opened 2018-03-23
Encryption code cleanup to isolate provider

bhariharan:
Some `BouncyCastle` functionality is being removed in `Android P`. Refactored the code to help us pivot from `BouncyCastle` and use `Conscrypt` instead if we want in the future.

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
@@ -26,7 +26,6 @@
  */
 package com.salesforce.androidsdk.analytics.security;
 
-import android.content.Context;
 import android.text.TextUtils;
 import android.util.Base64;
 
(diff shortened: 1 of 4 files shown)


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -58,67 +55,8 @@
...
-    private static String bestCipherAvailable;
-    private static boolean isFileSystemEncrypted;
-
-    /**
-     * Initializes this module.
-     *
-     * @param ctx Context.
-     * @return True - if the cryptographic module was successfully initialized, False - otherwise.
-     */
-    public static boolean init(Context ctx) {
```
bhariharan: Most of this code was legacy code that's not required anymore for 2 reasons:
1. File system encryption is available by default even with our `minApi`, i.e. `API 21`.
2. The reason we had hardcoded `BouncyCastle` in here was that it wasn't the default or even available on some devices prior to `Eclair` or `Froyo` (`Android 2.x`), which we haven't supported in years now.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -58,67 +55,8 @@
...
-        try {
-            Mac.getInstance(MAC_TRANSFORMATION, "BC");
-        } catch (GeneralSecurityException e) {
-            SalesforceAnalyticsLogger.e(ctx, TAG, "No MAC transformation available", e);
-            return false;
-        }
-        return true;
-    }
-
-    private static Cipher getBestCipher() throws GeneralSecurityException {
```
bhariharan: Same reasons as listed above. The best cipher is the one provided by the OS now, i.e. `Conscrypt` on `Android P` and higher.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -58,67 +55,8 @@
...
-        }
-        return cipher;
-    }
-
-    /**
-     * Checks if the file system is encrypted.
-     *
-     * @return True - if file system encryption is available and active, False - otherwise.
-     */
-    public static boolean isFileSystemEncrypted() {
```
bhariharan: This is always going to return `true` and it's not used anywhere or have any value anymore.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -302,8 +240,8 @@ public static String hash(String data, String key) {
             // Signs with SHA-256.
             byte [] keyBytes = key.getBytes(UTF8);
             byte [] dataBytes = data.getBytes(UTF8);
-            Mac sha = Mac.getInstance(MAC_TRANSFORMATION, "BC");
```
bhariharan: Use the default from the OS now, it's likely better than `BouncyCastle`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -454,9 +454,6 @@ public static void initInternal(Context context) {
         // Upgrades to the latest version.
         SalesforceSDKUpgradeManager.getInstance().upgrade();
 
-        // Initializes the encryption module.
-        Encryptor.init(context);
```
bhariharan: This made sense in the `Eclair` and `Froyo` days, but I don't think it's required anymore.

wmathurin:
Is some kind of migration needed when moving from 6.1 (or older) to 6.2?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -451,7 +390,21 @@ public static String decryptBytes(byte[] data, byte[] key, byte[] iv) {
...
+            cipher = Cipher.getInstance(PREFER_CIPHER_TRANSFORMATION, getEncryptionProvider());
+        } catch (Exception e) {
+            SalesforceAnalyticsLogger.e(null, TAG,
+                    "No cipher transformation available", e);
+        }
+        return cipher;
+    }
+
+    private static String getEncryptionProvider() {
+        return BOUNCY_CASTLE;
```
bhariharan: Added `Bouncy Castle` as the default back in here.

bhariharan:
Tested upgrade scenarios with passcode. Works fine.

PR #1660, opened 2018-03-29
Adding userinfo endpoint, changing cheap refresh API calls

khawkins:
Analogous to https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2479, changing the "check session" API endpoint to the [user info endpoint](https://help.salesforce.com/articleView?id=remoteaccess_using_userinfo_endpoint.htm), which averts errors due to a user not having API access configured (a common Communities use case).

Android's a little tricker than the iOS work, because `RestRequest` has no context to user-based endpoint configuration.  This approach would add an abstraction to allow the conceptual endpoint to be optionally specified.

**Note:** This is a work in progress.  I didn't want to go any further until the approach was validated.


khawkins:
Yep, that was my next step.  I just wanted to make sure I wasn't going in an undesired direction, before moving on.  Stay tuned!

khawkins:
I added tests and (hopefully) rounded out everything else.  Can you have another look?

bhariharan:
👍

PR #1662, opened 2018-04-04
Fixes for tests

bhariharan:
Calls to the token endpoint should use `loginUrl`, not `instanceUrl`.


bhariharan:
Tests are failing due to some profile related config on that org. Will investigate.

PR #1667, opened 2018-04-12
Fix hybrid unit tests

brandonpage:
Fix for:
test[testRegisterWithSpec] - com.salesforce.androidsdk.phonegap.SmartStoreJSTest
test[testUpsertConcurrentEntries] - com.salesforce.androidsdk.phonegap.SmartStoreLoadJSTest

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartStorePlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartStorePlugin.java
@@ -458,7 +458,7 @@ public class SmartStorePlugin extends ForcePlugin {
 			smartStore.registerSoup(soupName, indexSpecs);
 		}
 
-		callbackContext.success(soupName);
+		callbackContext.success(soupSpec != null ? soupSpec.getSoupName() : soupName);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SmartStorePlugin.java
```diff
@@ -458,7 +458,7 @@ private void registerSoup(JSONArray args, CallbackContext callbackContext) throw
 			smartStore.registerSoup(soupName, indexSpecs);
 		}
 
-		callbackContext.success(soupName);
+		callbackContext.success(soupSpec != null ? soupSpec.getSoupName() : soupName);
```
brandonpage: Thanks for this Wolf!

libs/test/SalesforceHybridTest/src/com/salesforce/androidsdk/phonegap/SmartStoreLoadJSTest.java
```diff
@@ -63,7 +64,9 @@
 
     @BeforeClass
     public static void runJSTestSuite() throws InterruptedException {
-        JSTestCase.runJSTestSuite(JS_SUITE, data(), 60);
+        // FIXME Update this when we stop using ARM Emulators
+        int timeout = Build.SUPPORTED_ABIS[0].contains("armeabi") ? 180 : 60;
+        JSTestCase.runJSTestSuite(JS_SUITE, data(), timeout);
```
bhariharan: Why not just bump it up to `180` across the board?

brandonpage: Didn't want to force a 3x timeout on multiple tests for the devices that definitely don't need it.  But since it seems unlikely these tests will break in that exact way I will simplify it to just 180.

PR #1669, opened 2018-04-12
Removing unused import

bhariharan:


brandonpage:
@bhariharan whoops, Thanks!

PR #1670, opened 2018-04-12
Updating sf__login.xml to no longer place webview in a scrollview and fit height

smcnulty-sfdc:
Current layout design causes blank space to appear in the view if the content doesn't fill the entire height.  

Change allows the webview to fill the entire height of the viewport and control it's own scroll instead of placing it inside of a scrollview.

### DIFF
--- a/libs/SalesforceSDK/res/layout/sf__login.xml
+++ b/libs/SalesforceSDK/res/layout/sf__login.xml
@@ -5,16 +5,10 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
-    <ScrollView android:layout_width="match_parent"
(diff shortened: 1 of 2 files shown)


bhariharan:
@smcnulty-sfdc What happens to the `IDP button` below it, if it's enabled? If the login webview takes the parent's height, the IDP button will be always hidden.

bhariharan:
@smcnulty-sfdc Revert your last commit back to `RelativeLayout`. It seems to work fine with your changes.

bhariharan:
before

smcnulty-sfdc:
@bhariharan adjusted layout to be linear, gave the webview a weight so it will auto-adjust size to fit and give space for the button when it is visible.

PR #1674, opened 2018-04-13
Un-registering from GCM push notifications only when the last account logs out

bhariharan:
`InstanceID` should be deleted only when the last user account logs out.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -890,30 +890,15 @@ public class SalesforceSDKManager {
         }
 	}
 
-    /**
-     * Unregisters from push notifications for both GCM (Android) and SFDC, and waits either for
-     * unregistration to complete or for the operation to time out. The timeout period is defined
-     * in PUSH_UNREGISTER_TIMEOUT_MILLIS. 
-     *
-     * If timeout occurs while the user is logged in, this method attempts to unregister the push
-     * unregistration receiver, and then removes the user's account.
-     *
-     * @param clientMgr ClientManager instance.
-     * @param showLoginPage True - if the login page should be shown, False - otherwise.
-     * @param refreshToken Refresh token.
-     * @param loginServer Login server.
-     * @param account Account instance.
-     * @param frontActivity Front activity.
-     */
     private void unregisterPush(final ClientManager clientMgr, final boolean showLoginPage,
     		final String refreshToken, final String loginServer,
-            final Account account, final Activity frontActivity) {
+            final Account account, final Activity frontActivity, boolean isLastAccount) {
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1031,10 +1003,11 @@ public void logout(Account account, Activity frontActivity, final boolean showLo
 		 * if the refresh token is available.
 		 */
 		final UserAccount userAcc = getUserAccountManager().buildUserAccount(account);
+		int numAccounts = mgr.getAccountsByType(getAccountType()).length;
     	if (PushMessaging.isRegistered(context, userAcc) && refreshToken != null) {
     		loggedOut = false;
     		unregisterPush(clientMgr, showLoginPage, refreshToken,
-    				loginServer, account, frontActivity);
+    				loginServer, account, frontActivity, (numAccounts == 1));
```
bhariharan: `if (numAccounts == 1)`, that means this is the last account being logged out.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -105,31 +101,32 @@ public static void register(Context context, UserAccount account) {
...
-            threadPool.execute(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        // avoid running on main thread
-                        instanceID.deleteInstanceID();
-                    } catch (IOException e) {
-                        SalesforceSDKLogger.e(TAG, "Error deleting InstanceID", e);
+            // Deletes InstanceID only if there are no other logged in accounts.
+            if (isLastAccount) {
```
bhariharan: Removing `InstanceID` only if this is the last account. This ensures that tokens for the other logged in users remain valid.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -376,25 +367,6 @@ private static String getSharedPrefFile(UserAccount account) {
...
     }
 
-    /**
-     * Returns the service info associated with an intent.
-     *
-     * @param context Context.
-     * @param intent Intent.
-     * @return Service info.
-     */
-    private static ServiceInfo getServiceInfo(Context context, Intent intent) {
```
bhariharan: Unused private method.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -286,23 +286,18 @@ private String registerSFDCPushNotification(String registrationId,
     	return null;
     }
 
-    private boolean unregisterSFDCPushNotification(String registeredId,
+    private void unregisterSFDCPushNotification(String registeredId,
```
bhariharan: We weren't using the return value.

PR #1676, opened 2018-04-16
Handling whitespace at the end

bhariharan:
This PR removes additional whitespace from hostname and label.


Songroid:
As mentioned in the chat, we may need to handle cases other than spaces as well. Maybe on par with `HttpUrl.parse(String url)` from okhttp3?

bhariharan:
Extra validation can be considered as a feature addition in the future. For now, let's rely on valid input.

PR #1678, opened 2018-04-18
Fixing Toast not showing for error messages

bhariharan:
The `Toast` message wasn't showing because we were attempting to display it from a background thread. We now explicitly call it on the main UI thread.


bhariharan:
@brandonpage FYI.

PR #1689, opened 2018-05-01
Adding new sync target for object metadata using describe API

bhariharan:
This target uses the old `describe` API, not the new UI SDK `object-info` API.


libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTest.java
```diff
@@ -144,6 +149,20 @@ public void testSyncDownWithoutOverwrite() throws Exception {
...
     }
 
+    /**
+     * Test for sync down with metadata target.
+     */
+    @Test
+    public void testSyncDownForMetadataTarget() throws Exception {
+
+        // Builds metadata sync down target and performs sync.
+        trySyncDown(MergeMode.LEAVE_IF_CHANGED, new MetadataSyncDownTarget(Constants.ACCOUNT), ACCOUNTS_SOUP);
```
wmathurin: In a real app, one would probably setup a soup to hold all the metadata.
Should you read it back and check a few keys on the json blob?

bhariharan: Sure - I can enhance this test to add a few keys.

bhariharan: Enhanced the test to verify the `keyPrefix` and `label`, which should be enough to ensure we're getting the correct object metadata.

PR #1691, opened 2018-05-03
Adding sync down target for object layout

bhariharan:
This uses the new `UI SDK` API endpoint.

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/CacheManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/CacheManager.java
@@ -56,7 +56,9 @@ import java.util.Map;
  * metadata, object layouts and MRU objects from a simple cache.
  *
  * @author bhariharan
+ * @deprecated Will be removed in Mobile SDK 7.0. Use {@link SyncManager} instead for synching data.
  */
+@Deprecated
 public class CacheManager {
 
     private static final String TAG = "SmartSync: CacheManager";
@@ -77,7 +79,9 @@ public class CacheManager {
      * This enum defines different possible cache policies.
      *
      * @author bhariharan
+     * @deprecated Will be removed in Mobile SDK 7.0.
      */
+    @Deprecated
     public enum CachePolicy {
         IGNORE_CACHE_DATA, // Ignores cache data and always loads from server.
         RELOAD_AND_RETURN_CACHE_ON_FAILURE, // Always reloads data and returns cache data only on failure.
@@ -93,7 +97,9 @@ public class CacheManager {
      *
      * @param account User account.
      * @return Instance of this class.
+     * @deprecated Will be removed in Mobile SDK 7.0.
      */
+    @Deprecated
     public static synchronized CacheManager getInstance(UserAccount account) {
         return getInstance(account, null);
     }
@@ -104,7 +110,9 @@ public class CacheManager {
      * @param account User account.
      * @param communityId Community ID.
      * @return Instance of this class.
+     * @deprecated Will be removed in Mobile SDK 7.0.
      */
+    @Deprecated
     public static synchronized CacheManager getInstance(UserAccount account, String communityId) {
         if (account == null) {
             account = SmartStoreSDKManager.getInstance().getUserAccountManager().getCurrentUser();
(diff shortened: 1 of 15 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/CacheManager.java
```diff
@@ -93,7 +97,9 @@
      *
      * @param account User account.
      * @return Instance of this class.
+     * @deprecated Will be removed in Mobile SDK 7.0.
      */
+    @Deprecated
```
wmathurin: Won't deprecating the class deprecate all its methods?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
```diff
@@ -0,0 +1,136 @@
...
+        return target;
+    }
+
+    @Override
+    public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws IOException, JSONException {
+        final RestRequest request = RestRequest.getRequestForObjectLayout(syncManager.apiVersion,
+                objectType, layoutType);
+        final RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
+        final JSONArray records = new JSONArray();
+        records.put(response.asJSONObject());
```
wmathurin: What's the shape of the response? What indexes should the soup they sync into have?
Should we provide some higher level (helper code) to setup that soup and the metadata syncs?
That could be the new MetaDataManager: the app would register interest for certain object type and we would create the soup and syncs and have methods to refresh the metadata. We could also have helper method to read it back from the store. Also, since the shape is known, we could have a typed object that is built from the json.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/CacheManager.java
bhariharan: Not AFAIK. It only marks the class deprecated. The methods have to be individually marked deprecated.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
bhariharan: For layouts, the shape can change based on the type of layout. I hesitate to strong-type this because more layouts could be added. The previous implementation had some strong typing since it was locked down on search layout.

bhariharan: As for `describe`, it's really hard to tell which fields an app is interested in. The field list could be drastically different based on the use case of the app. If the app knows what fields they're looking for, it might be easier for the app to specify the soup index specs.

wmathurin: There would be one soup for object infos and another for layouts.

For the object infos, there would be one row per object type which in the ui sdk object info response body is called apiName: so the object infos soup would only need a index on that field.

For layouts, should would the app retrieve the layout by id or should we add the object type in there and allow them to retrieve layout by object type+layout type+mode - in which case, those fields should be indexed.

wmathurin: The shape of the layout info and object info is well defined.
See
* https://developer.salesforce.com/docs/atlas.en-us.uiapi.meta/uiapi/ui_api_responses_object_info.htm
* https://developer.salesforce.com/docs/atlas.en-us.uiapi.meta/uiapi/ui_api_responses_record_layout.htm

So we could base our classes on that.
Unlike data, metadata is quite static.

bhariharan: Cool. I could add a thin helper layer on top of this that sets up the soup and provides utils to sync layout and metadata (the new `MetadataManager`). I'll do that in the next PR.

bhariharan:
@wmathurin Any other feedback? Can I make the changes you suggested in the next PR so that it's easier to review?

PR #1692, opened 2018-05-14
Fixed crashes when the custom url is invalid

Songroid:
We had top crashers from customers trying to input urls containing white spaces in the middle.
Fixed by using okhttp3's HttpUrl.parse() to detect invalid urls.


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/CustomServerUrlEditor.java
```diff
@@ -93,7 +95,7 @@ public void onClick(View v) {
 					return;
 				}
 				final String val = validateInput(R.id.sf__picker_custom_url);
-				if (val == null) {
+				if (val == null || HttpUrl.parse(val) == null) {
```
wmathurin: Shouldn't that move into validateInput?

Songroid: That's also shared with the label string (name).

bhariharan: Yup - should move inside the following section in `validateInput`:
```
		if (editId == R.id.sf__picker_custom_url) {
			isInvalidValue = !URLUtil.isHttpsUrl(etVal.toString());
			if (isInvalidValue) {
				Toast.makeText(context, getString(R.string.sf__invalid_server_url),
						Toast.LENGTH_SHORT).show();
			}
		}
```

wmathurin: We are doing url validation [here](https://github.com/Songroid/SalesforceMobileSDK-Android/blob/7efe3f31932aa565f52160538ced2de22d34e006/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/CustomServerUrlEditor.java#L176) - maybe we should do all url validation there.

PR #1694, opened 2018-05-15
Fix for parent-child sync up bug

wmathurin:
Android equivalent of https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2522

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
@@ -1020,32 +1020,27 @@ public class SmartStore  {
      */
     private void projectIndexedPath(JSONObject soupElt, ContentValues contentValues, IndexSpec indexSpec) {
         Object value = project(soupElt, indexSpec.path);
-        switch (indexSpec.type) {
-        case integer:
-            Long longValToUse = null;
-            try {
-                longValToUse = ((Number) value).longValue();
-            }
-            catch (Exception e) {
-                // Ignore and use the null value
-                SmartStoreLogger.e(TAG, "Unexpected error", e);
-            }
-            contentValues.put(indexSpec.columnName, longValToUse);
-            break;
-        case string:
-        case full_text:
-            contentValues.put(indexSpec.columnName, value != null ? value.toString() : null); break;
-        case floating:
-            Double doubleValToUse = null;
-            try {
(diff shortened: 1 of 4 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncUpTarget.java
```diff
@@ -304,6 +302,17 @@ else if (statusCode == HttpURLConnection.HTTP_NOT_FOUND) {
                     record.put(LOCAL, true);
                     record.put(LOCALLY_CREATED, true);
 
+                    // We need a re-run
+                    needReRun = true;
+                }
+            }
+            // Handling remotely deleted parent
+            else if(isEntityDeleted(response)) {
```
wmathurin: Heart of the fix

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncTest.java
```diff
@@ -820,4 +820,64 @@ public void testSyncUpWithLocallyCreatedChildrenRecords() throws Exception {
...
+
+    /**
+     * Create account on server, sync down
+     * Remotely delete account
+     * Create contacts locally, associates them with the account and run sync up
+     * Check smartstore and server afterwards
+     * The account should be recreated and the contacts should be associated to the new account id
+     */
+    @Test
+    public void testSyncUpWithLocallyCreatedChildrenRemotelyDeletedParent() throws Exception {
```
wmathurin: New test

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -1020,32 +1020,27 @@ private void projectIndexedPaths(JSONObject soupElt, ContentValues contentValues
...
-            try {
-                doubleValToUse = ((Number) value).doubleValue();
-            }
-            catch (Exception e) {
-                // Ignore and use the null value
-                SmartStoreLogger.e(TAG, "Unexpected error", e);
-            }
-            contentValues.put(indexSpec.columnName, doubleValToUse); break;
-        }
+
```
wmathurin: Unrelated to the bug.
More concise and less noisy projectIndexedPath method
Previous implementation would throw/log exception if integer or floating field was null

PR #1695, opened 2018-05-17
Creating high level utility to sync layout data with cache policies (LayoutSyncManager)

bhariharan:
- Introduces a data model encapsulation for layout data.
- Introduces a new manager class that makes it easy to sync layout data and returns a strongly-typed object for ease of use. This manager takes care of indexing the data, storing and retrieving from `SmartStore`.
- Introduces 3 simple cache policies that can be specified while retrieving data.
- Tests for the new manager class.

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
@@ -31,6 +31,11 @@ import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.smartstore.store.SmartStore;
 import com.salesforce.androidsdk.smartsync.app.SmartSyncSDKManager;
 import com.salesforce.androidsdk.smartsync.model.Layout;
+import com.salesforce.androidsdk.smartsync.target.LayoutSyncDownTarget;
+import com.salesforce.androidsdk.smartsync.target.SyncDownTarget;
+import com.salesforce.androidsdk.smartsync.util.SmartSyncLogger;
+import com.salesforce.androidsdk.smartsync.util.SyncOptions;
+import com.salesforce.androidsdk.smartsync.util.SyncState;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -46,7 +51,9 @@ import java.util.Set;
  */
 public class LayoutSyncManager {
 
+    private static final String SOUP_NAME = "layouts";
     private static final String FEATURE_LAYOUT_SYNC = "LY";
+    private static final String TAG = "LayoutSyncManager";
 
     private static Map<String, LayoutSyncManager> INSTANCES = new HashMap<>();
 
@@ -154,18 +161,64 @@ public class LayoutSyncManager {
     }
 
     /**
-     * Fetches layout data using the specified mode and triggers the supplied callback once complete.
+     * Fetches layout data for the specified object type and layout type using the specified
+     * mode and triggers the supplied callback once complete.
      *
+     * @param objectType Object type.
+     * @param layoutType Layout type. Defaults to "Full" if null is passed in.
      * @param mode Fetch mode. See {@link Mode} for available modes.
      * @param syncCallback Layout sync callback.
      */
-    public void fetchLayout(Mode mode, LayoutSyncCallback syncCallback) {
-
+    public void fetchLayout(String objectType, String layoutType, Mode mode,
+                            LayoutSyncCallback syncCallback) {
+        switch (mode) {
+            case CACHE_ONLY:
+                fetchFromCache(objectType, layoutType, syncCallback, false);
+                break;
+            case CACHE_FIRST:
+                fetchFromCache(objectType, layoutType, syncCallback, true);
+                break;
+            case SERVER_FIRST:
+                fetchFromServer(objectType, layoutType, syncCallback);
+                break;
+        }
     }
 
     private LayoutSyncManager(SmartStore smartStore, SyncManager syncManager) {
         this.smartStore = smartStore;
         this.syncManager = syncManager;
+        initializeSoup();
+    }
+
+    private void fetchFromServer(final String objectType, final String layoutType,
+                                 final LayoutSyncCallback syncCallback) {
+        final SyncDownTarget target = new LayoutSyncDownTarget(objectType, layoutType);
+        final SyncOptions options = SyncOptions.optionsForSyncDown(SyncState.MergeMode.OVERWRITE);
+        try {
+            syncManager.syncDown(target, options, SOUP_NAME, new SyncManager.SyncUpdateCallback() {
+
+                @Override
+                public void onUpdate(SyncState sync) {
+                    if (SyncState.Status.DONE.equals(sync.getStatus())) {
+                        fetchFromCache(objectType, layoutType, syncCallback, false);
+                    }
+                }
+            });
+        } catch (Exception e) {
+            SmartSyncLogger.e(TAG, "Exception occurred while syncing layout data", e);
+        }
+    }
+
+    private void fetchFromCache(String objectType, String layoutType,
+                                LayoutSyncCallback syncCallback, boolean fallbackOnServer) {
+        // TODO: SmartStore calls.
+        if (fallbackOnServer) {
+            fetchFromServer(objectType, layoutType, syncCallback);
+        }
+    }
+
+    private void initializeSoup() {
+        // TODO: Register soup if it doesn't exist.
     }
 
     /**
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
@@ -166,30 +166,12 @@ public class LayoutSyncManager {
     }
 
     /**
-     * Fetches layout data and triggers the supplied callback once complete. This call attempts
-     * to fetch data from the cache first. If the cache is empty, it hits the server for data.
+     * Fetches layout data using the specified mode and triggers the supplied callback once complete.
      *
+     * @param mode Fetch mode. See {@link Mode} for available modes.
      * @param syncCallback Layout sync callback.
      */
-    public void fetchLayout(LayoutSyncCallback syncCallback) {
-
-    }
-
-    /**
-     * Fetches layout data from the cache and triggers the supplied callback once complete.
-     *
-     * @param syncCallback Layout sync callback.
-     */
-    public void fetchLayoutFromCache(LayoutSyncCallback syncCallback) {
-
-    }
-
-    /**
-     * Fetches layout data from the server and triggers the supplied callback once complete.
-     *
-     * @param syncCallback Layout sync callback.
-     */
-    public void fetchLayoutFromServer(LayoutSyncCallback syncCallback) {
+    public void fetchLayout(Mode mode, LayoutSyncCallback syncCallback) {
 
     }
 
@@ -198,7 +180,6 @@ public class LayoutSyncManager {
         this.restClient = restClient;
     }
 
-    private
     /**
      * Callback interface for layout sync.
      *
@@ -213,5 +194,19 @@ public class LayoutSyncManager {
          * @param layout Layout.
          */
         void onSyncComplete(String objectType, Layout layout);
+    }
+
+    /**
+     * Enum for available data fetch modes.
+     *
+     * CACHE_ONLY - Fetches data from the cache and returns null if no data is available.
+     * CACHE_FIRST - Fetches data from the cache and falls back on the server if no data is available.
+     * SERVER_FIRST - Fetches data from the server and falls back on the cache if the server doesn't
+     * return data. The data fetched from the server is automatically cached.
+     */
+    public enum Mode {
+        CACHE_ONLY,
+        CACHE_FIRST,
+        SERVER_FIRST
     }
 }
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
@@ -40,7 +40,6 @@ import com.salesforce.androidsdk.smartsync.util.SyncOptions;
 import com.salesforce.androidsdk.smartsync.util.SyncState;
 
 import org.json.JSONArray;
-import org.json.JSONObject;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -233,8 +232,12 @@ public class LayoutSyncManager {
                     onSyncComplete(objectType, syncCallback, null);
                 }
             } else {
-                final JSONObject result = results.optJSONObject(0);
-                onSyncComplete(objectType, syncCallback, Layout.fromJSON(result));
+                final JSONArray resultArr = results.optJSONArray(0);
+                if (resultArr == null || resultArr.length() == 0) {
+                    onSyncComplete(objectType, syncCallback, null);
+                } else {
+                    onSyncComplete(objectType, syncCallback, Layout.fromJSON(resultArr.optJSONObject(0)));
+                }
             }
         } catch (Exception e) {
             SmartSyncLogger.e(TAG, "Exception occurred while reading layout data from the cache", e);
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
@@ -61,8 +61,8 @@ public class LayoutSyncManager {
             "} WHERE {" + SOUP_NAME + ":sobjectType} = '%s' and {" + SOUP_NAME + ":layoutType} = '%s'";
     private static final String TAG = "LayoutSyncManager";
     private static final IndexSpec[] INDEX_SPECS = new IndexSpec[] {
-        new IndexSpec("sobjectType", SmartStore.Type.string),
-        new IndexSpec("layoutType", SmartStore.Type.string)
+        new IndexSpec("sobjectType", SmartStore.Type.json1),
+        new IndexSpec("layoutType", SmartStore.Type.json1)
     };
 
     private static Map<String, LayoutSyncManager> INSTANCES = new HashMap<>();
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
@@ -232,12 +232,8 @@ public class LayoutSyncManager {
                     onSyncComplete(objectType, syncCallback, null);
                 }
             } else {
-                final JSONArray resultArr = results.optJSONArray(0);
-                if (resultArr == null || resultArr.length() == 0) {
-                    onSyncComplete(objectType, syncCallback, null);
-                } else {
-                    onSyncComplete(objectType, syncCallback, Layout.fromJSON(resultArr.optJSONObject(0)));
-                }
+                onSyncComplete(objectType, syncCallback,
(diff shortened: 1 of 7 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * Fetches layout data for the specified object type and layout type using the specified
+     * mode and triggers the supplied callback once complete.
+     *
+     * @param objectType Object type.
+     * @param layoutType Layout type. Defaults to "Full" if null is passed in.
+     * @param mode Fetch mode. See {@link Mode} for available modes.
+     * @param syncCallback Layout sync callback.
+     */
+    public void fetchLayout(String objectType, String layoutType, Mode mode,
```
bhariharan: Public API that allows the caller to specify where to fetch data from based on cache policy passed in.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+                }
+            }
+        } catch (Exception e) {
+            SmartSyncLogger.e(TAG, "Exception occurred while reading layout data from the cache", e);
+        }
+    }
+
+    private void onSyncComplete(String objectType, LayoutSyncCallback syncCallback, Layout layout) {
+        if (syncCallback != null) {
+            syncCallback.onSyncComplete(objectType, layout);
```
bhariharan: Triggers callback once the operation is complete.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+            smartStore.registerSoup(SOUP_NAME, INDEX_SPECS);
+        }
+    }
+
+    /**
+     * Callback interface for layout sync.
+     *
+     * @author bhariharan
+     */
+    public interface LayoutSyncCallback {
```
bhariharan: Callback interface for completion of operation.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+
+    /**
+     * Enum for available data fetch modes.
+     *
+     * CACHE_ONLY - Fetches data from the cache and returns null if no data is available.
+     * CACHE_FIRST - Fetches data from the cache and falls back on the server if no data is available.
+     * SERVER_FIRST - Fetches data from the server and falls back on the cache if the server doesn't
+     * return data. The data fetched from the server is automatically cached.
+     */
+    public enum Mode {
```
bhariharan: Cache modes.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -137,7 +136,7 @@ public static synchronized SyncManager getInstance(UserAccount account, String c
      */
     public static synchronized SyncManager getInstance(UserAccount account, String communityId, SmartStore smartStore) {
         if (account == null) {
-            account = SmartStoreSDKManager.getInstance().getUserAccountManager().getCurrentUser();
+            account = SmartSyncSDKManager.getInstance().getUserAccountManager().getCurrentUser();
```
bhariharan: Unrelated fix.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
```diff
@@ -0,0 +1,398 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents the layout of a Salesforce object.
+ *
+ * @author bhariharan
+ * @see <a href="https://developer.salesforce.com/docs/atlas.en-us.uiapi.meta/uiapi/ui_api_responses_record_layout.htm">https://developer.salesforce.com/docs/atlas.en-us.uiapi.meta/uiapi/ui_api_responses_record_layout.htm</a>
+ */
+public class Layout {
```
bhariharan: Data model of a layout object. Pretty much from the UI SDK public doc. I have added links here to that public doc. It's pretty well-defined.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
```diff
@@ -0,0 +1,398 @@
...
+    private List<LayoutSection> sections;
+    private JSONObject rawData;
+
+    /**
+     * Creates an instance of this class from its JSON representation.
+     *
+     * @param object JSON object.
+     * @return Instance of this class.
+     */
+    public static Layout fromJSON(JSONObject object) {
```
bhariharan: Just 1 public API to create an object from JSON.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+        return layout;
+    }
+
+    /**
+     * Returns the ID of this layout.
+     *
+     * @return ID of this layout.
+     */
+    public String getId() {
```
bhariharan: Getters for the fields available.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/SalesforceObject.java
```diff
@@ -36,9 +36,7 @@
  * This class represents a typical Salesforce object.
  *
  * @author bhariharan
- * @deprecated Will be removed in Mobile SDK 7.0.
```
bhariharan: I had accidentally deprecated this class. This is meant to stick around. Only `SalesforceObjectType` goes away.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
```diff
@@ -93,6 +94,10 @@ public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws I
         final RestRequest request = RestRequest.getRequestForObjectLayout(syncManager.apiVersion,
                 objectType, layoutType);
         final RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
+        final JSONObject responseJSON = response.asJSONObject();
```
bhariharan: Adding `objectType` while storing in `SmartStore`.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManagerTest.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Tests for {@link LayoutSyncManager}.
+ *
+ * @author bhariharan
+ */
+@RunWith(AndroidJUnit4.class)
+@MediumTest
+public class LayoutSyncManagerTest extends ManagerTestCase {
```
bhariharan: Test class for the new manager. Tests the different cache modes.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManagerTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    private static final String COMPACT = "Compact";
+    private static final String ACCOUNT = "Account";
+
+    private LayoutSyncManager layoutSyncManager;
+    private LayoutSyncCallbackQueue layoutSyncCallbackQueue;
+
+    private static class LayoutSyncCallbackQueue implements LayoutSyncManager.LayoutSyncCallback {
```
bhariharan: Blocking queue implementation to wait for callback.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTest.java
```diff
@@ -74,7 +74,7 @@
 import static java.util.Collections.singletonList;
 
 /**
- * Test class for SyncState.
+ * Test class for SyncManager.
```
bhariharan: Unrelated typo fix.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+ */
+public class LayoutSyncManager {
+
+    private static final String SOUP_NAME = "layouts";
+    private static final String FEATURE_LAYOUT_SYNC = "LY";
+    private static final String QUERY = "SELECT {" + SOUP_NAME + ":_soup} from {" + SOUP_NAME +
+            "} WHERE {" + SOUP_NAME + ":objectType} = '%s' and {" + SOUP_NAME + ":layoutType} = '%s'";
+    private static final String TAG = "LayoutSyncManager";
+    private static final IndexSpec[] INDEX_SPECS = new IndexSpec[] {
+        new IndexSpec("objectType", SmartStore.Type.string),
```
wmathurin: Should we use type json1? Maybe at some point in the future we will deprecate other index types - or maybe we will need to add indexes to that table, when using json1, index changes can be done very quickly (because it's pure sql).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+            final JSONArray results = smartStore.query(querySpec, 0);
+            if (results == null || results.length() == 0) {
+                if (fallbackOnServer) {
+                    fetchFromServer(objectType, layoutType, syncCallback);
+                } else {
+                    onSyncComplete(objectType, syncCallback, null);
+                }
+            } else {
+                final JSONArray resultArr = results.optJSONArray(0);
+                if (resultArr == null || resultArr.length() == 0) {
```
wmathurin: When would that happen? And shouldn't we then check fallbackOnServer.
In other words should those conditions move to line 228?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
```diff
@@ -0,0 +1,398 @@
...
+    public String getLayoutType() {
+        return layoutType;
+    }
+
+    /**
+     * Returns the mode of this layout.
+     *
+     * @return Mode of this layout.
+     */
+    public String getMode() {
```
wmathurin: Should that return an enum type?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
```diff
@@ -0,0 +1,398 @@
...
+    public String getId() {
+        return id;
+    }
+
+    /**
+     * Returns the layout type of this layout.
+     *
+     * @return Layout type of this layout.
+     */
+    public String getLayoutType() {
```
wmathurin: Should that return an enum type?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
```diff
@@ -39,12 +39,13 @@
...
 /**
  * Sync down target for object layouts. This uses the '/ui-api/layout' API to fetch object layouts.
+ * The easiest way to use this sync target is through {@link com.salesforce.androidsdk.smartsync.manager.LayoutSyncManager}.
  *
  * @author bhariharan
  */
 public class LayoutSyncDownTarget extends SyncDownTarget {
 
-    public static final String SOBJECT_TYPE = "sobjectType";
+    public static final String OBJECT_TYPE = "objectType";
```
wmathurin: We are using sobjectType in MRUSyncDownTarget - should we stay consistent with that?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * Provides an easy way to fetch layout data using {@link com.salesforce.androidsdk.smartsync.target.LayoutSyncDownTarget}.
+ * This class handles creating a soup, storing synched data and reading it into
+ * a meaningful data structure, i.e. {@link com.salesforce.androidsdk.smartsync.model.Layout}.
+ *
+ * @author bhariharan
+ */
+public class LayoutSyncManager {
+
+    private static final String SOUP_NAME = "layouts";
```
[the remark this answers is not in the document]
bhariharan: Sure - I can change that.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
bhariharan: Okay - I can take a look at that.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
bhariharan: Nope - this is not our cache mode. This is the mode in the payload of the layout response itself, could be `Create`, `Update`, etc. It represents record display mode I think.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
bhariharan: Nope. This is coming from the response itself. Besides, we don't have an enum of layout types - we leave it up to the UI SDK doc for that with a link. That way more types can be added in the future and no client-side changes or mutations would be required.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
bhariharan: Sure - I can change that. I changed it to match the `describe` response. I think that's `objectType`.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
bhariharan: Nope - this would probably happen when the server returned an error of some kind, or the object doesn't have a layout or something like that (I don't know if that's a real possibility).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
wmathurin: Is there a fixed set of possible values?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
wmathurin: If there is a fixed set of possible values, it would be better represented with an enum.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
wmathurin: Not sure I follow. If there is an error during sync down, nothing should be saved to the table so the first if should be entered.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
bhariharan: Right now it's `Compact` or `Full` I think, but it's not explicitly called out in the UI API doc, nor is it an `enum` on the server. Besides, the user already knows this since the user passes this into his/her call (`layoutType` is an argument). Right now, this class is a 1:1 representation of the layout response and uses the same data types as mentioned in the UI API doc. I'd prefer not to deviate from that. If the server side doc for the API was an `enum` of a specified set of values, we could change it, but for now, I feel like it's best to mirror the UI API doc 1:1.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
bhariharan: The UI API doc does not call out whether it's a fixed set of values or not. I'd prefer to mirror the API response data model per UI API doc and avoid mutation on the client. If the server doc is updated to specify a set of possible values (`enum`), we can update this to `enum` at that time.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
bhariharan: True. I guess this condition should never occur or would occur only if the response was a valid empty payload. Is that even possible?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -0,0 +1,288 @@
...
+ * Provides an easy way to fetch layout data using {@link com.salesforce.androidsdk.smartsync.target.LayoutSyncDownTarget}.
+ * This class handles creating a soup, storing synched data and reading it into
+ * a meaningful data structure, i.e. {@link com.salesforce.androidsdk.smartsync.model.Layout}.
+ *
+ * @author bhariharan
+ */
+public class LayoutSyncManager {
+
+    private static final String SOUP_NAME = "sfdcLayouts";
+    private static final String FEATURE_LAYOUT_SYNC = "LY";
```
bhariharan: Also adding new feature flag for this.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/model/Layout.java
wmathurin: Ok.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
bhariharan: Okay - will remove it.

bhariharan:
Removed [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1695/commits/d9a39eea8d334f0395c5084fa74afb5472d3418e).

wmathurin:
Another possible enhancement. 

Right now the config file that defines the syncs has only one top level key supported (it's "syncs"), for an example see: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/usersyncs.json

Maybe we could support configuring meta-data synching right there.

For instance, you could have 
```javascript
{  
   "objectInfos":[  
      "Contact",
      "Account"
   ], // => setup object info sync for Contact and Account
   "layouts":[  
      "Contact":[  
         "Full"
      ], // => setup layout sync for Full layout for Contact
      "Account":[  
         "Compact",
         "Full"
      ] // => setup layout sync for Full and Compact layouts for Account
   ],
   "syncs":[  
      {  
         "syncName":"syncDownContacts",
         "syncType":"syncDown",
         "soupName":"contacts",
         "target":{  
            "type":"soql",
            "query":"SELECT FirstName, LastName, Title, MobilePhone, Email, Department, HomePhone FROM Contact LIMIT 10000"
         },
         "options":{  
            "mergeMode":"OVERWRITE"
         }
      }
   ]
}
```

Thoughts?

bhariharan:
Good idea. We could do that. Should we evaluate usage of this and implement that as a 2nd iteration? ;-)

bhariharan:
My next PR will have an equivalent helper for metadata.

wmathurin:
Usage might be low because of discoverability - having it in the config might help on that front.

PR #1696, opened 2018-05-18
Capturing error during sync up on record

wmathurin:
If a record fails to sync up, the server response is saved in a `__last_error__` field on the record.
NB: That field will be cleared in subsequent successful sync up

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncUpTarget.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncUpTarget.java
@@ -84,6 +84,9 @@ public class SyncUpTarget extends SyncTarget {
     protected List<String> createFieldlist;
     protected List<String> updateFieldlist;
 
+    // Last sync error
+    protected String lastError;
+
     /**
      * Build SyncUpTarget from json
      *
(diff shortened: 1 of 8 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncUpTarget.java
```diff
@@ -176,6 +199,11 @@ public String createOnServer(SyncManager syncManager, JSONObject record, List<St
     protected String createOnServer(SyncManager syncManager, String objectType, Map<String, Object> fields) throws IOException, JSONException {
         RestRequest request = RestRequest.getRequestForCreate(syncManager.apiVersion, objectType, fields);
         RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
+
+        if (!response.isSuccess()) {
+            lastError = response.asString();
```
wmathurin: I chose to save the value in an instance variable to avoid changing the signatures of a bunch of methods.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/manager/SyncManagerTest.java
```diff
@@ -274,6 +274,63 @@ public void testReSyncByNameWithWrongName() throws Exception {
...
     }
 
+    /**
+     * Create a few records - some with bad names (too long or empty)
+     * Sync up
+     * Make sure the records with bad names are still marked as locally created and have the last error field populated
+     * @throws Exception
+     */
+    @Test
+    public void testSyncUpWithErrors() throws Exception {
```
wmathurin: Test for "regular" sync up.

libs/test/SmartSyncTest/src/com/salesforce/androidsdk/smartsync/target/ParentChildrenSyncTest.java
```diff
@@ -880,4 +880,95 @@ public void testSyncUpWithLocallyCreatedChildrenRemotelyDeletedParent() throws E
...
     }
+
+    /**
+     * Create accounts and contacts on server, sync down
+     * Update some of the accounts and contacts - using bad names (too long) for some
+     * Sync up
+     * Check smartstore and server afterwards
+     */
+    @Test
+    public void testSyncUpWithErrors() throws Exception {
```
wmathurin: Test for parent-child sync up.

PR #1700, opened 2018-05-22
Saving ID fields in the payload

bhariharan:
This gives a primary key for `SmartStore`.

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
@@ -29,6 +29,7 @@ package com.salesforce.androidsdk.smartsync.target;
 import com.salesforce.androidsdk.rest.RestRequest;
 import com.salesforce.androidsdk.rest.RestResponse;
 import com.salesforce.androidsdk.smartsync.manager.SyncManager;
+import com.salesforce.androidsdk.smartsync.util.Constants;
 
 import org.json.JSONArray;
 import org.json.JSONException;
@@ -47,6 +48,7 @@ public class LayoutSyncDownTarget extends SyncDownTarget {
 
     public static final String SOBJECT_TYPE = "sobjectType";
     public static final String LAYOUT_TYPE = "layoutType";
+    public static final String ID_FIELD_VALUE = "%s-%s";
 
     private String objectType;
     private String layoutType;
@@ -97,6 +99,7 @@ public class LayoutSyncDownTarget extends SyncDownTarget {
         final JSONObject responseJSON = response.asJSONObject();
         if (responseJSON != null) {
             responseJSON.put(SOBJECT_TYPE, objectType);
+            responseJSON.put(Constants.ID, String.format(ID_FIELD_VALUE, objectType, layoutType));
         }
         final JSONArray records = new JSONArray();
         records.put(response.asJSONObject());
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/MetadataSyncDownTarget.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/MetadataSyncDownTarget.java
@@ -29,6 +29,7 @@ package com.salesforce.androidsdk.smartsync.target;
 import com.salesforce.androidsdk.rest.RestRequest;
 import com.salesforce.androidsdk.rest.RestResponse;
 import com.salesforce.androidsdk.smartsync.manager.SyncManager;
+import com.salesforce.androidsdk.smartsync.util.Constants;
 
 import org.json.JSONArray;
 import org.json.JSONException;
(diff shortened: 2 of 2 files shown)


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/MetadataSyncDownTarget.java
```diff
@@ -86,6 +87,10 @@ public JSONObject asJSON() throws JSONException {
     public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws IOException, JSONException {
         final RestRequest request = RestRequest.getRequestForDescribe(syncManager.apiVersion, objectType);
         final RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
+        final JSONObject responseJSON = response.asJSONObject();
+        if (responseJSON != null) {
+            responseJSON.put(Constants.ID, objectType);
```
wmathurin: The soup we store that data in has an index on Id already?

bhariharan: I thought we did some work to auto-add `Id` and `LastModifiedDate` to the index specs if it was missing? I'll check.

bhariharan: I have created the high-level sync helper for metadata yet. The query on layout is based on `objectType` and `layoutType` already.

bhariharan: The default `externalId` is `Id`, so this should work.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
```diff
@@ -58,11 +59,10 @@
     private static final String SOUP_NAME = "sfdcLayouts";
     private static final String FEATURE_LAYOUT_SYNC = "LY";
     private static final String QUERY = "SELECT {" + SOUP_NAME + ":_soup} FROM {" + SOUP_NAME +
-            "} WHERE {" + SOUP_NAME + ":sobjectType} = '%s' AND {" + SOUP_NAME + ":layoutType} = '%s'";
+            "} WHERE {" + SOUP_NAME + ":" + Constants.ID + "} = '%s-%s'";
     private static final String TAG = "LayoutSyncManager";
     private static final IndexSpec[] INDEX_SPECS = new IndexSpec[] {
-        new IndexSpec("sobjectType", SmartStore.Type.json1),
```
wmathurin: Should we keep those indexes if someone wanted to do lookup by object types for instance?
(With json1, it doesn't add columns to the table - so it has little impact on storage)

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
```diff
@@ -98,7 +98,6 @@ public JSONArray startFetch(SyncManager syncManager, long maxTimeStamp) throws I
         final RestResponse response = syncManager.sendSyncWithSmartSyncUserAgent(request);
         final JSONObject responseJSON = response.asJSONObject();
         if (responseJSON != null) {
-            responseJSON.put(SOBJECT_TYPE, objectType);
```
wmathurin: Should we keep that, so that you don't have to parse that value out of the id field?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
bhariharan: We could, but there's little value in that though since the API callback already has `objectType` and the API calls have `objectType` too. I figured I'd try to keep the stored payload as close to the original as possible.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
bhariharan: We don't parse it out of the `Id` field. It's available in the callback since `objectType` is an instance variable.

bhariharan: See [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/blob/dev/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java#L182) - we have `objectType` and we pass it back in the callback.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/LayoutSyncManager.java
bhariharan: See [here](https://github.com/bhariharan/SalesforceMobileSDK-Android/blob/dev/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java#L53) - `objectType` is an instance variable in the sync down target so it's always available anyway.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/LayoutSyncDownTarget.java
wmathurin: I was thinking - if one was to read directly from the db - maybe that’s not a valid use case.

bhariharan:
Added a test for row count [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1700/commits/71edac6b02962e760d22a622b08a586b3a985112). Will do the same on `iOS` too.

PR #1705, opened 2018-05-25
Removing outdated library loader

bhariharan:
There were a few devices that would throw `UnsatisfiedLinkError` while attempting to load `SQLCipher` libs a few years ago. This was primarily an issue on devices running `KitKat` and lower (see [this StackOverflow post](http://stackoverflow.com/questions/18111739/why-do-some-android-phones-cause-our-app-to-throw-an-java-lang-unsatisfiedlinker)). Since our `minApi` is now `Lollipop`, we don't need this workaround anymore.


agarwasg:
👍

PR #1716, opened 2018-06-08
PushService is now subclassable

chrlie:
This allows for easier subclassing of `PushService` by overriding the methods that create requests for registration and unregistration. In order to specify that the subclass service be used, it has to be specified in the manifest (I've added an example in the Javadoc) and it has to be set with `SalesforceSDKManager.setPushServiceType`. This is because the service is started programmtically in a static method in PushMessaging and we need to know the correct class.

 I've also added a callback method for receiving the current registration status. This has been kept protected for subclassing purposes.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -33,6 +33,7 @@ import android.accounts.AccountManagerFuture;
 import android.app.Activity;
 import android.app.AlertDialog;
 import android.content.BroadcastReceiver;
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -161,6 +163,7 @@
     private AdminSettingsManager adminSettingsManager;
     private AdminPermsManager adminPermsManager;
     private PushNotificationInterface pushNotificationInterface;
+    private Class<? extends PushService> pushServiceType;
```
bhariharan: Shouldn't this be initialized to `PushService.class` by default?

PR #1717, opened 2018-06-08
Migrating from IntentService to JobIntentService for better job scheduling

bhariharan:
Starting with `Android Oreo`, implicit receivers triggering an `IntentService` are no longer allowed in the background. This restriction comes in place only if `targetApiVersion > 26`. Within the SDK there are a few classes that use `IntentService`. However, it is rare that these services are triggered from the background. They are usually deterministically triggered by code when the app is in the foreground.

`Android` has added a new type of service called `JobIntentService` that makes handling any job scheduling requests super easy. This is available in the support library and is backward compatible with old `Android` versions too. Based on the `targetApiVersion` of the app, it automatically determines whether to use `startService()` or `JobScheduler` (introduced in `API 26`) to run the request. `JobScheduler` is more optimal, conserves battery life and is the way forward.

References:
- [JobIntentService](https://developer.android.com/reference/android/support/v4/app/JobIntentService).
- [Background Execution in Oreo](https://developer.android.com/about/versions/oreo/background).

I attended a couple of sessions on `JobScheduler` at `Google I/O` this year. It does some pretty cool stuff and abstracts the gory details away from the caller.

### DIFF
--- a/libs/SalesforceSDK/build.gradle
+++ b/libs/SalesforceSDK/build.gradle
@@ -17,6 +17,7 @@ dependencies {
     api project(':libs:SalesforceAnalytics')
     api 'com.squareup.okhttp3:okhttp:3.9.0'
     api 'com.google.android.gms:play-services-gcm:12.0.1'
+    api 'com.android.support:support-compat:26.1.0'
     api 'com.android.support:customtabs:26.1.0'
     androidTestImplementation 'com.android.support.test:runner:1.0.1'
     androidTestImplementation 'com.android.support.test:rules:1.0.1'
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
@@ -41,7 +41,7 @@ import com.salesforce.androidsdk.accounts.UserAccountManager;
 public class AnalyticsPublisherService extends JobIntentService {
 
     private static final String ACTION_PUBLISH = "com.salesforce.androidsdk.analytics.action.ANALYTICS_PUBLISH";
-    private static final int JOB_ID = 1000;
+    private static final int JOB_ID = 81;
 
     /**
      * Starts this service to publish stored events. If the service is already
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
@@ -26,9 +26,9 @@
  */
 package com.salesforce.androidsdk.analytics;
 
-import android.app.IntentService;
 import android.content.Context;
 import android.content.Intent;
+import android.support.v4.app.JobIntentService;
 
 import com.salesforce.androidsdk.accounts.UserAccount;
 import com.salesforce.androidsdk.accounts.UserAccountManager;
@@ -38,17 +38,10 @@ import com.salesforce.androidsdk.accounts.UserAccountManager;
  *
  * @author bhariharan
  */
-public class AnalyticsPublisherService extends IntentService {
+public class AnalyticsPublisherService extends JobIntentService {
 
     private static final String ACTION_PUBLISH = "com.salesforce.androidsdk.analytics.action.ANALYTICS_PUBLISH";
-    private static final String TAG = "AnalyticsPublisherService";
-
-    /**
-     * Default constructor.
-     */
-    public AnalyticsPublisherService() {
-        super(TAG);
-    }
+    private static final int JOB_ID = 1000;
 
     /**
      * Starts this service to publish stored events. If the service is already
@@ -59,11 +52,11 @@ public class AnalyticsPublisherService extends IntentService {
     public static void startActionPublish(Context context) {
         final Intent intent = new Intent(context, AnalyticsPublisherService.class);
         intent.setAction(ACTION_PUBLISH);
-        context.startService(intent);
+        enqueueWork(context, AnalyticsPublisherService.class, JOB_ID, intent);
     }
 
     @Override
-    protected void onHandleIntent(Intent intent) {
+    protected void onHandleWork(Intent intent) {
         if (intent != null) {
             final String action = intent.getAction();
             if (ACTION_PUBLISH.equals(action)) {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
@@ -27,14 +27,13 @@
 package com.salesforce.androidsdk.push;
 
 import android.app.AlarmManager;
-import android.app.IntentService;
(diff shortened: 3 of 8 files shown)


libs/SalesforceSDK/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
     api project(':libs:SalesforceAnalytics')
     api 'com.squareup.okhttp3:okhttp:3.9.0'
     api 'com.google.android.gms:play-services-gcm:12.0.1'
+    api 'com.android.support:support-compat:26.1.0'
```
bhariharan: Adding the app compat library. Lotsa useful stuff in here, including `JobIntentService`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
```diff
@@ -38,17 +38,10 @@
...
     private static final String ACTION_PUBLISH = "com.salesforce.androidsdk.analytics.action.ANALYTICS_PUBLISH";
-    private static final String TAG = "AnalyticsPublisherService";
-
-    /**
-     * Default constructor.
-     */
-    public AnalyticsPublisherService() {
-        super(TAG);
-    }
+    private static final int JOB_ID = 81;
```
bhariharan: Uniquely identifies this job within this app.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
```diff
@@ -59,11 +52,11 @@ public AnalyticsPublisherService() {
     public static void startActionPublish(Context context) {
         final Intent intent = new Intent(context, AnalyticsPublisherService.class);
         intent.setAction(ACTION_PUBLISH);
-        context.startService(intent);
+        enqueueWork(context, AnalyticsPublisherService.class, JOB_ID, intent);
```
bhariharan: `enqueueWork` does the heavy lifting of whether to invoke `startService()` on older devices or use `JobScheduler` on newer devices.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
```diff
@@ -59,11 +52,11 @@ public AnalyticsPublisherService() {
...
     }
 
     @Override
-    protected void onHandleIntent(Intent intent) {
+    protected void onHandleWork(Intent intent) {
```
bhariharan: `JobIntentService` has an abstract method called `onHandleWork` that does the work (like `onHandleIntent` in an `IntentService`).

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -119,22 +119,11 @@ static void runIntentInService(Intent intent) {
             WAKE_LOCK.acquire(WAKE_LOCK_TIMEOUT_IN_MILLIS);
         }
         intent.setClassName(context, SalesforceSDKManager.getInstance().getPushServiceType().getName());
-        final ComponentName name = context.startService(intent);
-        if (name == null) {
-            SalesforceSDKLogger.w(TAG, "Could not start GCM service");
-        }
+        enqueueWork(context, SalesforceSDKManager.getInstance().getPushServiceType(), JOB_ID, intent);
```
bhariharan: Replaced any explicit call to `startService()` with `enqueueWork()` instead.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -204,17 +194,18 @@ private void performRegistrationChange(boolean register, UserAccount userAccount
...
+                SFDCRegistrationRetryAlarmReceiver.class);
         if (account == null) {
 			final Bundle bundle = new Bundle();
 			bundle.putString(PushMessaging.ACCOUNT_BUNDLE_KEY, PushMessaging.ALL_ACCOUNTS_BUNDLE_VALUE);
 			retryIntent.putExtra(PushMessaging.ACCOUNT_BUNDLE_KEY, bundle);
         } else {
             retryIntent.putExtra(PushMessaging.ACCOUNT_BUNDLE_KEY, account.toBundle());
         }
-        final PendingIntent retryPIntent = PendingIntent.getBroadcast(context,
+        final PendingIntent retryPIntent = PendingIntent.getBroadcast(SalesforceSDKManager.getInstance().getAppContext(),
```
bhariharan: Getting context from `SalesforceSDKManager` directly now.

bhariharan:
@chrlie FYI....this is what I was talking about earlier.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -119,22 +119,11 @@ static void runIntentInService(Intent intent) {
...
-        }
+        enqueueWork(context, SalesforceSDKManager.getInstance().getPushServiceType(), JOB_ID, intent);
     }
 
-	/**
-	 * Default constructor.
-	 */
-	public PushService() {
-		super(TAG);
-		context = SalesforceSDKManager.getInstance().getAppContext();
```
wmathurin: Why not keep that?

bhariharan: Because there's no such constructor in the super class. It could be invoked statically.

RoyalLachinov:
Hi guys,

My app crashes when I get push notification(App is in background). By the way, I use JobService instead of IntentService. In both cases, IntentService and JobService and JobIntentService,I got same crash

   java.lang.RuntimeException: Unable to start receiver com.salesforce.androidsdk.push.PushBroadcastReceiver: java.lang.IllegalStateException: Not allowed to start service Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x1000010 pkg=com.apm.acto cmp=com.apm.acto/com.salesforce.androidsdk.push.PushService (has extras) }: app is in background uid UidRecord{497a3ca u0a476 RCVR idle change:idle|uncached procs:1 seq(0,0,0)}


     Caused by: java.lang.IllegalStateException: Not allowed to start service Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x1000010 pkg=com.apm.acto cmp=com.apm.acto/com.salesforce.androidsdk.push.PushService (has extras) }: app is in background uid UidRecord{497a3ca u0a476 RCVR idle change:idle|uncached procs:1 seq(0,0,0)}

bhariharan:
@RoyalLachinov Have you tried using the latest code from `dev`?

RoyalLachinov:
do you mean, latest SDK ?

bhariharan:
Latest code from the `dev` branch (unreleased version).

RoyalLachinov:
No I did not, I only  changed current SDK versio to 6.2.0. But in this case, I got many crashes.

RoyalLachinov:
which spesific code I have to use from dev to solve push notificaiton crash ?

PR #1736, opened 2018-06-29
Adding build number to app version

bhariharan:
We already do this on `iOS`. Adding this to `Android` to bring it to parity.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -1135,20 +1135,44 @@ public class SalesforceSDKManager {
 
     /**
      * Returns a user agent string based on the Mobile SDK version. The user agent takes the following form:
-     *   SalesforceMobileSDK/{salesforceSDK version} android/{android OS version} appName/appVersion {Native|Hybrid} uid_{device id}
+     * SalesforceMobileSDK/{salesforceSDK version} android/{android OS version} appName/appVersion {Native|Hybrid} uid_{device id}
      *
      * @return The user agent string to use for all requests.
      */
     public final String getUserAgent() {
     	return getUserAgent("");
     }
-    
+
+    /**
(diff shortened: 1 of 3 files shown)


bhariharan:
@wmathurin Mentioning you, so this PR doesn't get lost in the flurry of e-mails from `GitHub` today. ;-)

bhariharan:
<b>Before:</b>
```SalesforceMobileSDK/7.0.0.dev android mobile/8.0.0 (Android SDK built for x86) SmartSyncExplorer/1.0 Native uid_1e5300489bca48f2 ftr_AI.MU.SY.UA.US```

<b>After:</b>
```SalesforceMobileSDK/6.2.0.dev android mobile/8.0.0 (Android SDK built for x86) SmartSyncExplorer/1.0(1) Native uid_1e5300489bca48f2 ftr_AI.MU.SY.UA.US```

The only difference is the build number in parenthesis, i.e. `SmartSyncExplorer/1.0 ` (vs) `SmartSyncExplorer/1.0(1)`.

**Comment by bhariharan** [comment]
Any feedback?

**Comment by wmathurin** [comment]
Don't devs change the version code whenever they change the version name?
What's the value of reporting both?

**Comment by bhariharan** [comment]
Typically it's used to report build number, so for a patch build, the dev may not bump up `versionName`, but just `versionNumber`. We were already doing this on `iOS`, but weren't on `Android`. `S1` wants it for their beta builds to get build number.

File: libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java [hunk 1]
```diff
@@ -1134,34 +1135,55 @@ private void notifyLogoutComplete(boolean showLoginPage) {
...
+                SDK_VERSION, Build.VERSION.RELEASE, Build.MODEL, appName, getAppVersion(),
                 appTypeWithQualifier, uid, TextUtils.join(".", features));
     }
 
+    /**
+     * Returns the app version of the app.
+     *
+     * @return App version.
+     */
+    public String getAppVersion() {
```
bhariharan: @wmathurin Done here. Pulled it into a common reusable method.

PR #1742, opened 2018-07-09
Removing deprecated MetadataManager, CacheManager and related classes

bhariharan:
Also removed the associated tests and fixed some warnings.


bhariharan:
All tests passed locally.

PR #1745, opened 2018-07-10
Parallelize nightly lib runs

brandonpage:


brandonpage:
@bhariharan Remind me that I need to have you create CircleCI contexts before this is merged.

PR #1747, opened 2018-07-10
Cleaning up upon test completion

bhariharan:
Some of these tests weren't cleaning up after.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
@@ -390,6 +390,9 @@ public class UserAccount {
      * Fetches this user's profile photo from the server and stores it in the cache.
      */
     public void downloadProfilePhoto() {
+        if (photoUrl == null) {
+            return;
+        }
         final File file = getProfilePhotoFile();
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -390,6 +390,9 @@ public Bitmap getProfilePhoto() {
      * Fetches this user's profile photo from the server and stores it in the cache.
      */
     public void downloadProfilePhoto() {
```
wmathurin: That code came in 6.2, could it throw exception for apps on 6.2?

bhariharan: Nope. `photoUrl` can never really be `null`. Our dummy test account doesn't have that field populated locally. I could add that field in the dummy account.

wmathurin: Agreed, the dummy test account should be more realistic.

bhariharan: Will do.

PR #1749, opened 2018-07-10
Misc fixes for SmartStore inspector

wmathurin:
SmartStore inspector was still referring soup_names
It has been renamed to soup_attrs a long time ago
No longer getting query from resource file (bad idea)
Instead building the query using the constants from SmartStore.java

Also showing more rows by default (100 instead of 10)
Showing details soup info if there are 100 soups or less (instead of 10)

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/ui/SmartStoreInspectorActivity.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/ui/SmartStoreInspectorActivity.java
@@ -63,6 +63,7 @@ import org.json.JSONObject;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 
 public class SmartStoreInspectorActivity extends Activity implements AdapterView.OnItemSelectedListener {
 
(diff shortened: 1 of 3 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/ui/SmartStoreInspectorActivity.java
```diff
@@ -276,7 +280,7 @@ public void onSoupsClick(View v) {
 		}
 
 		if (names.size() > 10) {
-			queryText.setText(getString(R.string.sf__inspector_soups_query));
+			queryText.setText(SOUPS_QUERY);
```
wmathurin: We didn't notice until now, because this query is only used when there are more than 10 soups.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/ui/SmartStoreInspectorActivity.java
```diff
@@ -96,6 +96,10 @@
 	private String lastAlertMessage;
 	private JSONArray lastResults;
 
+	// Default queries
+	private String SOUPS_QUERY = String.format("select %s from %s", SmartStore.SOUP_NAME_COL, SmartStore.SOUP_ATTRS_TABLE);
```
bhariharan: Use `String.format(Locale.US, ...) to avoid the `Lint` warning.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/ui/SmartStoreInspectorActivity.java
```diff
@@ -96,6 +96,10 @@
...
+	private String INDICES_QUERY = String.format("select %s, %s, %s from %s", SmartStore.SOUP_NAME_COL, SmartStore.PATH_COL, SmartStore.COLUMN_TYPE_COL, SmartStore.SOUP_INDEX_MAP_TABLE);
```
bhariharan: Use `String.format(Locale.US, ...) to avoid the `Lint` warning.

PR #1754, opened 2018-07-13
Some Gradle dependency cleanup

bhariharan:
- Upgrades test and support library versions to the latest.
- Removes some obsolete Gradle steps we had.


bhariharan:
Got to the bottom of the fake `SmartSync` failures. It's because the emulator doesn't have an SD card.

PR #1763, opened 2018-08-07
Fix hybrid remote session refresh case for SAML/SSO login flows

bhariharan:
If the org uses a SAML/SSO provider for login, it is possible to customize the SAML/SSO login URL to drop all arguments from the URL. Besides, the login redirect is beyond the control of the Salesforce IDP at this point and hence, the login redirect when a session timeout occurs will not contain `ec=301` or `startURL`. This leaves the user staring at a login page within the hybrid `WebView` that does nothing. Luckily, we can fetch the login URL in this case from `.well-known` and the container needs to do 2 things to solve this problem.
- We need to parse the SSO URLs from `.well-known` (which we already fetch) and check if these URLs are hit as part of the redirects that indicate session timeout.
- If this occurs, load the `startURL`.
This PR addresses these issues.

cc: @khawkins @chrlie 

The equivalent `iOS` PR is [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2616).

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
@@ -39,7 +39,6 @@ import com.salesforce.androidsdk.auth.HttpAccess.NoNetworkException;
 import com.salesforce.androidsdk.config.BootConfig;
 import com.salesforce.androidsdk.phonegap.app.SalesforceHybridSDKManager;
 import com.salesforce.androidsdk.phonegap.util.SalesforceHybridLogger;
-import com.salesforce.androidsdk.rest.ApiVersionStrings;
 import com.salesforce.androidsdk.rest.ClientManager;
 import com.salesforce.androidsdk.rest.ClientManager.AccountInfoNotFoundException;
 import com.salesforce.androidsdk.rest.ClientManager.RestClientCallback;
@@ -385,6 +384,7 @@ public class SalesforceDroidGapActivity extends CordovaActivity implements Sales
             public void onSuccess(RestRequest request, RestResponse response) {
                 SalesforceHybridLogger.i(TAG, "refresh callback - refresh succeeded");
                 runOnUiThread(new Runnable() {
+
                     @Override
                     public void run() {
                         /*
@@ -394,7 +394,7 @@ public class SalesforceDroidGapActivity extends CordovaActivity implements Sales
                         SalesforceDroidGapActivity.this.client = SalesforceDroidGapActivity.this.clientManager.peekRestClient();
                         setSidCookies();
                         loadVFPingPage();
-                        final String frontDoorUrl = getFrontDoorUrl(url, true);
+                        final String frontDoorUrl = getFrontDoorUrl(url, BootConfig.isAbsoluteUrl(url));
                         loadUrl(frontDoorUrl);
                     }
                 });
@@ -491,6 +491,7 @@ public class SalesforceDroidGapActivity extends CordovaActivity implements Sales
      * @return Front-doored URL.
      */
     public String getFrontDoorUrl(String url, boolean isAbsUrl) {
+
         /*
          * We need to use the absolute URL in some cases and relative URL in some
          * other cases, because of differences between instance URL and community
@@ -498,13 +499,12 @@ public class SalesforceDroidGapActivity extends CordovaActivity implements Sales
          * URL to use is in the 'resolveUrl' method in 'ClientInfo'.
          */
         url = (isAbsUrl ? url : client.getClientInfo().resolveUrl(url).toString());
-
-        HttpUrl frontDoorUrl = HttpUrl.parse(client.getClientInfo().getInstanceUrlAsString() + "/secur/frontdoor.jsp?").newBuilder()
+        final HttpUrl frontDoorUrl = HttpUrl.parse(client.getClientInfo().getInstanceUrlAsString()
+                + "/secur/frontdoor.jsp?").newBuilder()
                 .addQueryParameter("sid", client.getAuthToken())
                 .addQueryParameter("retURL", url)
                 .addQueryParameter("display", "touch")
                 .build();
-
         return frontDoorUrl.toString();
     }
 
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
@@ -26,17 +26,20 @@
  */
 package com.salesforce.androidsdk.phonegap.ui;
 
+import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.SystemClock;
 import android.view.KeyEvent;
 import android.view.View;
 import android.webkit.CookieManager;
+import android.webkit.URLUtil;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.auth.HttpAccess.NoNetworkException;
 import com.salesforce.androidsdk.config.BootConfig;
+import com.salesforce.androidsdk.config.LoginServerManager;
 import com.salesforce.androidsdk.phonegap.app.SalesforceHybridSDKManager;
 import com.salesforce.androidsdk.phonegap.util.SalesforceHybridLogger;
 import com.salesforce.androidsdk.rest.ClientManager;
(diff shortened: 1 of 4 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -80,13 +80,15 @@
...
     private BootConfig bootconfig;
+    private AuthConfigUtil.MyDomainAuthConfig authConfig;
 
     // Web app loaded?
     private boolean webAppLoaded = false;
 
     public SalesforceDroidGapActivity() {
         super();
         delegate = new SalesforceActivityDelegate(this);
+        authConfig = AuthConfigUtil.getMyDomainAuthConfig(SalesforceHybridSDKManager.getInstance().getLoginServerManager().getSelectedLoginServer().url);
```
bhariharan: Fetching it here so that it'll do it only once (when the hybrid app is loaded).

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
```diff
@@ -394,7 +406,7 @@ public void run() {
                         SalesforceDroidGapActivity.this.client = SalesforceDroidGapActivity.this.clientManager.peekRestClient();
                         setSidCookies();
                         loadVFPingPage();
-                        final String frontDoorUrl = getFrontDoorUrl(url, true);
+                        final String frontDoorUrl = getFrontDoorUrl(url, BootConfig.isAbsoluteUrl(url));
```
bhariharan: This shouldn't be always `true`. This parameter specifies if the passed in URL is an absolute URL or not.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceWebViewClientHelper.java
```diff
@@ -119,39 +121,56 @@ public static boolean hasCachedAppHome(Context ctx) {
...
         final Map<String, String> params = UriFragmentParser.parse(uri);
     	final String ec = params.get("ec");
     	int ecInt = (ec != null ? Integer.parseInt(ec) : -1);
     	final String startURL = params.get("startURL");
-        if ((ecInt == HttpURLConnection.HTTP_MOVED_PERM
-    			|| ecInt == HttpURLConnection.HTTP_MOVED_TEMP)
-    			&& startURL != null) {
-    		return startURL;
-    	} else {
+        if (ecInt == HttpURLConnection.HTTP_MOVED_PERM
```
bhariharan: The existing condition remains the same - `ec=301` or `ec=302`.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceWebViewClientHelper.java
```diff
@@ -119,39 +121,56 @@ public static boolean hasCachedAppHome(Context ctx) {
...
+    			|| ecInt == HttpURLConnection.HTTP_MOVED_TEMP) {
+            if (!TextUtils.isEmpty(startURL)) {
+                return startURL;
+            } else {
+                return BootConfig.getBootConfig(ctx).getStartPage();
```
bhariharan: Beefed this up a little more since `startURL` might not exist in some cases.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceWebViewClientHelper.java
```diff
@@ -119,39 +121,56 @@ public static boolean hasCachedAppHome(Context ctx) {
...
+            }
+        } else if (isSamlLoginRedirect(ctx, url)) {
```
bhariharan: This is the new condition that checks if there's a SAML redirect.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceWebViewClientHelper.java
```diff
@@ -119,39 +121,56 @@ public static boolean hasCachedAppHome(Context ctx) {
...
     }
+
+    private static boolean isSamlLoginRedirect(Context ctx, String url) {
+        if (ctx instanceof SalesforceDroidGapActivity) {
+            final AuthConfigUtil.MyDomainAuthConfig authConfig = ((SalesforceDroidGapActivity) ctx).getAuthConfig();
+            if (authConfig != null) {
+                final List<String> ssoUrls = authConfig.getSsoUrls();
+                if (ssoUrls != null && ssoUrls.size() > 0) {
+                   for (String ssoUrl : ssoUrls) {
+                       int paramsIndex = ssoUrl.indexOf("?");
```
bhariharan: The SSO URL itself could have additional crap in it. All we care about is the base URL + path.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/AuthConfigUtil.java
```diff
@@ -100,6 +107,19 @@ public MyDomainAuthConfig(JSONObject authConfig) {
                 if (mobileSDK != null) {
                     browserLoginEnabled = mobileSDK.optBoolean(USE_NATIVE_BROWSER_KEY);
                 }
+                final JSONArray samlProviders = authConfig.optJSONArray(SAML_PROVIDERS_KEY);
```
bhariharan: Parsing out the SSO URLs.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/util/AuthConfigUtilTest.java
```diff
@@ -68,6 +69,23 @@ public void testBrowserBasedLoginEnabled() {
         Assert.assertTrue("Browser based login should be enabled", authConfig.isBrowserLoginEnabled());
     }
 
+    @Test
+    public void testGetSSOUrls() {
```
bhariharan: Tests for the new values.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
wmathurin: What if login is deferred, wouldn't that code run too soon?

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceWebViewClientHelper.java
wmathurin: Couldn't we use https://developer.android.com/reference/android/net/Uri?

bhariharan: I could, but that would involve more work to parse the string, append base + path, etc. This seemed simpler because we get the URL as a string and all we want to do is chop off any params if they exist.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
bhariharan: It doesn't matter. In both cases, it runs only once in the lifecycle of this view controller. Also, the endpoint is open to all, so it has nothing to do with auth state of the app. Even if we deferred this call to after login, it would still run only once.

wmathurin: What if you select a different login server and are using deferred login, wouldn't you be using the wrong auth config then?

bhariharan: Selecting a different login server in the container would lead to this component being recreated. The selected login server should correspond with the selected one at creation time of this VC in a 1:1 manner always.

wmathurin: Is that true for an hybrid local app?
Doesn't the app continue after deferred login without its state being reset?

bhariharan: It's true for both apps because the only way through UI to switch servers today is through the login screen and when we tear down the login screen, we reload this activity.

bhariharan:
End-to-end testing complete. Works well.

PR #1771, opened 2018-08-15
Update to API 28

brandonpage:
Update docker image to use the latest Android dev tools.  Actually restore cache for performance.  Skip build entirely on merge by limiting non-nightly builds to CircleCI pr branches.


brandonpage:
Don't merge.  I believe there is an issue with the env/danger token

PR #1774, opened 2018-08-16
Replacing deprecated Assert with new signature

bhariharan:
This fixes 9 test failures in `SmartStore`. There are 2 more failures which will be fixed in my next PR to replace `BouncyCastle` with `Conscrypt` (the failures have nothing to do with `SmartStore` itself).


bhariharan:
The failure was `java.lang.AssertionError: Use assertEquals(expected, actual, delta) to compare floating-point numbers`. The extra parameter that I added in the new signature is `delta`.

PR #1775, opened 2018-08-17
Using BiometricPrompt for fingerprint auth on Android P (DO NOT MERGE)

bhariharan:
`FingerprintManager` is deprecated in `Android P` and has been replaced with `BiometricPrompt`, which is the OS-provided blessed UI for any biometric authentication. Right now it defaults to fingerprint, but in the future it would automatically support `Face ID`, etc. out of the box.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
@@ -35,10 +35,12 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.hardware.biometrics.BiometricPrompt;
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
```diff
@@ -446,18 +446,64 @@ public void onClick(DialogInterface dialog,
...
+     * Displays the dialog provided by the OS for biometric authentication
+     * using {@link android.hardware.biometrics.BiometricPrompt}.
+     */
+    @TargetApi(VERSION_CODES.P)
+    protected void showBiometricDialog() {
+
+        /*
+         * TODO: Remove this check once minAPI >= 28.
+         */
+        if (VERSION.SDK_INT >= VERSION_CODES.P) {
```
bhariharan: Here's the new `Android P` stuff. The actions are the same as the old fingerprint code.

PR #1788, opened 2018-09-22
Avoid exception from multiple initializeApp calls

mdeblase:
Firebase can only intitalizeApp once per App name.  Default App name is used here so there will be an exception thrown when a second MSDK login attempts to register for push. 

https://firebase.google.com/docs/reference/android/com/google/firebase/FirebaseApp#initializeApp(android.content.Context,%20com.google.firebase.FirebaseOptions)


jamesward:
@mdeblase I've invited you to the org: https://github.com/forcedotcom

Once accepted, refresh the CLA bot: https://cla.salesforce.com/status/forcedotcom/SalesforceMobileSDK-Android/pull/1788

PR #1792, opened 2018-09-27
Caching current user

wmathurin:
NB: only returning the cached value from a new method (getCachedCurrentUser()). getCachedCurrentUser() should only be used by callers that only care about the user identity (org id, user id) and not the tokens (since they could be outdated).

Locally on simulator, getCurrentUser() runs on average in 34 ms while getCachedCurrent() runs in 0.14 ms.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -317,6 +317,7 @@ public class UserAccountManager {
 		if (user.equals(curUser)) {
 			return;
 		}
+		clearCachedCurrentUser(curUser);
 		final ClientManager cm = new ClientManager(context, accountType,
 				SalesforceSDKManager.getInstance().getLoginOptions(), true);
 		final Account account = cm.getAccountByName(user.getAccountName());
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -132,6 +132,7 @@ public class UserAccountManager {
 	 * @param orgId Org ID.
 	 */
 	public void storeCurrentUserInfo(String userId, String orgId) {
+		clearCachedCurrentUser();
 		final SharedPreferences sp = context.getSharedPreferences(CURRENT_USER_PREF,
 				Context.MODE_PRIVATE);
         final Editor e = sp.edit();
@@ -186,13 +187,10 @@ public class UserAccountManager {
 	}
 
 	/**
-	 * Get rid of cached current user account if it matches the passed userAccount.
-	 * @param userAccount
+	 * Get rid of cached current user account
 	 */
-	public void clearCachedCurrentUser(UserAccount userAccount) {
-		if (cachedCurrentUserAccount != null && cachedCurrentUserAccount.equals(userAccount)) {
-			cachedCurrentUserAccount = null;
-		}
+	public void clearCachedCurrentUser() {
+		cachedCurrentUserAccount = null;
 	}
 
 	/**
@@ -317,7 +315,6 @@ public class UserAccountManager {
 		if (user.equals(curUser)) {
 			return;
 		}
-		clearCachedCurrentUser(curUser);
 		final ClientManager cm = new ClientManager(context, accountType,
 				SalesforceSDKManager.getInstance().getLoginOptions(), true);
 		final Account account = cm.getAccountByName(user.getAccountName());
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -102,6 +102,7 @@ public class UserAccountManager {
 	private Context context;
 	private AccountManager accountManager;
 	private String accountType;
+	private UserAccount cachedCurrentUserAccount;
 
 	/**
 	 * Returns a singleton instance of this class.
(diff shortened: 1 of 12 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -167,7 +168,31 @@ public String getStoredOrgId() {
...
+	/**
+	 * Returns a cached value of the current user.
+	 *
+	 * NB: The oauth tokens might be outdated
+	 *     Should be used by methods that only care about the current user's identity (org id, user id etc)
+	 *     Is faster than getCurrentUser()
+	 *
+	 * @return Current user that's logged in (with potentially outdated oauth tokens)
+	 */
+	public UserAccount getCachedCurrentUser() {
```
bhariharan: Should we maintain this at this layer, or move this to the layer that needs it, i.e. plugins? If we provide this here, we will need to keep things in sync for any user account state changes and all components will be affected.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -167,7 +168,31 @@ public String getStoredOrgId() {
...
+		return cachedCurrentUserAccount != null ? cachedCurrentUserAccount : getCurrentUser() /* will populate cachedCurrentUserAccount */ ;
+	}
+
+	/**
+	 * Get rid of cached current user account if it matches the passed userAccount.
+	 * @param userAccount
+	 */
+	public void clearCachedCurrentUser(UserAccount userAccount) {
```
bhariharan: Should this be `public`? A component that calls this incorrectly could invalidate the cache for every other component.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -761,6 +761,7 @@ private void cleanUp(Activity frontActivity, Account account) {
     protected void cleanUp(UserAccount userAccount) {
         SalesforceAnalyticsManager.reset(userAccount);
         RestClient.clearCaches(userAccount);
+        UserAccountManager.getInstance().clearCachedCurrentUser(userAccount);
```
bhariharan: We also need to clear the cache when user switching occurs.

bhariharan: Also needs to be updated on token refresh.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
wmathurin: I thought about doing that at first, but it would mean keeping (or getting) a reference to the SmartStorePlugin from SalesforceHybridSDKManager.cleanUp(UserAccount ua). There is no clean way to do that.

Also we would have to do the same work in SalesforceReactSDKManager.cleanUp(UserAccount xx) for SmartStoreReactBridge. So it would be duplicated code.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
wmathurin: It wouldn't be incorrect. If there is no cached value, getCachedCurrentUser() simply calls getCachedCurrentUser().

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
wmathurin: We don't care about the token refresh case. The javadoc for getCachedCurrentUser() cleary states that it should only be called by methods that only care about the identity of the current user (org id, user id) and that oauth tokens might be outdated.

I'll look at user switching.

bhariharan: True, but it's low cost to update this when the token changes - an intent is broadcast. If we include that, this cached account should always be in sync and can be used in more places.

bhariharan: If we do this, do we need a new API? Can we simply use the existing current user API while it maintains the cached value internally in an instance variable?

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
bhariharan: All you need is a broadcast receiver for this.

wmathurin: I need a Context to register the broadcast receiver. Also isn't asynchronous, couldn't a call to getCurrentUser() pick up the cached value when it shouldn't?

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -292,6 +317,7 @@ public void switchToUser(UserAccount user, int userSwitchType, Bundle extras) {
 		if (user.equals(curUser)) {
 			return;
 		}
+		clearCachedCurrentUser(curUser);
```
bhariharan: This only handles the switch to existing user use-case. You also need to call this in the switch to new user use-case. I think a better place to handle this would be in `public void storeCurrentUserInfo(String userId, String orgId)` in `UserAccountManager`. Any change to current user ends up there, including new user and current user. This call can be moved there to handle both cases.

wmathurin: Hmm but the only caller of storeCurrentUserInfo(...) is this method `switchToUser(UserAccount user, int userSwitchType, Bundle extras)`, how do the store values get updated in the switch to new user use-case?

PR #1793, opened 2018-09-27
Change errorJSON type to JSONObject

booker0108:


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -362,8 +362,20 @@ public void run() {
                 } catch (Exception e) {
                     SmartSyncLogger.e(TAG, "Exception thrown in runSync", e);
 
-                    //Set error message to sync state
-                    sync.setError(e.getMessage());
+                    //Get error from API respond
```
bhariharan: Replace `respond` with `response`.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -362,8 +362,20 @@ public void run() {
...
+                    String errorMessage = e.getMessage();
+                    if(errorMessage != null){
+                        try {
+                            //Parse respond string to JSONArray
```
bhariharan: Replace `respond` with `response`.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -362,8 +362,20 @@ public void run() {
...
+                            JSONArray errorInJSON = new JSONArray(errorMessage);
+
+                            //Set error message from first object in JSONArray to sync state
+                            sync.setError(errorInJSON.getJSONObject(0));
+                        } catch (JSONException jsonException) {
+                            jsonException.printStackTrace();
```
bhariharan: Replace this line with:
```
SmartSyncLogger.e(TAG, "Exception thrown while parsing JSON", jsonException);
```

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
booker0108: Resolved in 3d45638d9478cc7560c9ebcce4677a7c761fc26b

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
booker0108: Resolved in 3d45638d9478cc7560c9ebcce4677a7c761fc26b

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
booker0108: Resolved in 3d45638d9478cc7560c9ebcce4677a7c761fc26b

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -362,8 +362,20 @@ public void run() {
                 } catch (Exception e) {
                     SmartSyncLogger.e(TAG, "Exception thrown in runSync", e);
 
-                    //Set error message to sync state
-                    sync.setError(e.getMessage());
+                    //Get error from API response
+                    String errorMessage = e.getMessage();
```
wmathurin: During sync down, if the API call returns an error, it ends up stringified in the error message (see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java#L137)).

But during sync up, we don't stop the sync operation because one record could not be created/updated/deleted on the server. Instead we store the error on that record (see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncUpTarget.java#L203)). During a sync up, the sync will fail only if the sync target throws some exceptions e.g. if the server can't be reached. In those cases the exception won't have a stringified json message.

(Furthermore, on iOS, our network library doesn't give us the raw error response, so we wouldn't be able to provide the json error there even for sync down.)

**I recommend, we keep the code before (which stores the error as a string on the sync state object).**

wmathurin:
Closing since error don't always have a stringified json body (see my comment above for a longer explanation).

PR #1794, opened 2018-10-03
Moving to RN 0.56.1

wmathurin:
Moving to RN 0.56.1, first tried 0.57.1, but ran into an issue (see https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/commit/1add4778589616feafb9500b91fd5ed005510a0e)

All react tests are passing

### DIFF
--- a/libs/SalesforceReact/build.gradle
+++ b/libs/SalesforceReact/build.gradle
@@ -18,7 +18,7 @@ apply plugin: 'com.jfrog.bintray'
 
 dependencies {
   api project(':libs:SmartSync')
-  api 'com.facebook.react:react-native:0.55.4'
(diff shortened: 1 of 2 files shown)


wmathurin:
@bhariharan @brandonpage 
All tests passing, but danger is failing with 

> All com.android.support libraries must use the exact same version specification (mixing versions can lead to runtime crashes). Found versions 27.1.1, 26.1.0. Examples include com.android.support:customtabs:27.1.1 and com.android.support:animated-vector-drawable:26.1.0

Is this error new? Does it matter?

bhariharan:
@wmathurin Nope - don't worry about it. That's more of a warning. Unfortunately, `React` is using an older version of that library, which we override at our layer.

PR #1808, opened 2018-11-01
Check run config servers when initializing

mdeblase:
Prevents default selected server in shared pref from being incorrect when run config server list is present during first launch initialization.


bhariharan:
@mdeblase Do you have the latest from `dev` on your fork? CI is indicating a build error that I fixed last week on `dev`.

PR #1811, opened 2018-11-02
Replacing deprecations in tests

bhariharan:
Replaced deprecated classes with their new equivalents. No change in code - just test imports and `Gradle` imports.


bhariharan:
@smcnulty-sfdc

PR #1813, opened 2018-11-16
Fixing crash in dev dialog for unauthenticated users

bhariharan:
Calling `super()` here crashes within `MyCommunities` since most of our use cases are unauthenticated and the list of users would be empty.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -1311,10 +1311,6 @@ public class SalesforceSDKManager {
      * @return Dev info (list of name1, value1, name2, value2 etc) to show in DevInfoActivity
      */
     public List<String> getDevSupportInfos() {
-        final UserAccount cachedCurrentUser = getUserAccountManager().getCachedCurrentUser();
-        final String currentUserString = (cachedCurrentUser == null) ? "" : usersToString(cachedCurrentUser);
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1311,25 +1311,26 @@ public void setDevSupportEnabled(boolean isDevSupportEnabled) {
      * @return Dev info (list of name1, value1, name2, value2 etc) to show in DevInfoActivity
      */
     public List<String> getDevSupportInfos() {
```
wmathurin: `usersToString` already handles null.
The problem is the `toArray()` call against a possible `null` value.
We could add a second `usersToString` method that takes a List parameter.
```java
    private String usersToString(List<UserAccount> userAccounts) {
        return userAccounts == null ? usersToString(userAccounts.toArray(new UserAccount[0]));
    }
```
Then we could replace
```java
                "Authenticated Users", usersToString(getUserAccountManager().getAuthenticatedUsers().toArray(new UserAccount[0]))
```
With
```java
                "Authenticated Users", usersToString(getUserAccountManager().getAuthenticatedUsers())
```

wmathurin: The new method should be more like:
```java
    private String usersToString(List<UserAccount> userAccounts) {
        return usersToString(userAccounts == null ? null : userAccounts.toArray(new UserAccount[0]));
    }
```

PR #1814, opened 2018-11-17
Saving user account before passcode creation (if any)

wmathurin:
Passcode is no longer used to encrypt user account data and therefore user account can be saved without waiting for passcode to be created
=> it simplifies the code
=> flow was broken in case of app being killed during passcode creation flow, on restart, the passcode creation flow would resume, but once completed, since no account had been saved, user would get a login screen

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
@@ -385,9 +385,7 @@ public class LoginActivity extends AccountAuthenticatorActivity
 
 	@Override
 	protected void onActivityResult(int requestCode, int resultCode, Intent data) {
-		if (requestCode == PasscodeManager.PASSCODE_REQUEST_CODE && resultCode == Activity.RESULT_OK) {
-			webviewHelper.onNewPasscode();
-		} else if (requestCode == SPRequestHandler.IDP_REQUEST_CODE) {
+		if (requestCode == SPRequestHandler.IDP_REQUEST_CODE) {
             spRequestHandler.handleIDPResponse(resultCode, data);
         } else {
 	        super.onActivityResult(requestCode, resultCode, data);
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
@@ -188,26 +188,6 @@ public class OAuthWebviewHelper implements KeyChainAliasCallback {
     	webview.loadUrl("about:blank");
     }
 
-    /**
-     * Method called by login activity when it resumes after the passcode activity
-     *
-     * When the server has a mobile policy requiring a passcode, we start the passcode activity after completing the
-     * auth flow (see onAuthFlowComplete).
-     * When the passcode activity completes, the login activity's onActivityResult gets invoked, and it calls this method
-     * to finalize the account creation.
-     */
-    public void onNewPasscode() {
-
-    	/*
-    	 * Re-encryption of existing accounts with the new passcode is taken
-    	 * care of in the 'Confirm Passcode' step in PasscodeActivity.
-    	 */
-        if (accountOptions != null) {
-            final UserAccount addedAccount = addAccount();
-            callback.finish(addedAccount);
-        }
-    }
-
     /** Factory method for the WebViewClient, you can replace this with something else if you need to */
     protected WebViewClient makeWebViewClient() {
     	return new AuthWebViewClient();
@@ -590,6 +570,9 @@ public class OAuthWebviewHelper implements KeyChainAliasCallback {
                 mgr.getAdminPermsManager().setPrefs(id.customPermissions, account);
             }
 
+            // Save the user account
+            addAccount(account);
+
             // Screen lock required by mobile policy.
             if (id.screenLockTimeout > 0) {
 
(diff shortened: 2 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -385,9 +385,7 @@ public void onPickServerClick(View v) {
 
 	@Override
 	protected void onActivityResult(int requestCode, int resultCode, Intent data) {
-		if (requestCode == PasscodeManager.PASSCODE_REQUEST_CODE && resultCode == Activity.RESULT_OK) {
-			webviewHelper.onNewPasscode();
```
wmathurin: The onNewPasscode method was taking care of creating the user account. Now the user account will already have been created.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -590,40 +570,26 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
...
+            addAccount(account);
+
             // Screen lock required by mobile policy.
             if (id.screenLockTimeout > 0) {
 
                 // Stores the mobile policy for the org.
                 final PasscodeManager passcodeManager = mgr.getPasscodeManager();
                 passcodeManager.storeMobilePolicyForOrg(account, id.screenLockTimeout * 1000 * 60, id.pinLength);
                 passcodeManager.setTimeoutMs(id.screenLockTimeout * 1000 * 60);
-                boolean changeRequired = passcodeManager.setMinPasscodeLength((Activity) getContext(), id.pinLength);
```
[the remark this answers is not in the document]
bhariharan: I think we still need this for the multi-user scenario. The reason why we don't consider login complete until passcode creation is complete is because there's logic here to match the new policy against the old one (if an extra user logs in) and pick the best combination of the two (or three or more) and pick a new length and timeout. If this happens, we need to kick off the change passcode flow, which can only be initialized from here - the passcode manager has no knowledge of this. Also, it needs to be enforced right after this user logs in (new security policy).

wmathurin:
I'll look into. There might be a way to handle that use case in a more graceful manner.

bhariharan:
This is one of those bugs that IMO could expose more bugs if we attempt to fix it. The premise of our auth flow on Android is that it's transactional to the point of passcode creation. Until passcode creation is complete, the auth flow isn't complete. This has helped us avoid many corner case bugs that come up on iOS with state management, where the two states are decoupled. How many people are going to go through the login flow and background the app without creating a passcode? Given the number of passcode users, and given the frequency of logging in on mobile (typically one time), the frequency of this use case happening in production is pretty low. Also, the effect of the bug isn't so bad - the user has to log in again and complete the flow. There are many other parts of the system that rely on the number of users in `AccountManager` at different stages and any change would cause ripple effects there because we would have an extra user now before passcode creation. I would recommend Never'ing this bug - it's a P4 anyway. My $0.02.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -75,14 +75,14 @@
...
-    public static final int PASSCODE_REQUEST_CODE = 777;
-
-    // Key used to specify that a longer passcode needs to be created.
-    public static final String CHANGE_PASSCODE_KEY = "change_passcode";
+    // Key in preference to indicate passcode change is required.
+    private static final String KEY_PASSCODE_CHANGE_REQUIRED= "passcode_change_required";
 
     // Key in preference for failed attempts
-    private static final String FAILED_ATTEMPTS = "failed_attempts";
+    private static final String KEY_FAILED_ATTEMPTS = "failed_attempts";
```
wmathurin: Renaming to be consistent with the other keys

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -75,14 +75,14 @@
     // Key in preference for the passcode length.
     private static final String KEY_PASSCODE_LENGTH = "passcode_length";
 
-    // Request code used to start passcode activity
-    public static final int PASSCODE_REQUEST_CODE = 777;
-
-    // Key used to specify that a longer passcode needs to be created.
-    public static final String CHANGE_PASSCODE_KEY = "change_passcode";
```
wmathurin: Passcode activity asks passcode manager directly - no longer passing the flag through the intent.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -182,6 +192,7 @@ private void storeMobilePolicy(Context context) {
             Editor e = sp.edit();
             e.putInt(KEY_TIMEOUT, timeoutMs);
             e.putInt(KEY_PASSCODE_LENGTH, minPasscodeLength);
+            e.putBoolean(KEY_PASSCODE_CHANGE_REQUIRED, passcodeChangeRequired);
```
wmathurin: Saving to preference so that a restart is properly handled.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -462,34 +479,22 @@ public int getMinPasscodeLength() {
...
-            this.minPasscodeLength = minPasscodeLength;
-
-            /*
-             * This needs to happen only if a passcode exists, in order to trigger
-             * the 'Change Passcode' flow. Otherwise, we simply need to update
-             * the minimum length in memory. The 'Create Passcode' flow is
-             * triggered later from OAuthWebviewHelper.
-             */
             if (hasStoredPasscode(ctx)) {
-                showLockActivity(ctx, true);
```
wmathurin: The next SalesforceActivity that gets resumed will ask the passcode manager if the app needs to be locked - that's when the screen will be shown.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -590,40 +570,30 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
...
-                    passcodeManager.lockIfNeeded((Activity) getContext(), true);
-                } else if (!changeRequired) {
-
-                    // If a passcode change is required, the lock screen will have already been set in setMinPasscodeLength.
-                    final UserAccount addedAccount = addAccount();
-                    callback.finish(addedAccount);
-                }
+                // NB setMinPasscodeLength(...)
+                //    If there was a passcode and the length is increased, the passcode manager will remember that a passcode change is required
+                //    The next SalesforceActivity to resume, will cause the locking screen to popup in passcode change mode
```
wmathurin: We are not triggering the passcode screen here. Instead we leave the passcode manager in a state that means, that next time a SalesforceActivity resumes, the lock screen will be shown (with the change passcode flow).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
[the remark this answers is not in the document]
wmathurin: The solution is to have the passcode manager know that a passcode change is required (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1814/commits/b9ff5a9d52abd707de49299d6a0f2d2889d96e1b).
No longer showing passcode activity from random places (like OAuthWebViewHelper or PasscodeManager's setMinPasscodeLength) instead capturing the state and letting the next SalesforceActivity that resumes and that asks the PasscodeManager if app should be locked trigger the showing of the PasscodeActivity.

wmathurin:
More tests / changes in the way ;-)

PR #1817, opened 2018-11-27
Do not reload the Cordova WebView when foregrounding

chrlie:
This behavior occurs when the `RestClient` is not null but not in the unauthenticated case.


chrlie:
I’m not sure if you want this to go in as-is, as it will change the current unauthenticated experience that people may be relying on. An alternative would be a toggle:

```java
protected void enableReloadOnForeground(boolean shouldReload) {
    this.shouldReload = shouldReload;
}

@Override
public void onResume(RestClient c) {
    ...
    if (client == null) {
            if (!webAppLoaded && shouldReload) {
                onResumeNotLoggedIn();
                ...
}
```

PR #1819, opened 2018-11-30
Stripped non-printing, non-ASCII chars from user agent 

chrlie:
HTTP headers support Latin-1 only. See [this okhttp PR](https://github.com/square/okhttp/issues/891)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -1039,12 +1039,9 @@ public class SalesforceSDKManager {
     public String getUserAgent(String qualifier) {
         String appName = provideAppName();
         String appTypeWithQualifier = getAppType() + qualifier;
-        String unencoded = String.format("SalesforceMobileSDK/%s android mobile/%s (%s) %s/%s %s uid_%s ftr_%s",
+        return String.format("SalesforceMobileSDK/%s android mobile/%s (%s) %s/%s %s uid_%s ftr_%s",
                 SDK_VERSION, Build.VERSION.RELEASE, Build.MODEL, appName, getAppVersion(),
                 appTypeWithQualifier, uid, TextUtils.join(".", features));
-
-        // Strip out all non-printing non-ASCII characters
-        return unencoded.replaceAll("\\P{ASCII}|\\P{Print}", "");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1029,9 +1029,12 @@ public String getUserAgent(String qualifier) {
             SalesforceSDKLogger.w(TAG, "Package info could not be retrieved", e);
         }
         String appTypeWithQualifier = getAppType() + qualifier;
-        return String.format("SalesforceMobileSDK/%s android mobile/%s (%s) %s/%s %s uid_%s ftr_%s",
+        String unencoded = String.format("SalesforceMobileSDK/%s android mobile/%s (%s) %s/%s %s uid_%s ftr_%s",
                 SDK_VERSION, Build.VERSION.RELEASE, Build.MODEL, appName, getAppVersion(),
                 appTypeWithQualifier, uid, TextUtils.join(".", features));
+
+        // Strip out all non-printing non-ASCII characters
```
wmathurin: Those characters probably come from the appname. 
They might not be readable but if we strip them, it will change the app name in our logs - and impact our stats.
What problem are they causing?

bhariharan: @wmathurin 
```
1. The MobileSDK is getting the app name
2. One of our customers is not using a regular quotation mark, but rather \u026F, which is `'`
3. Android is okay with this (which explains why the build worked on the pipeline before) but OkHTTP isn’t okay with it - it doesn’t like unencoded unicode
4. Trying to send a `User-agent` header with this value fails
so the fix is in the MobileSDK and it’s to encode the user agent properly, which requires ISO-8859-1 encoding (ASCII)
5. The customer can't login.
```
Root cause is here -> `https://github.com/square/okhttp/issues/891`.

wmathurin: @bhariharan Are you saying we should encode it better or just strip the characters?

chrlie: After talking to Bharath, I've taken a less aggressive approach and moved the logic to get the app's display name into an overrideable method. This allows me to fix this at the app level.

PR #1821, opened 2018-12-08
Fix NPE in debug menu

bhariharan:
We're using `getCachedCurrentUser()` to get the current user and passing that in via `varargs`. This will be `null` in the unauthenticated use case. We need a `null` check here.


bhariharan:
@wmathurin Tiny bug fix discovered in the unauthenticated use case.

PR #1827, opened 2018-12-14
Dev

wmathurin:


libs/SalesforceAnalytics/AndroidManifest.xml
```diff
@@ -2,8 +2,8 @@
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
 	package="com.salesforce.androidsdk.analytics" 
-	android:versionCode="62"
-	android:versionName="7.0.0.dev">
+	android:versionCode="63"
+	android:versionName="7.1.0">
```
bhariharan: @wmathurin The ones in the manifest should be `7.1.0.dev`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -109,7 +109,7 @@
     /**
      * Current version of this SDK.
      */
-    public static final String SDK_VERSION = "7.0.0.dev";
+    public static final String SDK_VERSION = "7.0.0";
```
bhariharan: Should be `7.1.0.dev`.

libs/SalesforceHybrid/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.salesforce.androidsdk.hybrid"
     android:versionCode="62"
-    android:versionName="7.0.0.dev">
+    android:versionName="7.0.0">
```
bhariharan: Why does this say `7.0.0`?

PR #1839, opened 2019-01-21
Upgrading to SQLCipher 4.0.1

wmathurin:
Taking the path of least resistance, opening every database with 3.x settings (except of kdf iter, where we are still using 2.x value for performance reason).

We should evaluate if 4.x settings have performance benefits (e.g. the larger page size). 
If they do, we will have to either:
- have code to try 4.x settings first then fall back on 3.x settings if that fails (to open older databases),
- or we should have code (or an API) for migrating older databases to 4.x.

### DIFF
--- a/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTest.java
+++ b/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTest.java
@@ -102,6 +102,14 @@ public class SmartStoreTest extends SmartStoreTestCase {
 	}
 
 	/**
+	 * Checking sqlcipher version
+	 */
+	@Test
+	public void testSQLCipherVersion() {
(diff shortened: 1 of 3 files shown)


libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreTest.java
```diff
@@ -101,6 +101,14 @@ public void testCompileOptions() {
         Assert.assertTrue("ENABLE_JSON1 flag not found in compile options", compileOptions.contains("ENABLE_JSON1"));
 	}
 
+	/**
+	 * Checking sqlcipher version
+	 */
+	@Test
+	public void testSQLCipherVersion() {
+		Assert.assertEquals("Wrong sqlcipher version", "4.0.1 community", store.getSQLCipherVersion());
```
wmathurin: We've had that test on iOS for years.

PR #1840, opened 2019-01-22
Improved setversion.sh that takes care of build.gradle files and generate_doc.sh

wmathurin:
Also changed testJavaScriptPluginVersionsWithNewVersion so it no longer needs to be changed with every release

### DIFF
--- a/libs/test/SalesforceHybridTest/src/com/salesforce/androidsdk/phonegap/JavaScriptPluginVersionTest.java
+++ b/libs/test/SalesforceHybridTest/src/com/salesforce/androidsdk/phonegap/JavaScriptPluginVersionTest.java
@@ -144,7 +144,13 @@ public class JavaScriptPluginVersionTest {
 	 */
     @Test
 	public void testJavaScriptPluginVersionsWithNewVersion() {
-		for (String version : new String[] {"7.2.0", "7.3.0", "8.0.0"}) {
+		String[] versionParts = SalesforceSDKManager.SDK_VERSION.split("\\.");
+		String nextMajor = String.join(".","" + (Integer.parseInt(versionParts[0])+1), versionParts[1], versionParts[2]);
+		String nextMinor = String.join(".",versionParts[0], "" + (Integer.parseInt(versionParts[1])+1), versionParts[2]);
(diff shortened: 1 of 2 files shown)


libs/test/SalesforceHybridTest/src/com/salesforce/androidsdk/phonegap/JavaScriptPluginVersionTest.java
```diff
@@ -144,7 +144,13 @@ public void testJavaScriptPluginVersionsWithCurrentVersion() {
 	 */
     @Test
 	public void testJavaScriptPluginVersionsWithNewVersion() {
-		for (String version : new String[] {"7.2.0", "7.3.0", "8.0.0"}) {
+		String[] versionParts = SalesforceSDKManager.SDK_VERSION.split("\\.");
```
wmathurin: That way we don't need to manually change the test with every release.

setversion.sh
```diff
@@ -88,12 +103,21 @@ else
     VERSION_SUFFIXED=${OPT_VERSION}
 fi
 
+SHORT_VERSION=`echo ${OPT_VERSION} | cut -d. -f1,2`
```
wmathurin: E.g. if version is 7.1.0, short version should be 7.1

setversion.sh
```diff
@@ -88,12 +103,21 @@ else
...
 
 echo -e "${YELLOW}*** SETTING VERSION NAME TO ${OPT_VERSION}, VERSION CODE TO ${OPT_CODE}, IS DEV = ${OPT_IS_DEV} ***${NC}"
 
 echo "*** Updating package.json ***"
 update_package_json "./package.json" "${OPT_VERSION}"
 
+echo "*** Updating build.gradle files ***"
+update_build_gradle "./libs/SalesforceAnalytics/build.gradle" "${OPT_VERSION}"
```
bhariharan: One more file to add -> top-level `build.gradle` at the root of the repo. That doesn't have a `vcsTag` though. It looks like this:
```
allprojects {
  group = 'com.salesforce.mobilesdk'
  version = '7.1.0'
  repositories {
    maven {
      // All of React Native (JS, Android binaries) is installed from npm
      url "$rootProject.projectDir/libs/SalesforceReact/node_modules/react-native/android"
    }
    google()
    jcenter()
  }
}
```

PR #1844, opened 2019-02-07
BatchingSyncUpTarget for Android

wmathurin:
1) Repurposed `AdvancedSyncUpTarget` to support batching (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/89803e5ab2ea7d51999aef676387ffb41f37cf87)
2) Sync manager changes were pretty mininmal (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/78fe3f4a1629cec15d98e74b296fee6745f7eff0)
2) For tests, moved existing sync up tests to a new class `SyncUpTargetTest` and made sure, there was only one place in the whole class where the sync up target was actually built. Added a subclass `BatchingSyncUpTargetTest` which creates `BatchingSyncUpTarget`'s instead of `SyncUpTarget`'s. So all the existing sync up tests are now also run with a batching sync up target. Using batch size 2 to test going to the server more than once during a sync (since existing tests deal with 3-5 dirty records).


libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/BatchingSyncUpTarget.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Construct SyncUpTarget from json
+     * @param target
+     * @throws JSONException
+     */
+    public BatchingSyncUpTarget(JSONObject target) throws JSONException {
+        super(target);
+        this.soupName = target.getString(SOUP_NAME);
+        this.maxBatchSize = target.getInt(MAX_BATCH_SIZE);
```
bhariharan: This should also have a check for max of 25 like the constructor above, i.e. `Math.min(...)`.

PR #1846, opened 2019-02-15
Upgrading to Cordova 8.0.0

bhariharan:
The release is in voting stage right now and should go out in about 48 hours.


bhariharan:
`SmartSyncExplorerHybrid` is working fine locally. The tests are failing here because of expired tokens.

PR #1850, opened 2019-02-20
Reduced memory usage during smartstore native queries

wmathurin:
**Problem:**
With the [deserialization optimization](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1798), we routed native smartstore query through queryAsString: we get the full result in a StringBuffer, turn it into a String, then turn that into a JSONArray. So at one point, we have 3 copies on the stack. For large result sets, that cause a out of memory exceptions. 

**To reproduce the OOMs:**
Add the following parameters to SmartStoreLoadTest (and SmartStoreLoadExternalStorageTest)
{"UpsertQuery1StringIndex1000fields1000characters", Type.string, NUMBER_ENTRIES, 1000, 1000, 1}

**Solution:**
Native smartstore query are back to building the result JSONArray incrementally. 
Both query(...) and queryAsString(...) now calls runQuery(...) which expects either a JSONArray or StringBuffer and getDataFromRow is back.
NB: The optimization is still there fore hybrid queries.

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
@@ -797,9 +797,9 @@ public class SmartStore  {
      * @throws JSONException
 	 */
 	public JSONArray query(QuerySpec querySpec, int pageIndex) throws JSONException {
-		StringBuilder resultBuilder = new StringBuilder();
-		queryAsString(resultBuilder, querySpec, pageIndex);
-		return new JSONArray(resultBuilder.toString());
+		JSONArray resultAsArray = new JSONArray();
+		runQuery(resultAsArray, null, querySpec, pageIndex);
+		return resultAsArray;
 	}
 	/**
 	 * Run a query given by its query Spec, only returned results from selected page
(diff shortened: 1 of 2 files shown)


bhariharan:
Compilation error from `CircleCI`:
```
> Task :libs:SalesforceReact:compileDebugJavaWithJavac FAILED
/home/circleci/SalesforceMobileSDK-Android/libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/SmartStoreReactBridge.java:226: error: unreported exception JSONException; must be caught or declared to be thrown
		JSONObject result = storeCursor.getData(smartStore);
```

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -822,33 +829,54 @@ public void queryAsString(StringBuilder resultBuilder, QuerySpec querySpec, int
 			Cursor cursor = null;
 			try {
 				cursor = DBHelper.getInstance(db).limitRawQuery(db, sql, limit, querySpec.getArgs());
-				resultBuilder.append("[");
+
```
agarwasg: @wmathurin  I wondering if we need to use StringBuffer here instead of StringBuilder? I do see that you are synchronizing on DB object. Maybe that is enough.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -822,33 +829,54 @@ public void queryAsString(StringBuilder resultBuilder, QuerySpec querySpec, int
 			Cursor cursor = null;
 			try {
 				cursor = DBHelper.getInstance(db).limitRawQuery(db, sql, limit, querySpec.getArgs());
```
agarwasg: @wmathurin do we need to set a HARD limit on query? We know that it crashes with 1000 objects. Thoughts on that?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
wmathurin: StringBuffer has the same methods as the StringBuilder , but each method in StringBuffer is synchronized that is StringBuffer is thread safe. We don't need to synchronize here - so no reason to use the slower class.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
wmathurin: It really depends on what you have in the database. So app developer should use a value that makes sense for them.
If you have "reasonable" data e.g. short strings, numbers etc, then you can go a lot further than 1000. If you have "giant" JSON blobs, it's a different story.

PR #1852, opened 2019-02-20
Only one getDataFromRow method (handles both JSONArray and StringBuil…

wmathurin:
…der)


libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreLoadExternalStorageTest.java
```diff
@@ -60,6 +60,7 @@ protected void registerSoup(SmartStore store, String soupName, IndexSpec[] index
     @Parameterized.Parameters(name = "{0}")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][]{
+//                {"UpsertQuery1StringIndex1000fields1000characters", Type.string, NUMBER_ENTRIES, 1000, 1000, 1}, // to push memory utilization
```
agarwasg: Any reason we have commented this out? Is it too slow to keep this test around?

wmathurin: It takes minute even with hardware acceleration. (Not surprising since it inserts 1000 entries with 1000 fields of 1000 characters each - so more than 1 GB total).

PR #1863, opened 2019-03-11
Consolidating KeyStore functionality in a single class

bhariharan:
This pull request consolidates all the existing `KeyStore` functionality in a single place.

Next steps:
- Add methods for generating `ECC` keys.
- Migrate stored key from `SharedPrefs` to `KeyStore`.
- Upgrade code.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
@@ -26,13 +26,27 @@
  */
 package com.salesforce.androidsdk.security;
 
+import android.os.Build;
+import android.security.KeyPairGeneratorSpec;
+import android.security.keystore.KeyGenParameterSpec;
+import android.security.keystore.KeyProperties;
+import android.util.Base64;
+
+import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
 import java.io.IOException;
+import java.math.BigInteger;
+import java.security.KeyPairGenerator;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.PublicKey;
 import java.security.cert.CertificateException;
+import java.util.Calendar;
+
+import javax.security.auth.x500.X500Principal;
 
 /**
  * This class provides utilities to interact with the Android KeyStore.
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -151,57 +137,39 @@ public static String getSHA256Hash(String privateKey) {
...
     /**
-     * Generates a keypair and returns the public key.
+     * Generates an RSA keypair and returns the public key.
      *
      * @param name Alias of the entry in which the generated key will appear in Android KeyStore.
+     * @param length Key length.
      * @return RSA public key.
+     * @deprecated Will be removed in Mobile SDK 8.0. Use {@link KeyStoreWrapper} instead.
      */
     public static synchronized PublicKey getRSAPublicKey(String name, int length) {
```
bhariharan: I marked these `public` methods `deprecated` and moved the implementation inside `KeyStoreWrapper`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -259,53 +227,4 @@ private static String getAddendum(String name) {
         final String suffix = TextUtils.isEmpty(name) ? "" : name;
         return String.format(Locale.US, ADDENDUM, suffix);
     }
-
-    private static void createRSAKeysIfNecessary(String name, int length) {
```
bhariharan: Moved inside `KeyStoreWrapper`.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/security/SalesforceKeyGeneratorTest.java
```diff
@@ -90,31 +87,4 @@ public void testGetEncryptionKey() {
         Assert.assertEquals("Encryption keys with the same name should be the same", id1, id1Again);
         Assert.assertNotSame("Encryption keys with different names should be different", id1, id2);
     }
-
-    @Test
-    public void testGetRSAPublicString() {
```
bhariharan: Moved into the other test class.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+                INSTANCE.keyStore = INSTANCE.loadKeyStore();
+            } catch (Exception e) {
+                SalesforceSDKLogger.e(TAG, "Could not load KeyStore", e);
+            }
+        }
+        return INSTANCE;
+    }
+
+    /**
+     * Generates an RSA keypair and returns the public key.
```
wmathurin: Could we use Elliptic Curve encryption instead of RSA: it's what we use on iOS / key generation is faster / stronger for the same number of bits  (a 256 bit ECC key is equivalent to RSA 3072 bits).

bhariharan: This is existing functionality for encrypted push notifications. I will be adding methods for ECC in the next PR as I mentioned in the description.

PR #1864, opened 2019-03-12
Adding elliptical curve (EC) crypto implementation

bhariharan:
This PR adds support for generating named `EC` key pairs with key length of `256`. It also adds support for `StrongBox Keymaster` where available (on some devices `API 28` and above).

What's next:
- Encryption and decryption methods that use `EC` keys.
- Encrypting our `AES` encryption key with `EC` keys and storing them in the `KeyStore`.
- Refactoring the code to NOT cache the encryption key in memory and fetch it from the `KeyStore` instead.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
@@ -58,6 +58,8 @@ public class KeyStoreWrapper {
 
     private static final String ANDROID_KEYSTORE = "AndroidKeyStore";
     private static final String RSA = "RSA";
+    private static final String EC = "EC";
+    private static final int EC_KEY_LENGTH = 256;
     private static final String TAG = "KeyStoreWrapper";
 
     private static KeyStoreWrapper INSTANCE;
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
```diff
@@ -123,42 +113,107 @@ public synchronized String getRSAPublicString(String name, int length) {
...
-                            .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1)
-                            .build());
+                            .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1);
+
+                    /*
+                     * TODO: Remove this check once minVersion > 28.
+                     */
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                        try {
+                            keyGenParameterSpecBuilder.setIsStrongBoxBacked(true);
```
bhariharan: This is the code that requests a `StrongBox Keymaster` implementation where available. This will give us a trusted standalone hardware where the keys are stored. More details [here](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder.html?hl=en#setIsStrongBoxBacked(boolean)).

PR #1866, opened 2019-03-12
Refactoring usage of encryption and decryption methods for optimized performance

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -220,10 +220,11 @@ public Account getCurrentAccount() {
         	if (account != null) {
 
         		// Reads the user ID and org ID from account manager.
+				final String encryptionKey = SalesforceSDKManager.getEncryptionKey();
```
wmathurin: Should encryptionKey be a string a better typed object?

bhariharan: Could be. I can consider that in the next PR. This PR is just replacing existing methods for now.

PR #1868, opened 2019-03-14
Removing memory caching and adding upgrade steps

bhariharan:
Testing upgrade steps with sample app.

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
@@ -236,8 +236,8 @@ public class Encryptor {
         try {
 
             // Signs with SHA-256.
-            byte [] keyBytes = key.getBytes(StandardCharsets.UTF_8);
-            byte [] dataBytes = data.getBytes(StandardCharsets.UTF_8);
+            byte[] keyBytes = key.getBytes(StandardCharsets.UTF_8);
+            byte[] dataBytes = data.getBytes(StandardCharsets.UTF_8);
             Mac sha;
 
             /*
@@ -250,7 +250,7 @@ public class Encryptor {
             }
             final SecretKeySpec keySpec = new SecretKeySpec(keyBytes, sha.getAlgorithm());
             sha.init(keySpec);
-            byte [] sig = sha.doFinal(dataBytes);
+            byte[] sig = sha.doFinal(dataBytes);
 
             // Encodes with Base64.
             return Base64.encodeToString(sig, Base64.NO_WRAP);
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
@@ -284,17 +284,7 @@ public class Encryptor {
      * @return Encrypted data.
      */
     public static byte[] encryptWithRSABytes(PublicKey publicKey, String data) {
-        if (publicKey == null || TextUtils.isEmpty(data)) {
-            return null;
-        }
-        try {
-            final Cipher cipher = Cipher.getInstance(RSA_PKCS1);
-            cipher.init(Cipher.ENCRYPT_MODE, publicKey);
-            return cipher.doFinal(data.getBytes());
-        } catch (Exception e) {
-            SalesforceAnalyticsLogger.e(null, TAG, "Error during asymmetric encryption using RSA", e);
-        }
-        return null;
+        return encryptWithPublicKey(publicKey, data, RSA_PKCS1);
     }
 
     /**
@@ -325,18 +315,7 @@ public class Encryptor {
      * @return Decrypted data.
      */
     public static byte[] decryptWithRSABytes(PrivateKey privateKey, String data) {
-        if (privateKey == null || TextUtils.isEmpty(data)) {
-            return null;
-        }
-        try {
-            final Cipher cipher = Cipher.getInstance(RSA_PKCS1);
-            cipher.init(Cipher.DECRYPT_MODE, privateKey);
-            byte[] decodedBytes = Base64.decode(data.getBytes(),Base64.NO_WRAP | Base64.NO_PADDING);
-            return cipher.doFinal(decodedBytes);
-        } catch (Exception e) {
-            SalesforceAnalyticsLogger.e(null, TAG, "Error during asymmetric decryption using RSA", e);
-        }
-        return null;
+        return decryptWithPrivateKey(privateKey, data, RSA_PKCS1);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -72,6 +76,21 @@ protected synchronized void upgradeAccMgr() {
...
         // Update shared preference file to reflect the latest version.
         writeCurVersion(ACC_MGR_KEY, SalesforceSDKManager.SDK_VERSION);
+
+        /*
+         * If the installed version < v7.1.0, we need to store the current
+         * user's user ID and org ID in a shared preference file, to
+         * support fast user switching.
+         */
+        try {
+            final String majorVersionNum = installedVersion.substring(0, 3);
```
bhariharan: Upgrade hookup.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -57,17 +58,14 @@
...
-    private static final String ADDENDUM = "addendum_%s";
+    private static final String ENCRYPTED_ID_SHARED_PREF_KEY = "encrypted_id_%s";
+    private static final String KEYSTORE_ALIAS = "com.salesforce.androidsdk.security.KEYPAIR";
     private static final String UTF8 = "UTF-8";
     private static final String SHA1 = "SHA-1";
     private static final String SHA256 = "SHA-256";
     private static final String SHA1PRNG = "SHA1PRNG";
     private static final String AES = "AES";
 
-    private static Map<String, String> UNIQUE_IDS = new HashMap<>();
```
bhariharan: Removed memory caching.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -168,42 +160,67 @@ public static synchronized String getRSAPublicString(String name, int length) {
...
+    public static PrivateKey getRSAPrivateKey(String name, int length) {
         return KeyStoreWrapper.getInstance().getRSAPrivateKey(name, length);
     }
 
-    private static void generateEncryptionKey(String name) {
+    /**
+     * Upgrades the keys stored in SharedPrefs to encrypted keys. This is a one-time
+     * migration step that's run while upgrading to Mobile SDK 7.1.
+     */
+    public synchronized static void upgradeTo7Dot1() {
```
bhariharan: Upgrade steps. We basically read the old keys, encrypt them and store them back as encrypted with RSA.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -284,17 +284,7 @@ public static String encryptWithRSA(PublicKey publicKey, String data) {
...
-        }
-        try {
-            final Cipher cipher = Cipher.getInstance(RSA_PKCS1);
-            cipher.init(Cipher.ENCRYPT_MODE, publicKey);
-            return cipher.doFinal(data.getBytes());
-        } catch (Exception e) {
-            SalesforceAnalyticsLogger.e(null, TAG, "Error during asymmetric encryption using RSA", e);
-        }
-        return null;
+        return encryptWithPublicKey(publicKey, data, RSA_PKCS1);
```
wmathurin: Doesn't the key know its own algorithm? https://developer.android.com/reference/java/security/Key.html#getAlgorithm()

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -168,42 +160,67 @@ public static synchronized String getRSAPublicString(String name, int length) {
...
+        final SharedPreferences prefs = SalesforceSDKManager.getInstance().getAppContext().getSharedPreferences(SHARED_PREF_FILE, 0);
+        final Map<String, ?> prefContents = prefs.getAll();
+        if (prefContents != null) {
+            final Set<String> keys = prefContents.keySet();
+            for (final String key : keys) {
+                if (key != null && key.startsWith("id")) {
+                    final String value = prefs.getString(key, null);
+                    if (value != null) {
+                        final PublicKey publicKey = KeyStoreWrapper.getInstance().getECPublicKey(KEYSTORE_ALIAS);
```
wmathurin: Should it be `KeyStoreWrapper.getInstance().getRSAPublicKey`

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -168,42 +160,67 @@ public static synchronized String getRSAPublicString(String name, int length) {
...
+                        final String encryptedValue = Encryptor.encryptWithRSA(publicKey, value);
```
wmathurin: If the key knows its algorithm, could that call just be `Encryptor.encrypt(publicKey, value)`

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
bhariharan: That method call would return "RSA", not "RSA/ECB/PKCS1Padding".

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
bhariharan: Nice catch. Too quick on the auto-complete.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
bhariharan: We can't do that - not all algorithms are supported in `Android`. The `Cipher` class on `Android` has a limited subset that's supported. For example, `EC` isn't supported.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
wmathurin: Maybe we could our own key object wrapped around Android's and Encryptor would consume that.

bhariharan: Maybe we could, but I don't think it would add too much value, not at this time anyway. We still need this API the way it is because S1 uses this for push notification encryption/decryption. There's not much we would add to `PublicKey` or `PrivateKey` right now. I would prefer leaving this as is for now. Perhaps we can re-evaluate if there is more extensive usage of keys.

bhariharan:
@brandonpage @wmathurin Testing is complete. Everything looks great!

Tests that I ran:
- `GET` actions.
- `POST` actions.
- Token refresh.
- Force killing the app and restarting.
- Offline edit and sync.
- Upgrading from an app running `Mobile SDK 7.0` to `Mobile SDK 7.1` for key migration (and then running the above tests).

API versions of emulators that I ran the tests against:
- `API 28`.
- `API 26`.

Apps that were run:
- `RestExplorer`.
- `SmartSyncExplorer`.

Merging now.

PR #1870, opened 2019-03-18
Pull test creds from env

brandonpage:
- CI should seamlessly pickup creds
- Quick setup is documented internally
- External developers should supply their own creds to run tests locally.


.circleci/setTestCreds.rb
```diff
@@ -0,0 +1,24 @@
+boot_configs = %w(libs/test/SalesforceAnalyticsTest/res/values/bootconfig.xml
```
wmathurin: Could you add #!/usr/bin/ruby

wmathurin:
What's the idea when running the tests locally? Set the env variables manually then run your ruby script then make sure to not to commit the locally modified files? Sounds cumbersome.

PR #1871, opened 2019-03-20
Make the SDK respect connected app biometric attribute.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -174,8 +178,9 @@ public int getPasscodeLengthForOrg(UserAccount account) {
      * @param account UserAccount instance.
      * @param timeout Timeout value, in ms.
      * @param passLen Minimum passcode length.
+     * @param bioAllowed If biometric Unlock is Allowed by connected App
      */
-    public void storeMobilePolicyForOrg(UserAccount account, int timeout, int passLen) {
+    public void storeMobilePolicyForOrg(UserAccount account, int timeout, int passLen, boolean bioAllowed) {
```
bhariharan: Don't change this public API. Instead, add a new overloaded public API with the same name and the additional parameter and call that new API from this passing in `true` as the default value. Switch out all internal calls to the old method with the new one. Mark this deprecated too for removal in `8.0`.

```
     * @deprecated Will be removed in Mobile SDK 8.0. Use {@link blah} instead.
```

bhariharan: Semantic versioning prevents us from removing this directly even though it's unlikely anyone's using this.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -185,6 +190,7 @@ public void storeMobilePolicyForOrg(UserAccount account, int timeout, int passLe
         final Editor e = sp.edit();
         e.putInt(KEY_TIMEOUT, timeout);
         e.putInt(KEY_PASSCODE_LENGTH, passLen);
+        e.putBoolean(KEY_BIOMETRIC_ALLOWED, bioAllowed);
```
bhariharan: You'll need an upgrade step to add the default value here with the value `true`. Wait for my PR to go in before adding that though - you'll get the upgrade module for free.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
brandonpage: 👍

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -174,6 +178,9 @@ public int getPasscodeLengthForOrg(UserAccount account) {
      * @param account UserAccount instance.
      * @param timeout Timeout value, in ms.
      * @param passLen Minimum passcode length.
+     *
+     * @deprecated Will be removed in Mobile SDK 8.0.
+     * Use {@link PasscodeManager#storeMobilePolicyForOrg(UserAccount, int, int, boolean)} instead.
      */
     public void storeMobilePolicyForOrg(UserAccount account, int timeout, int passLen) {
     	if (account == null) {
```
bhariharan: Instead of duplicating this code, just replace this method body with:
```
storeMobilePolicyForOrg(account, timeout, passLen, true);
```

PR #1872, opened 2019-03-20
Sync manager cleans up interrupted syncs state at startup

wmathurin:
* Added index (on status) in syncs soup
* Changed indexes on syncs soup to json1
* Added clean up method (and calling it at SyncManager creation)
* Added tests

### DIFF
--- a/libs/SmartSync/AndroidManifest.xml
+++ b/libs/SmartSync/AndroidManifest.xml
@@ -5,10 +5,6 @@
     android:versionCode="63"
     android:versionName="7.1.0.dev">
 
-    <application>
-        <service android:name="com.salesforce.androidsdk.analytics.AnalyticsPublisherService"
-            android:enabled="false" />
-    </application>
-
+    <application />
 
 </manifest>
--- a/libs/SmartSync/AndroidManifest.xml
+++ b/libs/SmartSync/AndroidManifest.xml
@@ -5,5 +5,10 @@
     android:versionCode="63"
     android:versionName="7.1.0.dev">
 
-    <application />
+    <application>
+        <service android:name="com.salesforce.androidsdk.analytics.AnalyticsPublisherService"
+            android:enabled="false" />
+    </application>
+
+
 </manifest>
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
@@ -94,6 +94,7 @@ public class SyncManager {
         this.smartStore = smartStore;
         this.restClient = restClient;
         SyncState.setupSyncsSoupIfNeeded(smartStore);
+        SyncState.cleanupSyncsSoupIfNeeded(smartStore);
     }
 
     /**
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncState.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncState.java
@@ -27,6 +27,8 @@
 package com.salesforce.androidsdk.smartsync.util;
 
 import com.salesforce.androidsdk.smartstore.store.IndexSpec;
+import com.salesforce.androidsdk.smartstore.store.QuerySpec;
+import com.salesforce.androidsdk.smartstore.store.SmartSqlHelper;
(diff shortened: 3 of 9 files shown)


libs/SmartSync/AndroidManifest.xml
```diff
@@ -5,5 +5,10 @@
     android:versionCode="63"
     android:versionName="7.1.0.dev">
 
-    <application />
+    <application>
+        <service android:name="com.salesforce.androidsdk.analytics.AnalyticsPublisherService"
```
wmathurin: Disabling ailtn in this test suite (we should do the same in other test suites).

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -94,6 +94,7 @@ private SyncManager(SmartStore smartStore, RestClient restClient) {
         this.smartStore = smartStore;
         this.restClient = restClient;
         SyncState.setupSyncsSoupIfNeeded(smartStore);
+        SyncState.cleanupSyncsSoupIfNeeded(smartStore);
```
wmathurin: The cleanup call

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncState.java
```diff
@@ -82,19 +84,20 @@
...
+    	if (store.hasSoup(SYNCS_SOUP) && store.getSoupIndexSpecs(SYNCS_SOUP).length == 3) {
 			return;
 		}
 
 		final IndexSpec[] indexSpecs = {
-			new IndexSpec(SYNC_TYPE, SmartStore.Type.string),
-			new IndexSpec(SYNC_NAME, SmartStore.Type.string)
+			new IndexSpec(SYNC_TYPE, SmartStore.Type.json1),
+			new IndexSpec(SYNC_NAME, SmartStore.Type.json1),
+			new IndexSpec(SYNC_STATUS, SmartStore.Type.json1)
```
wmathurin: New index / index changes

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncState.java
```diff
@@ -82,19 +84,20 @@
...
 		};
 
 		// Syncs soup exists but doesn't have all the required indexes
 		if (store.hasSoup(SYNCS_SOUP)) {
 			try {
-				store.alterSoup(SYNCS_SOUP, indexSpecs, false);
+				store.alterSoup(SYNCS_SOUP, indexSpecs, true /* reindexing to json1 is quick */);
```
wmathurin: reIndexing when going to json1 is really fast since it's just a bunch of create index calls instead of having to read all rows back to memory / extracting indexed paths / saving back to db.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SyncState.java
```diff
@@ -105,6 +108,28 @@ public static void setupSyncsSoupIfNeeded(SmartStore store) {
...
+	 * At startup, no sync could be running already
+	 * If a sync is in the running state, we change it to failed
+	 * @param store
+	 */
+	public static void cleanupSyncsSoupIfNeeded(SmartStore store) {
+		try {
+			QuerySpec runningSyncQuery = QuerySpec.buildSmartQuerySpec(String.format("select {%1$s:%2$s} from {%1$s} where {%1$s:%3$s} = '%4$s'", SYNCS_SOUP, SmartSqlHelper.SOUP, SYNC_STATUS, Status.RUNNING.name()), Integer.MAX_VALUE);
+			JSONArray rows = store.query(runningSyncQuery, 0);
+			for (int i=0; i<rows.length(); i++) {
+				JSONObject sync = rows.getJSONArray(i).getJSONObject(0);
```
wmathurin: Mark running job as failed

libs/SmartSync/AndroidManifest.xml
bhariharan: Good idea. Could you add them as part of this PR?

wmathurin:
@bhariharan Good idea. It looks like I put it in the wrong manifest (lib manifest instead of test app manifest).
I noticed the issue while trying to run the new tests that don't need valid credentials (the ailtn code however was trying to call up to the server and crash the test).

Issue #1787, opened 2018-09-19
SoqlSyncDownTarget with subquery issue


jpttrindade:
1. Version of Mobile SDK Used: 6.0.0
2. Issue found in Native App or Hybrid App: Native App
3. OS Version: Android 8.0 
4. Device: Moto G6

5. Steps to reproduce: 
i) Create a soql syncdown to any object (like Opportunity) with a subquery as filter.   
Sample: SELECT Id where OwnerId in (select Id from Account)
ii) run resync
iii) run cleanResyncGhosts from SyncManager to this object

6. Actual behavior:
A error of "MALFORMED_QUERY" occurs because the handler of filter(subquery in this case). 
At the SoqlSyncDownTarget class and his getSoqlForRemoteIds methods, the 'split' in atcual query not support subquery sintaxe because put away part of subquery that contains  'from' string;
```
Sample:  
array =  "select X from Y where a = (select M from N)".split('from') will to return a array with 3 values
1. array[0] = 'select X '
2. array[1] =  ' Y where a = (select M '
3. array[2] = ' N)'
 and method get only array[1] to build final query to request:
```

**SNAPSHOT:**
```
final StringBuilder soql = new StringBuilder("SELECT ");
        soql.append(getIdFieldName());
        soql.append(" FROM ");
        final String[] fromClause = getQuery(0).split("([ ][fF][rR][oO][mM][ ])");
        soql.append(fromClause[1]);
        return soql.toString();
```

7. Expected Behavior:
At the SoqlSyncDownTarget class and his getSoqlForRemoteIds methods the query split should result e only 2 parts, where only first 'from' is handled to support subqueries.

```
(...) final String[] fromClause 
 = getQuery(0).split("([ ][fF][rR][oO][mM][ ])", 2); // note second parameter
where, fromClause is:
fromClause[0] =  'select X '
fromClause[1] = ' Y where a = (select M from N)'

// fromClause[1] is entire second part of query.
```

8. Error Log:

```
{
        "message": "\nAccount WHERE Id in (SELECT AccountId\n                                    ^\nERROR at Row:1:Column:52\nunexpected token: '<EOF>'",
        "errorCode": "MALFORMED_QUERY"
      }
```

wmathurin:
Fixing with https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1873
[...]

PR #1873, opened 2019-03-22
Handling subqueries when computing soql for remote ids

wmathurin:
Fix for https://github.com/forcedotcom/SalesforceMobileSDK-Android/issues/1787

### DIFF
--- a/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java
+++ b/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java
@@ -179,12 +179,23 @@ public class SoqlSyncDownTarget extends SyncDownTarget {
     }
 
     protected String getSoqlForRemoteIds() {
+	    // Full query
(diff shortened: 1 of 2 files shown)


wmathurin:
More work needed

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SoqlSyncDownTarget.java
```diff
@@ -179,12 +180,33 @@ public JSONArray continueFetch(SyncManager syncManager) throws IOException, JSON
...
+
+            if (afterFrom) {
+                soql.append(" ").append(token);
+            }
+
+            if (token.startsWith("(")) {
+                depth++;
+            } else if (token.endsWith(")")) {
+                depth--;
+            } else if (depth == 0 && afterFrom == false && token.toLowerCase().equals("from")) {
```
bhariharan: Wouldn't this just be `!afterFrom` instead of `afterFrom == false`? Also, we should probably change `toLowerCase()` to `toLowerCase(Locale.US)`.

PR #1874, opened 2019-03-22
Adding StrongBox Keymaster protection for Android P

bhariharan:


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
```diff
@@ -228,8 +229,22 @@ private synchronized void createKeysIfNecessary(String algorithm, String name, i
...
+
+                    /*
+                     * TODO: Remove this check once minVersion > 28.
+                     */
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                        try {
+                            keyGenParameterSpecBuilder.setIsStrongBoxBacked(true);
+                            kpg.initialize(keyGenParameterSpecBuilder.build());
+                            kpg.generateKeyPair();
+                        } catch (StrongBoxUnavailableException sb) {
```
bhariharan: Unfortunately, there's no way to check if `StrongBox Keymaster` is available before generating the key. So, if it's unavailable, we have to catch the exception, set it back to `false` and re-generate the keys (they will fail the first time if unavailable).

wmathurin: Is upgrade already handled?
I'm on a device with strong box, I upgrade the app from a version using sdk 7.0 to a version using sdk 7.1. Will I get a strong box back master key then?

bhariharan: Yes - upgrade steps don't change. This code is run only once - when the keys are generated the first time (during upgrade or new app installation). Also, `StrongBox` availability never changes - either a device has it (and will always have it), or it doesn't have it (and will never have it).

Also, yes - if you upgrade on a `Pixel 3` from `7.0` to `7.1`, you'll get a key that lives on the `Titan Chip StrongBox`.

wmathurin: 👍

bhariharan: More info on `StrongBox` and what algorithms are available there -> `https://developer.android.com/training/articles/keystore?hl=en#HardwareSecurityModule`.

PR #1877, opened 2019-03-28
test_credentials.json for Android

wmathurin:
NB: still in the process of validating that all tests are passing


libs/SalesforceSDK/src/com/salesforce/androidsdk/util/test/TestCredentials.java
```diff
@@ -24,15 +24,20 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-package com.salesforce.androidsdk.smartsync;
+package com.salesforce.androidsdk.util.test;
```
wmathurin: One one TestCredentials.java file living in the util.test package.
It reads the credentials from test_credentials.json.
test_credentials.json is under shared/test, but we have links to it in the assets folder of every test project.

shared/test/test_credentials.json.sample
```diff
@@ -0,0 +1,27 @@
...
+
+"test_redirect_uri":"__INSERT_REMOTE_ACCESS_CALLBACK_URL_HERE__",
+
+"refresh_token":"__INSERT_TOKEN_HERE__",
+ 
+"instance_url":"__INSERT_INSTANCE_URL_HERE__",
+
+"identity_url":"__INSERT_IDENTITY_URL_HERE__"
+
+"organization_id": "__INSERT_ORG_ID_HERE__",
```
wmathurin: Note the extra fields.

To create a test_credentials.json use the [credsHelper](https://github.com/forcedotcom/SalesforceMobileSDK-Shared/tree/dev/credshelper)

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/test/TestCredentials.java
```diff
@@ -51,18 +56,25 @@
...
+
+            API_VERSION = ApiVersionStrings.getVersionNumber(ctx);
+            ACCOUNT_TYPE = ctx.getString(R.string.account_type);
+            ORG_ID = json.getString("organization_id");
+            USERNAME = json.getString("username");
+            ACCOUNT_NAME = json.getString("display_name");
+            USER_ID = json.getString("user_id");
+            LOGIN_URL = json.getString("test_login_domain");
+            INSTANCE_URL = json.getString("instance_url");
+            COMMUNITY_URL = json.optString("community_url", INSTANCE_URL /* in case the test_credentials.json was obtained for a user / org without community setup */);
```
wmathurin: If you generate test_credentials.json by login to an org without community setup, it won't have community_url.
In that case we just use the instance_url.

libs/test/SalesforceReactTest/src/com/salesforce/androidsdk/reactnative/util/ReactTestActivity.java
```diff
@@ -39,23 +40,23 @@
...
+    static String photoUrl = TestCredentials.PHOTO_URL;
+    static String clientId = TestCredentials.CLIENT_ID;
 
     @Override
     protected ClientManager buildClientManager() {
         final ClientManager clientManager = super.buildClientManager();
-        final ClientManager.LoginOptions loginOptions = SalesforceSDKManager.getInstance().getLoginOptions();
         clientManager.createNewAccount(accountName, username, refreshToken, authToken, instanceUrl,
-                loginUrl, identityUrl, loginOptions.getOauthClientId(), orgId, userId,
+                loginUrl, identityUrl, clientId, orgId, userId,
```
wmathurin: That's important - loginOptions.getOauthClientId() returns the client id from the bootconfig / which might not match the actual client id in test_credentials.json => not only refresh will fail, but the refresh token will be invalidated!

brandonpage:
Looking really good so far!

wmathurin:
Locally was able to run all tests.
All passed except for a SalesforceKeyGeneratorTest:testGetUniqueId and a couple of react native tests.

libs/test/SalesforceReactTest/src/com/salesforce/androidsdk/reactnative/util/ReactTestActivity.java
bhariharan: So this is what was breaking our tests all this while?

shared/test/test_credentials.json.sample
```diff
@@ -0,0 +1,27 @@
+{
+
+"access_token":"__NOT_REQUIRED__",
+ 
+"test_client_id":"__INSERT_REMOTE_ACCESS_CLIENT_KEY_HERE__",
+
+"test_login_domain":"test.salesforce.com",
```
bhariharan: Shouldn't this be `__INSERT_LOGIN_URL_HERE__`?

PR #1878, opened 2019-03-28
Issues with unauthenticated requests in hybrid / react native app/tests

wmathurin:
1) Hybrid / react pull test credentials in the clientManager built by the react / hybrid test activity but unauthenticated rest client were built with the clientManager obtained from SalesforceSDKManager
2) Hybrid / react net plugin didn't not expect end point to be null, but it could be null when hitting other servers (the path will then contain the full url).

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -198,7 +198,7 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
         final JSONObject arg0 = args.optJSONObject(0);
         if (arg0 != null) {
             final RestRequest.RestMethod method = RestRequest.RestMethod.valueOf(arg0.optString(METHOD_KEY));
-            final String endPoint = arg0.optString(END_POINT_KEY, "");
+            final String endPoint = arg0.optString(END_POINT_KEY);
             final String path = arg0.optString(PATH_KEY);
             final String queryParamString = arg0.optString(QUERY_PARAMS_KEY);
             JSONObject queryParams = new JSONObject();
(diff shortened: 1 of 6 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -198,7 +198,7 @@ private RestRequest prepareRestRequest(JSONArray args) throws UnsupportedEncodin
         final JSONObject arg0 = args.optJSONObject(0);
         if (arg0 != null) {
             final RestRequest.RestMethod method = RestRequest.RestMethod.valueOf(arg0.optString(METHOD_KEY));
-            final String endPoint = arg0.optString(END_POINT_KEY);
+            final String endPoint = arg0.optString(END_POINT_KEY, "");
```
wmathurin: null value causes NPE at line 235

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -244,7 +244,7 @@ private RestClient getRestClient(boolean doesNotRequireAuth) {
             return null;
         }
         if (doesNotRequireAuth) {
-            return SalesforceHybridSDKManager.getInstance().getClientManager().peekUnauthenticatedRestClient();
+            return currentActivity.buildClientManager().peekUnauthenticatedRestClient();
```
wmathurin: Tests use an activity that override buildClientManager() to build a ClientManager using test credentials.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/SalesforceNetReactBridge.java
```diff
@@ -145,7 +145,7 @@ private RestRequest prepareRestRequest(ReadableMap args) throws UnsupportedEncod
 
         // Parse args
         RestRequest.RestMethod method = RestRequest.RestMethod.valueOf(args.getString(METHOD_KEY));
-        String endPoint = args.getString(END_POINT_KEY);
+        String endPoint = args.isNull(END_POINT_KEY) ? "" : args.getString(END_POINT_KEY);
```
wmathurin: Null value causes a NPE line 171.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
bhariharan: `optString` is supposed to return empty string (NOT `null`) based on documentation [here](https://developer.android.com/reference/org/json/JSONObject?hl=en#optString(java.lang.String)). Are you sure it's coming back with `null`?

PR #1879, opened 2019-03-29
Stop/resume support in SmartSync

wmathurin:
1) Moved code in `SyncManager` that was run on a worker thread to separate classes: `SyncTask` (the super class - a `Runnable`) and the subclasses: `SyncDownTask`, `SyncUpTask`, `AdvanvedSyncUpTask` and `CleanSyncGhostsTask`
2) `SyncManager` has a map of sync id to `SyncTask` for active syncs (instead of a set of sync ids)
3) New `SyncState.Status`: `stopped`
4) `SyncManager` has new methods: `stop()`, `isStopped()`, `isStopping()`, `resume(boolean restartStoppedSyncs, SyncUpdateCallback callback)`, `checkIfStopRequested()`: `SyncManager` has now a state (accepting_syncs, stopping, stopped) which get sets accordingly. All the sync tasks check it regularly by calling `checkIfStopRequested()` which throws a `SyncManagerStoppedException` if stop was requested. The run methods of `SyncTask` catch that exception and update the sync to `STOPPED`.
5) New method in `SyncDownTarget`: `isSyncDownSortedByLatestModification()`, when true, the maxTimeStamp is updated in the sync state as the sync runs, otherwise it's only set once the sync completes => that way, if we resume a sync that was stopped in flight, we don't need to refetch everything
6) New class for SOQL manipulation: `SFSDKSoqlMutator`.
7) `SOQLSyncDownTarget` attempts to add a `order by last-modified-field`, if it succeeds its `isSyncDownSortedByLatestModification()` method returns `true`.
8) `ParentChildrenSyncDownTarget` now use `order by last-modified-field`.
9) Resuming a stopped sync is like doing a `reSync()` with one difference, records at `maxTimeStamp` will be refetched (so instead of looking for records with `timestamp > maxTimeStamp`, we look for records with `timestamp >= maxTimeStamp`)
10) At startup, syncs found in the `running` state are changed to the `stopped` state (instead of the `failed` state)


wmathurin:
NB: I have tests for stopping/resuming a sync down. But I need to add tests for stopping syncUp's and cleanResyncGhosts.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/CleanSyncGhostsTask.java
```diff
@@ -0,0 +1,92 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Runnable class responsible for running clean sync ghosts operation
+ */
+public class CleanSyncGhostsTask extends SyncTask {
+
+    // Constants
+    private static final String TAG = "SyncManager";
```
bhariharan: `TAG` should be `CleanSyncGhostsTask`.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -196,6 +201,77 @@ public static synchronized void reset(UserAccount account) {
...
+     */
+    public boolean isStopping() {
+        return stopRequested && activeSyncs.size() > 0;
+    }
+
+    /**
+     * @return true if stop was requested and there no syncs are active anymore
+     */
+    public boolean isStopped() {
+        return stopRequested && activeSyncs.size() == 0;
```
bhariharan: Once we stop it, wouldn't we reset the value of `stopRequested`, i.e. `stopRequested = false`? Should this condition be just `return activeSyncs.size() == 0`, i.e. no active syncs running?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
```diff
@@ -77,13 +69,24 @@
...
     private static Map<String, SyncManager> INSTANCES = new HashMap<String, SyncManager>();
 
-    // Members
-    private Set<Long> runningSyncIds = new HashSet<Long>();
-    public final String apiVersion;
+    // Keeping track of active syncs (could be waiting for thread or running on a thread)
+    private Map<Long, SyncTask> activeSyncs = new HashMap<>();
+
+    // Flag set when stop is requested
+    private boolean stopRequested = false;
```
bhariharan: Instead of having a `boolean` based system, should we consider using an `enum` for states -> `STOP_REQUESTED`, `STOPPING_IN_PROGRESS`, `STOP_COMPLETE`, `RESUME_REQUESTED`, `RESUME_IN_PROGRESS`, and then finally, `SYNC_COMPLETE`? If we use `RESUME`, we should probably pair it with `PAUSE` for clarity.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: The idea was once the SyncManager is stopped, it won't accept new jobs (and will try to wrap up existing jobs as quickly as possible). The app has to call resume to allow jobs to be submitted again. If I flip the flag back, jobs submission won't be prevented.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: That sounds good. I'm not sure about about the SYNC_COMPLETE state, because once you resume, new jobs could be submitted, so there might always be sync in flights. I don't know if we want to separately track syncs submitted before the stop from the one submitted after the resume.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: Gotcha.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
bhariharan: Should we call it `PAUSE` instead of `STOP`, since we have `RESUME`?

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncTask.java
```diff
@@ -0,0 +1,153 @@
...
+import com.salesforce.androidsdk.smartsync.util.SmartSyncLogger;
+import com.salesforce.androidsdk.smartsync.util.SyncState;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Abstract super class of runnable classes responsible for running syncs
+ */
+public abstract class SyncTask implements Runnable {
```
bhariharan: We don't have to change this now, but in the future we might want to consider using `AsyncTask` here instead of `Runnable`. `AsyncTask` is more Android lifecycle friendly and it's easier to get callbacks and make UI changes if required on the main thread directly.

wmathurin: Looking at https://developer.android.com/topic/performance/threads.html, it says:
> For this reason, we suggest that you only use AsyncTask to handle work items shorter than 5ms in duration.
So not quite our use case.

I also considered using java.concurrent classes like FutureTask. However, cancellation relies on Thread.interrupt() (see https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/java/util/concurrent/FutureTask.java#L174) but that doesn't play nice with OkHttp (tons of issues around it if you google for it).

bhariharan: Cool! 👍

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncManager.java
wmathurin: I started by calling it `pause` but then it's not a perfect `pause` (I'm restarting the sync and reducing repeated work by leveraging the maxTimeStamp). For SOQL sync down a true pause/resume would leverage the `nextRecordsUrl` of the started SOQL request. The problem with that approach is that that URL is not available for ever, so resuming could fail if it happens too long after the pause. For that reason, I decided to keep it simple and leverage what we already have (maxTimeStamp) with the only change being that maxTimeStamp is updated as the sync runs (instead of simply at the end) if the sync's target returns true in for isSyncDownSortedByLatestModification().

Maybe for symmetry, I should rename `resume` to `restart`?

bhariharan: Up to you - I don't feel strongly either way.

libs/SmartSync/src/com/salesforce/androidsdk/smartsync/util/SOQLMutator.java
```diff
@@ -0,0 +1,207 @@
+/*
+ * Copyright (c) 2014-present, salesforce.com, inc.
```
bhariharan: `2019`.

PR #1889, opened 2019-04-12
Passcode UI updates

brandonpage:
- Fix UI not changing after first unlock on app upgrade
- Fix keyboard focus
- Don't allow passcode length to be lowered
- Fix passcode does not match reentering when creating

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
@@ -616,12 +616,14 @@ public class PasscodeManager  {
      */
     public void setPasscodeLength(Context ctx, int passcodeLength) {
     	if (passcodeLength > this.passcodeLength) {
-            if (hasStoredPasscode(ctx)) {
+            if (hasStoredPasscode(ctx) && passcodeLengthKnown) {
                 this.passcodeChangeRequired = true;
             }
+
+            this.passcodeLength = passcodeLength;
     	}
-        this.passcodeLength = passcodeLength;
-    	this.passcodeLengthKnown = true;
+
+        this.passcodeLengthKnown = true;
         storeMobilePolicy(ctx);
     }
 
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -616,12 +616,14 @@ public void setMinPasscodeLength(Context ctx, int minPasscodeLength) {
...
     public void setPasscodeLength(Context ctx, int passcodeLength) {
     	if (passcodeLength > this.passcodeLength) {
-            if (hasStoredPasscode(ctx)) {
+            if (hasStoredPasscode(ctx) && passcodeLengthKnown) {
                 this.passcodeChangeRequired = true;
             }
+
+            this.passcodeLength = passcodeLength;
     	}
-        this.passcodeLength = passcodeLength;
```
brandonpage: @bhariharan Why was passcode length requirement allowed to be lowered?

bhariharan: Passcode length could be lowered if you log out of an account with a longer length requirement, when there are 2 accounts with passcode enabled.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
```diff
@@ -642,4 +647,11 @@ private void clearUi() {
         biometricBox.setVisibility(View.GONE);
         fingerImage.setVisibility(View.GONE);
     }
+
+    private void hideKeyboard() {
+        InputMethodManager imm = (InputMethodManager) this.getSystemService(Activity.INPUT_METHOD_SERVICE);
+        if (this.passcodeField != null) {
```
bhariharan: `this` is implicit. You don't need to explicitly add it.

brandonpage: Whoops.  Made the change locally, I will commit it when UX feedback comes back.

PR #1906, opened 2019-05-15
Fixing issue with re-login after logging out

bhariharan:
This fixes an issue with re-login after logging out for users with push notifications enabled.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -169,6 +169,7 @@ public class SalesforceSDKManager {
     private boolean browserLoginEnabled;
     private String idpAppURIScheme;
     private boolean idpAppLoginFlowActive;
+    private int numPostPushUnregisterCalls;
 
     /**
      * PasscodeManager object lock.
@@ -815,6 +816,7 @@ public class SalesforceSDKManager {
             @Override
             public void onReceive(Context context, Intent intent) {
                 if (PushMessaging.UNREGISTERED_ATTEMPT_COMPLETE_EVENT.equals(intent.getAction())) {
+                    numPostPushUnregisterCalls++;
                     postPushUnregister(this, clientMgr, showLoginPage,
                     		refreshToken, loginServer, account, frontActivity);
                 }
@@ -838,6 +840,7 @@ public class SalesforceSDKManager {
                     // Waits for half a second at a time.
                     SystemClock.sleep(500);
                 }
+                numPostPushUnregisterCalls++;
                 postPushUnregister(pushUnregisterReceiver, clientMgr, showLoginPage,
                 		refreshToken, loginServer, account, frontActivity);
             }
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -848,12 +851,24 @@ private void postPushUnregister(BroadcastReceiver pushReceiver,
...
+         */
+        if (numPostPushUnregisterCalls <= 1) {
+            try {
+                context.unregisterReceiver(pushReceiver);
+            } catch (Exception e) {
+                SalesforceSDKLogger.e(TAG, "Exception occurred while un-registering", e);
+            }
+            removeAccount(clientMgr, showLoginPage, refreshToken, loginServer, account, frontActivity);
+        } else {
+            numPostPushUnregisterCalls = 0;
```
wmathurin: I'm not sure it will work in all cases. For instance:
* What if broadcast receiver does `numPostPushUnregisterCalls++`, followed by `postPushUnregister()` which gets to the else and therefore sets `numPostPushUnregisterCalls` back to 0.
At that point, the background thread that was sleeping, does `numPostPushUnregisterCalls++`, followed by `postPushUnregister()`, the `if (numPostPushUnregisterCalls <= 1)` will be entered a second time then.
* What if broadcast receiver does `numPostPushUnregisterCalls++`, and the background thread does its `numPostPushUnregisterCalls++`, before `postPushUnregister()` runs. In that case the `if (numPostPushUnregisterCalls <= 1)` will never be entered.

bhariharan: Yeah, it doesn't cover all cases. The foolproof way to do this to solve the single user and multi-user scenarios is to have a `boolean` per `UserAccount`. Only then would we be able to support this use case across multiple users, because resetting the `boolean` at the end is always going to be unpredictable (we don't know how long to wait). If we had 1 `boolean` per `UserAccount`, we could reset it when the user logs back in. Let me take a stab at that.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -844,16 +847,28 @@ public void run() {
...
+        /*
+         * This method can be triggered from 2 places - either when the push un-registration call
+         * completes, or when the server times out. Since they're on different threads, we don't
+         * know when either of those calls returns. However, the code should be executed only
+         * once. We also need to make sure that the value is reset once both calls come in,
+         * to ensure that push un-registration works in the multi-user scenario.
+         */
+        final Boolean postPushCalledBool = postPushCalled.get(account);
+        if (postPushCalledBool == null || !postPushCalledBool) {
+            postPushCalled.put(account, true);
```
bhariharan: @wmathurin This should work better. This ensures that this code is executed only once, regardless on when/where the call comes from. It also ensures that the multi-user scenario works, since the `boolean` is per `UserAccount`.

Songroid:
Can we get this PR merged?

bhariharan:
@Songroid Could you test this new fix?

PR #1922, opened 2019-06-03
Bringing back encryption key caching

wmathurin:
(getEncryptionKey is prohibitively slow without it

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -40,11 +40,11 @@ import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.SecureRandom;
-import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.crypto.KeyGenerator;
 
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -94,7 +97,12 @@ public static String getUniqueId(String name, int length) {
      * @return Encryption key.
      */
     public static String getEncryptionKey(String name) {
-        return generateEncryptionKey(name);
+        String encryptionKey = CACHED_ENCRYPTION_KEYS.get(name);
```
wmathurin: Bringing back the code we had in 7.0 and before

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -94,7 +97,12 @@ public static String getUniqueId(String name, int length) {
...
+        if (encryptionKey == null) {
+            encryptionKey = generateEncryptionKey(name);
```
wmathurin: NB: despite the name it's not actually generating a new encryption key every time.
It checks if we have a stored unique id and derive the key from it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
ktanna: Thanks for bringing this back. Should the map be a ConcurrentHashMap? I guess this was working pre 7.1, so it is ok.

PR #1926, opened 2019-06-10
Add whitelist urls

bbirman:


libs/SalesforceHybrid/res/xml/config.xml
```diff
@@ -12,11 +12,17 @@
     <allow-navigation href="https://localhost" />
     <allow-navigation href="https://*.force.com" />
     <allow-navigation href="https://*.salesforce.com" />
+    <allow-navigation href="https://*.visualforce.com" />
+    <allow-navigation href="https://*.documentforce.com" />
+    <allow-navigation href="https://*.salesforce-communities.com" />
```
bhariharan: What is `salesforce-communities.com`? I've never heard of it. Are you sure we own that?

wmathurin: See spring 18 release notes: https://releasenotes.docs.salesforce.com/en-us/spring18/release-notes/rn_vf_instance_names_removed.htm

PR #1929, opened 2019-06-14
Fixing setversion.sh

wmathurin:
Was not setting tag correctly in SalesforceReact's package.json


wmathurin:
NB: The package.json for SalesforceReact in master is not correct but it doesn't matter since nobody consumes RN for Android from the main repo.

PR #1936, opened 2019-06-21
Moving to SQLCipher 4.2.0

wmathurin:
All tests passing


wmathurin:
@bhariharan it’s a bigger story (because of migration). We have a story for assessing perf gain of moving to 4.x.

PR #1943, opened 2019-07-01
Adding pragma to upgrade db to sqlcipher 4.x

ktanna:


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -397,15 +397,17 @@ public static boolean smartStoreExists(Context ctx, String dbNamePrefix,
...
 			// => should open 2.x databases without any migration
 			database.execSQL("PRAGMA cipher_default_kdf_iter = 4000");
 		}
 
+		/**
+		 * Need to migrate for SqlCipher 4.x
+		 * @param database db being processed
+		 */
 		public void postKey(SQLiteDatabase database) {
+			database.rawExecSQL("PRAGMA cipher_migrate");
```
wmathurin: I wonder if that set the kdf_iter to 256000 (4.x default) - in the past we noticed that a high kdf_iter made opening the database noticeably slower. I'll do some testing.

wmathurin: I just did the work for [iOS](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2925) and checked the run time settings and upgrade scenarios. Everything looks fine.

wmathurin:
Merging now. Will send another PR to add the test I added on iOS.

PR #1954, opened 2019-07-25
Moving to React Native 0.60.4

wmathurin:


build.gradle
```diff
@@ -13,9 +13,15 @@ allprojects {
   group = 'com.salesforce.mobilesdk'
   version = '7.3.0'
   repositories {
+    mavenLocal()
```
bhariharan: You may not need `mavenLocal()` since you have explicit `maven` URLs declared below.

PR #1957, opened 2019-08-03
Fix SmartSync tests. 

brandonpage:


brandonpage:
Tests failed to API rate limiting.  How do we avoid this?

PR #1965, opened 2019-08-18
Fixing a corner case bug/regression with logout from the background

bhariharan:
In some corner cases, an app might want to log out from the background, i.e. with no `Activity` involved. Some examples are:
- from `Settings` screen.
- from a background `Service`.
- from a `Security` screen (where we don't want the screen to be dismissed).

We have always had a flag to support this, but it appears to have regressed at some point, rendering the flag useless. When this flag is set to `false`, we should not dismiss the `Activity` passed in, if any. This PR fixes this issue (and a few warnings).

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -716,8 +716,9 @@ public class SalesforceSDKManager {
      *
      * @param frontActivity Front activity.
      * @param account Account.
+     * @param shouldDismissActivity Dismisses current activity if true, does nothing otherwise.
      */
-    private void cleanUp(Activity frontActivity, Account account) {
+    private void cleanUp(Activity frontActivity, Account account, boolean shouldDismissActivity) {
         final UserAccount userAccount = UserAccountManager.getInstance().buildUserAccount(account);
 
         // Clean up in this process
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -116,9 +116,6 @@
      */
     private static final String CLEANUP_INTENT_ACTION = "com.salesforce.CLEANUP";
 
-    // Receiver for CLEANUP_INTENT_ACTION broadcast
-    private CleanupReceiver cleanupReceiver;
```
bhariharan: Can be a local variable - `Android Studio` warning.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -139,20 +136,19 @@
     private static final String DEFAULT_APP_DISPLAY_NAME = "Salesforce";
     private static final String INTERNAL_ENTROPY = "6cgs4f";
     private static final String TAG = "SalesforceSDKManager";
-    protected static String AILTN_APP_NAME;
```
bhariharan: Fixing scope for a few variables - they have setters.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -728,8 +725,8 @@ private void cleanUp(Activity frontActivity, Account account) {
 
         final List<UserAccount> users = getUserAccountManager().getAuthenticatedUsers();
 
-        // Finishes front activity if specified, and if this is the last account.
-        if (frontActivity != null && (users == null || users.size() <= 1)) {
+        // Finishes front activity if specified, if this is the last account.
+        if (shouldDismissActivity && frontActivity != null && (users == null || users.size() <= 1)) {
```
bhariharan: This is the fix - we should call `finish()` on the activity only if the flag is `true`, not always.

PR #1969, opened 2019-08-22
Adding API to fetch OpenID token

bhariharan:
This PR adds a new API to fetch OpenID token. Using this API requires an OpenID token to be configured on the connected app. It also needs the "openid" scope to be added on the client and the server.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -115,6 +115,7 @@ public class OAuth2 {
     private static final String CUSTOM_PERMISSIONS = "custom_permissions";
     private static final String SFDC_COMMUNITY_ID = "sfdc_community_id";
     private static final String SFDC_COMMUNITY_URL = "sfdc_community_url";
+    private static final String ID_TOKEN = "id_token";
     private static final String AND = "&";
     private static final String EQUAL = "=";
     private static final String QUESTION = "?";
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -441,6 +440,28 @@ private static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAcc
...
+     *
+     * @param loginServer Login server.
+     * @param clientId Client ID.
+     * @param refreshToken Refresh token.
+     * @return OpenID token.
+     */
+    public static String getOpenIDToken(String loginServer, String clientId, String refreshToken) {
+        String idToken = null;
+        try {
+            final TokenEndpointResponse tr = refreshAuthToken(HttpAccess.DEFAULT,
```
bhariharan: Currently, we have to use the `/token` endpoint to fetch OpenID token, but we might have a dedicated endpoint in the future. At such a time, we can swap out the implementation here to use that new endpoint, keeping the Mobile SDK API the same.

libs/test/SalesforceSDKTest/res/values/bootconfig.xml
```diff
@@ -6,6 +6,7 @@
     <string-array name="oauthScopes">
         <item>api</item>
         <item>web</item>
+        <item>openid</item>
```
bhariharan: Adding `openid` scope to the test project.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/OAuth2Test.java
```diff
@@ -291,6 +288,16 @@ public void testCallIdentityService() throws IOException, OAuthFailedException,
...
-        Assert.assertEquals("Wrong biometricUnlockAllowed returned", true, id.biometricUnlockAlowed);
+        Assert.assertTrue("Wrong biometricUnlockAllowed returned", id.biometricUnlockAlowed);
 	}
+
+    /**
+     * Testing getOpenIDToken.
+     */
+	@Test
+	public void testGetOpenIDToken() {
+        final String openIdToken = OAuth2.getOpenIDToken(TestCredentials.LOGIN_URL,
```
bhariharan: Test for new token. I have added `openid` scope to our test's connected app.

libs/test/SalesforceSDKTest/res/values/bootconfig.xml
brandonpage: @rwhitley I assume this will change documentation in a few places.

bhariharan: @brandonpage Not really, no. It doesn't change existing documentation. This scope is only required for apps that need `openid`. It will be a part of the new OpenID token documentation.

brandonpage:
👍 All tests passed.  Job fails for a known reason, will send a PR to fix that today.

bhariharan:
All tests passed locally and on `CircleCI`:
```
Your job ran 181 tests in unknown with 0 failures
Slowest test: com.salesforce.androidsdk.rest.RestClientTest testUpdateWithIfUnmodifiedSince (took 2.46 seconds).
```
Looks like the `Code Coverage` step is failing (unrelated to this PR).

PR #1981, opened 2019-09-17
Adding support for Android Q

bhariharan:
This PR:
- Bumps up `targetApiVersion` to `29`.
- Bumps up `minApiVersion` to `23`.
- Code refactoring to handle `TODO:` and `FIXME:` tags around conditional code to handle older versions prior to `23`.
- Removing methods that were marked `@deprecated` for `Mobile SDK 8.0` removal.


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/PasscodeActivity.java
```diff
@@ -373,11 +374,6 @@ protected boolean onSubmit(String enteredPasscode) {
                 }
             }
             return true;
-
-        case Change:
```
brandonpage: Good catch.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -261,18 +262,11 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
 		if (oldVersion == 1) {
 			SmartStore.createLongOperationsStatusTable(db);
 		}
-
```
wmathurin: If we are getting rid of this block, shouldn't we also get rid of the one above (line 261) ?

bhariharan: Yes - I could.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
```diff
@@ -227,33 +226,17 @@ private synchronized void createKeysIfNecessary(String algorithm, String name, i
...
-                         * Disabling StrongBox for now, since it's too slow on Pixel 3
-                         * and Pixel 3 XL.
-                         *
-                         * TODO: Re-enable in Mobile SDK 7.2.
-                         */
-                        //keyGenParameterSpecBuilder.setIsStrongBoxBacked(true);
-                        keyGenParameterSpecBuilder.setIsStrongBoxBacked(false);
-                        kpg.initialize(keyGenParameterSpecBuilder.build());
-                        kpg.generateKeyPair();
-                    } catch (StrongBoxUnavailableException sb) {
```
wmathurin: No need for the catch anymore?

bhariharan: Nope - we don't use `StrongBox` anymore based on Google's recommendation.

bhariharan:
Looking into test failures.

bhariharan:
All tests are passing now (was an unrelated token issue on `CircleCI`). Merging.

PR #1995, opened 2019-10-19
Replacing deprecated removeAccount() calls with removeAccountExplicitly

bhariharan:
We have a custom `AuthenticatorService` that we use to interact with `AccountManager`. As a result, we use `addAccountExplicitly()` to add accounts to `AccountManager`. However, for removal, we use `removeAccount()`, instead of `removeAccountExplicitly()`. This is because `removeAccountExplicitly()` was introduced in `API 22`, and until this upcoming release, our minimum was `API 21`. Now, our minimum is `API 23`, and we can switch to this new, better-suited API for our account removal. Switching to `removeAccountExplicitly()` is required, because it fixes a bug in our code when `DISALLOW_MODIFY_ACCOUNTS` is set by an MDM provider. Basically, when this flag is set, it allows only programmatic addition and removal of accounts, i.e. through `addAccountExplicitly()` and `removeAccountExplicitly()`. In our case, addition works, but removal fails, leaving the user stuck unable to log out of the app. This PR fixes that bug. See [this doc](https://developer.android.com/reference/android/os/UserManager.html#DISALLOW_MODIFY_ACCOUNTS) for more information.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -937,17 +937,12 @@ public class SalesforceSDKManager {
     					clientMgr.removeAccounts(accounts);
     				}
     				clientMgr.removeAccount(accounts[accounts.length - 1]);
-                    notifyLogoutComplete(showLoginPage);
-    			} else {
-    				notifyLogoutComplete(showLoginPage);
     			}
-    		} else {
-    			notifyLogoutComplete(showLoginPage);
     		}
     	} else {
     	    clientMgr.removeAccount(account);
-            notifyLogoutComplete(showLoginPage);
     	}
+        notifyLogoutComplete(showLoginPage);
     	isLoggingOut = false;
 
     	// Revokes the existing refresh token.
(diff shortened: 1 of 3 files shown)


bhariharan:
Tested using our sample apps - logout works fine, as expected.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -938,28 +936,17 @@ private void removeAccount(ClientManager clientMgr, final boolean showLoginPage,
...
-    				clientMgr.removeAccountAsync(accounts[accounts.length - 1],
-    						new AccountManagerCallback<Boolean>() {
-
-    	    			@Override
-    	    			public void run(AccountManagerFuture<Boolean> arg0) {
-    	    				notifyLogoutComplete(showLoginPage);
-    	    			}
-    	    		});
+    				clientMgr.removeAccount(accounts[accounts.length - 1]);
+                    notifyLogoutComplete(showLoginPage);
```
brandonpage: If we call this in the positive and ever single negative (4 times total) why not just put it after the conditional line ~951?

bhariharan: Good catch. Will do.

PR #1997, opened 2019-10-23
Fixing logout bug with multiple users

bhariharan:
We switched logout from async to sync. As a result, in the multi-user scenario, logging out from an account was causing issues. This is because we attempt to make a `peekRestClient()` call before `isLoggingOut` is reset indicating completion of logout. This fixes the sequence of resetting the flag before sending the intent that logout is complete.


bhariharan:
The change to switch logout from async to sync was made [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1995).

bhariharan:
Test failures are unrelated - API request limit exceeded.

PR #2009, opened 2019-11-13
Improve dark theme APIs.

brandonpage:
- Remove disable and force APIs in favor of just setting theme.
- Remove need to send an activity to check if dark theme should be shown.
- Add tests.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -167,7 +167,7 @@ public class SalesforceSDKManager {
     private Theme theme =  Theme.SYSTEM_DEFAULT;
 
     /**
-     * The theme to be used to style the SDK.
+     * Available Mobile SDK style themes.
      */
     public enum Theme {
         LIGHT,
@@ -1408,7 +1408,7 @@ public class SalesforceSDKManager {
     }
 
     /**
-     * Returns if dark theme should be displayed.  The value is retrieved from the OS, is no value is set.
+     * Indicates whether dark theme should be displayed.  The value is retrieved from the OS, if no value is set.
      * @see SalesforceSDKManager#setTheme
      *
      * @return             True if dark theme should be displayed, otherwise false.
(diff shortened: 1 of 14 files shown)


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTest.java
```diff
@@ -101,15 +102,45 @@ public void testDefaultAiltnAppName() {
...
         compareAiltnAppNames(getDefaultAppName());
     }
 
+    /**
+     * Test the default theme value.
+     */
+    @Test
+    public void testDefaultTheme() {
+        SalesforceSDKTestManager.init(getInstrumentation().getTargetContext(), MainActivity.class);
+        Assert.assertFalse(SalesforceSDKTestManager.getInstance().isDarkTheme());
```
wmathurin: If I run that test on a device that uses dark theme, it will fail.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1400,32 +1408,26 @@ private Object getBuildConfigValue(Context context, String fieldName) {
...
 
     /**
-     * Forces dark theme for all supported API levels
-     * @param forced     True if dark theme should be force enabled.
+     * Sets the theme for the SDK.  This value only persists as long as the instance of SalesforceSDKManager.
+     * @param theme     The theme to use.
      */
-    public synchronized void forceDarkTheme(boolean forced) {
-        forceDarkTheme = forced;
+    public synchronized void setTheme(Theme theme) {
```
brandonpage: This replaces `disableDarkTheme` and `forceDarkTheme`.  This allows the developer to present this option to the user and change it whenever they like.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTest.java
brandonpage: Good catch, the default for a new device might even change in the future.  I have updated the test.

PR #2011, opened 2019-11-20
Make custom server url input more lenient.

brandonpage:
-  `https://<url>` / `http://<url>` / `<url>` are now all acceptable inputs for custom server url and get added as proper https urls.  

-  Replaced the existing way of verifying label and url (using removed edit server functionality) with a check that the radio button with the proper label and url is displayed in the list.
-  Add tests for no protocol url and rest button.

### DIFF
--- a/native/NativeSampleApps/test/RestExplorerTest/src/com/salesforce/samples/restexplorer/ServerPickerActivityTest.java
+++ b/native/NativeSampleApps/test/RestExplorerTest/src/com/salesforce/samples/restexplorer/ServerPickerActivityTest.java
@@ -131,6 +131,25 @@ public class ServerPickerActivityTest {
     }
 
     /**
+     * Test an invalid valid URL is not entered or saved.
+     *
+     * @throws Throwable
+     */
+    @Test
+    public void testInvalidUrl() throws Throwable {
+        String label = "Invalid URL";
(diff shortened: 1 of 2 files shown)


native/NativeSampleApps/test/RestExplorerTest/src/com/salesforce/samples/restexplorer/ServerPickerActivityTest.java
```diff
@@ -113,39 +127,60 @@ public void testAddCustomInstance() throws Throwable {
...
-                activity.getCustomServerUrlEditor().getDialog().isShowing());
-        url = "https://valid.url.com";
+    public void testAddHttpUrl() throws Throwable {
+        String label = "My http URL";
+        String httpUrl = "http://invalid.url.com";
+        String httpsUrl = "https://invalid.url.com";
+        addCustomUrl(label, httpUrl);
+        checkServerAdded(label, httpsUrl);
+    }
+
```
wmathurin: Maybe one more test where you try to add a URL that has a protocol that is neither https nor http and you would check that nothing got added / that an error is shown?

brandonpage:
@bhariharan Just double checked, Raj fixed that issue some time ago on iOS.

PR #2028, opened 2020-02-04
Get communityId from UserAccount instance

jiwansvmx:
Even if the logged in user is a community user, default userstore created by the following
call
```java
SalesforceReactSDKManager.getInstance().setupUserStoreFromDefaultConfig();
```
has a filename format of 
```
smartstore_{orgID}_{userID}_internal.db
```
This diff fixes the above so the filenames have the following format 
```
smartstore_{orgID}_{userID}_{communityID}.db
```


jiwansvmx:
I can certainly use what you suggested but regarding

> If they had created a store before the upgrade, it won't be found after upgrade (because of the path change) and a new store will be created instead.

isn't that the case right now for community users anyway ? Doesn't it already not find the default user store db and create a new one ? At least that's what I am noticing.

bhariharan:
I agree with Wolf. This API is meant to point to the internal DB for that user. For a community user (for a community-scoped DB), please use the other API.

PR #2030, opened 2020-02-10
Fixing a crash in SmartSyncExplorer and some warnings

bhariharan:


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/ui/SmartStoreInspectorActivity.java
```diff
@@ -148,7 +148,7 @@ private void readExtras() {
 		isGlobal = bundle == null || !bundle.containsKey(IS_GLOBAL_STORE) || bundle.getBoolean(IS_GLOBAL_STORE) || !hasUser;
 		// dbName is set to DBOpenHelper.DEFAULT_DB_NAME
 		//   if no bundle, or no value for dbName in bundle
-		dbName = bundle == null || !bundle.containsKey(DB_NAME) ? DBOpenHelper.DEFAULT_DB_NAME : bundle.getString(DB_NAME);
+		dbName = bundle == null || !bundle.containsKey(DB_NAME) ? DBOpenHelper.DEFAULT_DB_NAME : bundle.getString(DB_NAME, DBOpenHelper.DEFAULT_DB_NAME);
```
bhariharan: This is the fix for the crash. The function to create the intent to launch this activity requires `dbName`, so it has to be set to `null`. In such cases, the value for `dbName` will be set to `null` and cause issues throughout this activity. This adds a default value if the explicitly assigned value in `null`.

PR #2035, opened 2020-02-19
Replacing Persistence Solution for Instrumentation Events with Paper

bhariharan:
- Currently, a single instrumentation event is stored in a single file. As a result, each write creates a new file, and in order to check if max limit of events has been reached, we look up the number of files created.
- As the number of events increases, this approach isn't the best. What we need is a simple key-value store that is fast.
- [Paper](https://github.com/pilgr/Paper) is an open source library that offers this functionality.
- I replaced the existing implementation using `Paper` instead.
- <b>Good news:</b> no public APIs were broken as part of this exercise, so we aren't breaking `semver`.
- Only the internals of reads and writes have been updated.

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
@@ -50,7 +50,7 @@ import io.paperdb.Paper;
  */
 public class EventStoreManager {
 
-    private static final String FILENAME = "Event_Store";
+    private static final String FILENAME = "event_store";
     private static final String TAG = "EventStoreManager";
 
     private Context context;
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
@@ -36,13 +36,11 @@ import com.salesforce.androidsdk.analytics.util.SalesforceAnalyticsLogger;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.FileReader;
-import java.io.FilenameFilter;
 import java.util.ArrayList;
 import java.util.List;
+
+import io.paperdb.Book;
+import io.paperdb.Paper;
 
 /**
  * Provides APIs to store events in an encrypted store on the filesystem.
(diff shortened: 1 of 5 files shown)


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
```diff
@@ -303,28 +238,4 @@ private String encrypt(String data) {
     private String decrypt(String data) {
         return Encryptor.decrypt(data, encryptionKey);
     }
-
-    /**
-     * This class acts as a filter to identify only the relevant event files.
-     *
-     * @author bhariharan
-     */
-    private static class EventFileFilter implements FilenameFilter {
```
bhariharan: We don't need this anymore - there's only one file.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
```diff
@@ -236,64 +224,11 @@ public int getMaxEvents() {
...
     private boolean shouldStoreEvent() {
-        final List<File> files = getAllFiles();
-        int fileCount = 0;
-        if (files != null) {
-            fileCount = files.size();
-        }
-        return (isLoggingEnabled && (fileCount < maxEvents));
-    }
-
-    private InstrumentationEvent fetchEvent(File file) {
```
bhariharan: We don't need this anymore - fetching an event is straightforward.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
```diff
@@ -52,13 +50,12 @@
  */
 public class EventStoreManager {
 
+    private static final String FILENAME = "Event_Store";
```
brandonpage: Is first character capital and underscores the filename paradigm?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
```diff
@@ -158,16 +156,9 @@ public InstrumentationEvent fetchEvent(String eventId) {
...
      */
-    public boolean deleteEvent(String eventId) {
-        if (TextUtils.isEmpty(eventId)) {
-            SalesforceAnalyticsLogger.e(context, TAG, "Invalid event ID supplied: " + eventId);
-            return false;
-        }
-        final String filename = eventId + filenameSuffix;
-        final File file = new File(rootDir, filename);
-        return file.delete();
+    public void deleteEvent(String eventId) {
```
brandonpage: SemVer?  You could `return !book.contains(eventId);` but I doubt anyone will ever care.  Probably not worth fixing.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
brandonpage: 🎉

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
bhariharan: Fixed in #2036

PR #2045, opened 2020-03-06
Enhancing LayoutSyncManager to support more fields and use cases

bhariharan:
- Details [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/3157).
- This should now support everything listed under `UI API` [here](https://developer.salesforce.com/docs/atlas.en-us.uiapi.meta/uiapi/ui_api_resources_record_layout.htm).
- Added new tests for all the new functionality/fields.

### DIFF
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
@@ -26,6 +26,8 @@
  */
 package com.salesforce.androidsdk.mobilesync.app;
 
+import com.salesforce.androidsdk.mobilesync.manager.LayoutSyncManager;
+import com.salesforce.androidsdk.mobilesync.util.MobileSyncLogger;
 import com.salesforce.androidsdk.smartstore.app.SmartStoreUpgradeManager;
 
 /**
@@ -39,6 +41,7 @@ public class MobileSyncUpgradeManager extends SmartStoreUpgradeManager {
      * Key in shared preference file for mobile sync version.
      */
     private static final String MOBILE_SYNC_KEY = "mobile_sync_version";
+    private static final String TAG = "MobileSyncUpgradeManager";
 
     private static MobileSyncUpgradeManager INSTANCE = null;
 
@@ -71,6 +74,17 @@ public class MobileSyncUpgradeManager extends SmartStoreUpgradeManager {
 
         // Update shared preference file to reflect the latest version.
         writeCurVersion(MOBILE_SYNC_KEY, MobileSyncSDKManager.SDK_VERSION);
+
+        // If the installed version < v8.2.0, we need to delete the old layout SmartStore file.
+        try {
+            final String majorVersionNum = installedVersion.substring(0, 3);
+            double installedVerDouble = Double.parseDouble(majorVersionNum);
+            if (installedVerDouble < 8.2) {
+                upgradeTo8Dot2();
+            }
+        } catch (Exception e) {
+            MobileSyncLogger.e(TAG, "Failed to parse installed version.");
+        }
     }
 
     /**
@@ -80,5 +94,13 @@ public class MobileSyncUpgradeManager extends SmartStoreUpgradeManager {
      */
     public String getInstalledSobjectVersion() {
         return getInstalledVersion(MOBILE_SYNC_KEY);
+    }
+
+    /*
+     * TODO: Remove this in Mobile SDK 9.0.
+     */
+    private void upgradeTo8Dot2() {
+        LayoutSyncManager.getInstance().getSmartStore().dropAllSoups();
+        LayoutSyncManager.reset();
     }
 }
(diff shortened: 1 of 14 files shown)


libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/LayoutSyncManager.java
```diff
@@ -59,7 +59,7 @@
 
     static final String SOUP_NAME = "sfdcLayouts";
     static final String QUERY = "SELECT {" + SOUP_NAME + ":_soup} FROM {" + SOUP_NAME +
```
wmathurin: What if the sync down target was created before 8.2? That query won't find it.

libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfigTest.java
```diff
@@ -177,19 +176,19 @@ public void testRefreshSyncDownFromConfig()  throws JSONException {
     }
 
     @Test
-    public void testLayoutSyncDownFromConfig()  throws JSONException {
+    public void testLayoutSyncDownFromConfig() throws JSONException {
```
wmathurin: We should probably update:
- the schema https://github.com/forcedotcom/SalesforceMobileSDK-Package/blob/master/shared/syncs.schema.json#L124
- the example syncs configs: https://github.com/forcedotcom/SalesforceMobileSDK-Package/blob/master/shared/example.usersyncs.json#L55 and https://github.com/forcedotcom/SalesforceMobileSDK-Package/blob/master/shared/example.usersyncs.json#L55 and https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/MobileSync/MobileSyncTestApp/usersyncs.json#L55

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/LayoutSyncManager.java
bhariharan: Yes, to avoid the upgrade step, I figured we'd pay the one-time cost of an extra sync. It'll fallback on server and hydrate the cache if someone switches from the old API to the new one.

libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfigTest.java
bhariharan: Will do. I had forgotten that we had added these.

bhariharan:
@wmathurin I've incorporated your feedback.
- [usersyncs.json has been updated](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2045/commits/49268b44ab245c15af0488ae0c038fac7dac2fea).
- [Upgrade steps have been added](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2045/commits/78a566df0cb47cbd7d14d8543abde1ed4b1031db).
I'll add the other PRs (for `iOS` and `Package`) after this gets merged.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
```diff
@@ -81,4 +95,12 @@ protected synchronized void upgradeSObject() {
     public String getInstalledSobjectVersion() {
         return getInstalledVersion(MOBILE_SYNC_KEY);
     }
+
+    /*
+     * TODO: Remove this in Mobile SDK 9.0.
+     */
+    private void upgradeTo8Dot2() {
+        LayoutSyncManager.getInstance().getSmartStore().dropAllSoups();
```
wmathurin: Why remove all the soups? Should you simply delete the soups that layout sync sank into?
Alternatively, could you write the query to work with the old keys and new keys for layouts?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestRequestTest.java
```diff
@@ -57,7 +57,10 @@
 	
 	private static final String TEST_API_VERSION = "v99.0";
 	private static final String TEST_OBJECT_TYPE = "testObjectType";
-	private static final String LAYOUT_TYPE_COMPACT = "Compact";
+	private static final String TEST_LAYOUT_TYPE_COMPACT = "Compact";
+	private static final String TEST_FORM_FACTOR_MEDIUM = "Medium";
+	private static final String TEST_MODE_EDIT = "Edit";
+	private static final String TEST_RECORD_TYPE_ID = "test_record_type_id";
     private static final String TEST_OTHER_OBJECT_TYPE = "testOtherObjectType";
```
brandonpage: bad indent

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
bhariharan: Ah yes, you're right. Copy-paste error from the tests ;-) I will tweak this to drop just the `layout` soup.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestRequestTest.java
bhariharan: Seems like some `Github` weirdness.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
bhariharan: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2045/commits/cf1f6224e34d2e313726b94e1452f9b07dd0cb9f).

PR #2046, opened 2020-03-17
Encrypting output stream and decrypting input stream classes

wmathurin:
New classes for the `SalesforceAnalytics` library
* `EncrypterOutputStream`: stream to write an encrypted file 
* `DecrypterInputStream`: stream to read an encrypted file 
* `WatchableStream`: interface for streams that can be watched (e.g. when close), useful for instance to time operations when the stream is being read by code outside the sdk

Also added test suite `EncrypterStreamTest`.
NB: Unfortunately encrypted files created the "old" way (using `Encryptor`) cannot be read using a `DecrypterInputStream` because Encryptor base 64 encodes the encrypted bytes. It would be nice to switch to the stream classes for all our encrypted files needs (e.g. external storage for `smartstore`) - for that we need migration code.

### DIFF
--- a/libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/security/EncrypterStreamTest.java
+++ b/libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/security/EncrypterStreamTest.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright (c) 2020-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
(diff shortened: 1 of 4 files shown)


libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/security/EncrypterStreamTest.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * Test that writes an encrypted file using EncrypterOutputStream and reads it back using
+     * FileInputStream and Encryptor.decrypt
+     *
+     * <p>Note: won't work because Encryptor.decrypt expect the encrypted content base 64 encoded
+     */
+    // @Test
+    public void testWriteThroughStreamAndReadWithEncryptor() {
```
wmathurin: Can't write with Encryptor and read with the new stream classes.
Keeping that test around until I add the code to facilitate the migration (maybe as part of this PR).

PR #2050, opened 2020-03-31
File based key value store

wmathurin:
Not done yet: cleanup key value stores when user logs out

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
@@ -126,7 +126,7 @@ public class SmartStoreSDKManager extends SalesforceSDKManager {
     protected void cleanUp(UserAccount userAccount) {
         if (userAccount != null) {
             // NB if database file was already deleted, we still need to call DBOpenHelper.deleteDatabase to clean up the DBOpenHelper cache
-            DBOpenHelper.deleteDatabase(getAppContext(), userAccount);
+            DBOpenHelper.deleteAllDatabases(getAppContext(), userAccount);
         } else {
             DBOpenHelper.deleteAllUserDatabases(getAppContext());
         }
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
@@ -465,17 +465,71 @@ public class SmartStoreSDKManager extends SalesforceSDKManager {
     }
 
     /**
+     * Return whether there is a key value store with given name for current user
+     */
+    public boolean hasKeyValueStore(String storeName) {
+        return hasKeyValueStore(storeName, getUserAccountManager().getCachedCurrentUser(), null);
+    }
+
+    /**
+     * Return whether there is a key value store with given name for given user
+     */
+    public boolean hasKeyValueStore(String storeName, UserAccount account) {
+        return hasKeyValueStore(storeName, account, null);
+
+    }
+
+    /**
+     * Return whether there is a key value store with given name for given user / community id
+     */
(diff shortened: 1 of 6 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
```diff
@@ -424,4 +422,39 @@ public void onSelected() {
         ));
         return devSupportInfos;
     }
+
+    /**
+     * Get key value store with given name for current user
+     * @param storeName store name
+     * @return a KeyValueEncryptedFileStore
+     */
+    public KeyValueEncryptedFileStore getKeyValueStore(String storeName) {
```
wmathurin, line 431: Convenience methods to get named key value stores for current user / some user / some user + community.
Do we need to have global key value stores also?

Not done: the methods to get all key value stores of a user / to clean them up.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -0,0 +1,195 @@
...
+    private final File storeDir;
+
+    /**
+     * Constructor
+     *
+     * @param parentDir parent directory for key value store
+     * @param storeName name for key value store
+     * @param encryptionKey encryption key for key value store
+     */
+    public KeyValueEncryptedFileStore(File parentDir, String storeName, String encryptionKey) {
```
wmathurin, line 57: If one calls that constructor directly, the key value store might not be properly managed by Mobile SDK (e.g. if the kv store is stored in some random location etc).
Maybe I should not keep that method public?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreSDKManager.java
bhariharan: If it's not too much effort, it would be nice to add the "global" option as well. Keeps the feature set at parity with everything else in `SmartStore`.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
bhariharan: Yeah, maybe best to not make it `public`, otherwise the user suffix may not be appended properly and the parent directory could be something else.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param key Unique identifier.
+     * @param value Value to be persisted.
+     * @return True - if successful, False - otherwise.
+     */
+    public boolean saveValue(String key, String value) {
+        if (!isKeyValid(key, "saveValue")) {
+            return false;
+        }
+        if (value == null) {
+            SmartStoreLogger.w(TAG, "saveValue: Invalid value supplied: " + key);
```
bhariharan, line 77: Shouldn't this be `value`, in this log line? Also, is empty string a valid `value`? If not, `TextUtils.isEmpty(value)` might be a better check.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/app/SmartStoreSDKManagerTest.java
```diff
@@ -0,0 +1,120 @@
...
+        SmartStoreSDKManager.initNative(context, null);
+        manager = SmartStoreSDKManager.getInstance();
+    }
+
+    /**
+     * Using getKeyValueStore / hasKeyValueStore / removeKeyValueStore / removeAllKeyValueStore
+     * with a single user and single store
+     */
+    @Test
+    public void testKeyValueStoreOperationsWithOneUserOneStore() throws JSONException {
```
wmathurin, line 75: More tests coming.

PR #2052, opened 2020-04-23
Upgrade to SHA-256 Keys from SHA-1 and from AES/CBC to AES/GCM Encryption

bhariharan:
What's in this PR:
- Uses `SHA-256` instead of `SHA-1` per security recommendations.
- Added support for `AES/GCM/NoPadding` while maintaining support for `AES/CBC/PKCS5Padding`.
- Initial wiring for upgrade steps.
- Adding APIs to fetch legacy `SHA-1` keys for apps to re-encrypt using new `SHA-256` keys.
- Verified that encryption and decryption are working as expected with newly installed apps, and that the cipher used is `AES/GCM/NoPadding`.
- Verified that the new key generated is `AES-256` as expected.
- Upgrade steps for `SalesforceAnalytics`, `SalesforceSDK`, `SmartStore` and `MobileSync` are complete. 

Testing:
- Automated tests are passing.
- Fresh installations of sample apps are running properly and using the new encryption key and new encryption and decryption modes successfully.
- Upgrade testing of `RestExplorer` was successful. Analytics data and account data were successfully migrated.
- Upgrade testing of `MobileSyncExplorer` was successful. `SmartStore` data was successfully migrated.


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -166,15 +195,38 @@ public synchronized static void upgradeTo7Dot1() {
         }
     }
 
+    /**
+     * Upgrades the keys stored in SharedPrefs to encrypted keys. This is a one-time
+     * migration step that's run while upgrading to Mobile SDK 8.2.
+     *
+     * @deprecated Will be removed in Mobile SDK 9.0.
+     */
+    public synchronized static void upgradeTo8Dot2() {
```
bhariharan, line 204: The idea here is to prefix existing keys with `legacy_` and generate new keys with the same supplied "key" (overloaded term) ;-)

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -166,15 +195,38 @@ public synchronized static void upgradeTo7Dot1() {
         }
     }
 
+    /**
+     * Upgrades the keys stored in SharedPrefs to encrypted keys. This is a one-time
+     * migration step that's run while upgrading to Mobile SDK 8.2.
+     *
+     * @deprecated Will be removed in Mobile SDK 9.0.
```
wmathurin, line 202: We should keep that method longer for apps that upgrade from major to major.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -110,6 +111,32 @@ public static String getEncryptionKey(String name) {
...
     }
 
+    /**
+     * Returns the legacy encryption key. This should be called only as a means to migrate to the new key.
+     *
+     * @param name Unique name associated with this legacy encryption key.
+     * @return Legacy encryption key.
+     * @deprecated Will be removed in Mobile SDK 9.0.
+     */
+    public static String getLegacyEncryptionKey(String name) {
```
wmathurin, line 121: Why do we need the name for? Doesn't Android generate a single key named with the value of INTERNAL_ENTROPY?

bhariharan: Yes - for internal use cases, but apps might have supplied some other value for `name`. If they need to upgrade, they'll need to retrieve the old key.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
bhariharan: Sure - I'll bump it up to `10.0`.

bhariharan:
Next up - rewriting `Encryptor` to use `AES/GCM/NoPadding` instead of `AES/CBC/PKCS5Padding` and then upgrade steps.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -92,53 +94,91 @@ public static Cipher getDecryptingCipher(String encryptionKey, byte[] iv)
     }
 
     private static Cipher getDecryptingCipher(byte[] keyBytes, byte[] iv)
-        throws InvalidAlgorithmParameterException, InvalidKeyException {
-        final Cipher cipher = getBestCipher();
+            throws InvalidAlgorithmParameterException, InvalidKeyException {
+        final Cipher cipher = getBestCipher(AES_GCM_CIPHER);
+        final SecretKeySpec skeySpec = new SecretKeySpec(keyBytes, cipher.getAlgorithm());
+        final IvParameterSpec ivSpec = new IvParameterSpec(iv);
+        cipher.init(Cipher.DECRYPT_MODE, skeySpec, ivSpec);
```
wmathurin, line 101: Don't you need a GCMParameterSpec ? See https://gist.github.com/praseodym/f2499b3e14d872fe5b4a

bhariharan: Yes, I haven't gotten there yet.

bhariharan: We need to use `SecretKeySpec` because we have raw bytes for the key. We don't need `GcmParameterSpec` as long as we update `AAD` in the `Cipher` directly.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -169,11 +211,11 @@ public static String decrypt(byte[] data, String key, byte[] iv) {
     }
 
     /**
-     * Encrypts data with key using AES-128.
+     * Encrypts data with key using AES/GCM/NoPadding.
      *
      * @param data Data.
      * @param key Base64 encoded 128 bit key or null (to leave data unchanged).
```
wmathurin, line 217: Should say Base64 encoded 256 bit key

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -185,12 +227,12 @@ public static String encrypt(String data, String key) {
...
```
wmathurin, line 233: Should say base64 encoded 256 bit key

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -52,93 +52,135 @@
...
+     */
+    public static String decrypt(byte[] data, String key) {
+        return decrypt(data, key, new byte[12]);
+    }
+
+    /**
+     * Decrypts data with key using AES/GCM/NoPadding.
+     *
+     * @param data Data.
+     * @param key Base64 encoded 128 bit key or null (to leave data unchanged).
```
wmathurin, line 171: Are we not going to 256 bit keys?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -52,93 +52,135 @@
...
+    public static String legacyDecrypt(String data, String key) {
         return decrypt(data, key, new byte[16]);
     }
 
     /**
-     * Decrypts data with key using AES-128.
+     * Decrypts data with key using AES/GCM/NoPadding.
      *
      * @param data Data.
      * @param key Base64 encoded 128 bit key or null (to leave data unchanged).
```
wmathurin, line 136: Should be 256 instead of 128 bit key.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -171,10 +199,10 @@ private synchronized static String generateEncryptionKey(String name) {
         try {
             final String keyString = getUniqueId(name);
             byte[] secretKey = keyString.getBytes(StandardCharsets.UTF_8);
-            final MessageDigest md = MessageDigest.getInstance(SHA1);
+            final MessageDigest md = MessageDigest.getInstance(SHA256);
             secretKey = md.digest(secretKey);
-            byte[] dest = new byte[16];
-            System.arraycopy(secretKey, 0, dest, 0, 16);
+            byte[] dest = new byte[32];
```
wmathurin, line 204: Why are we changing upgradeTo7Dot1, shouldn't the move from 128 to 256 happen in upgradeTo8Dot2?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -52,93 +52,135 @@
...
+     * Decrypts data with key using AES/CBC/PKCS5Padding. Should be used only for
+     * migration of encrypted data from an older version to Mobile SDK 8.2.
      *
      * @param data Data.
      * @param key Key.
      * @return Decrypted data.
+     * @deprecated Will be removed in Mobile SDK 10.0.
      */
-    public static String decrypt(byte[] data, String key) {
+    public static String legacyDecrypt(byte[] data, String key) {
```
wmathurin, line 152: Could you add a legacyEncrypt (could be package level)? That will make writing upgrade tests easier.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
bhariharan: Yes - I need to update all the comments.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/DecrypterInputStream.java
```diff
@@ -47,7 +47,7 @@ public DecrypterInputStream(FileInputStream inputStream, String encryptionKey)
             throws GeneralSecurityException, IOException {
         // First byte should be iv length
         int ivLength = inputStream.read();
-        if (ivLength != 16 && ivLength != 32) {
+        if (ivLength != 12 && ivLength != 32) {
```
bhariharan, line 50: IV length for GCM mode is 12.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -430,22 +472,33 @@ public static String decryptBytes(byte[] data, byte[] key, byte[] iv) {
...
         // Grabs the encrypted body after the init vector prefix.
         int meatLen = length - iv.length;
         int meatOffset = iv.length;
         byte[] meat = new byte[meatLen];
         System.arraycopy(data, meatOffset, meat, 0, meatLen);
-        final Cipher cipher = getDecryptingCipher(key, iv);
+        Cipher cipher;
+
+        // AES/GCM has an IV of length 12 bytes, whereas AES/CBC has an IV of length 16 bytes.
+        if (iv.length == 12) {
```
bhariharan, line 486: These conditions will go away in `Mobile SDK 10.0` when we remove the upgrade steps.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -430,22 +472,33 @@ public static String decryptBytes(byte[] data, byte[] key, byte[] iv) {
...
 
-    private static Cipher getBestCipher() {
+    private static Cipher getBestCipher(String cipherMode) {
         Cipher cipher = null;
         try {
-            cipher = Cipher.getInstance(PREFER_CIPHER_TRANSFORMATION, getLegacyEncryptionProvider());
+            if (AES_GCM_CIPHER.equals(cipherMode)) {
+                cipher = Cipher.getInstance(cipherMode);
+            } else {
+                cipher = Cipher.getInstance(cipherMode, getLegacyEncryptionProvider());
```
bhariharan, line 500: We can get rid of `BouncyCastle` then as well.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -383,6 +383,30 @@ private SalesforceAnalyticsManager(UserAccount account) {
...
 
+    /**
+     * One time upgrade steps from older versions to Mobile SDK 8.2. Only for internal use!
+     *
+     * @param account User account.
+     * @param context Context.
+     * @deprecated Will be removed in Mobile SDK 10.0.
+     */
+    public static synchronized void upgradeTo8Dot2(UserAccount account, Context context) {
+        final String filenameSuffix = (account != null) ? account.getCommunityLevelFilenameSuffix()
```
bhariharan, line 394: Migrates analytics/instrumentation data.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -127,4 +136,31 @@ protected String getInstalledVersion(String key) {
...
+     */
+    protected void upgradeTo8Dot2() {
+        ClientManager.upgradeTo8Dot2();
+        migrateAnalyticsData();
+    }
+
+    private void migrateAnalyticsData() {
+        final List<UserAccount> userAccounts = UserAccountManager.getInstance().getAuthenticatedUsers();
+
+        // Migrating an unauthenticated user's analytics data.
```
bhariharan, line 153: Handles unauthenticated user stores here as well.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -240,6 +241,88 @@ public RestClient peekRestClient(Account acc) {
         }
     }
 
+    /**
+     * One time upgrade steps from older versions to Mobile SDK 8.2. Only for internal use!
+     *
+     * @deprecated Will be removed in Mobile SDK 10.0.
+     */
+    public static synchronized void upgradeTo8Dot2() {
```
bhariharan, line 249: Migration steps for account data stored in `AccountManager`. The key here is to NOT call any `UserAccountManager` APIs or anything else that will call `encrypt` or `decrypt` behind the scenes.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreUpgradeManager.java
```diff
@@ -82,4 +89,21 @@ protected synchronized void upgradeSmartStore() {
...
+
+    @Override
+    protected void upgradeTo8Dot2() {
+        super.upgradeTo8Dot2();
+        final Map<String, DBOpenHelper> dbOpenHelperMap = DBOpenHelper.getOpenHelpers();
+        if (dbOpenHelperMap != null) {
+            final Collection<DBOpenHelper> dbOpenHelpers = dbOpenHelperMap.values();
+            for (final DBOpenHelper dbOpenHelper : dbOpenHelpers) {
+                if (dbOpenHelper != null) {
+                    final String oldEncryptionKey = SmartStoreSDKManager.getLegacyEncryptionKey();
```
bhariharan, line 101: Migration steps for all `SmartStore` data. Uses `PRAGMA REKEY` behind the scenes, available in `SQLCipher`.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -492,34 +490,35 @@ public boolean removeExternalBlobsDirectory(String soupTableName) {
...
-	 * Re-encrypts the files on external storage with the new key. If external storage is not enabled for any table in the db, this operation is ignored.
+	 * Re-encrypts the files on external storage with the new key. If external storage is not
+	 * enabled for any table in the db, this operation is ignored.
 	 *
-	 * @param db DB containing external storage (if applicable).
+	 * @param db DB containing external storage (if applicable).O
 	 * @param oldKey Old key with which to decrypt the existing data.
 	 * @param newKey New key with which to encrypt the existing data.
 	 */
 	public static void reEncryptAllFiles(SQLiteDatabase db, String oldKey, String newKey) {
```
wmathurin, line 500: What if this method is called to change key (both using the new encryption scheme)?
Instead of calling Encryptor.legacyDecrypt on line 519, shouldn't you decide based on the key length or something like that?

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreExternalStorageTest.java
```diff
@@ -91,28 +89,6 @@ public void testRegisterSoupWithExternalStorageAndJSON1() {
         Assert.assertFalse("Register soup call should have failed", store.hasSoup(OTHER_TEST_SOUP));
 	}
 
-	/**
-	 * Ensure data is still accessible after changing key
-	 */
-    @Test
-	public void testChangeKey() throws JSONException {
```
wmathurin, line 98: Why kill that test?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
bhariharan: Do we want to support that? I can add that in. However, it would be nice to keep this internal, so apps don't call it incorrectly - goes with the theme of keeping all encryption-related functionality internally handled by the SDK.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreExternalStorageTest.java
bhariharan: I didn't want to add public APIs for `legacyEncrypt` - it adds too many public APIs that apps might call and there's very little value in adding it because a simple manual test exposes whether it works or not - there's no halfway. Since there's no `legacyEncrypt`, I'd have to hardcode that in this test - not sure if there's enough value in it, but I can add it back if we want to.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
wmathurin: It's called from the public changeKey method so it should handle any key change (with and without encryption scheme change). If it's only for upgrade purposes, it should have a different name and be called from a upgrade only change key method.
If we can recognize the encryption change by inspecting the key we can have the method handle all cases.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
bhariharan: I don't think we need a `TODO:` here - I added a deprecation marker for `getLegacyDecryptingCipher`, so when we remove it in `10.0`, this will raise a compilation error and we will be forced to remove this then.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
bhariharan: Same comment as above.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
bhariharan: Makes sense. Let me look into it.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -412,7 +454,7 @@ public static String decryptBytes(byte[] data, byte[] key, byte[] iv) {
 
     private static byte[] generateInitVector() throws NoSuchAlgorithmException {
```
brandonpage, line 455: It appears this is only ever called once, when we first generate the encryption key.  Is that right?  Since GCM (unlike AES) has the two-time-pad/many-time-pad problem shouldn't we be generating a new IV every time?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -342,6 +342,33 @@ public static synchronized void deleteAllDatabases(Context ctx, UserAccount user
 		}
 	}
 
+	/**
+	 * One time upgrade steps from older versions to Mobile SDK 8.2. Only for internal use!
+	 *
+	 * @deprecated Will be removed in Mobile SDK 10.0.
+	 */
+	public static void upgradeTo8Dot2() {
```
bhariharan, line 350: It's important that this lives inside `DBOpenHelper` and be executed from within here - otherwise the call is too late, i.e. the call can't be after `getWritableDatabase()` since that will result is attempting to open it with the wrong key.

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/store/SmartStoreExternalStorageTest.java
bhariharan: @wmathurin Brought back the test below. No code changes - just moved down in the file.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -492,34 +519,55 @@ public boolean removeExternalBlobsDirectory(String soupTableName) {
...
 								}
 								br.close();
-								result = Encryptor.decrypt(json.toString(), oldKey);
+
+								/*
+								 * If the key length is 24, then it's the old key (16 bytes for the
+								 * key and 8 bytes for the IV. If the key length is 44, then it's the
+								 * new key (32 bytes for the key and 12 bytes for the IV).
+								 */
+								if (oldKey.getBytes().length == 24) {
```
bhariharan, line 564: @wmathurin Added a check here based on key size - `legacyDecrypt()` is applied to the old key size and `decrypt()` is applied to the new key size. Brought back the test as well.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
bhariharan: @brandonpage Actually, it looks like a new IV is generated every time we want to encrypt.

It's called from any call to `encrypt(data, key)`:
```
    /**
     * Encrypts data with key using AES/GCM/NoPadding.
     *
     * @param data Data.
     * @param key Base64 encoded 256 bit key or null (to leave data unchanged).
     * @return Encrypted data.
     */
    public static String encrypt(String data, String key) {
        try {
            return encrypt(data, key, generateInitVector());
        } catch (Exception ex) {
            SalesforceAnalyticsLogger.w(null, TAG, "Error during encryption", ex);
        }
        return null;
    }
```
And each call to it generates a new IV:
```
    private static byte[] generateInitVector() throws NoSuchAlgorithmException {
        final SecureRandom random = SecureRandom.getInstance(SHA1PRNG);
        byte[] iv = new byte[12];
        random.nextBytes(iv);
        return iv;
    }
```
It also gets called from `getEncryptingCipher(key)`, but that also generates a new IV every time:
```
    /**
     * Returns initialized cipher for encryption with an IV automatically generated.
     *
     * @param encryptionKey Encryption key.
     * @return Initialized cipher.
     */
    public static Cipher getEncryptingCipher(String encryptionKey)
            throws NoSuchAlgorithmException, InvalidAlgorithmParameterException, InvalidKeyException {
        final byte[] keyBytes = Base64.decode(encryptionKey, Base64.DEFAULT);
        return getEncryptingCipher(keyBytes, generateInitVector());
    }
```

brandonpage: Ah, I missed that `generateInitVector()` was called in called at the top level `encrypt` (since it wasn't changed in this PR).  👍

bhariharan:
@wmathurin This PR is ready for merging. I've tested all upgrade scenarios of native and hybrid apps, both with and without `SmartStore` - all the data is being migrated properly. I brought back the test, and amended the `changeKey` method to go by key length as well.

PR #2054, opened 2020-04-23
Adding a changeEncryptionKey method to key value store

wmathurin:
That will make key rotations easier.

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
@@ -38,17 +38,20 @@ import com.salesforce.androidsdk.util.ManagedFilesHelper;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.security.GeneralSecurityException;
 
 /** Key-value store backed by file system */
 public class KeyValueEncryptedFileStore  {
 
     private static final String TAG = KeyValueEncryptedFileStore.class.getSimpleName();
     public static final int MAX_STORE_NAME_LENGTH = 96;
-    private final String encryptionKey;
+    private String encryptionKey;
     private final File storeDir;
 
     public static final String KEY_VALUE_STORES = "keyvaluestores";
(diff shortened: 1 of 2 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -285,6 +279,47 @@ public String getStoreName() {
         return storeDir.getName();
     }
 
+    /**
+     * Change encryption key
+     * All files are read/decrypted with old key and encrypted/written back with new key
+     * @param newEncryptionKey
+     * @return true if successful
+     */
+    public boolean changeEncryptionKey(String newEncryptionKey) {
```
bhariharan, line 288: Should this be a public method? I don't know if we want this functionality to be a part of the public API. I would prefer handling this situation through upgrade managers in the library internally like we do for the other parts of the SDK. It allows for centralized handling and keeps the public API footprint clean.

wmathurin: The kv-store is not too strongly coupled with the SDK:
- for instance the public constructor will accept any directory (instead of computing it from the current user) 
- and is also given an encryption key (instead of doing SalesforceSDKManager.getEncryptionKey()).

If one wants a "managed" kv-store (one that goes way when the user logs out), then one should use the SmartStoreSDKManager factory methods for getting a user or global kv-store. Those will be created in the right location and will use the SalesforceSDKManager's encryption key.

Because the key is already exposed in the public API, I think it makes sense to have the key changing operation be public also.

bhariharan: Are some of those things something we should reconsider?
- Not scoping per user assumes that the caller will do the right thing. How does cleanup get handled when a user is logged out?
- Regarding encryption keys, it would be nice to standardize and use `SalesforceSDKManager.getEncryptionKey()`. This ensures that everything is held to a high security key standard as supplied by Mobile SDK. From a security standpoint, we're trying to get everyone to use the Mobile SDK's encryption key - otherwise we're finding that keys are weaker when left to the caller, like `FeedSDK`. Also, every other part of Mobile SDK that needs encryption is using an internal encryption key, so this would be a bit of an anti-pattern to the rest of the SDK. I'd prefer to line this up with the rest of `SmartStore` where the encryption key is internally-managed by the SDK.

wmathurin: Having user references / SalesforceSDKManager calls inside the class makes it harder to mock / test / or use in apps that are trying to abstract the SDK away.

PR #2055, opened 2020-04-23
Secure streams don't create their own cipher or initialization vector - they call Encryptor instead

wmathurin:


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/DecrypterInputStream.java
```diff
@@ -37,37 +36,23 @@
...
-    private static final String PREFER_CIPHER_TRANSFORMATION = "AES/CBC/PKCS5Padding";
-
     private InputStream cipherInputStream;
     private List<Watcher> watchers;
 
     public DecrypterInputStream(FileInputStream inputStream, String encryptionKey)
             throws GeneralSecurityException, IOException {
-
-        byte[] iv = new byte[16];
+        int ivLength = inputStream.read();
```
wmathurin, line 48: Writing iv length so that we can handle iv length changes in the future.
The other option would be to pass the expected length as one of the constructor parameter.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/DecrypterInputStream.java
```diff
@@ -37,37 +36,23 @@
...
+        byte[] iv = new byte[ivLength];
         inputStream.read(iv);
-        Cipher cipher = getCipher(encryptionKey, Cipher.DECRYPT_MODE, iv);
+        Cipher cipher = Encryptor.getDecryptingCipher(encryptionKey, iv);
```
wmathurin, line 51: We need a single place where cipher are being created -> Encryptor.java. Will make future changes easier.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/EncrypterOutputStream.java
```diff
@@ -42,28 +40,19 @@
...
     private List<Watcher> watchers;
 
     public EncrypterOutputStream(FileOutputStream outputStream, String encryptionKey)
             throws GeneralSecurityException, IOException {
-        final Cipher cipher =
-                DecrypterInputStream.getCipher(
-                        encryptionKey, Cipher.ENCRYPT_MODE, generateInitVector());
+        final Cipher cipher = Encryptor.getEncryptingCipher(encryptionKey);
+        final byte[] iv = cipher.getIV();
+        outputStream.write(iv.length);
```
wmathurin, line 50: Data in the file is prefixed by iv length followed by iv.

libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/security/EncryptorTest.java
```diff
@@ -139,6 +146,53 @@ public void testEncryptDecryptWithDifferentKeys() {
...
+	 * Check cipher returned by Encryptor.getEncryptingCipher
+	 */
+	@Test
+	public void testGetEncryptingCipher()
+		throws InvalidAlgorithmParameterException, NoSuchAlgorithmException, InvalidKeyException {
+    	Cipher cipher = Encryptor.getEncryptingCipher(makeKey("my-key"));
+    	Assert.assertEquals("Wrong algorithm", "AES/CBC/PKCS5Padding", cipher.getAlgorithm());
+		Assert.assertEquals("Wrong iv length", 16, cipher.getIV().length);
+		Assert.assertEquals("Wrong mode", 16, cipher.getBlockSize());
+	}
```
wmathurin, line 159: Can't find a way to see the mode the cipher was initialized with. But the third test below makes sure encryption/decryption work as expected.

wmathurin:
@bhariharan This PR should probably ahead of yours.
New secure streams no longer setup their own cipher or iv, instead they call new utility methods in Encryptor.java. Those methods will be the only ones to change when we change our encryption scheme.

PR #2056, opened 2020-04-24
Failing fast if first byte of file is not correct iv length

wmathurin:


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/DecrypterInputStream.java
```diff
@@ -45,7 +45,12 @@
 
     public DecrypterInputStream(FileInputStream inputStream, String encryptionKey)
             throws GeneralSecurityException, IOException {
+        // First byte should be iv length
         int ivLength = inputStream.read();
+        if (ivLength != 16 && ivLength != 32) {
```
wmathurin, line 50: Without that check it would fail later (probably in the getDecryptingCipher method) but the error could be hard to make sense of.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/DecrypterInputStream.java
```diff
@@ -45,7 +45,12 @@
...
+            throw new IOException("Can't decrypt file: incorrect iv length found in file: " + ivLength);
+        }
+        // Next bytes should be iv
         byte[] iv = new byte[ivLength];
```
wmathurin, line 54: ivLength is an int, but it's the result of reading a single byte. So it won't exceed 255. So this line will not cause out of memory exception. Still it's better to fail fast with a meaningful message.

PR #2057, opened 2020-04-24
Moving to sqlcipher 4.3 - all tests passing.

wmathurin:


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -128,7 +128,7 @@
     public static synchronized void changeKey(SQLiteDatabase db, String oldKey, String newKey) {
     	synchronized(db) {
 	        if (newKey != null && !newKey.trim().equals("")) {
-	            db.execSQL("PRAGMA rekey = '" + newKey + "'");
+	            db.query("PRAGMA rekey = '" + newKey + "'");
```
wmathurin, line 131: Getting an error when calling with db.execSQL (in sqlcipher 4.3.0, pragma returns ok).

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -1412,10 +1412,10 @@ public void delete(String soupName, Long[] soupEntryIds, boolean handleTx) {
 	            db.beginTransaction();
 	        }
 	        try {
-	            db.delete(soupTableName, getSoupEntryIdsPredicate(soupEntryIds), (String []) null);
+	            db.delete(soupTableName, getSoupEntryIdsPredicate(soupEntryIds), new String[0]);
```
wmathurin, line 1415: delete no longer likes null for the third arg (tries to do a .length on it without checking for null).

wmathurin:
No kdf iterations changes. We have tests checking the runtime settings.

PR #2059, opened 2020-04-30
Moving to react native 0.62.2

wmathurin:
Next stop: updating the templates.


wmathurin:
The failing test is passing consistently locally so I don't think the React Native upgrade caused a regression. I'll merge to be able to move forward with the work. We should however figure out why CI fails.

PR #2061, opened 2020-05-04
Rename onSuccess to onResponse, deprecated for removal in 9.0.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -351,8 +369,10 @@ public void onFailure(Call call, IOException e) {
 
 			@Override
 			public void onResponse(Call call, Response response) throws IOException {
+            	// TODO: Remove onSuccess in 9.0.
 			    callback.onSuccess(restRequest, new RestResponse(response));
```
wmathurin, line 373: If someone implements both, could that be an issue?
The response should not be consumed twice.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -120,8 +120,12 @@ protected void sendRequest(JSONArray args, final CallbackContext callbackContext
             }
             restClient.sendAsync(request, new RestClient.AsyncRequestCallback() {
 
+                // TODO: Remove onSuccess for 9.0 Release.
                 @Override
-                public void onSuccess(RestRequest request, RestResponse response) {
+                public void onSuccess(RestRequest request, RestResponse response) {}
```
bhariharan, line 125: There's no point "deprecating" this interface method (the customer still has to implement the new callback that we have introduced which makes it a breaking change). I'd make an exception and simply rename it to `onResponse`.

bhariharan:
Based on our offline discussion, should we close this PR, @brandonpage and @wmathurin? No one's really asking for the rename.

PR #2066, opened 2020-05-18
Add convenience methods for notification requests

bbirman:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestRequest.java
```diff
@@ -709,6 +714,103 @@ public static RestRequest getRequestForSObjectTree(String apiVersion, String obj
...
+     *                         Required if `seen` not provided.
+     * @param seen             Marks notifications as seen (true) or unseen (false). If null, field won't be updated.
+     *                         Required if `read` not provided.
+     */
+    public static RestRequest getRequestForNotificationsUpdate(String apiVersion, List<String> notificationIds, Date before, Boolean read, Boolean seen) {
+        final Map<String, Object> parameters = new HashMap<>();
+        if (notificationIds != null) {
+            parameters.put("notificationIds", notificationIds);
+        }
+        if (before != null) {
```
brandonpage, line 801: Should we check that they haven't supplied both `notificationIds` and `before`?  Or does the API handle that?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestRequest.java
```diff
@@ -168,6 +172,7 @@ public String getPath(Object... args) {
 		}
 	}
 
+	private static final DateFormat ISO8601_DATE_FORMAT = new SimpleDateFormat ("yyyy-MM-dd'T'HH:mm:ss.SSSZ", Locale.US);
```
wmathurin, line 175: Constant also define in MobileSync: https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/util/Constants.java#L82

Maybe we should have it in one place?

bhariharan: We could make this `public` and reuse it in `MobileSync` since that library depends on this one anyway.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestRequest.java
bhariharan: Nah - I think we rely on the doc for that (both ours and the server side doc). If the user supplies both, the endpoint will throw an error.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestRequest.java
bbirman: Would we deprecate the MobileSync one? If we have to work around semver I was also thinking about making another constants class in SalesforceSDK with the date that the existing MobileSync class could extend

bhariharan: Nah - that's too much work. Just update the constant in `MobileSync` to point to this one:
```
public static final DateFormat TIMESTAMP_FORMAT = RestRequest. ISO8601_DATE_FORMAT
```
That doesn't violate `semver`.

bbirman: Cool 👍

PR #2068, opened 2020-05-26
Allowed customization of the unauthenticated start page used by SalesforceDroidGapActivity

chrlie:
This PR allows subclasses of `SalesforceDroidGapActivity` to override the unauthenticated URL passed to Cordova at runtime. If this method is not overridden, the default value from `BootConfig` is still used for backwards compatibility.

This is the only usage of `unauthenticatedStartPage` in the SDK, so synchronization of the value with `BootConfig` itself shouldn't be needed.


chrlie:
@bhariharan it looks like an unrelated test has timed out. Can the checks be re-run?

bhariharan:
@chrlie I kicked off another test run. I'll get this merged after that test run.

PR #2083, opened 2020-06-26
Move to v49.0

wmathurin:


bbirman:
I think there's one more in strings.xml and that's why the notification tests are failing

PR #2084, opened 2020-06-29
Upgrade Steps for Analytics Data in 8.2

bhariharan:
Follow up to [this PR](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2052/). The upgrade steps in that PR would only apply to `SApp` (because of cherry-picking). It would not apply to any other `Mobile SDK` client because I switched the persistent store from single file per event to using `Paper` in the `8.2` timeframe. So, these upgrade steps take events from the old files and move them over to `Paper`, and delete the old files. Tested and verified that it works with our sample apps.

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
@@ -238,9 +238,6 @@ public class EventStoreManager {
     }
 
     private String decrypt(String data) {
-        if (encryptionKey.getBytes().length == 24) {
-            return Encryptor.legacyDecrypt(data, encryptionKey);
-        }
         return Encryptor.decrypt(data, encryptionKey);
     }
 }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
@@ -35,6 +35,7 @@ import com.salesforce.androidsdk.accounts.UserAccount;
 import com.salesforce.androidsdk.analytics.manager.AnalyticsManager;
 import com.salesforce.androidsdk.analytics.model.DeviceAppAttributes;
 import com.salesforce.androidsdk.analytics.model.InstrumentationEvent;
+import com.salesforce.androidsdk.analytics.security.Encryptor;
 import com.salesforce.androidsdk.analytics.store.EventStoreManager;
 import com.salesforce.androidsdk.analytics.transform.AILTNTransform;
 import com.salesforce.androidsdk.analytics.transform.Transform;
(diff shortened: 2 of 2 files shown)


libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
```diff
@@ -238,9 +238,6 @@ private String encrypt(String data) {
     }
 
     private String decrypt(String data) {
-        if (encryptionKey.getBytes().length == 24) {
```
bhariharan, line 241: We don't need this anymore because there is no `PaperDB` prior to `8.2`, so this will never get called with the old code.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -477,4 +455,95 @@ public void run() {
...
+        final String filenameSuffix = (account != null) ? account.getCommunityLevelFilenameSuffix()
+                : UNAUTH_INSTANCE_KEY;
+        final File rootDir = context.getFilesDir();
+        final List<File> oldEventFiles = getAllEventFiles(rootDir, filenameSuffix);
+        final List<InstrumentationEvent> oldEvents = new ArrayList<>();
+        for (final File file : oldEventFiles) {
+            final InstrumentationEvent event = fetchEvent(file);
+            if (event != null) {
+                oldEvents.add(event);
+                file.delete();
```
bhariharan, line 476: Deletes the old file after extracting the event from it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -477,4 +455,95 @@ public void run() {
...
+                SalesforceSDKLogger.e(TAG, "Exception occurred while attempting to convert to JSON", e);
+            }
+        }
+        return event;
+    }
+
+    /*
+     * TODO: Exists only for upgrade steps to 8.2. Remove this in Mobile SDK 10.0.
+     */
+    private static List<File> getAllEventFiles(File rootDir, String fileSuffix) {
```
bhariharan, line 519: File filter to get the old files. Will be removed when the upgrade steps are removed in `10.0`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -477,4 +455,95 @@ public void run() {
...
+            }
+        }
+        final SalesforceAnalyticsManager sfAnalyticsManager = new SalesforceAnalyticsManager(account);
+        sfAnalyticsManager.getEventStoreManager().storeEvents(oldEvents);
+    }
+
+    /*
+     * TODO: Exists only for upgrade steps to 8.2. Remove this in Mobile SDK 10.0.
+     */
+    private static InstrumentationEvent fetchEvent(File file) {
```
bhariharan, line 486: Fetching the event from the old file. Will be removed in `10.0` when the upgrade steps are removed.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -477,4 +455,95 @@ public void run() {
...
+        String eventString = null;
+        final StringBuilder json = new StringBuilder();
+        try {
+            final BufferedReader br = new BufferedReader(new FileReader(file));
+            String line;
+            while ((line = br.readLine()) != null) {
+                json.append(line).append('\n');
+            }
+            br.close();
+            eventString = Encryptor.legacyDecrypt(json.toString(),
```
bhariharan, line 500: Uses `legacyDecrypt` and the old encryption key to read the event.

PR #2090, opened 2020-07-14
Fixing flappers in EventStoreManagerTest

wmathurin:


libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/store/EventStoreManagerTest.java
```diff
@@ -193,22 +193,28 @@ public void testDeleteMultipleEvents() throws Exception {
...
         genEvents.add(event1);
         genEvents.add(event2);
+        genEvents.add(event3);
         storeManager.storeEvents(genEvents);
         final List<InstrumentationEvent> eventsBeforeDel = storeManager.fetchAllEvents();
         Assert.assertNotNull("List of events stored should not be null", eventsBeforeDel);
-        Assert.assertEquals("Number of events stored should be 2", 2, eventsBeforeDel.size());
-        Assert.assertTrue("Stored event should be the same as generated event", event1.equals(eventsBeforeDel.get(0)));
-        Assert.assertTrue("Stored event should be the same as generated event", event2.equals(eventsBeforeDel.get(1)));
+        Assert.assertEquals("Number of events stored should be 3", 3, eventsBeforeDel.size());
```
wmathurin, line 206: Not assuming that fetchAllEvents() which calls book.getAllKeys() which relies on bookFolder.list(new FilenameFilter() {...}) returns events in the order they were inserted.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -1039,7 +1039,7 @@ public void testUpdateSeenNotifications() throws Exception {
      */
     private RestResponse getStreamTestResponse() throws IOException {
         final RestResponse response = restClient.sendSync(RestRequest.getRequestForResources(TestCredentials.API_VERSION));
-        Assert.assertEquals("Response code should be HTTP OK", response.getStatusCode(), HttpURLConnection.HTTP_OK);
+        Assert.assertEquals("Response code should be HTTP OK", HttpURLConnection.HTTP_OK, response.getStatusCode());
```
wmathurin, line 1042: Unrelated change Assert.assertEquals expects message, expected, actual not message, actual, expected.

libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/store/EventStoreManagerTest.java
bhariharan: Interesting! So, the previous implementation of flat files guaranteed order because each event was in its own file, if I'm not mistaken, but `PaperDB` doesn't, which makes sense.

PR #2092, opened 2020-07-14
Better handling of bad or missing sobject type during sync up

wmathurin:
Fixes the infinite loop reported here: forcedotcom/SalesforceMobileSDK-ReactNative#168

### DIFF
--- a/libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/target/SyncUpTargetTest.java
+++ b/libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/target/SyncUpTargetTest.java
@@ -198,6 +198,86 @@ public class SyncUpTargetTest extends SyncManagerTestCase {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/target/SyncUpTargetTest.java
```diff
@@ -197,6 +197,86 @@ else if (name.equals("")) {
...
     }
 
+    /**
+     * Create a few records - some with no sobject type
+     * Sync up
+     * Make sure the records with no sobject type fail to sync and that last error reflects the problem
+     * @throws Exception
+     */
+    @Test
+    public void testSyncUpWithNoType() throws Exception {
```
wmathurin, line 207: New tests:
- no type
- bad typ
Those tests run for both SyncUpTarget and BatchSyncUpTarget.
Before the code change, the test would repro the infinite looping.

PR #2095, opened 2020-07-15
Rest error dictionary

bbirman:
For discussion here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS-Hybrid/pull/138#discussion_r453828871

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -125,12 +125,12 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
                     try {
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            JSONObject responseObject = new JSONObject();
-                            responseObject.put("allHeaders", new JSONObject(response.getAllHeaders()));
+                            final JSONObject responseObject = new JSONObject();
+                            responseObject.put("headers", new JSONObject(response.getAllHeaders()));
                             responseObject.put("statusCode", response.getStatusCode());
                             responseObject.put("body", response.asString());
 
-                            JSONObject errorObject = new JSONObject();
+                            final JSONObject errorObject = new JSONObject();
                             errorObject.put("response", responseObject);
                             callbackContext.error(errorObject.toString());
                         }
@@ -172,7 +172,7 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
 
                 @Override
                 public void onError(Exception exception) {
-                    JSONObject errorObject = new JSONObject();
+                    final JSONObject errorObject = new JSONObject();
                     try {
                         errorObject.put("error", exception.getMessage());
                     } catch (JSONException jsonException) {
@@ -182,7 +182,7 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
                 }
             });
         } catch (Exception exception) {
-            JSONObject errorObject = new JSONObject();
+            final JSONObject errorObject = new JSONObject();
             try {
                 errorObject.put("error", exception.getMessage());
             } catch (JSONException jsonException) {
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -125,9 +125,14 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
                     try {
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            JSONObject errorObj = new JSONObject();
-                            errorObj.putOpt("response", response.fullResponseAsJSONObject());
-                            callbackContext.error(errorObj.toString());
+                            JSONObject responseObject = new JSONObject();
+                            responseObject.put("allHeaders", new JSONObject(response.getAllHeaders()));
+                            responseObject.put("statusCode", response.getStatusCode());
+                            responseObject.put("body", response.asString());
+
+                            JSONObject errorObject = new JSONObject();
+                            errorObject.put("response", responseObject);
+                            callbackContext.error(errorObject.toString());
                         }
                         // Binary response
                         else if (returnBinary) {
@@ -167,13 +172,23 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
 
                 @Override
                 public void onError(Exception exception) {
-                    String jsonString = "{error: \"" + exception.getMessage() + "\"}";
-                    callbackContext.error(jsonString);
+                    JSONObject errorObject = new JSONObject();
+                    try {
+                        errorObject.put("error", exception.getMessage());
+                    } catch (JSONException jsonException) {
+                        SalesforceHybridLogger.e(TAG, "Error creating error object", jsonException);
+                    }
+                    callbackContext.error(errorObject.toString());
                 }
             });
         } catch (Exception exception) {
-            String jsonString = "{error: \"" + exception.getMessage() + "\"}";
-            callbackContext.error(jsonString);
+            JSONObject errorObject = new JSONObject();
+            try {
+                errorObject.put("error", exception.getMessage());
+            } catch (JSONException jsonException) {
+                SalesforceHybridLogger.e(TAG, "Error creating error object", jsonException);
+            }
+            callbackContext.error(errorObject.toString());
         }
     }
 
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -125,7 +125,9 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
                     try {
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            callbackContext.error(response.asString());
+                            JSONObject errorObj = new JSONObject();
+                            errorObj.putOpt("response", response.fullResponseAsJSONObject());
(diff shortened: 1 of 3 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -165,11 +167,13 @@ else if (response.asBytes().length > 0) {
 
                 @Override
                 public void onError(Exception exception) {
-                    callbackContext.error(exception.getMessage());
+                    String jsonString = "{error: \"" + exception.getMessage() + "\"}";
```
bbirman, line 170: Constructing the string without JSONObject because this is called when json parsing fails and avoiding another try/catch

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
```diff
@@ -189,6 +190,24 @@ public JSONObject asJSONObject() throws JSONException, IOException {
...
 	}
 
+	/**
+	 * JSONObject including the response body, status code and headers.
+	 *
+	 * @return JSONObject for full response
+	 * @throws JSONException
+	 * @throws IOException
+	 */
+	public JSONObject fullResponseAsJSONObject() throws JSONException, IOException {
```
bhariharan, line 200: This extra method is a bit weird. The method `asJsonObject()` is meant to be a JSON representation of this object. I'd rather we use that method. The receiver can extract all the fields from that or parse it however they want.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -125,7 +125,9 @@ public void onSuccess(RestRequest request, RestResponse response) {
                     try {
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            callbackContext.error(response.asString());
+                            JSONObject errorObj = new JSONObject();
+                            errorObj.putOpt("response", response.fullResponseAsJSONObject());
```
bhariharan, line 129: Use `response.asJsonObject()` instead. Also, use `put()` instead of `putOpt()`, `null` as a value is fine.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/SalesforceNetReactBridge.java
```diff
@@ -109,7 +109,9 @@ public void onSuccess(RestRequest request, RestResponse response) {
 
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            errorCallback.invoke(response.asString());
+                            JSONObject errorObj = new JSONObject();
```
bhariharan, line 112: Same comment as above.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
bhariharan: Could we revert changes to this file and go with `asJsonObject()` instead?

bbirman: `asJsonObject` just has the response body but not status code & headers. Do you want me to move this part to a helper method in the bridge and plugin classes?

bhariharan: Do we need `status code` and `headers` in the error callback? If we really need it, then a helper method in the bridge and plugin classes would be less confusing IMO.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
wmathurin: If the exception message has ", couldn't we end up with an invalid json string?
It would be safer to do 
```
JSONObject errorJson = new JSONObject(); 
errorJson.put("error", exception.getMessage())
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
wmathurin: The code and headers will probably have more value in the case of failed requests. So this method will be more valuable than the existing one.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
```diff
@@ -189,6 +190,24 @@ public JSONObject asJSONObject() throws JSONException, IOException {
...
+		if (fullResponseAsJSONObject == null) {
+			JSONObject jsonObject = new JSONObject();
+			jsonObject.put("headers", new JSONObject(getAllHeaders()));
```
wmathurin, line 203: 👍 
Could we have iOS and Android full json response look similar? It will help hybrid and react native apps (they won't need platform specific parsing code).

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
bbirman: That's what I started with but since put can throw an exception, what do you think should happen if it fails?
```
JSONObject errorJson = new JSONObject();
try {
    errorJson.put("error", e.getMessage());
} catch (JSONException ex) {
                            
}
```

wmathurin: Just log it - I don't think it can actually happen since we are putting a string.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
bbirman: I'm thinking standardize on this. iOS also has some extra fields on NSURLResponse like `suggestedFilename` which I'm thinking we can still add if someone wants to go deeper on the iOS side but the shared things would match
```
{
     "response": {
         "body": XXX,
         "statusCode": XXX,
         "allHeaders":XXX,
         // Other derived NSURLResponse fields
      },
     "error": XXX 
}
```

bhariharan: Yep, that format LGTM.

bbirman: Once this is set, I'll update the iOS side

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -125,7 +125,14 @@ public void onSuccess(RestRequest request, RestResponse response) {
                     try {
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            callbackContext.error(response.asString());
+                            JSONObject responseObject = new JSONObject();
+                            responseObject.put("allHeaders", new JSONObject(response.getAllHeaders()));
```
bhariharan, line 129: Could we use the key `"headers"` instead?

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -125,7 +125,14 @@ public void onSuccess(RestRequest request, RestResponse response) {
...
+                            responseObject.put("statusCode", response.getStatusCode());
+                            responseObject.put("body", response.asString());
```
bhariharan, line 131: Shouldn't this value should be `response.asJsonObject()`?

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/bridge/SalesforceNetReactBridge.java
```diff
@@ -109,7 +110,14 @@ public void onSuccess(RestRequest request, RestResponse response) {
 
                         // Not a 2xx status
                         if (!response.isSuccess()) {
-                            errorCallback.invoke(response.asString());
+                            JSONObject responseObject = new JSONObject();
+                            responseObject.put("allHeaders", new JSONObject(response.getAllHeaders()));
```
bhariharan, line 114: Same comments as above.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
wmathurin: Body might not be json - so giving the raw body (string) is probably safer.

bhariharan: Okay, fair enough. We can leave this as-is.

bbirman: I forgot that on iOS we check to see if it's a JSON object or array and then default to a string if not, I'll update this to match

bhariharan:
Looks like `ForceJSTestSuite.testParseUrl` is failing after the submodule update ([https://app.circleci.com/pipelines/github/forcedotcom/SalesforceMobileSDK-Android/401/workflows/d413af70-dc1a-443e-be76-2711b2a4dc0f/jobs/10808/tests](full test report)). Could this be related to [this change](https://github.com/forcedotcom/SalesforceMobileSDK-Shared/pull/523/), @wmathurin?

wmathurin:
@bhariharan Probably - I'll take a look.

wmathurin:
@bhariharan That should fix the tests (https://github.com/forcedotcom/SalesforceMobileSDK-Shared/pull/524).

PR #2106, opened 2020-08-04
Adding new API for an easy way to refresh and persist auth tokens

bhariharan:
Currently, there's an easy way to refresh tokens from the server, but no single API that does the refresh and handles persisting the new token as well. This adds a convenient way of refreshing and persisting the response. This is useful for apps like `Trailhead GO` that make their own requests using `OkHttpClient`, but need an easy way to refresh tokens in the SDK.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -40,6 +40,7 @@ import com.salesforce.androidsdk.app.Features;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.auth.AuthenticatorService;
 import com.salesforce.androidsdk.rest.ClientManager;
+import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -58,6 +59,7 @@ public class UserAccountManager {
 	private static final String CURRENT_USER_PREF = "current_user_info";
 	private static final String USER_ID_KEY = "user_id";
 	private static final String ORG_ID_KEY = "org_id";
+	private static final String TAG = "UserAccountManager";
 
 	public static final String USER_SWITCH_INTENT_ACTION = "com.salesforce.USERSWITCHED";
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -349,15 +351,6 @@ public void switchToNewUser(String jwt, String url) {
         switchToNewUserWithOptions(options);
 	}
 
-	private void switchToNewUserWithOptions(Bundle options) {
```
bhariharan, line 352: Just moved this to the bottom of this file.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -554,4 +547,36 @@ public UserAccount getUserFromOrgAndUserId(String orgId, String userId) {
...
+
+	/**
+	 * Attempts to refresh the access token for this user by making an API call
+	 * to the "/token" endpoint. If the call succeeds, the new token is persisted.
+	 * If the call fails and the refresh token is no longer valid, the user is logged out.
+	 * This should NOT be called from the main thread because it makes a network request.
+	 *
+	 * @param userAccount User account whose token should be refreshed. Use 'null' for current user.
+	 */
+	public synchronized void refreshToken(UserAccount userAccount) {
```
bhariharan, line 559: A simple new convenience method.

wmathurin: Do we need/have the same on iOS?

bhariharan: There is a functionally equivalent API on iOS, but it doesn't live in SFUserAccountManager on iOS. It lives in SFSDKAuthHelper, if I'm not mistaken. Should we add one on SFUserAccountManager on iOS as well (we could simply call the existing one)? Or leave it as is?

qingqingliu: Thanks @bhariharan !

ahetawal-salesforce: Hey @bhariharan is this a synchronous network call, or is there a call back which client needs to listen to after token refresh is successful. 
Should this method return the new token or the client still needs to call the getAuthToken after this call synchronously ?

bhariharan: Yes, it is a synchronous network call. I have mentioned that in the `Javadoc` for this method above. After this, you need to read the new token by updating your instance of `UserAccount` and calling `UserAccount.getAuthToken()`.

PR #2115, opened 2020-08-14
Fix keyboard not showing on passcode screen after biometric cancel.

brandonpage:


brandonpage:
@bhariharan Upon further inspection I was wrong.  This fixes the issue with no side effects.  The code we talked about earlier will be removed with other work planned for 9.0.

bhariharan:
Okay, sounds good. If it's not breaking `semver` and is internal, we could remove it now. If not, maybe we should add a deprecation marker for it.

PR #2122, opened 2020-09-22
Replacing stream encryption/decryption with in-memory solution

bhariharan:
Turns out `CipherInputStream` is a lot slower with `AES-GCM`. While this solution is not particularly good from a memory standpoint, we will need to employ this as a workaround for now, until we figure out how we can achieve acceptable performance with a streaming solution.

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
@@ -211,6 +211,30 @@ public class Encryptor {
     }
 
     /**
+     * Decrypts data with key using using AES/GCM/NoPadding. The data is not Base64 encoded.
+     *
+     * @param data Data.
+     * @param key Key.
+     * @return Decrypted data.
+     */
+    public static byte[] decryptWithoutBase64Encoding(byte[] data, String key) {
+        if (TextUtils.isEmpty(key)) {
+            return data;
+        }
+        try {
+
+            // Decodes with Base64.
+            byte[] keyBytes = Base64.decode(key, Base64.DEFAULT);
+
+            // Decrypts with AES.
+            return decrypt(data, data.length, keyBytes, new byte[12]);
+        } catch (Exception ex) {
+            SalesforceAnalyticsLogger.w(null, TAG, "Error during decryption", ex);
+        }
+        return null;
+    }
+
+    /**
      * Encrypts data with key using AES/GCM/NoPadding.
      *
      * @param data Data.
@@ -290,6 +314,28 @@ public class Encryptor {
             byte[] keyBytes = Base64.decode(key, Base64.DEFAULT);
             byte[] dataBytes = data.getBytes(StandardCharsets.UTF_8);
             return Base64.encode(encrypt(dataBytes, keyBytes, iv), Base64.DEFAULT);
+        } catch (Exception ex) {
+            SalesforceAnalyticsLogger.w(null, TAG, "Error during encryption", ex);
+        }
+        return null;
+    }
+
+    /**
+     * Encrypts data with key using AES/GCM/NoPadding. The data is not Base64 encoded.
+     *
+     * @param data Data.
+     * @param key Key.
+     * @return Encrypted data.
+     */
+    public static byte[] encryptWithoutBase64Encoding(byte[] data, String key) {
+        if (TextUtils.isEmpty(key)) {
+            return data;
+        }
+        try {
+
+            // Encrypts with our preferred cipher.
+            byte[] keyBytes = Base64.decode(key, Base64.DEFAULT);
+            return encrypt(data, keyBytes, generateInitVector());
         } catch (Exception ex) {
             SalesforceAnalyticsLogger.w(null, TAG, "Error during encryption", ex);
         }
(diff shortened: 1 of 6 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -344,21 +338,34 @@ private boolean isKeyValid(String key, String operation) {
         return files == null ? new File[0] : files;
     }
 
-    InputStream getStream(File file, String encryptionKey) throws IOException, GeneralSecurityException {
-        FileInputStream f = new FileInputStream(file);
-        DecrypterInputStream inputStream = new DecrypterInputStream(f, encryptionKey);
-        return inputStream;
+    InputStream getStream(File file, String encryptionKey) throws IOException {
```
wmathurin, line 341: You won't be able to read back values written with kv-store from 8.2 because Encryptor.encrypt/decrypt does base 64 encoding/decoding (which is a waste if you don't need a string back). See [encryptBytes](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java#L290) and [decryptBytes](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java#L200). That's the reasons why the tests [testWriteWithEncryptorAndReadThroughStream and testWriteThroughStreamAndReadWithEncryptor](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/security/EncrypterStreamTest.java#L84) are commented out.
Maybe we could have encrypt and decrypt methods that take bytes / return bytes and don't base 64 encode or decode.

wmathurin (changes requested):

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
bhariharan: Right - I deliberately left out the upgrade case, I don't think this is being used currently. I will look into adding the vanilla `encryptBytes` and `decryptBytes` methods though.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -210,6 +210,30 @@ public static String decrypt(byte[] data, String key, byte[] iv) {
...
     }
 
+    /**
+     * Decrypts data with key using using AES/GCM/NoPadding. The data is not Base64 encoded.
+     *
+     * @param data Data.
+     * @param key Key.
+     * @return Decrypted data.
+     */
+    public static byte[] decryptWithoutBase64Encoding(byte[] data, String key) {
```
bhariharan, line 220: Added new methods for encryption and decryption that takes `byte[]` and returns `byte[]` without `Base64` encoding and decoding.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -138,15 +137,19 @@ public boolean saveValue(String key, String value) {
...
         }
-
-        long startNanoTime = System.nanoTime();
-        try (FileOutputStream f = new FileOutputStream(getFileForKey(key));
-                EncrypterOutputStream outputStream = new EncrypterOutputStream(f, encryptionKey)) {
-            outputStream.write(value.getBytes(StandardCharsets.UTF_8));
-            return true;
+        try {
+            final FileOutputStream f = new FileOutputStream(getFileForKey(key));
+            byte[] encryptedContent = Encryptor.encryptWithoutBase64Encoding(value.getBytes(), encryptionKey);
```
bhariharan, line 145: Using these new methods in this class now, so upgrades should work.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
bhariharan: Using new method now.

libs/test/SalesforceAnalyticsTest/src/com/salesforce/androidsdk/analytics/security/EncryptorTest.java
```diff
@@ -198,6 +198,22 @@ public void testEncryptDecryptWithCipher()
 		Assert.assertEquals("Recovered text should match original", originalText, recoveredText);
 	}
 
+	/**
+	 * Encrypting/decrypting data with ciphers returned by Encryptor.encryptWithoutBase64Encoding and
+	 * Encryptor.decryptWithoutBase64Encoding.
+	 */
+	@Test
+	public void testEncryptDecryptWithoutBase64Encoding() {
```
bhariharan, line 206: Added a new test here for the new methods that were added.

wmathurin (approved):

PR #2126, opened 2020-09-28
Fixing file.close() and adding comments about encryption/decryption

bhariharan:
Some cleanup to ensure that `file.close()` will always be called and the stream is reliably closed.


wmathurin (commented):
Should we use the more modern syntax https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

bhariharan:
@wmathurin That's `JDK 8`, right? Would that work on `API 23`?

wmathurin:
@bhariharan not sure - let's go with try { } finally { } for now ;-)

wmathurin (approved):

PR #2134, opened 2020-10-09
Fix Key-Value Encrypted Store dropping new lines.

brandonpage:


wmathurin (changes requested):
https://stackoverflow.com/questions/309424/how-do-i-read-convert-an-inputstream-into-a-string-in-java provides 11 ways to convert a stream to a string.
Some convert different line breaks to one.

It looks like the fastest and correct in all cases JDK based solution is solution 8:
```java
ByteArrayOutputStream result = new ByteArrayOutputStream();
byte[] buffer = new byte[1024];
int length;
while ((length = inputStream.read(buffer)) != -1) {
    result.write(buffer, 0, length);
}
// StandardCharsets.UTF_8.name() > JDK 7
return result.toString("UTF-8");
```

wmathurin (approved):
Could you check to see if we are doing the stream to string stuff the wrong way anywhere else in the codebase?

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -206,13 +207,13 @@ public String getValue(String key) {
...
-            while ((line = reader.readLine()) != null) {
-                out.append(line);
+            ByteArrayOutputStream value = new ByteArrayOutputStream();
+            byte[] buffer = new byte[READ_BUFFER_LENGTH];
+            int length;
+            while ((length = inputStream.read(buffer)) != -1) {
+                value.write(buffer, 0, length);
             }
-            return out.toString();
+            return value.toString("UTF-8");
```
bhariharan, line 216:
```suggestion
            return value.toString(StandardCharsets.UTF_8);
```

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -56,6 +56,7 @@
 
     private static final String TAG = KeyValueEncryptedFileStore.class.getSimpleName();
     public static final int MAX_STORE_NAME_LENGTH = 96;
+    private final int READ_BUFFER_LENGTH = 1024;
```
bhariharan, line 59:
```suggestion
    private static final int READ_BUFFER_LENGTH = 1024;
```

wmathurin (approved):

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -469,6 +475,24 @@ public static String decryptBytes(byte[] data, byte[] key, byte[] iv) {
         return null;
     }
 
+    public static String getStringFromStream(InputStream stream) throws IOException {
```
bhariharan, line 478: Add `Javadoc` since they're `public` methods.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -469,6 +475,24 @@ public static String decryptBytes(byte[] data, byte[] key, byte[] iv) {
...
+        ByteArrayOutputStream output = new ByteArrayOutputStream();
+        byte[] buffer = new byte[READ_BUFFER_LENGTH];
+        int length;
+        while ((length = stream.read(buffer)) != -1) {
+            output.write(buffer, 0, length);
+        }
+        stream.close();
```
bhariharan, line 485: If an exception is thrown, the stream won't be closed. Put this in a `finally` block.

bhariharan (changes requested):

bhariharan (approved):

PR #2141, opened 2020-11-23
Allow the authenticator service's icon to be rebranded.

chrlie:
This PR allows the account icon provided by the authenticator service (visible in the Settings app) to be overridden at the app level. `<account-authenticator>` seems to be outside the responsibility of variant/manifest merging, so I've chosen to use an overridable string.


bhariharan (changes requested):
@chrlie We don't need this. The idea is that the app binary would replace `sf__icon` with the drawable/icon of its choice, and it would be automatically picked.

PR #2151, opened 2020-12-24
Update RestResponse.java

shu22ren:
AsBytes method call consume has give responseAsString assignment!!!

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
@@ -170,7 +170,10 @@ public class RestResponse {
 	public String asString() throws IOException {
 		if (responseAsString == null) {
 			byte[] bytes = asBytes(); // will also compute responseCharSet
-			responseAsString = new String(bytes, responseCharSet);
+			if(responseAsString==null){
+				responseAsString = new String(bytes, responseCharSet);
+			}
+			
 		}
 		return responseAsString;
 	}


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
```diff
@@ -170,7 +170,10 @@ public String getContentType() {
 	public String asString() throws IOException {
 		if (responseAsString == null) {
 			byte[] bytes = asBytes(); // will also compute responseCharSet
-			responseAsString = new String(bytes, responseCharSet);
+			if(responseAsString==null){
```
bhariharan, line 173: I'm not sure I understand what the intent of this is. The `if (responseAsString == null)` check is already being performed in `line 171`, and calling `asBytes()` doesn't affect this (`asBytes()` only returns `responseAsBytes`, it doesn't mutate the value of `responseAsString`). This check is redundant and doesn't perform anything as far as I can tell.

bhariharan (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java
shu22ren:
```
JsonObject jsonObject = (JsonObject) jsonParser.parse(response.asString());

public String asString() throws IOException {
	if (responseAsString == null) {
		byte[] bytes = asBytes(); // will also compute responseCharSet
		///////////////////////////
		asBytes() 151 line call consume()()
		consume()() 119 line
		if (responseAsBytes != null && responseAsBytes.length > 0) {
			responseAsString = new String(responseAsBytes, responseCharSet);
		}
		///////////////////////////
		responseAsString = new String(bytes, responseCharSet); //double new String
	}
	return responseAsString;
}
```


I have a lot of large amount of data downloading

bhariharan: For a large dataset, you probably want to use `Jackson` to stream parse the JSON response, instead of calling `asString()` and using `JSONObject`.

PR #2152, opened 2021-01-05
[BUGFIX] Using app context instead of activity context to avoid memory leak

mnaturel-niji:
## Description
In `SalesforceActivityDelegate`, we replace the usage of Activity Context to build `ClientManager` since it may lead to memory leak when the activity is destroyed. We use Application Context instead.

## Motivation and Context
Using [LeakCanary](https://square.github.io/leakcanary/), we can detect memory leak when the related Activity is destroyed. The reference is kept by the `ClientManager` and its underlaying dependencies.

## How Has This Been Tested?
- Configure [LeakCanary](https://square.github.io/leakcanary/) on your project which imports the Salesforce SDK 
- Start an activity which uses `SalesforceActivityDelegate` such as `SalesforceActivity`
- Quit the activity
- Check the leak does not exist anymore


wmathurin (approved):
Thanks @mnaturel-niji

bhariharan (approved):
Thanks!

PR #2159, opened 2021-02-04
Do not instantiate smartstore if there are no syncs or store configs

wmathurin:
Issue was only with hybrid apps (SmartStoreSDKManager, MobileSyncSDKManager and by extension SalesforceReactSDKManager were doing the check already).


rwhitley (commented):
LGTM

bhariharan (approved):

PR #2166, opened 2021-02-25
Key value store that keeps recently accessed values in a in-memory lru cache for faster access

wmathurin:
* Added class MemCacheKeyValueStore (wraps around regular key value store)
* Added interface KeyValueStore (implemented by both KeyValueEncryptedFileStore and MemCacheKeyValueStore)
* Added tests
* Move smartstore tests from package androidsdk.store to androidsdk.smartstore.store to allow access to default members

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/MemCachedKeyValueStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/MemCachedKeyValueStore.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2021-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of salesforce.com, inc. nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission of salesforce.com, inc.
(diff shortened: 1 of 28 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/MemCachedKeyValueStore.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public InputStream getStream(String key) {
+        String value = getValue(key);
+        return value == null ?  null : new ByteArrayInputStream(value.getBytes(StandardCharsets.UTF_8));
+    }
+
+    @Override
+    public boolean saveValue(String key, String value) {
+        if (keyValueStore.saveValue(key, value)) {
```
bhariharan, line 71: Could be simplified to something like the following, maybe:
```
return (keyValueStore.saveValue(key, value) ? memCache.put(key, value) : false);
```

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/MemCachedKeyValueStore.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public boolean saveStream(String key, InputStream stream) throws IOException {
+        String value = Encryptor.getStringFromStream(stream);
+        return saveValue(key, value);
+    }
+
+    @Override
+    public boolean deleteValue(String key) {
+        if (keyValueStore.deleteValue(key)) {
```
bhariharan, line 87: Maybe similar simplification like above.

bhariharan (approved):

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/MemCachedKeyValueStore.java
brandonpage: I think it's fine as is.  `memCache.put(key, value)` returns the previously stored value, not boolean.

brandonpage (approved):
👍  LGTM.  Maybe before 9.1 goes out I can add a button to the KV Store Inspector to print LRU Cache snapshot on MemeCache Stores.

PR #2168, opened 2021-02-27
Adding keySet() method to key store

wmathurin:
- Storing values in `<hash(key)>.value` and keys in `<hash<(key)>.key` => `keySet()` simply gets all the `*.key` files and read their content.
- To be backwards compatible, also added a version file (version is now 2, stores without a version file will be considered to be version 1). Operations behavior depends on store version => `keySet()` throws an `UnsupportedOperationException` for a v1 store.
- Added tests for new operations and also to make sure v1 stores are still supported.

TODO (in a separate PR/story): enhance kv store inspector to take advantage of `keySet()` e.g. the key input field could work like a search field with the list below showing all the key/value fields matching the partially typed key.

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
@@ -161,7 +161,7 @@ public class KeyValueEncryptedFileStore implements KeyValueStore {
             return false;
         }
         try {
-            encryptStringToFile(getKeyFile(key), key, encryptionKey);
+            if (kvVersion >= 2) encryptStringToFile(getKeyFile(key), key, encryptionKey);
             encryptStringToFile(getValueFile(key), value, encryptionKey);
             return true;
         } catch (Exception e) {
@@ -188,7 +188,7 @@ public class KeyValueEncryptedFileStore implements KeyValueStore {
             return false;
         }
         try {
-            encryptStringToFile(getKeyFile(key), key, encryptionKey);
+            if (kvVersion >=2) encryptStringToFile(getKeyFile(key), key, encryptionKey);
             encryptStreamToFile(getValueFile(key), stream, encryptionKey);
             return true;
         } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -144,28 +160,13 @@ public boolean saveValue(String key, String value) {
...
-            byte[] encryptedContent = Encryptor.encryptWithoutBase64Encoding(value.getBytes(), encryptionKey);
-            if (encryptedContent != null) {
-                f.write(encryptedContent);
-                return true;
-            } else {
-                return false;
-            }
-        } catch (IOException e) {
-            SmartStoreLogger.e(TAG, "IOException occurred while saving value to filesystem", e);
+            if (kvVersion >= 2) encryptStringToFile(getKeyFile(key), key, encryptionKey);
```
bhariharan, line 164: High level design question: Instead of maintaining a version here that's separate from the Mobile SDK version, could we use `SmartStoreUpgradeManager` to perform a one-time upgrade step between `9.0` and `9.1` like we do with the other components? You'll get the version information for free since we already write the current version to a file, and apps will automatically be upgraded because upgrade steps are wired into the `init` call.

wmathurin: We can't upgrade. In 9.0, we one-way hash the keys to produce the file names. There is no way to get the keys back.

wmathurin: For the version number, I used https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java#L62 as inspiration.

bhariharan: Gotcha. Is there a long term plan to remove the old code, i.e. `V1`?

bhariharan (approved):

wmathurin:
We should stop supporting v1 kv stores at some point. Maybe 10.0?

PR #2183, opened 2021-03-24
Adding Some Enhancements and Fixing Some Issues in Cookie Management

bhariharan:
This PR fixes a few issues with the hybrid webview.
- We don't need `loadVFPingPage()` anymore. It's inefficient with the hidden webview loading, and doesn't really do anything much anymore, since the `frontdoor` hops we do cover all the scenarios we need. It also slows down load time in hybrid apps.
- I removed `setSidCookies()` as well. It's not required anymore for a few reasons. One reason is because `frontdoor` does its hops and sets the cookies where required. The second reason is because `Google` discourages cookie manipulation by apps in `Chromium` based webviews. The third, and most important reason is that it was actually causing issues with certain community URLs, because when the URL changes the old URL has the old `sid` and the new URL has the new `sid`. This causes an `Internal Server Error` which took a while to track down. Removing manual cookie management solves all these issues, with `frontdoor` handling setting of cookies wherever they need to be set.
- I deprecated the cookie management `public` APIs for removal in `Mobile SDK 10.0`. I don't think we should be in the business of managing cookies manually anymore. This is in line with Google's recommendations as well.

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java
@@ -29,11 +29,7 @@ package com.salesforce.androidsdk.phonegap.ui;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.view.KeyEvent;
-import android.view.View;
-import android.webkit.CookieManager;
 import android.webkit.URLUtil;
-import android.webkit.WebView;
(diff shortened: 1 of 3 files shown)


brandonpage (approved):

bhariharan:
Tests are passing locally when I change `"test_login_domain" : "login.salesforce.com"` to `"test_login_domain" : "https://login.salesforce.com"`. How do I make this change to `CircleCI`, @brandonpage?

bhariharan:
All the hybrid tests are passing locally as well. I think the credentials on `CircleCI` need to be updated, i.e. `test_credentials`.

PR #2189, opened 2021-04-24
Getting Ready for Bintray / jCenter Being Sunset

bhariharan:
- Removes the publishing module from our library's Gradle files.
- Removes the dependency on the Bintray publish plugin.
- Keeps `jCenter()` in a few places because `Cordova` and `PaperDB` still need it.
- I'll get to removing the remaining instances when I figure out an alternate fetch strategy for `Cordova` and `PaperDB` (we have till early 2022 for this).
- I'll re-add publishing code when we migrate to `mavenCentral()` (probably in a future release).


brandonpage (approved):

wmathurin (approved):
We should get rid of the jCenter related next steps [here](https://github.com/forcedotcom/SalesforceMobileSDK-Package/blob/dev/release/release.js#L173).

bhariharan:
@wmathurin Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Package/pull/232).

PR #2190, opened 2021-04-27
Disable Jetifier

jasonschroeder-sfdc:
There are no more dependencies on support libraries, everything is using
AndroidX.
Bye Bye Jetifier! :wave:

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,3 @@
 cdvCompileSdkVersion=30
 cdvMinSdkVersion=23
 android.useAndroidX=true
-android.enableJetifier=true


gradle.properties
```diff
@@ -1,4 +1,4 @@
 cdvCompileSdkVersion=30
 cdvMinSdkVersion=23
 android.useAndroidX=true
-android.enableJetifier=true
+android.enableJetifier=false
```
bhariharan, line 4: Can we remove this line? I assume the default is `false`.

bhariharan: Also, does this work for a Cordova-generated project?

jasonschroeder-sfdc: I wanted to be really explicit, so I was thinking of just setting it to `false`. But I can certainly remove it.

I don't know about cordova - I can say that all automated tests are passing.

bhariharan: Let's just remove it :-)

bhariharan (approved):

PR #2198, opened 2021-06-04
Attempts to fetch the auth config will now send a broadcast after they complete.

chrlie:
This change allow sequential app-level code to be built around the `FetchAuthConfigTask` (or any other usage of `getMyDomainAuthConfig`)

For instance, one can listen for broadcasts in a subclass of SalesforceDroidGapActivity:

```kotlin
class MainActivity: SalesforceDroidGapActivity {
   private val receiver = object : BroadcastReceiver() {
      override fun onReceive(context: Context?, intent: Intent?) {
         if (intent != null && intent.hasExtra(WAS_REQUEST_SUCCESSFUL_EXTRA)) {
            onAuthConfigComplete(intent.getBooleanExtra(WAS_REQUEST_SUCCESSFUL_EXTRA, false))
         }
      }
   }
   
   override fun onStart() {
      // Register a receiver before getMyDomainAuthConfig is called in onResume()
      registerReceiver(receiver, IntentFilter(AUTH_CONFIG_COMPLETE_INTENT_ACTION))
   }
   
   private fun onAuthConfigComplete() { ... }
}
```

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/util/AuthConfigUtil.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/util/AuthConfigUtil.java
@@ -26,8 +26,10 @@
  */
 package com.salesforce.androidsdk.util;
 
+import android.content.Intent;
 import android.text.TextUtils;
 
+import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.auth.HttpAccess;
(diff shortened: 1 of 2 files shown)


wmathurin (approved):

chrlie:
@wmathurin it looks like the tests are failing because `test_credentials.json` is empty

bhariharan (approved):

wmathurin:
test_credentials.json is empty in github. But on CircleCI we generate it from an environment variable before running tests.

PR #2203, opened 2021-06-30
Fixing bug with Android for Work where addition of accounts is disabled by the admin

bhariharan:
`Android for Work` has a relatively new  policy allowing admins to disable users from adding accounts. However, apps should still be allowed to add user accounts, even if this policy is disabled. The correct way to handle this is by ensuring that apps don't call `addAccount` directly, and only call `addAccountExplicitly`. This PR ensures that `Mobile SDK` is compliant with this policy, by only using `addAccountExplicitly` as the API is designed to be used.


brandonpage (approved):

brandonpage:
There is a comment on line 509 that says: 
``` 
// There is a bug in AccountManager::addAccountExplicitly() that sometimes causes user data to not be
// saved when the user data is passed in through that method. The work-around is to call setUserData()
// for all the user data manually after passing in empty user data into addAccountExplicitly().
```

Has this been fixed?

bhariharan:
@brandonpage We set them individually. I haven't seen that happen recently.

PR #2207, opened 2021-07-15
Add gradle files to publish to MavenCentral

brandonpage:


libs/MobileSync/build.gradle
```diff
@@ -62,3 +67,11 @@ android {
     abortOnError false
   }
 }
+
+ext {
```
wmathurin, line 71: We need to update setversions.sh [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/setversion.sh#L62) to change the publish version.

brandonpage: Will do.  While I'm at it, should I just update the version in `publish-module.gradle` since all of them will always be the same?

wmathurin (approved):

publish/publish.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
```
wmathurin, line 2: Must it be run from the root dir?

brandonpage: I was running `./publish/publish.sh` but I could add an if check and cd if necessary.

wmathurin: That's fine, I just wanted [this](https://github.com/forcedotcom/SalesforceMobileSDK-Package/pull/239) to be correct

PR #2216, opened 2021-08-19
Key value store operations should not throw exception even if there a…

wmathurin:
…re bad key files

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
@@ -43,6 +43,7 @@ import java.io.FileOutputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -269,9 +270,13 @@ public void deleteAll() {
                 file.delete();
             }
         } else {
-            for (String key : keySet()) {
-                SmartStoreLogger.i(TAG, "deleting key :" + key);
-                deleteValue(key);
+            for (File file : safeListFiles(KEY_SUFFIX)) {
```
wmathurin, line 273: Faster this way - keySet() reads all the key files

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/KeyValueEncryptedFileStore.java
```diff
@@ -302,7 +307,7 @@ public void deleteAll() {
     /** @return number of entries in the store. */
     @Override
     public int count() {
-        return kvVersion == 1 ? safeListFiles(null /* all */).length : safeListFiles(VALUE_SUFFIX).length;
+        return kvVersion == 1 ? safeListFiles(null /* all */).length : keySet().size();
```
wmathurin, line 310: count() should return the count of good key files

bhariharan (approved):

PR #2217, opened 2021-08-24
Fix CI build issue preventing tests from running.

brandonpage:


wmathurin (approved):

bhariharan (approved):
Should we just perform the upgrade now and  remove `jcenter()` in `9.2`, @wmathurin @brandonpage?

brandonpage:
@bhariharan We should do it 9.2 but there is more pressing work to be done this week.

PR #2227, opened 2021-09-07
Replace Passcode with Screen Lock

brandonpage:
Sorry for the enormous PR, but it had to be done.

- All Passcode related public API has been deprecated for removal in 10.0.
- Passcode is replace with the OS's Screen Lock.  This approach is far simpler for the end user and will be much easier to maintain over time.  
- Removed pin length requirement.  If customers want to enforce more strict policy than Android's minimum they can do this via MDM.
- Removed timeout requirement.  The app locks on background consistently like users expect.  
- Removed biometric override from connected app.  
- Removed auth attempts, this is handled by the OS now.
- No more passcode lock-in until you uninstall or logout all users.  When the last user that requires screen lock is removed, the lock is too.  Logout button on screen lock removes all (and only) users that require the lock.
- ScreenLockActivity is simple lock screen that displays the app icon, a setup button if applicable, a retry auth button if applicable, logout button if applicable and an OS provided error message if applicable.  

As far as I can tell there is no way to actually write activity tests that require keyguard/biometric but someone please correct me if I am wrong on that.


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -929,31 +963,9 @@ public void logout(Account account, Activity frontActivity, final boolean showLo
...
 
-    	/*
-    	 * Removes the existing account, if any. 'account == null' does not
-    	 * guarantee that there are no accounts to remove. In the 'Forgot Passcode'
-    	 * flow there could be accounts to remove, but we don't have them, since
-    	 * we don't have the passcode hash to decrypt them. Hence, we query
-    	 * AccountManager directly here and remove the accounts for the case
-    	 * where 'account == null'. If AccountManager doesn't have accounts
-    	 * either, then there's nothing to do.
-    	 */
```
brandonpage, line 942: I believe this code block has not been needed since we stopped using passcode as entropy.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -150,4 +156,30 @@ private void migrateAnalyticsData() {
             }
         }
     }
+
+    private void upgradeTo9Dot2() {
```
brandonpage, line 160: Upgrade steps.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -555,17 +553,7 @@ public IdServiceResponse(Response response) {
...
-                    screenLockTimeout = parsedResponse.getJSONObject(MOBILE_POLICY).getInt(SCREEN_LOCK);
-                    if (customAttributes != null) {
-                        String bioAttribute = customAttributes.optString(BIOMETRIC_UNLOCK).toLowerCase(Locale.US);
-                        if (bioAttribute.equals("false")) {
-                            biometricUnlockAllowed = false;
-                            SalesforceSDKLogger.i(TAG, "Biometric Unlock disabled by connected app.");
-                        }
-                    }
-                }
+                mobilePolicy = parsedResponse.has(MOBILE_POLICY);
```
brandonpage, line 556: We now only care that the admin as enabled the policy on the connected app.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
```diff
@@ -0,0 +1,135 @@
...
+import com.salesforce.androidsdk.util.EventsObservable;
+
+import java.util.List;
+
+/**
+ * Manages if the app should be locked.
+ *
+ * @author bpage
+ */
+public class ScreenLockManager {
```
brandonpage, line 48: Simplified Screen Lock Manager.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+import com.salesforce.androidsdk.accounts.UserAccount;
+import com.salesforce.androidsdk.accounts.UserAccountManager;
+import com.salesforce.androidsdk.app.SalesforceSDKManager;
+
+import java.util.List;
+
+/**
+ * Locks the app behind OS provided authentication.
+ */
+public class ScreenLockActivity extends FragmentActivity {
```
brandonpage, line 71: New Screen Lock Activity.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+                break;
+            case BiometricManager.BIOMETRIC_SUCCESS:
+                resetUI();
+
+                /*
+                 * This is necessary due to an Android bug that can't be fixed.
+                 * https://issuetracker.google.com/issues/145231213
+                 *
+                 * TODO: Remove when min API > 29.
+                 */
```
brandonpage, line 179: So much for "One Biometric API Over all Android" 🙄

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -147,9 +151,9 @@
     private LoginOptions loginOptions;
     private final Class<? extends Activity> mainActivityClass;
     private Class<? extends Activity> loginActivityClass = LoginActivity.class;
-    private Class<? extends PasscodeActivity> passcodeActivityClass = PasscodeActivity.class;
+    private Class<? extends ScreenLockActivity> screenLockActivityClass = ScreenLockActivity.class;
```
bhariharan, line 154: Remove this. Now that it's handled by the OS, we don't want apps to customize this anymore. They can tweak the colors and branding  by overriding it in XML. The only reason we allowed `PasscodeActivity` to be overridden was  for legacy reasons  specific  to `SalesforceApp`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -176,9 +180,9 @@
     }
 
     /**
-     * PasscodeManager object lock.
+     * ScreenLockManager object lock.
      */
-    private final Object passcodeManagerLock = new Object();
+    private final Object screenLockManagerLock = new Object();
```
bhariharan, line 185: I  don't  think this is  needed anymore.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -400,20 +404,39 @@ public static void initNative(Context context, Class<? extends Activity> mainAct
...
+     */
+    public Class<? extends PasscodeActivity> getPasscodeActivity() { return PasscodeActivity.class; }
+
+    /**
+     * Sets a custom ScreenLock activity class to be used instead of the default class.
+     * The custom class must subclass {@link ScreenLockActivity}.
+     *
+     * @param activity Subclass of ScreenLockActivity.
+     */
+    public void setScreenLockActivity(Class<? extends ScreenLockActivity> activity) {
```
bhariharan, line 427: Remove this. Read reason above. We don't want to  expose a  setter  anymore.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -400,20 +404,39 @@ public static void initNative(Context context, Class<? extends Activity> mainAct
...
+    }
+
+    /**
+     * Returns the descriptor of the ScreenLock activity class that's currently in use.
+     *
+     * @return ScreenLock activity class descriptor.
      */
-    public Class<? extends PasscodeActivity> getPasscodeActivity() {
-    	return passcodeActivityClass;
+    public Class<? extends ScreenLockActivity> getScreenLockActivity() {
```
bhariharan, line 438: Remove this.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -512,17 +535,26 @@ public synchronized void setPushServiceType(Class<? extends PushService> type) {
...
+     * Returns the ScreenLock manager that's associated with SalesforceSDKManager.
      *
-     * @return PasscodeManager instance.
+     * @return ScreenLockManager instance.
      */
-    public PasscodeManager getPasscodeManager() {
-    	synchronized (passcodeManagerLock) {
-            if (passcodeManager == null) {
-                passcodeManager = new PasscodeManager(context);
+    public ScreenLockManager getScreenLockManager() {
```
bhariharan, line 551: Do  we still need this? Remove it if possible.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
bhariharan: Add `TODO:` here to remove this  in `Mobile SDK 10.0`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/PasscodeManager.java
```diff
@@ -49,141 +49,56 @@
  *
  * @author wmathurin
  * @author bhariharan
+ *
+ * @deprecated Will be removed in Mobile SDK 10.0.  Use {@link ScreenLockManager} instead.
```
bhariharan, line 53: Lol - this class is worthless  at this point ;-)

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public static final String MOBILE_POLICY_PREF = "mobile_policy";
```
bhariharan, line 50: Are these constants used  outside this class? Make them `private` otherwise.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+        SharedPreferences globalPrefs = ctx.getSharedPreferences(MOBILE_POLICY_PREF, Context.MODE_PRIVATE);
+        if (screenLockRequired) {
+            globalPrefs.edit().putBoolean(SCREEN_LOCK, true).apply();
+        }
+    }
+
+    public boolean onResume() {
+        boolean locked = shouldLock && readMobilePolicy();
+        if (locked) {
```
bhariharan, line 69: This is confusing - "if locked, lock"? What are we doing here?

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+        // If locked, do nothing - when the app gets unlocked we will be back here.
+        return !locked;
+    }
+
+    public void onPause() {
+        setShouldLock(true);
+    }
+
+    public void setShouldLock(boolean shouldLock) {
```
bhariharan, line 81: Can this be made `private`?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+import com.salesforce.androidsdk.accounts.UserAccountManager;
+import com.salesforce.androidsdk.app.SalesforceSDKManager;
+
+import java.util.List;
+
+/**
+ * Locks the app behind OS provided authentication.
+ */
+public class ScreenLockActivity extends FragmentActivity {
+    public static final int SCREEN_LOCK_REQUEST_CODE = 777;
```
bhariharan, line 72: Can the code be `private`?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+        logoutButton = findViewById(R.id.sf__screen_lock_logout_button);
+        logoutButton.setOnClickListener(v -> logoutScreenLockUsers());
+        actionButton = findViewById(R.id.sf__screen_action_button);
+        ImageView appIcon = findViewById(R.id.sf__app_icon);
+
+        try {
+            Drawable icon = getPackageManager().getApplicationIcon(getApplicationInfo().packageName);
+            appIcon.setImageDrawable(icon);
+        } catch (PackageManager.NameNotFoundException e) {
+            Log.e(TAG, "Unable to retrieve host app icon.  NameNotFoundException: " + e.getMessage());
```
bhariharan, line 104: Use `SalesforceSDKLogger`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+         * presentAuth again after the user has come back from security settings to ensure they
+         * actually set up a secure lock screen (pin/pattern/password/etc) instead of swipe or none.
+         */
+        if (requestCode == SETUP_REQUEST_CODE) {
+            presentAuth();
+        }
+
+        /*
+         * Get the results of KeyguardManager on API 29.
+         * TODO: TODO: Remove when min API > 29.
```
bhariharan, line 125: Remove the extra `TODO:`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+        BiometricManager biometricManager = BiometricManager.from(this);
+
+        switch (biometricManager.canAuthenticate(getAuthenticators())) {
+            case BiometricManager.BIOMETRIC_ERROR_NO_HARDWARE:
+            case BiometricManager.BIOMETRIC_ERROR_SECURITY_UPDATE_REQUIRED:
+            case BiometricManager.BIOMETRIC_ERROR_UNSUPPORTED:
+            case BiometricManager.BIOMETRIC_STATUS_UNKNOWN:
+                // This should never happen.
+                String error = getString(R.string.sf__screen_lock_error);
+                Log.e(TAG, "Biometric manager cannot authenticate. " + error);
```
bhariharan, line 147: Use  `SalesforceSDKLogger`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -0,0 +1,296 @@
...
+                break;
+            case BiometricManager.BIOMETRIC_ERROR_HW_UNAVAILABLE:
+                setErrorMessage(getString(R.string.sf__screen_lock_error_hw_unavailable));
+                break;
+            case BiometricManager.BIOMETRIC_ERROR_NONE_ENROLLED:
+                setErrorMessage(getString(R.string.sf__screen_lock_setup_required, appName));
+
+                /*
+                 * Prompts the user to setup OS screen lock and biometric.
+                 * TODO: TODO: Remove when min API > 29.
```
bhariharan, line 158: Extra `TODO:`.

bhariharan (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
brandonpage: I believe we do since we still need to set mobile policy in `OAuthWebviewHelper` and update on unlock from `ScreenLockActivity`.  Unfortunately both of those are in the `ui` package so it needs to remain public.  Also we also need it for user cleanup in `SalesforceSDKManager` so I don't think it makes sense to move it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
brandonpage: Removed the request code as we don't use it and the manager class that would actually use it can't be extended with my last commit.

bhariharan (approved):
LGTM! 👍🏻

bhariharan (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
wmathurin: Could you rewrite as
```java
public boolean onResume() {
  if (shouldLock && readMobilePolicy()) {
    lock();
    return false; 
  } else {
    // If locked, do nothing - when the app gets unlocked we will be back here.
    return true;
  }
```
Maybe add a javadoc for the meaning of the boolean returned?

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
brandonpage: Unfortunately it is needed in the activity.  But since we removed the SalesforceSDKManager getter it shouldn't be a problem.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
brandonpage: Hey are used in the ScreenLockActivity.

bhariharan (approved):

PR #2232, opened 2021-09-15
Fix ScreenLock issues

brandonpage:
- Improve when ScreenLock is shown by using lifecycle apis and fix associated issues.  
- Cleanup LockScreenManager API.   
- Fix an upgrade issue related to future logouts.


wmathurin (commented):
LGTM 
So the lifecycle event is handled before any activities are resumed?

bhariharan (approved):

wmathurin (approved):

PR #2239, opened 2021-09-28
Adding CI jobs for Android 30 and 31

wmathurin:


.circleci/config.yml
```diff
@@ -518,14 +565,102 @@ workflows:
...
+    triggers:
+      - schedule:
+          cron: "0 3 * * 6"
+          filters:
+            branches:
+              only:
+                - dev
+    jobs:
+      - run-tests:
+          name: "SalesforceAnalytics API 31"
```
brandonpage, line 578: I think this will fail because Firebase Test Lab does not support API 31 yet

bhariharan (approved):

.circleci/config.yml
wmathurin: Ok, I'll comment out 31 for now.

PR #2248, opened 2021-11-08
Removing Deprecated Methods and Classes for 10.0

bhariharan:
- Removed deprecated encryption/decryption methods.
- Removed upgrade steps for `8.2` - it has been long enough.
- Removed old passcode classes that aren't used or referenced anywhere anymore.


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1223,19 +1162,11 @@ public void run() {
                         new AlertDialog.Builder(frontActivity)
                                 .setItems(
                                         devActions.keySet().toArray(new String[0]),
-                                        new DialogInterface.OnClickListener() {
-                                            @Override
-                                            public void onClick(DialogInterface dialog, int which) {
-                                                devActionHandlers[which].onSelected();
-                                                devActionsDialog = null;
-                                            }
+                                        (dialog, which) -> {
```
bhariharan, line 1165: Now that we need `Java 8` minimum, lambda functions should be fine.

brandonpage (approved):

PR #2250, opened 2021-11-16
Update minSDKVersion from API 23 to 24.

brandonpage:
I tried to fix/update everything API 23/24 related that I could find.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -389,7 +389,6 @@ workflows:
   # Friday 8 PM  - API 29
   # Friday 9 PM  - API 27
   # Friday 10 PM - API 25
-  # Friday 11 PM - API 23
   test-api-30:
     triggers:
       - schedule:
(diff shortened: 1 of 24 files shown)


brandonpage:
@bhariharan Is there any work here to be done around around hardware backed keystore?  Android 6 added a lot of security features not present beforehand.

.circleci/config.yml
```diff
@@ -739,46 +738,3 @@ workflows:
           test_apk: *salesforceReact-apk-path
           api_level: 25
 
-  test-api-23:
```
brandonpage, line 742: Side note:  This file has no business being 700+ lines long.  I will come back to this and fix it up with matrix jobs and/or CircleCI's new monorepo support.

libs/SalesforceSDK/build.gradle
```diff
@@ -11,34 +11,22 @@ dependencies {
     api project(':libs:SalesforceAnalytics')
     api 'com.squareup.okhttp3:okhttp:3.12.12'
     api 'com.google.firebase:firebase-messaging:20.1.0'
```
brandonpage, line 13: The reasons we haven't updated `okhttp` and `firebase-messagine` to the latest are unrelated from API version, correct?

bhariharan:
@brandonpage Are we cleared to bump this up? Have we run external comms and checked with SApp?

bhariharan:
Also, you need to update the tests/test projects too.

libs/SalesforceSDK/build.gradle
bhariharan: Yes, they're unrelated, and we shouldn't bump them up yet.

bhariharan:
I'll take a look at the keystore stuff after this goes in.

brandonpage:
> @brandonpage Are we cleared to bump this up? Have we run external comms and checked with SApp?

I haven't reached out to them but from looking at their repos it seems SApp and SFS are both min 24 already.  I will double check.

bhariharan:
Fair enough. Communities is still on `23`, but I'm okay bumping that up.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceWebViewClient.java
```diff
@@ -90,21 +90,14 @@ public SalesforceWebViewClient(SalesforceWebViewEngine parentEngine) {
...
-    }
-
     @Override
 	public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
-		return shouldOverrideUrlLoading(view, request.getUrl().toString());
+    	String url = request.getUrl().toString();
+		if (SalesforceWebViewClientHelper.shouldOverrideUrlLoading(ctx, view, url)) {
+			return true;
+		} else {
+			return super.shouldOverrideUrlLoading(view, url);
```
bhariharan, line 99:
```suggestion
			return super.shouldOverrideUrlLoading(view, request);
```

bhariharan (changes requested):
Just one change to make.

bhariharan (approved):

PR #2253, opened 2021-11-29
Batch read and publish of Instrumentation Events

npars:
- Batch read and publish of instrumentation events to reduce memory
  footprint

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
@@ -111,7 +111,7 @@ public class EventStoreManager {
     }
 
     /**
-     * Returns a specific event stored on the filesystem.
+     * Returns a specific event stored on the filesystem for that unique identifier.
      *
      * @param eventId Unique identifier for the event.
      * @return Event.
@@ -143,7 +143,7 @@ public class EventStoreManager {
     }
 
     /**
-     * Returns all the events stored on the filesystem for that unique identifier.
+     * Returns all the events stored on the filesystem.
      *
      * @return List of events.
      */
@@ -158,8 +158,8 @@ public class EventStoreManager {
     }
 
     /**
-     * Returns all the events stored on the filesystem for that unique identifier.
-     * Does not fetch them all at once, but
+     * Streams all the events stored on the filesystem. Will load each
+     * event into memory one at a time to decrease memory impact.
      *
      * @return Iterable of events.
      */
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
@@ -37,6 +37,8 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 
 import io.paperdb.Book;
@@ -146,15 +148,42 @@ public class EventStoreManager {
      * @return List of events.
      */
     public List<InstrumentationEvent> fetchAllEvents() {
-        final List<String> eventIds = book.getAllKeys();
         final List<InstrumentationEvent> events = new ArrayList<>();
-        for (final String eventId : eventIds) {
-            final InstrumentationEvent event = fetchEvent(eventId);
+        for (InstrumentationEvent event : iterateAllEvents()) {
             if (event != null) {
                 events.add(event);
             }
         }
         return events;
+    }
+
+    /**
+     * Returns all the events stored on the filesystem for that unique identifier.
+     * Does not fetch them all at once, but
+     *
+     * @return Iterable of events.
+     */
+    public Iterable<InstrumentationEvent> iterateAllEvents() {
+        return new Iterable<InstrumentationEvent>() {
+            private final List<String> eventIds = book.getAllKeys();
+
+            @Override
+            public Iterator<InstrumentationEvent> iterator() {
+                return new Iterator<InstrumentationEvent>() {
+                    private final Iterator<String> eventIdIterator = eventIds.iterator();
+
+                    @Override
+                    public boolean hasNext() {
+                        return eventIdIterator.hasNext();
+                    }
+
+                    @Override
+                    public InstrumentationEvent next() {
+                        return fetchEvent(eventIdIterator.next());
+                    }
+                };
+            }
+        };
     }
 
     /**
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -200,6 +204,13 @@ public static synchronized void setPublishFrequencyInHours(int publishFrequencyI
         }
     }
 
+    public static void setEventPublishBatchSize(int batchSize) {
```
bhariharan, line 207: Make this `synchronized`, and add some `Javadoc` for this method.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManagerTest.java
```diff
@@ -0,0 +1,102 @@
+package com.salesforce.androidsdk.analytics;
```
bhariharan, line 1: Add copyright. You can copy it in from one of the other classes.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/store/EventStoreManager.java
```diff
@@ -146,17 +148,44 @@ public InstrumentationEvent fetchEvent(String eventId) {
...
             if (event != null) {
                 events.add(event);
             }
         }
         return events;
     }
 
+    /**
+     * Returns all the events stored on the filesystem for that unique identifier.
+     * Does not fetch them all at once, but
```
bhariharan, line 162: This sentence seems incomplete.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -200,6 +204,13 @@ public static synchronized void setPublishFrequencyInHours(int publishFrequencyI
...
+        if (batchSize <= 0) {
```
bhariharan, line 208: I'd prefer that we rework this to not throw an exception. Instead, if input is `<=0`, simply return right away and continue to use the default of `100`. In `Javadoc` for this method, you can mention that valid input needs to be `> 0` if you want.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -370,6 +395,19 @@ public void addRemotePublisher(Class<? extends Transform> transformer,
         remotes.put(transformer, publisher);
     }
 
+    /**
+     * Removes a remote publisher to publish events to.
+     *
+     * @param transformer Transformer class.
+     */
+    public void removeRemotePublisher(Class<? extends Transform> transformer) {
```
bhariharan, line 403: Why do we need this? I'd prefer not adding this to our public API.

bhariharan (changes requested):
Also, could you test this against our sample apps?

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
npars: I can limit visibility. It's required because by default new instances `SalesforceAnalyticsManager` have the AILTN publisher added to them. In order to test this class the AILTN publisher needs to be removed.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -370,6 +402,19 @@ public void addRemotePublisher(Class<? extends Transform> transformer,
         remotes.put(transformer, publisher);
     }
 
+    /**
+     * Removes a remote publisher to publish events to.
+     *
+     * @param transformer Transformer class.
+     */
+    void removeRemotePublisher(Class<? extends Transform> transformer) {
```
bhariharan, line 410: Not being able to remove the remote publisher associated with a transformer is by design. You can either remove both the transformer and publisher, or keep them both. Could you remove this method? It won't be able to publish in your test scenario, since it's using an unauthenticated instance anyway.

bhariharan (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
npars: This method does remove both the transformer and the publisher, they are stored in a HashMap with the transformer acting as the key.

If I am not able to remove the default transformer + publisher I would need to build a full instrumentation payload which would make the tests much more verbose.

bhariharan: @npars I'd still prefer to remove this method and not add it just for testing. I'm okay with verbose tests. There should be other tests already that construct a full instrumentation payload. Feel free to copy from there.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManagerTest.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright (c) 2016-present, salesforce.com, inc.
```
bhariharan, line 2: Should be `2021-present`.

bhariharan (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
npars: I just tested with leaving `AILTNPublisher` active and it results in test failures since it tries to make an outbound network request, which rightly fails to succeed.

bhariharan: @npars Yes, since we're testing publish, I'm suggesting that we make it a true live network test. We have examples of this in our test suite. All our `RestClient` tests are going against a live server and making calls. You can run these tests against your test org. Look at `RestClientTest` on how to setup a live network test and run it.

npars: The goal of this test isn't to validate the actual network request, but to validate batching of metrics prior to sending to the publisher. Making this an integration test as you suggest would add a lot of complexity and make the test more difficult to maintain.

bhariharan: Alright, that's fair. I'll get this merged then.

bhariharan (approved):

PR #2260, opened 2021-12-09
Enables WAL in SmartStore

gkotula:
The SQLite Write Ahead Logging is enabled for all SmartStore DB objects.

## Overview

This is a small but impactful change to how the underlying SQLite DB objects function.  The benefits seem to be measurable, repeatable, and significant.  The head of my [sandbox/perf-testing](https://github.com/gkotula/SalesforceMobileSDK-Android/tree/sandbox/perf-testing) has a [simple upsert test](https://github.com/gkotula/SalesforceMobileSDK-Android/blob/f8b6740afc5a407f77fead8e02306d266804f3be/benchmarking/src/androidTest/java/com/salesforce/androidsdk/benchmarking/SmartStoreBenchmark.kt#L63) in a tight loop which sees a ~5x average speed increase with WAL enabled compared to without.

Further investigation for the performance impacts could be done, but between these results and the work that others have done, it seems clear that there is a very tangible performance benefit to enabling WAL.  It seems unlikely that cases would exist where enabling WAL will negatively impact performance, but further testing would be required to determine if there are such cases.

These runs were done using Jetpack's Benchmarking library which stabilizes CPU, thermals, JIT, and a host of other factors to make testing performance as repeatable as possible.  It also runs the test code 50 times and aggregates the data to generate a report.

### Test Runs

| Run # | WAL Enabled? | Median Runtime (ms) | Min Runtime (ms) | Max Runtime (ms) | Standard Deviation (ms) |
| :---: | :----------: | :-----------------: | :--------------: | :--------------: | :---------------------: |
|   1   |      ❌      |       1657 ms       |       952 ms     |      3075 ms     |         472 ms          |
|   2   |      ✅      |        255 ms       |       221 ms     |       513 ms     |          78 ms          |
|   3   |      ❌      |       1291 ms       |      1147 ms     |      1584 ms     |         106 ms          |
|   4   |      ✅      |        247 ms       |       218 ms     |       534 ms     |          62 ms          |

## Caveats

All tests were run locally and were shown to be passing, but there was one exception.  As you can see in the PR, there was one test that needed modification to pass, and this is confirmed to be caused by enabling WAL.  This change was required because of 1) the nature of the test, and 2) how WAL works at a low level.

The test is fairly simple, just ensuring the DB file increases in size with new records inserted, and this began failing after enabling WAL.  This is because WAL writes changes to a separate "WAL file" separate from the main DB file before checkpointing the changes back to the DB.  This checkpointing is scheduled automatically _once the WAL file reaches a certain **page** size_.  This test was not reaching that page threshold, and thus the test failed because the DB file never received the updates from the WAL file.

The fact that enabling WAL changed existing behavior should be called out as risk.  There could be subtle and difficult-to-diagnose bugs introduced into consumer apps with this change depending on how much they interact with SmartStore directly.

### DIFF
--- a/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/SmartStoreTest.java
+++ b/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/SmartStoreTest.java
@@ -1335,6 +1335,12 @@ public class SmartStoreTest extends SmartStoreTestCase {
 			JSONObject soupElt = new JSONObject("{'key':'abcd" + i + "', 'value':'va" + i + "', 'otherValue':'ova" + i + "'}");
 			store.create(TEST_SOUP, soupElt);
 		}
+
+		// With WAL enabled we must force a WAL checkpoint if we want the actual DB file to reflect the new content:
+		store.getDatabase()
+				.query("PRAGMA wal_checkpoint(FULL);")
+				.moveToNext();
+
         Assert.assertTrue("Database should be larger now", store.getDatabaseSize() > initialSize);
 	}
 
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
@@ -197,6 +197,11 @@ public class DBOpenHelper extends SQLiteOpenHelper {
 	}
 
 	@Override
+	public void onConfigure(final SQLiteDatabase db) {
+		db.enableWriteAheadLogging();
+	}
+
+	@Override
 	public void onCreate(SQLiteDatabase db) {
 
 		/*


libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/SmartStoreTest.java
```diff
@@ -1335,6 +1335,12 @@ public void testGetDatabaseSize() throws JSONException {
 			JSONObject soupElt = new JSONObject("{'key':'abcd" + i + "', 'value':'va" + i + "', 'otherValue':'ova" + i + "'}");
 			store.create(TEST_SOUP, soupElt);
 		}
+
+		// With WAL enabled we must force a WAL checkpoint if we want the actual DB file to reflect the new content:
+		store.getDatabase()
+				.query("PRAGMA wal_checkpoint(FULL);")
+				.moveToNext();
```
gkotula, line 1342: I felt that this was the most non-invasive way to fix the test, but this does expose some raw DB queries and knowledge about how SQLite works.  The alternative to getting this to pass is to perform enough writes to trigger a checkpoint, but that threshold is determined in the config stage and this test may not be able to know what that page threshold is set to.  This could lead to inconsistent test runs which is why I elected to force the checkpoint regardless of the page threshold.

wmathurin (approved):

wmathurin (commented):
Could you include your new benchmarking tests in the PR?

brandonpage (approved):
🎉

gkotula:
> Could you include your new benchmarking tests in the PR?

@wmathurin Absolutely!  How would you like the module structure?  In my sandbox I have the `benchmarking` module at the top level, parallel with `hybrid`, `libs`, and `native`.  I'm guessing that under `libs` would be more appropriate.

Also, what is the opinion on introducing Kotlin?  I could rewrite the tests in Java and avoid the new `classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.0'` line in the project-level `build.gradle` file.  Although, I would personally prefer to keep the Kotlin since the benchmarking module is completely independent from the actual SDK.

bhariharan:
@gkotula Could you put it under a package in `SalesforceSDKTest`?

bhariharan (approved):

wmathurin:
@bhariharan You probably meant [SmartStoreTest](https://github.com/forcedotcom/SalesforceMobileSDK-Android/tree/dev/libs/test/SmartStoreTest).

+1 for Kotlin.

Until we have Kotlin code in the libs themselves, we could keep add org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.0 to the classpath only for tests.

gkotula:
> @gkotula Could you put it under a package in SalesforceSDKTest?

The Jetpack Benchmarking library needs to run with debugging disabled at a manifest level.  Is that okay to enforce that for an SDK module?

bhariharan:
Yes, I meant `SmartStoreTest`. Also, agreed about scoping `Kotlin` to the test project for now.

@gkotula We can't disable debugging for the library, no. We could do it  for the test project though.

gkotula:
Oh I had my IDE in a mode that was hiding the true project structure, and I didn't realize there was an entirely separate source set for the `androidTest` in `SmartStore` called `SmartStoreTest`.  So it would be okay to add `debuggable="false"` to the `<application>` tag in that manifest?  That would disallow debugging all tests in the `SmartStoreTest` unless you flip that flag.

wmathurin:
May be we should have SmartStoreBenchmarkingTest as a sibling of SmartStoreTest.
Only SmartStoreBenchmarkingTest would have debuggable="false".

bhariharan:
Agree with Wolf about adding `SmartStoreBenchmarkingTest` as a sibling of `SmartStoreTest`.

gkotula:
Another thing to consider is future benchmarking for other modules.  Having the benchmarking as a separate module allows us to pull in things like `MobileSync` and benchmark those too.  It really just comes down to how we want to organize the benchmarks.  I.e. do we want all benchmarks consolidated into one module, or do we want benchmarking modules specific to each library?

bhariharan:
Maybe we should create a separate repo for benchmarking tests. That way we can keep it separate, and include all modules we want to benchmark. What say, @wmathurin and @gkotula?

wmathurin:
When things are in a separate repo, there are extra steps involved with every code change (first merge the lib change, then update the submodule in the second repo etc) which means that they are less likely to be run as often.

So maybe we could do what @gkotula suggested earlier in this thread: have a top directory for benchmarking.

gkotula:
Maybe we're creeping scope with this story?  Perhaps this is a larger conversation and we should just merge this and follow up with the perf testing module?

wmathurin:
Sure, you can use a different PR / story.

PR #2283, opened 2022-02-24
Mobile Sync Compose Template

gkotula:
🚧 This PR is a draft because there are still some minor TODOs that will be addressed before it's merged.  I'm opening it now to get feedback on the overall content.

## Overview

Creates a sample app to demonstrate how to use Jetpack Compose with MobileSync and the MSDK in general.  It is functionally very similar to the MobileSyncExplorer app where the only functionality it exposes is creating, viewing, editing, and deleting contacts.

### All UI is built with Jetpack Compose

This more or less requires specific app architecture that behaves nicely with Compose semantics.  Primarily, UI state objects and data model objects are immutable, and changes in the UI state are driven by a sort of state-machine structure.  UI events are handled serially and atomically in the View Model, creating new ui/model objects and emitting them to the Compose framework.

### First-class support for coroutines

While MSDK does not have any support for Kotlin coroutines, this sample demonstrates how to bridge between traditional callback-style and suspending-style.  Coroutines are fully embraced by Android, and wrapping the MSDK in suspending mechanics allows use of all modern Android development practices.

### Fully embraces recommended Android app architecture

https://developer.android.com/jetpack/guide

This app's architecture emerges from using "unidirectional data flow" (UDI) along with MVVM and "Repos" as the guidelines.

## The New Roles of Activities and Fragments

With Jetpack View Models and Compose, the Activity can almost be treated as just a container for the app window.  All the UI rendering, input handling, and data operations are handled by Compose + View Model.  Activities still are the only way to interact with the Android operating system through `Context` and `Intents`, but beyond that the Activity [should only be responsible for kicking off the Compose + View Model pattern](https://developer.android.com/jetpack/guide#separation-of-concerns).

This leads to the new role of Fragments.  Fragments are meant to be reusable pieces of UI + user interaction, but Jetpack Compose has the power to render anything to the Activity's window with reusable UI components defined as `@Composable` methods.  Consequently, Fragments are almost obsolete at this point, and this sample app does not use them.

Even Android's premier flexible UI demo app [Jetnews](https://github.com/android/compose-samples/tree/master/JetNews) only uses `Fragments` as wrappers for in-app destinations because they work with the existing Jetpack Navigation Component.  Within a destination, Jetnews implements all navigation, reusable UI, and data operations purely with Compose + View Models instead of Fragments.

## The Monolithic View Model

Just by lines of code, the `ContactsActivityViewModel` is by far the largest file in the project. Many of the issues that come with large classes with complex behavior are that they grow organically and are prone to bugs.  This large VM seems like a code smell, but upon closer inspection it is seen that all operations the VM performs are concise, encapsulated, and atomic.

There is a quote from the [architecture guidelines](https://developer.android.com/jetpack/guide/ui-layer/events#handle-viewmodel-events):

>Mapping UI actions to UI state is not always a simple process, but it does lead to simpler logic. [...]  In other words: don't think about what actions the UI needs to make; think about how those actions affect the UI state.

Each event that the VM handles is like its own little state machine, atomically inspecting the current state, deciding how the state should update from the event, and then transitioning into the new state.  This makes the VM naturally resistant to the problems that come with mutable shared state and many event entry points.

### Why not separate View Models for each component?

At first it seems natural to begin breaking up a monolith by encapsulating each component's behavior by its own View Model (e.g. something like a `ContactsListViewModel`); however, pursuing this path leads to its own set of problems stemming from flexible UI.

The device configuration is unknown to the View Model (as it should be), and thus the View Model has no idea how its state is going to be displayed to the user.  It is possible that every single component will be showing at the same time in the same window, and this is where the issues of per-component View Models arise.  Individual components in a flexible UI are _not_ decoupled, and because of this, using separate View Models is difficult to debug and extend.

For example, tapping a different contact from the list while the details pane has unsaved changes leads to a prompt for the user to discard their changes.  This interaction could not happen without some coupling between the list and details components, and this breaks the encapsulation benefits of using separate View Models.

It turns out it is simpler to assume that all components will be on the screen at the same time and design the View Model around that idea.


build.gradle
```diff
@@ -10,11 +6,23 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:7.1.1'
+    classpath 'com.android.tools.build:gradle:7.2.0-alpha07'
```
bhariharan, line 9: Do `7.1.2` - let's not ship with an `alpha` just yet.

build.gradle
```diff
@@ -10,11 +6,23 @@ buildscript {
...
     classpath 'io.github.gradle-nexus:publish-plugin:1.1.0'
   }
 }
 
+plugins {
+  id 'com.android.application' version '7.2.0-alpha07' apply false
```
bhariharan, line 15: Same here - `7.1.2`.

native/NativeSampleApps/MobileSyncCompose/.gitignore
```diff
@@ -0,0 +1,15 @@
+/build
```
bhariharan, line 1: You could just add this to the top level `.gitignore` if  you like - I think it's wildcarded anyway.

native/NativeSampleApps/MobileSyncCompose/src/main/java/com/salesforce/samples/mobilesynccompose/app/MobileSyncComposeApp.kt
```diff
@@ -0,0 +1,16 @@
+package com.salesforce.samples.mobilesynccompose.app
```
bhariharan, line 1: Add copyright in all the source files - you can copy them from elsewhere.

native/NativeSampleApps/MobileSyncCompose/src/main/java/com/salesforce/samples/mobilesynccompose/contacts/ContactsActivity.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    override fun onKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
+        return salesforceActivityDelegate.onKeyUp(keyCode, event) || super.onKeyUp(keyCode, event)
+    }
+
+    override fun onResume(client: RestClient?) {
+        vm.sync(syncDownOnly = true)
+    }
+
+    override fun onLogoutComplete() {
```
bhariharan, line 100: Remove - NO-OP.

settings.gradle
```diff
@@ -1,10 +1,18 @@
...
+    }
+}
 include 'libs:SalesforceAnalytics'
 include 'libs:SalesforceSDK'
 include 'libs:SmartStore'
 include 'libs:MobileSync'
 include 'libs:SalesforceHybrid'
 include 'libs:SalesforceReact'
 include 'native:NativeSampleApps:RestExplorer'
+include 'native:NativeSampleApps:MobileSyncCompose'
```
bhariharan, line 15: Could we rename this to `MobileSyncExplorerKotlin`?

bhariharan (changes requested):
First round of feedback. Also, could you rename the project to `MobileSyncExplorerKotlin`, instead of `MobileSyncCompose`?

native/NativeSampleApps/MobileSyncCompose/src/main/java/com/salesforce/samples/mobilesynccompose/contacts/ContactsActivity.kt
gkotula: Oh that's interesting, is that a policy of ours?  I find empty method bodies without explicitly commenting them as no-ops confusing.

settings.gradle
gkotula: `MobileSyncCompose` was the working title; it's no problem to change it 👍

gradle.properties
```diff
@@ -1,3 +1,17 @@
...
+# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+#
+# When configured, Gradle will run in incubating parallel mode.
+# This option should only be used with decoupled projects. More details, visit
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true
+#Fri Jan 14 14:58:47 MST 2022
 cdvMinSdkVersion=24
+cdvCompileSdkVersion=32
+org.gradle.jvmargs=-Xmx2048M -Dkotlin.daemon.jvm.options\="-Xmx2048M"
```
gkotula, line 16: I think this file was modified by the IDE and I would not be opposed to reverting it if people think it should be.

bhariharan: If you think you need those JVM args, then keep them.

gkotula: I think this makes the emulators more responsive, but it's dependent on the limitations of the machine.  I vaguely remember tweaking IDE memory settings toward the beginning of this endeavor, but I feel like this should be in a local config file that isn't committed.  I'm honestly leaning toward reverting this now that I'm combing through each modified file.

native/NativeSampleApps/MobileSyncCompose/src/main/java/com/salesforce/samples/mobilesynccompose/app/MobileSyncComposeApp.kt
gkotula: Would the notice need to also go in files generated from the Jetpack Compose "new project" template provided by Android Studio?  Several were not modified like the `Colors.kt`.

build.gradle
```diff
@@ -10,11 +6,23 @@ buildscript {
...
+
+  /* TODO: Re-evaluate the need to have this plugin declared at the top-level build.gradle.  With
+      this plugin only applied at the module-level, the Compose compiler fails with a very low level
+      Kotlin IR error, so we're keeping it here for now. gkotula 2022-01-13 */
+  id 'org.jetbrains.kotlin.android' version '1.6.10' apply false
```
gkotula, line 20: I just checked again today whether this needs to be at the top-level, and the build still fails if this is moved to the new template module's `build.gradle`.  So I guess it has to stay here for now.

native/NativeSampleApps/MobileSyncCompose/.gitignore
wmathurin: If we have it here, it will show up and the generated app also.

wmathurin: Never mind. I just realized the new app is in the main repo. 
If we want it available to customers, we need it to live in SalesforceMobileSDK-Templates.

wmathurin (changes requested):
If there are no changes to the SDK libraries, could you move this app to the SalesforceMobileSKD-Templates repo that way it would be one of the template customers could use as a starting point?

bhariharan:
Agreed - let's move this whole PR to the `Templates` repo.

gkotula:
> Agreed - let's move this whole PR to the `Templates` repo.

Sounds like a consensus!  Closing to re-target the PR.

PR #2294, opened 2022-03-31
Improve EventStoreManager.storeEvent() speed

wmathurin:
storeEvent() calls shouldStoreEvent() which calls getNumStoreEvents() which calls book.getAllKeys().size() which gets slower the more events are stored
Fix: keep a count in memory and only go to disk the first time around

Also added asserts validating getNumStoreEvents() in various cases (store / delete etc)


bhariharan (approved):
LGTM!

PR #2295, opened 2022-04-11
Removing un-necessary saveValue  in MemCachedKeyValueStore.getValue

wmathurin:


wmathurin:
The good news is that this class is not currently used by LSDK (LSDK has its own class that does the same thing).

PR #2297, opened 2022-04-11
Fix SalesforceSDK Tests.

brandonpage:


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManagerTest.java
```diff
@@ -86,6 +86,7 @@ public void testFailedBatchPublish() {
         assertEquals(3, manager.getEventStoreManager().getNumStoredEvents());
         assertEquals(1, TestPublisher.publishedEvents.size());
         assertEquals(2, TestPublisher.publishedEvents.get(0).length());
+        manager.getEventStoreManager().deleteAllEvents();
```
wmathurin, line 89: Should we do that in tear down?

brandonpage: I put it here because this test does not clean up after itself and it isn't needed elsewhere.  There are only two tests in the class though so I guess it doesn't make much difference.

wmathurin (approved):

bhariharan (approved):

PR #2299, opened 2022-04-13
Mobile Sync Kotlin Template

gkotula:
🚧  DRAFT 🚧 

There are a _lot_ of things to finish or clean up, so this PR should be treated as something to play around with while the rest of the TODOs are done.

> But wait, I thought you were doing TODOs this whole time?

Well, one TODO led to another and then I rewrote the whole thing (basically) 😅

Long story short, there were many complicating factors along the way and I spent a lot more time creating something that I think people could copy-paste into their projects and base entire app architecture from.

This template should be scalable, maintainable, feature-rich, and air-tight.  Some highlights are:

- Full abstraction of MobileSync via Repos
- Clean architecture of soup ELTs to concrete data models and back
- Total concurrency built in from the ground up at every layer
- Trend-setting how to build flexible UI with Jetpack Compose #NoFragments
- ... and more to come as I finish my writeups about my learnings 😁 

For now, play around with it, but know that it's buggy.


wmathurin:
I thought we agreed it should be a template rather than a sample app. Otherwise customers won't be able to use it as a starting point. See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2283#pullrequestreview-895483836

gkotula:
> I thought we agreed it should be a template rather than a sample app. Otherwise customers won't be able to use it as a starting point.

Ah you are correct, my mistake.  That slipped through the cracks.  Closing this and reopening against the template repo.

PR #2300, opened 2022-05-02
Force WAL checkpoint in getDatabaseSize.

brandonpage:
It stands to reason that if we have a public API to get DB size it should be accurate at the time of calling.  This also fixes a lot of Hybrid Tests.

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
@@ -241,6 +241,8 @@ public class SmartStore  {
      * Get database size
      */
     public int getDatabaseSize() {
+		// With WAL enabled we must force a WAL checkpoint if we want the actual DB file size.
+		getDatabase().query("PRAGMA wal_checkpoint(FULL);").moveToNext();
     	int size =  (int) (new File(getDatabase().getPath()).length()); // XXX That cast will be trouble if the file is more than 2GB
     	if (dbOpenHelper instanceof DBOpenHelper) {
     		size += ((DBOpenHelper) dbOpenHelper).getSizeOfDir(null);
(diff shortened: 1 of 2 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -241,6 +241,8 @@ public JSONObject getLastExplainQueryPlan() {
      * Get database size
      */
     public int getDatabaseSize() {
+		// With WAL enabled we must force a WAL checkpoint if we want the actual DB file size.
+		getDatabase().query("PRAGMA wal_checkpoint(FULL);").moveToNext();
     	int size =  (int) (new File(getDatabase().getPath()).length()); // XXX That cast will be trouble if the file is more than 2GB
```
brandonpage, line 246: Side note:  The 2GB file size limit was artificially imposed by Android back in the day but I believe it was lifted long ago to the real (FAT32) hard limit of 4GB.  Does anybody know when?  And is this actually a dangerous cast?

gkotula: `int` is signed, so it only goes up to +2^31 anyway.  While the cast may succeed, the size may be negative in those cases where it is bigger than ~2GB

gkotula:
From the docs, `PRAGMA wal_checkpoint(FULL)` blocks until there are no active DB writers and all readers are reading from the most recent snapshot.  Could that fact cause issues with consumers that expect this method to return quickly? 
 SmartStore currently synchronizes on the database instance for all operations anyway, so this may be a non-issue.

wmathurin (approved):

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
brandonpage: Whoops, I completely missed that. Thanks.

PR #2301, opened 2022-05-02
Removing completed syncs from active tasks before calling callback

wmathurin:
Otherwise, calling clean sync ghost in call back when done status detected will complain that sync is still running


wmathurin:
NB: all sync tests passing locally

gkotula (approved):
🎉

PR #2303, opened 2022-05-04
Lowering max ids per soql to avoid 414 uri too long error

wmathurin:
Also updated RefreshSyncDownTarget


libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/RefreshSyncDownTarget.java
```diff
@@ -57,49 +57,33 @@ public class RefreshSyncDownTarget extends SyncDownTarget {
     public static final String FIELDLIST = "fieldlist";
     public static final String SOBJECT_TYPE = "sobjectType";
     public static final String SOUP_NAME = "soupName";
-    public static final String COUNT_IDS_PER_SOQL = "coundIdsPerSoql";
+    public static final String COUNT_IDS_PER_SOQL = "countIdsPerSoql";
```
wmathurin, line 60: Fixing this typo means that existing refresh sync will run with the default count ids per soql after upgrade. Shouldn't be a big deal since few are using that sync down target in the wild and even fewer if any have overridden the count ids per soql.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/RefreshSyncDownTarget.java
```diff
@@ -57,49 +57,33 @@ public class RefreshSyncDownTarget extends SyncDownTarget {
...
     // NB: For each sync run - a fresh sync down target is created (by deserializing it from smartstore)
     // The following members are specific to a run
     // page will change during a run as we call start/continueFetch
     private boolean isResync = false;
     private int page = 0;
 
-    /**
-     * Return number of ids to pack in a single SOQL call
-     */
-    public int getCountIdsPerSoql() {
```
wmathurin, line 77: These methods were only there to support tests. Replacing them with a package level constructor where one can specify the count ids per soql.

bbirman (approved):

libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/manager/SyncManagerTest.java
```diff
@@ -702,140 +702,6 @@ public void testCleanResyncGhostsForSOSLTarget() throws Exception {
         deleteRecordsByIdOnServer(new HashSet<String>(Arrays.asList(accountIds[0])), Constants.ACCOUNT);
     }
 
-    /**
```
wmathurin, line 705: Moved refresh sync down target tests to their own class in the target sub package so that they can use the package level constructor which has a count ids per soql parameter.

wmathurin:
Again tests passing locally so merging.

PR #2304, opened 2022-05-04
Returning at most countIdsPerSoql records from startFetch / continueFetch

wmathurin:
To avoid too much memory usage.

If you have
```
typejA -> idA1 ... idAn
typeB -> idB1 ... idBn
etc
```
We are building a list that looks like `[idA1 ... idAn idB1 ... idBn ... ]`
And chunking that list, so if there are a lot of A's, we might do only A's in the first round, but if there are less than 500 A's we might do all the A's and some B's etc - but always no more than 500 records.

The new class `TypedIds` is a helper class that has methods for "chunking" and also for going back to the map of object type to list of ids.

### DIFF
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/BriefcaseSyncDownTarget.java
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/BriefcaseSyncDownTarget.java
@@ -181,12 +181,12 @@ public class BriefcaseSyncDownTarget extends SyncDownTarget {
 
     @Override
     public Set<String> getIdsToSkip(SyncManager syncManager, String soupName) throws JSONException {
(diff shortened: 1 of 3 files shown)


libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/BriefcaseSyncDownTarget.java
```diff
@@ -174,6 +180,16 @@ public int cleanGhosts(SyncManager syncManager, String soupName, long syncId) th
         return countGhosts;
     }
 
+    @Override
```
wmathurin, line 183: This is unrelated.
Without overriding this method a sync down with LEAVE_IF_CHANGED will not work correctly if the briefcase targets multiple object types.
(Need to add a test for that also).

bhariharan (approved):

PR #2305, opened 2022-05-06
Rest wrappers for collection create/update/upsert/retrieve and delete

wmathurin:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestRequest.java
```diff
@@ -836,6 +833,91 @@ public static RestRequest getRequestForPrimingRecords(String apiVersion, String
...
+	 * @param objectType    Type of the requested record.
+	 * @param objectIds     List of Salesforce IDs of the requested records.
+	 * @param fieldList     List of requested field names.
+	 *
+	 * @see <a href="https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_retrieve.htm">https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_retrieve.htm</a>
+	 */
+	public static RestRequest getRequestForCollectionRetrieve(String apiVersion, String objectType, List<String> objectIds, List<String> fieldList)
+		throws UnsupportedEncodingException, JSONException {
+		StringBuilder path = new StringBuilder(RestAction.SOBJECT_COLLECTION_RETRIEVE.getPath(apiVersion, objectType));
+		// Using a post body which is allowed by the end point and allows more ids to be sent up (2000 instead of ~800)
```
wmathurin, line 865: @bbirman we should use that instead of a "soql where ids in (...)" in briefcase sync down target. It allows for more ids and should be faster.

bhariharan (approved):

PR #2306, opened 2022-05-09
Using sObject collection retrieve in BriefcaseSyncDownTarget

wmathurin:
Using sObject collection retrieve instead of instead of SOQL where ids in xxx:
* we can get 2000 records per call since sObject retrieve uses POST body and therefore avoids 414 error URI too long


bhariharan (approved):

bbirman (approved):

wmathurin:
Tests passing locally.

PR #2315, opened 2022-05-23
First pass at fixing RestClientCallback for getRestClient()

gkotula:
Opening this for feedback on the solution for this issue.  Mainly, I am unsure how this may affect complex login flows including IDP and user switching in all customer use-cases.

The issue:  https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/247

Original PR that introduced the issue: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2203

The core problem is that launching the Login Activity from `Intent` does not get the callback for when the `RestClient` is ready, but we cannot use `AccountManager.addAccount(...)` since users may not have permission to add through the Android framework's `AccountManager`.

Luckily, `AccountManager` has an API for listening for changes to the accounts on the device, but care must be taken to properly remove the listener after the callback is invoked so as not to leak Activities.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -28,9 +28,8 @@ package com.salesforce.androidsdk.rest;
 
 import android.accounts.Account;
 import android.accounts.AccountManager;
-import android.accounts.AccountManagerCallback;
-import android.accounts.AccountManagerFuture;
 import android.accounts.NetworkErrorException;
+import android.accounts.OnAccountsUpdateListener;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -467,39 +515,6 @@ public void removeAccount(Account acc) {
         }
     }
 
-    /**
-     * Callback from either user account creation, or a call to getAuthToken, used
-     * by the Android account management components.
-     */
-    private class AccMgrCallback implements AccountManagerCallback<Bundle> {
```
gkotula, line 474: This was unused.

wmathurin (approved):

gkotula:
Closing due to edge case discovered while testing that would leak the account change listener if the LoginActivity is finished without modifying the accounts (e.g. if the user clicks back and the Activity pops from the task stack)

bhariharan:
@gkotula I'd prefer to leave the login code as-is for now. There are many different flows that need to be accounted for.

PR #2319, opened 2022-05-25
Fixing last error to be the same as in 10.0

wmathurin:


libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/BatchSyncUpTarget.java
```diff
@@ -232,6 +232,7 @@ protected RecordRequest buildRequestForRecord(JSONObject record,
     protected boolean updateRecordInLocalStore(SyncManager syncManager, String soupName, JSONObject record, SyncState.MergeMode mergeMode, Map<String, String> refIdToServerId, RecordResponse response, boolean isReRun) throws JSONException, IOException {
 
         boolean needReRun = false;
+        String lastError = response != null && response.errorJson != null ? response.errorJson.toString() : null;
```
brandonpage, line 235: NIT:  This would be a bit more readable with parentheses around the whole ternary condition.

brandonpage (approved):

PR #2328, opened 2022-06-20
Clean ghost tasks was never removed from active sync list

wmathurin:
As a result, running a sync down after running a sync down would throw the "Sync is running" error.


wmathurin:
Code is correct on iOS - see [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/MobileSync/MobileSync/Classes/Manager/SFCleanSyncGhostsTask.m#L63) and [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/MobileSync/MobileSync/Classes/Manager/SFCleanSyncGhostsTask.m#L69).

bhariharan (approved):

brandonpage (approved):

PR #2335, opened 2022-08-05
Add Screen Lock Timeout

brandonpage:


bhariharan (commented):
I thought we were going to remove timeout from the connected app interface instead? Are we sure we want to add this back, especially in a patch?

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/ScreenLockManager.java
```diff
@@ -137,18 +167,18 @@ public void cleanUp(UserAccount account) {
...
-
-    private boolean readMobilePolicy() {
+        long elapsedTime = System.currentTimeMillis() - lastLockedTimestamp;
         Context ctx = SalesforceSDKManager.getInstance().getAppContext();
         SharedPreferences sharedPrefs = ctx.getSharedPreferences(MOBILE_POLICY_PREF, Context.MODE_PRIVATE);
-        return sharedPrefs.getBoolean(SCREEN_LOCK, false);
+        boolean hasLock = sharedPrefs.getBoolean(SCREEN_LOCK, false);
+        int timeout = sharedPrefs.getInt(SCREEN_LOCK_TIMEOUT, 0);
+
+        return hasLock && (elapsedTime > timeout);
```
npars, line 181: Should this also be checking if `lastLockedTimestamp != 0`? `unlock()` sets lastLockedTimestamp to 0 but that would means that unlock would guarantee that `elapsedTime > timeout` would be true (unless someone set a timeout greater than 2022 years or something)

brandonpage: No.  I explicitly want this to return true when `lastLockedTimestamp` is 0 upon initial login.  If the connected app tells us the user requires the lock they should have to demonstrate it once to finish the login process.  `lastLockedTimestamp` should be captured on background any other time except cold boot, which should start locked.

npars: That makes sense to me, but I'm still unclear what the purpose of the `unlock()` function is. It looks like it will lock the app instead of unlocking, is that correct?

brandonpage: `unlock()` was a bad name for that method because it was only used to setup for the next `shouldLock` call _on unlock_.  I have updated the code path where this setup was needed and deprecated `unlock`.  Thanks for pointing this out.

bbirman (approved):

PR #2346, opened 2022-09-14
Deprecate getPushNotificationClientId.

brandonpage:
This is the only method that is for sure going away.  It seems unlikely that methods in `PushMessages.java` will change signature based on my initial testing.


bhariharan (approved):

wmathurin (approved):

brandonpage:
Screen Shot 2022-09-14 at 4 09 14 PM
I have never seen this before on Firebase 👀

PR #2360, opened 2022-09-23
Move to React Native 0.70.1

wmathurin:
Successfully ran `SalesforceReact` tests (but one flaps a lot).

### DIFF
--- a/libs/test/SalesforceReactTest/src/com/salesforce/androidsdk/reactnative/util/ReactNativeTestHost.java
+++ b/libs/test/SalesforceReactTest/src/com/salesforce/androidsdk/reactnative/util/ReactNativeTestHost.java
@@ -29,14 +29,10 @@ package com.salesforce.androidsdk.reactnative.util;
 
 import android.app.Application;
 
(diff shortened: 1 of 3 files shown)


libs/test/SalesforceReactTest/src/com/salesforce/androidsdk/reactnative/util/ReactNativeTestHost.java
```diff
@@ -87,27 +83,12 @@ protected ReactInstanceManager createReactInstanceManager() {
...
         return builder.build();
     }
-
-
-    /**
-     * Implementation of NativeModuleCallExceptionHandler used for testing (when using bundle)
-     *
-     * Marks the current test running is marked as failed if any javascript error takes place
-     */
-    static class NativeModuleCallExceptionTestHandler implements NativeModuleCallExceptionHandler {
```
wmathurin, line 106: That was only used when developing test to route the exception back to native land.
NativeModuleCallExceptionHandler no longer exists.

brandonpage (approved):

PR #2361, opened 2022-09-23
Fixes crash on upgrade

gkotula:
The core of this change is to move the `HttpAccess.init()` to before any upgrade steps since the upgrade steps may be fetching missing config data from the Org.

There was also a bug in determining which versions should perform the upgrade step, and so I whipped up a quick wrapper around our versioning system with easy-to-use comparison methods.  E.g.

```java
final String currentVersionStr = getInstalledAccMgrVersion();
if (SdkVersion.parseFromString(currentVersionStr)
        .isLessThan(new SdkVersion(9, 2, 0, /*isDev*/ false))
) {
    /* Upgrade from before 9.2.0 to 10.2.0 */
} else { ... }
```

`SdkVersion` is similar in usage to Kotlin data classes, but the generated methods must be included in the class definition instead of built transparently by the compiler.  The comparison semantics are declarative and easy to use 👍 

I have tested this to ensure that it fixes the bug that was crashing the upgrade steps, but I will continue to do testing across all the upgrade permutations to be absolutely sure this will work in the field.

### DIFF
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncSDKManager.java
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncSDKManager.java
@@ -64,9 +64,10 @@ public class MobileSyncSDKManager extends SmartStoreSDKManager {
     		INSTANCE = new MobileSyncSDKManager(context, mainActivity, loginActivity);
     	}
 
+		initInternal(context);
+
 		// Upgrade to the latest version.
 		MobileSyncUpgradeManager.getInstance().upgrade();
-		initInternal(context);
         EventsObservable.get().notifyEvent(EventType.AppCreateComplete);
 	}
 
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncSDKManager.java
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncSDKManager.java
@@ -64,11 +64,11 @@ public class MobileSyncSDKManager extends SmartStoreSDKManager {
     		INSTANCE = new MobileSyncSDKManager(context, mainActivity, loginActivity);
     	}
 
-		initInternal(context);
-
 		// Upgrade to the latest version.
 		MobileSyncUpgradeManager.getInstance().upgrade();
-        EventsObservable.get().notifyEvent(EventType.AppCreateComplete);
+
+		initInternal(context);
+		EventsObservable.get().notifyEvent(EventType.AppCreateComplete);
 	}
 
     /**
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
@@ -75,14 +75,7 @@ public class MobileSyncUpgradeManager extends SmartStoreUpgradeManager {
         // Update shared preference file to reflect the latest version.
         writeCurVersion(MOBILE_SYNC_KEY, MobileSyncSDKManager.SDK_VERSION);
 
-        try {
-            final SdkVersion installedVersion = SdkVersion.parseFromString(installedVersionStr);
-        } catch (Exception e) {
-            MobileSyncLogger.e(
-                    TAG,
-                    "Failed to parse installed version. Error message: " + e.getMessage()
-            );
-        }
+        // Compare SDK versions using SdkVersion class and add upgrade steps here as needed.
     }
 
     /**
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
@@ -26,6 +26,7 @@
  */
 package com.salesforce.androidsdk.mobilesync.app;
 
+import com.salesforce.androidsdk.app.SdkVersion;
 import com.salesforce.androidsdk.mobilesync.util.MobileSyncLogger;
 import com.salesforce.androidsdk.smartstore.app.SmartStoreUpgradeManager;
 
@@ -66,20 +67,21 @@ public class MobileSyncUpgradeManager extends SmartStoreUpgradeManager {
      * Upgrades mobile sync data from existing client version to the current version.
      */
     protected synchronized void upgradeSObject() {
-        final String installedVersion = getInstalledSobjectVersion();
-        if (installedVersion.equals(MobileSyncSDKManager.SDK_VERSION)) {
+        final String installedVersionStr = getInstalledSobjectVersion();
+        if (installedVersionStr.equals(MobileSyncSDKManager.SDK_VERSION)) {
             return;
         }
 
         // Update shared preference file to reflect the latest version.
         writeCurVersion(MOBILE_SYNC_KEY, MobileSyncSDKManager.SDK_VERSION);
 
-        // If the installed version < v8.2.0, we need to delete the old layout SmartStore file.
         try {
-            final String majorVersionNum = installedVersion.substring(0, 3);
-            double installedVerDouble = Double.parseDouble(majorVersionNum);
+            final SdkVersion installedVersion = SdkVersion.parseFromString(installedVersionStr);
         } catch (Exception e) {
-            MobileSyncLogger.e(TAG, "Failed to parse installed version.");
+            MobileSyncLogger.e(
+                    TAG,
+                    "Failed to parse installed version. Error message: " + e.getMessage()
+            );
         }
     }
 
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/app/SalesforceHybridSDKManager.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/app/SalesforceHybridSDKManager.java
@@ -83,10 +83,10 @@ public class SalesforceHybridSDKManager extends MobileSyncSDKManager {
     		INSTANCE = new SalesforceHybridSDKManager(context, mainActivity, loginActivity);
     	}
 
-        initInternal(context);
-
 		// Upgrade to the latest version.
 		SalesforceHybridUpgradeManager.getInstance().upgrade();
+        initInternal(context);
+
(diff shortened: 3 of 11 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -211,13 +206,12 @@ private void upgradeTo9Dot2() {
     }
 
     // TODO: Remove upgrade step in Mobile SDK 12.0
-    private void upgradeTo10Dot1Dot1() {
+    private void upgradeFromVersions9_2_0Thru10_1_1To10_1_1PasscodeFixes() {
```
gkotula, line 209: These names are cumbersome, but I was getting really confused by which method was doing what.  I think this more accurately describes what is happening, but it is only a cosmetic change.  I will not protest reverting these method names.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SdkVersion.java
```diff
@@ -0,0 +1,149 @@
...
+        return new SdkVersion(
+                Integer.parseInt(parts[0]),
+                Integer.parseInt(parts[1]),
+                Integer.parseInt(parts[2]),
+                parts.length == 4 // If we pass regex matching then we know the final part is "dev"
+        );
+    }
+
+    @Override
+    public int compareTo(SdkVersion o) {
```
gkotula, line 77: This method feels like it could be compacted, but this is actually the most optimal algorithm for comparing two versions.  First check the major version, then minor, etc.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
```diff
@@ -66,20 +67,21 @@ public void upgrade() {
...
+        final String installedVersionStr = getInstalledSobjectVersion();
+        if (installedVersionStr.equals(MobileSyncSDKManager.SDK_VERSION)) {
             return;
         }
 
         // Update shared preference file to reflect the latest version.
         writeCurVersion(MOBILE_SYNC_KEY, MobileSyncSDKManager.SDK_VERSION);
 
-        // If the installed version < v8.2.0, we need to delete the old layout SmartStore file.
         try {
```
wmathurin, line 78: What's the point of that try-catch? We are not doing anything with the calculated installedVersion.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncSDKManager.java
```diff
@@ -64,9 +64,10 @@ private static void init(Context context, Class<? extends Activity> mainActivity
     		INSTANCE = new MobileSyncSDKManager(context, mainActivity, loginActivity);
     	}
 
+		initInternal(context);
```
bhariharan, line 67: This call is meant to be after the upgrade step, by design. There's a lot of internal initialization that occurs there, and without running upgrade steps, it would lead to permanent corruption. For example, `SmartStore` DB opening calls. Let's move this back below.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/app/SalesforceHybridSDKManager.java
```diff
@@ -83,9 +83,10 @@ private static void init(Context context, Class<? extends Activity> mainActivity
     		INSTANCE = new SalesforceHybridSDKManager(context, mainActivity, loginActivity);
     	}
 
+        initInternal(context);
```
bhariharan, line 86: This call is meant to be after the upgrade step, by design. There's a lot of internal initialization that occurs there, and without running upgrade steps, it would lead to permanent corruption. For example, `SmartStore` DB opening calls. Let's move this back below.

libs/SalesforceReact/src/com/salesforce/androidsdk/reactnative/app/SalesforceReactSDKManager.java
```diff
@@ -72,9 +72,10 @@ private static void init(Context context, Class<? extends Activity> mainActivity
     		INSTANCE = new SalesforceReactSDKManager(context, mainActivity, loginActivity);
     	}
 
+		initInternal(context);
```
bhariharan, line 75: This call is meant to be after the upgrade step, by design. There's a lot of internal initialization that occurs there, and without running upgrade steps, it would lead to permanent corruption. For example, `SmartStore` DB opening calls. Let's move this back below.

bhariharan (changes requested):

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncSDKManager.java
gkotula: I looked at that code path and it didn't look like `initInternal()` did anything that would cause issues across all SDK manager classes.  If it really does need to be after the upgrade steps, we need to break the `HttpAccess.init()` out of `initInternal()`.  Do you foresee that causing issues if we have to revert this reordering?

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
gkotula: It was there before, but I could remove it if that's preferred.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SdkVersion.java
```diff
@@ -0,0 +1,149 @@
+package com.salesforce.androidsdk.app;
```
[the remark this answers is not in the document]
gkotula: Ah I always forget that, sorry!

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SdkVersion.java
```diff
@@ -0,0 +1,149 @@
...
+    @Override
+    public int compareTo(SdkVersion o) {
+        if (o == null) {
+            // non-null is before all null values in natural order
+            return -1;
+        }
+        if (o == this) { // reference compare
+            return 0;
+        }
+        if (this.major < o.major) {
```
wmathurin, line 85: Lines 85 .. 102 could be rewritten as (if `major` / `minor` / `patch` are changed to `Integer`'s)
```java
if (this.major != o.major) {
  return this.major.compareTo(o.major);
}
if (this.minor != o.minor) {
  return this.minor.compareTo(o.minor);
}
if (this.patch != o.patch) {
  return this.patch.compareTo(o.patch);
}
```

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/app/SmartStoreUpgradeManager.java
```diff
@@ -77,10 +78,12 @@ protected synchronized void upgradeSmartStore() {
 
         // If the installed version < v8.2.0, we need to migrate encryption keys.
         try {
-            final String majorVersionNum = installedVersion.substring(0, 3);
-            double installedVerDouble = Double.parseDouble(majorVersionNum);
+            final SdkVersion installedVersion = SdkVersion.parseFromString(installedVersionStr);
```
wmathurin, line 81: Same question as above, what's the point of that try-catch? We are not doing anything with the calculated installedVersion.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
gkotula: I make `SdkVersion` also throw `IllegalArgumentException` if the format of the version string is not recognized, so I think the pattern of doing the try-catch to parse the version is good.  But in these files where the try-catch is empty, it does not serve any purpose to keep them.  They were kept in there in previous versions, so I just left them this time too.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SdkVersion.java
gkotula: Ah I didn't even think about leveraging the `Integer` comparison methods!  Good catch, that makes this much cleaner.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.java
wmathurin: The `MobileSyncUpgradeManager`'s `upgrade` calls its super class `SmartStoreUpgradeManager`'s `upgrade` which calls its super class `SalesforceSDKUpgradeManager`'s `upgrade which does the same parsing in a try-catch. So I would say let's drop the empty try-catch in the subclasses that don't need to do anything with the calculated version.
We could even refactor the code a bit to do that parsing once, maybe make that installed version available in a member?

gkotula:
Okay, @bhariharan I just wound up putting the `HttpAccess.init()` in the upgrade method itself and moved the `initInternal()` calls back to after the upgrade steps.  Since upgrade steps usually don't hit the network, this seems like a decent middle ground for this weird upgrade.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -211,14 +206,14 @@ private void upgradeTo9Dot2() {
...
         final Context ctx = SalesforceSDKManager.getInstance().getAppContext();
         final SharedPreferences globalPrefs = ctx.getSharedPreferences(MOBILE_POLICY_PREF, Context.MODE_PRIVATE);
         if (globalPrefs.contains(SCREEN_LOCK)) {
             final UserAccountManager manager = SalesforceSDKManager.getInstance().getUserAccountManager();
             final List<UserAccount> accounts = manager.getAuthenticatedUsers();
-
             if (accounts != null) {
+                HttpAccess.init(ctx); // only needed because we need to hit the network for this upgrade step.
```
brandonpage, line 216: 👍  As discussed, this is probably the least disruptive solution.  It is a rare scenario that an upgrade step would require a network call.

brandonpage (approved):

bhariharan (approved):

PR #2370, opened 2022-12-10
Add Kotlin dependency.  Bump grade version and all other dependencies.

brandonpage:
This allows us to write Kotlin code in the SDK.  I have specified a Kotlin `apiVersion` and `languageVersion` of 1.6 so that is the _minimum_ version consumers must use.  This setting will throw errors if we use newer non-backwards compatible features within the SDK, but we are allowed to take advantage of the bugfixes and perf improvements associated with compiling against latest version of Kotlin (`1.7.21` in this case).  

~~TODO:  Catch the unhelpful error thrown by not specifying a Kotlin version in a pure Java project and re-throw with a message stating that Kotlin 1.6+ is required.~~

Edit: 
```
A problem occurred evaluating project ':libs:MobileSync'.
> Plugin with id 'org.jetbrains.kotlin.android' not found.
```
The above is the **Gradle** error that is thrown in a pure Java app.  AFAIK there is no way to catch this error so we should document it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,8 +10,9 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:7.2.1'
+    classpath 'com.android.tools.build:gradle:7.3.1'
     classpath 'io.github.gradle-nexus:publish-plugin:1.1.0'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.21"
   }
 }
 
@@ -31,5 +32,14 @@ allprojects {
     }
     google()
     mavenCentral()
+  }
+
+  // Ensure that we do not use newer language features that would make the SDK incompatible with
+  // apps that do not target the latest version of Kotlin.
+  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions {
+      apiVersion = "1.6"
+      languageVersion = "1.6"
+    }
   }
 }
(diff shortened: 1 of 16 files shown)


bhariharan (approved):

gkotula:
This is exciting!!

Quick clarification,

>I have specified a Kotlin `apiVersion` and `languageVersion` of 1.6 so that is the minimum version consumers must use.

That means we cannot use 1.7+ language features in the SDK itself, but if someone is using the 1.7 language features their app will compile fine?

brandonpage:
> This is exciting!!
> 
> Quick clarification,
> 
> > I have specified a Kotlin `apiVersion` and `languageVersion` of 1.6 so that is the minimum version consumers must use.
> 
> That means we cannot use 1.7+ language features in the SDK itself, but if someone is using the 1.7 language features their app will compile fine?

Yes, that is my understanding.  Kotlin supports this ["for three previous language and API versions in addition to the latest stable one"](https://kotlinlang.org/docs/kotlin-evolution.html#compatibility-flags).  And since new 1.X feature versions of Kotlin release every ~6 months I think we can treat this as a breaking chance and not update it again until Mobile SDK 12.0.

brandonpage:
> This is exciting!!
> 
> Quick clarification,
> 
> > I have specified a Kotlin `apiVersion` and `languageVersion` of 1.6 so that is the minimum version consumers must use.
> 
> That means we cannot use 1.7+ language features in the SDK itself, but if someone is using the 1.7 language features their app will compile fine?

Yes, that is my understanding.  Kotlin supports this ["for three previous language and API versions in addition to the latest stable one"](https://kotlinlang.org/docs/kotlin-evolution.html#compatibility-flags).  And since new 1.X feature versions of Kotlin release every ~6 months I think we can treat this as a breaking chance and not update it again until Mobile SDK 12.0.

wmathurin (approved):
👍

PR #2372, opened 2022-12-16
Update CustomServerUrlEditor to use Androidx DialogFragment. 

brandonpage:
Since the old and new `DialogFragment` are not API compatible it was easier to just rewrite `CustomServerUrlEditor` in Kotlin. 

I also made some minor UI tweaks around padding and rounding the corners to match the buttons:
Screen Shot 2022-12-15 at 3 50 52 PM

There is another deprecated `DialogFragment` instance in MobileSyncExplorer but to (easily) replace it would require a version of `SalesforceActivity` that extended `FragmentActivity` instead of `Activity`.


bhariharan (approved):
Here comes Kotlin! :-)

wmathurin (commented):
👍

PR #2380, opened 2023-01-18
Upgrade to Firebase Messaging 23.1.1

brandonpage:
This upgrade comes with a major breaking change since the way we previously initialized FCM has been removed.  Apps are now required to follow the recommended FCM setup of placing a `google-services.json` in the top level `app` directory; but continue to handle the rest of the setup and SFDC registration. This will be included in the updated documentation.  

Since the cleanup around this migration required touching a lot of files across the entire project I took the opportunity to get rid of _a lot_ of warnings (apologies reviewers).  

I will send corresponding PR's to the shared, CordovaPlugin and templates repos.


wmathurin (approved):
LGTM

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -383,12 +381,13 @@ static long getBackoff(Context context, UserAccount account) {
...
      * @param account User account.
      */
+    @SuppressWarnings("unused")
     static void setBackoff(Context context, long backoff, UserAccount account) {
         final SharedPreferences prefs = context.getSharedPreferences(getSharedPrefFile(account),
         		Context.MODE_PRIVATE);
         final Editor editor = prefs.edit();
         editor.putLong(BACKOFF, backoff);
-        editor.commit();
+        editor.apply();
```
bhariharan, line 390: I thought `apply()` was async (vs) `commit()` was sync, but I read more about it and it appears that `apply()` is probably more efficient since it's atomic to an in-memory cache, but async to disk. 👍🏻

bhariharan (commented):
👍🏻

PR #2382, opened 2023-01-27
Update dependencies.

brandonpage:


libs/SmartStore/build.gradle
```diff
@@ -10,14 +10,15 @@ apply plugin: 'org.jetbrains.kotlin.android'
 
 dependencies {
     api project(':libs:SalesforceSDK')
+    //noinspection GradleDependency -  Needs to line up with supported sqlcipher version.
     api 'androidx.sqlite:sqlite:2.0.1'
     api 'net.zetetic:android-database-sqlcipher:4.5.2'
-    implementation 'androidx.core:core-ktx:+'
```
brandonpage, line 15: **This is important.**  Not setting a version for `core-ktx` can cause duplicate headers in consuming apps.

libs/SmartStore/build.gradle
```diff
@@ -10,14 +10,15 @@ apply plugin: 'org.jetbrains.kotlin.android'
...
-    androidTestImplementation 'androidx.test:runner:1.4.0'
-    androidTestImplementation 'androidx.test:rules:1.4.0'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.3'
-    implementation 'com.google.android.material:material:1.4.0'
```
brandonpage, line 20: I thought there was some reason we were not updating material here but I couldn't find it in previous PRs/commits.  If there is some reason I will roll it back and comment so we know for the future.

bhariharan (approved):

PR #2384, opened 2023-02-06
Fix Firebase initialization.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -82,31 +80,34 @@ public class PushMessaging {
...
-                } catch (Exception e) {
-                    SalesforceSDKLogger.e(TAG, "Error during FCM registration", e);
-                }
-            });
-
-        } else {
-            registerSFDCPush(context, account);
+        boolean firebaseInitialized = initializeFirebaseIfNeeded(context);
+
+        if (firebaseInitialized) {
```
brandonpage, line 85: I chose not to just to `if (initializeFirebaseIfNeeded(context)` because it is not obvious that the return is true if firebase is initialized after the call, not if it need to be.  I did it this was in an attempt to not break public API that we hadn't deprecated.

If anyone feels strongly that the method should be renamed for better clarity or if there is a clean way to split it into two methods, let me know.

wmathurin: What happens if initializeFirebaseIfNeeded returns false? How will the caller no that we didn't do anything?

gkotula: I think changing a method return type is not binary-compatible, but if I had to guess I'd say that it is a very rare case that a consumer would just upgrade SDK binaries without rebuilding their app.  In all but that rare case, the change should be just fine.  I think keeping  the method name the same is perfectly fine since the return value is clearly documented, it is less migration for both us and the typical consumer, and this variable naming indeed helps the reader 👍

gkotula (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
gkotula: Oops, I made my last comment without seeing that Wolf had responded.

> How will the caller know that we didn't do anything?

I think that is taken care of with the JavaDoc comment.  We could simply add a little more detail to the method doc to be clearer about what a `false` return value means.  There is also the logger warning that comes with a `false` return value which is good enough in my opinion 🤷

brandonpage: `initializeFirebaseIfNeeded` logs a warning if it is unsuccessful. Google also logs (probably more specifically) but I didn't want to rely solely on that.  Is that what you mean?

wmathurin: But register does not return a boolean - so they can't programmatically react to the fact it failed, or is there another method to call to check if one is indeed registered?

brandonpage: AFAIK There isn't a way to check other than `FirebaseApp.getInstance()`... which throws if the app isn't initialized.  I think this is because initialization relies on the `google-services.json` file being present in the module which can't be checked at runtime.  

This is really an instance where the app is either setup correctly or it isn't.  Google throws because there is no scenario you should try to init if aren't setup but we have to try to init because we don't know if the app wants push or not.

wmathurin: Should we throw too, instead of logging the message "means that com.google.gms:google-services was not applied to your gradle project" - by failing fast, we might save devs some time debugging?

brandonpage: That would crash every app that doesn't want FCM push notifications.

brandonpage: Before we could simply check if they app had added their Firebase Project ID to the boot config and skip setup if they didn't.  We could achieve a similar thing by adding an API that flips a "yes I want to use push notifications" flag.  Or we just try to setup on login and log if it is unsuccessful (what the code above does).

wmathurin: Sounds good. We should probably tell Steven to document that part (check you log for error to see if you misconfigured).

wmathurin (approved):

PR #2394, opened 2023-04-10
Biometric Authentication

brandonpage:
- New connected app controlled "Biometric Authentication" lock screen that forces user re-authentication or biometric to unlock.  Auth token is not refreshed when the app is locked.
- Rewrite ScreenLockManager and implement BiometricAuthenticationManager as implementations of new AppLockManager class.
- SalesforceSDKManager returns ScreenLockManager and BiometricAuthenticationManager interfaces implementations instead internal classes.
- Biometric enrollment dialog is shown upon initial login.  Can be displayed again by the app.
- Native biometric unlock button shown by default but can be disabled by the app.

~TODO:~
- [x] Handle web scheme to prompt biometric.
- [x] Get accurate session timeout.
- [x] Revoke old sessions if user re-authenticates before session is expired.

<p float="left">
  
   
  
</p>


libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -47,7 +47,7 @@
             android:theme="@style/SalesforceSDK"
             android:launchMode="singleInstance"
             android:screenOrientation="portrait"
-            android:windowSoftInputMode="adjustResize"
+            android:windowSoftInputMode="adjustNothing"
```
brandonpage, line 50: This prevents the biometric login (or SP login) button from moving above they keyboard and blocking the UI if the user chooses to type into the webview.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -76,21 +95,30 @@
  *
  * The bulk of the work for this is actually managed by OAuthWebviewHelper class.
  */
-public class LoginActivity extends AccountAuthenticatorActivity
-        implements OAuthWebviewHelperEvents {
+public class LoginActivity extends FragmentActivity
```
brandonpage, line 98: `AccountAuthenticatorActivity` was deprecated and did not provide much functionality.  I have replicated everything it provided with the `accountAuthenticatorResult` and `mAccountAuthenticatorResponse` variables in this class.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.java
```diff
@@ -415,4 +473,109 @@ public void onReceive(Context context, Intent intent) {
             }
         }
     }
+
+    private void presentBiometric() {
```
brandonpage, line 477: The reset of the class is biometric code that is extremely similar to what is in `ScreenLockActivity` so I will consider creating a single biometric utility class they both use.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -455,7 +457,18 @@ protected class AuthWebViewClient extends WebViewClient {
...
+            // Remove the Biometric Login button from the Connected App allow denny screen.
+            if (url.contains("frontdoor.jsp")) {
+                final RelativeLayout parentView = (RelativeLayout) view.getParent();
+                if (parentView != null) {
+                    final Button button = parentView.findViewById(R.id.sf__bio_login_button);
+                    if (button != null) {
+                        button.setVisibility(View.INVISIBLE);
+                    }
+                }
+            }
```
brandonpage, line 469: There is likely a cleaner way to do this, will try to improve it.  Suggestions welcome.

wmathurin: I'll probably have to do something similar with the "Login with IDP" button.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -453,13 +462,31 @@ protected class AuthWebViewClient extends WebViewClient {
...
 
         @Override
         public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
             Uri uri = request.getUrl();
+
+            // Login webview embedded button has sent the signal to show the  biometric prompt.
+            if (uri.toString().equals(BIOMETRIC_PROMPT) && activity != null) {
+                ((LoginActivity) activity).presentBiometric();
+                return true;
+            }
```
brandonpage, line 488: This is untested, but _looks_ correct to me.

libs/SalesforceSDK/res/values/bootconfig.xml
```diff
@@ -7,4 +7,5 @@
         <item>api</item>
         <item>web</item>
     </string-array>
+    <string name="remoteAccessConsumerSecret"></string>
```
wmathurin, line 10: Hmm, I thought bundling the consumer secret in a mobile app was a no-no for security.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
wmathurin: So if you reload the web view from the menu, will the button reappear?

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/BiometricAuthenticationManager.kt
```diff
@@ -0,0 +1,149 @@
...
+            val options = SalesforceSDKManager.getInstance().loginOptions.asBundle()
+            val intent = Intent(ctx, SalesforceSDKManager.getInstance().loginActivityClass)
+            intent.addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP)
+            intent.addFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT)
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+            intent.putExtras(options)
+            ctx.startActivity(intent)
+            EventsObservable.get().notifyEvent(EventsObservable.EventType.AppLocked)
+
+            getAccountPrefs(user).edit().putString("token", user.refreshToken).apply()
```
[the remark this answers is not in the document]
brandonpage: Woah, this is something leftover from testing.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
brandonpage: `onPageFinished` gets called again if you reload from the menu.

wmathurin (approved):
LGTM

PR #2401, opened 2023-05-11
@W-10186872: Upgrade deprecated JobIntentService code

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🚀

This will migrate the SDK from Android's deprecated `JobIntentScheduler` to the new Android Jetpack [Background Tasks](https://developer.android.com/guide/background) library.

I've done quite a bit of testing to verify registration and de-registration, plus making sure the same callbacks are invoked if a custom `PushService` is supplied.  If there are other test cases out there though, do let me know.

Thanks!


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.kt
```diff
@@ -0,0 +1,82 @@
...
+ * This class is intended to be instantiated by the background tasks work
+ * manager.
+ *
+ * Use startActionPublish to enqueue an analytics publishing worker.
+ *
+ * @param context The Android context provided by the work manager
+ * @param workerParams The worker parameters provided by the work manager
+ * @see <a href='https://developer.android.com/guide/background'>Android Background Tasks</a>
+ */
+internal class AnalyticsPublishingWorker(
```
JohnsonEricAtSalesforce, line 49: @brandonpage, I'd be curious to get your early feedback on this draft Background Tasks `Worker` for analytics. This is the easier of the two objects have to migrate, so I started here.  Is this heading in the direction you imagined?

The app will run and fire this worker, even in this rough draft state.  I'm looking now for any enqueue options that we might want to consider beyond the defaults.

brandonpage: Sorry for the late response.  I took a look at the updated code and it looks very clean.  👍

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -489,7 +488,9 @@ private static ScheduledFuture createPublishHandler() {
 
             @Override
             public void run() {
-                AnalyticsPublisherService.startActionPublish(SalesforceSDKManager.getInstance().getAppContext());
+                AnalyticsPublishingWorker.Companion.startActionPublish(
+                        SalesforceSDKManager.getInstance().getAppContext()
+                );
             }
         };
         return scheduler.scheduleAtFixedRate(publishRunnable, 0, sPublishFrequencyInHours, TimeUnit.HOURS);
```
JohnsonEricAtSalesforce, line 496: @brandonpage - While I was really thinking this change through, doesn't it seem like this `ScheduledExecutorService` is potentially redundant if it's now running a background tasks `OneTimeWorkRequest`?  Since this wants to publish at periodic intervals, what if we replaced this with `PeriodicWorkRequest` at this level? Maybe I ought to take this change a little further.

Even just typing it out loud, I'm feeling that's what we want to do so we're not laying new scheduling code under the old one here 💡 I'll give it a whirl!

JohnsonEricAtSalesforce: That looks a lot better, even in the Java code since it doesn't have to worry about cancellation anymore.  See 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2401/commits/2ea90b39ec9aed414485805b7b67e0a72ddf94e9

libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -116,16 +116,5 @@
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
         </service>
-
-        <receiver android:name="com.salesforce.androidsdk.push.PushService$SFDCRegistrationRetryAlarmReceiver"
```
JohnsonEricAtSalesforce, line 120: All three of these declarations are no longer needed.  The alarm receiver was used by push service for the six days later registration retry, but now that's just a delay parameter when the work request is enqueued.

The services are also no longer needed since we're using Android background tasks.

libs/SalesforceSDK/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
     api 'com.google.firebase:firebase-messaging:23.1.1'
     api 'androidx.core:core:1.9.0'
     api 'androidx.browser:browser:1.4.0'
+    api 'androidx.work:work-runtime-ktx:2.8.1'
```
JohnsonEricAtSalesforce, line 17: A crux change is getting in the new Android background tasks library.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    override fun doWork() =
+
+        // Publish all stored analytics for the current user.
+        UserAccountManager
+            .getInstance()
+            .cachedCurrentUser
+            .let { userAccount ->
+
+                SalesforceAnalyticsManager
```
JohnsonEricAtSalesforce, line 69: This one line is the crux of what used to be in `AnalyticsPublisherService`.  I tested this interactively in the debugger to verify it fires at the same intervals as `dev`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublisherService.java
```diff
@@ -1,82 +0,0 @@
...
-/**
- * A service that publishes stored data when an intent is triggered.
- *
- * @author bhariharan
- *
- * @deprecated This class will be drastically altered or replaced in Mobile SDK 11.0 when the deprecated
- * {@link androidx.core.app.JobIntentService} base is replaced with a WorkManager Worker class.
- */
-@Deprecated
-public class AnalyticsPublisherService extends JobIntentService {
```
JohnsonEricAtSalesforce, line 45: See the new `AnalyticsPublishingWorker`, which replaces this entirely.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -53,11 +53,6 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Stream;
```
JohnsonEricAtSalesforce, line 60: Always nice to wave "goodbye!" to `java.util.concurrent` 👋🏻

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -76,7 +71,6 @@ public class SalesforceAnalyticsManager {
 
     private static Map<String, SalesforceAnalyticsManager> INSTANCES;
     private static boolean sPublishHandlerActive;
-    private static ScheduledFuture sScheduler;
```
JohnsonEricAtSalesforce, line 79: There are a few touch points in this source, but its great that Android background tasks support cancelling existing work before enqueuing the same item.  That alleviates this Java code from that task.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -483,15 +474,10 @@ private void resetAnalyticsPolicy() {
         e.commit();
     }
 
-    private static ScheduledFuture createPublishHandler() {
-        final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
JohnsonEricAtSalesforce, line 487: It also worked out well that we can eliminate the `JobIntentService` _and_ this `ScheduledExecutorService` with Android background tasks.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -58,17 +64,13 @@ public class PushMessaging {
 	// Public constants.
     public static final String UNREGISTERED_ATTEMPT_COMPLETE_EVENT = "com.salesforce.mobilesdk.c2dm.UNREGISTERED";
     public static final String UNREGISTERED_EVENT = "com.salesforce.mobilesdk.c2dm.ACTUAL_UNREGISTERED";
-    public static final String ACCOUNT_BUNDLE_KEY = "account_bundle";
```
JohnsonEricAtSalesforce, line 61: These constants are no longer needed.  I just noticed they are public.  Is that relevant?  Their use seemed really specific to user account push notification registration.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -191,9 +193,14 @@ public static String getAppNameForFirebase(Context context) {
...
-    public static void registerSFDCPush(Context context, UserAccount account) {
-        final Intent registrationIntent = new Intent(PushService.SFDC_REGISTRATION_RETRY_INTENT);
-        runPushService(context, account, registrationIntent);
+    public static void registerSFDCPush(
+            @NonNull Context context,
+            @Nullable UserAccount account) {
+        runPushService(
+                context,
+                account,
+                Register
```
JohnsonEricAtSalesforce, line 202: Note `Register` and `Deregister` are Kotlin enums.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.java
```diff
@@ -368,36 +388,6 @@ public static void clearRegistrationInfo(Context context, UserAccount account) {
...
 
-    /**
-     * Returns the last backoff time.
-     *
-     * @param context Context.
-     * @return Backoff time.
-     * @param account User account.
-     */
-    @SuppressWarnings("unused")
-    static long getBackoff(Context context, UserAccount account) {
```
JohnsonEricAtSalesforce, line 379: `backoff` was a package-level member.  I though it was part of the six-days-later push notifications registration retry.  However, that wasn't correct.  Then, I noticed it wasn't used.  Is that correct? I could restore it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushNotificationsRegistrationChangeWorker.kt
```diff
@@ -0,0 +1,107 @@
...
+    workerParams: WorkerParameters
+) : Worker(
+    context,
+    workerParams
+) {
+
+    override fun doWork(): Result {
+
+        // Fetch worker input data for registration action and user account.
+        val pushNotificationsRegistrationAction = PushNotificationsRegistrationAction.valueOf(
```
JohnsonEricAtSalesforce, line 62: The logic here is a Kotlin translation of the logic that was in the work method of push service.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushNotificationsRegistrationChangeWorker.kt
```diff
@@ -0,0 +1,107 @@
...
+        // Instantiate push notifications registrar...
+        val pushNotificationsRegistrar = getInstance().pushServiceType.newInstance()
+
+        // ...Determine scope of user accounts when...
+        when (userAccount) {
+
+            // ...The input data didn't provide a user account...
+            null ->
+                // ...Change push notification registration for all user accounts.
+                getInstance().userAccountManager.authenticatedUsers?.forEach { nextUserAccount ->
```
JohnsonEricAtSalesforce, line 78: The null/all users flow only seems applicable to the six-days-later push notifications registration retry.  I was able to verify that using a fifteen minute delay instead and a little logging.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -57,38 +61,24 @@
...
  * @author bhariharan
  * @author ktanna
- *
- * @deprecated This class will be drastically altered or replaced in Mobile SDK 11.0 when the deprecated
- * {@link androidx.core.app.JobIntentService} base is replaced with a WorkManager Worker class.
+ * @author Eric C. Johnson <Johnson.Eric@Salesforce.com>
  */
-@Deprecated
-public class PushService extends JobIntentService {
+public class PushService {
```
JohnsonEricAtSalesforce, line 75: A crux change is this type no longer extends the deprecated `JobIntentService`.  However, it's a public type and one existing users may have subclassed.  So, it lives on here in modified form with hopefully no or minimal breaking changes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -105,69 +95,42 @@ public class PushService extends JobIntentService {
...
-
-	@Override
-	protected void onHandleWork(Intent intent) {
-
-		/*
-		 * Grabs the extras from the intent, and determines based on the
-		 * bundle whether to perform the operation for all accounts or
-		 * just the specified account that's passed in.
-		 */
-		final Bundle bundle = intent.getBundleExtra(PushMessaging.ACCOUNT_BUNDLE_KEY);
```
JohnsonEricAtSalesforce, line 129: This is now in `PushNotificationsRegistrationChangeWorker`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -105,69 +95,42 @@ public class PushService extends JobIntentService {
...
+	/**
+	 * Enqueues a change to one or more user accounts' push notifications
+	 * registration as persistent work via Android background tasks.
+	 *
+	 * @param userAccount       The user account or null for all user accounts
+	 * @param action            The push notifications registration action
+	 * @param delayMilliseconds The amount of delay before the push registration
+	 *                          action is taken, such as in a retry scenario
+	 */
+	static void enqueuePushNotificationsRegistrationWork(
```
JohnsonEricAtSalesforce, line 107: This replaces `runIntentInService`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -105,69 +95,42 @@ public class PushService extends JobIntentService {
...
+		).setInitialDelay(
+				delayMilliseconds == null ? 0 : delayMilliseconds,
+				MILLISECONDS
+		).build();
+
+		workManager.enqueue(workRequest);
 	}
 
-	private void performRegistrationChange(boolean register, UserAccount userAccount) {
+	void performRegistrationChange(boolean register, UserAccount userAccount) {
```
JohnsonEricAtSalesforce, line 133: This moves to package level access to be invoked by `PushNotificationsRegistrationChangeWorker`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -179,25 +142,6 @@ private void performRegistrationChange(boolean register, UserAccount userAccount
 		}
 	}
 
-    private void scheduleSFDCRegistrationRetry(long when, UserAccount account) {
```
JohnsonEricAtSalesforce, line 182: This is replaced with a simple parameter in `enqueuePushNotificationsRegistrationWork`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -215,7 +159,19 @@ private void onRegistered(String registrationId, UserAccount account) {
     	} catch (Exception e) {
             SalesforceSDKLogger.e(TAG, "Error occurred during SFDC registration", e);
     	} finally {
-            scheduleSFDCRegistrationRetry(MILLISECONDS_IN_SIX_DAYS, null);
+			/*
```
JohnsonEricAtSalesforce, line 162: @brandonpage and I are wondering why this six-days-later registration retry exists.  While it's here though, I added some commentary to aid the next traveler in understanding it.

@wmathurin - Do you have any thoughts? Let me know if I can catch you up on my question.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -308,7 +265,6 @@ private String registerSFDCPushNotification(String registrationId, UserAccount a
             		}
             	} else if (res.getStatusCode() == HttpURLConnection.HTTP_NOT_FOUND) {
                     id = NOT_ENABLED;
-                    status = REGISTRATION_STATUS_FAILED;
```
JohnsonEricAtSalesforce, line 311: This is already initialized to this value.  Thanks, inspector!

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
```diff
@@ -397,27 +353,4 @@ private RestClient getRestClient(UserAccount account) {
     	}
     	return client;
     }
-
-    /**
-     * Broadcast receiver to retry SFDC push registration.
-     *
-     * @author ktanna
-     */
-    public static class SFDCRegistrationRetryAlarmReceiver extends BroadcastReceiver {
```
JohnsonEricAtSalesforce, line 406: This also goes to the dust bin courtesy of Android background tasks.

JohnsonEricAtSalesforce (commented):
Here's some code walk-through commentary 📣 Thanks for reading!

JohnsonEricAtSalesforce:
@brandonpage - After a busy day testing everything I could for push notifications, this is ready for final review with some code walkthrough comments.  Thanks for the `PushMessaging` conversion.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -0,0 +1,106 @@
...
+            publishHoursInterval: Long
+        ) = Builder(
+            AnalyticsPublishingWorker::class.java,
+            publishHoursInterval,
+            HOURS
+        ).build().also { publishAnalyticsPeriodicWorkRequest ->
+            runCatching {
+                getInstance(context)
+            }.getOrNull()?.enqueueUniquePeriodicWork(
+                "SalesforceAnalyticsPublishingPeriodicWork",
```
brandonpage, line 100: NIT:  Probably unnecessary but should this be a const on the companion in case it needs to get referenced again?

JohnsonEricAtSalesforce: I was on the verge of doing that, though for cancellation this provides the UUID of the unique job and that's probably what code outside this source would always use.  I made the new constant private to reflect that so cancellation is more likely to be done by id.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.java
brandonpage: @wmathurin Do we think any apps have extended this?  At a glance SFS never has and Salesforce app used to have an `EncryptedPushService` but they deleted it when we added that feature.  I doubt external apps have.  I am not sure there is much value in making the class internal though.

brandonpage (approved):
LGTM.

Make sure you run tests locally.  They aren't running in CI because this PR is from a branch on the repo and not a fork.

PR #2402, opened 2023-05-11
Login through web view can use "user agent flow" or "web server flow"

wmathurin:
- By default it will use "web server flow".
- By calling SalesforceSDKManager.getInstance().setUseWebServerAuthentication(false), an app can opt to use the (deprecated) user agent flow instead. 
- The selected flow is shown in the dev info screen.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
@@ -53,8 +53,7 @@ public class DevInfoActivity extends Activity {
         boolean isDarkTheme = SalesforceSDKManager.getInstance().isDarkTheme();
         setTheme(isDarkTheme ? R.style.SalesforceSDK_Dark : R.style.SalesforceSDK);
 
-        // Title / layout
-        getActionBar().setTitle(R.string.sf__dev_support_title);
+        // Layout
         setContentView(R.layout.sf__dev_info);
 
         // Setup list
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
@@ -31,6 +31,8 @@ import android.os.Bundle;
 import android.widget.ListView;
 import android.widget.SimpleAdapter;
 
+import androidx.appcompat.app.AppCompatActivity;
+
 import com.salesforce.androidsdk.R;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
```diff
@@ -53,8 +53,7 @@ protected void onCreate(Bundle savedInstanceState) {
         boolean isDarkTheme = SalesforceSDKManager.getInstance().isDarkTheme();
         setTheme(isDarkTheme ? R.style.SalesforceSDK_Dark : R.style.SalesforceSDK);
 
-        // Title / layout
-        getActionBar().setTitle(R.string.sf__dev_support_title);
```
wmathurin, line 57: getActionBar() throws a NPE after the recent code changes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -322,15 +322,18 @@ protected void showError(Exception exception) {
     public void loadLoginPage() {
         if (TextUtils.isEmpty(loginOptions.getJwt())) {
             loginOptions.setLoginUrl(getLoginUrl());
-            doLoadPage(false);
```
wmathurin, line 325: We don't need to pass a jwtFlow boolean, we can figure that out by seeing if there is a jwt token in the LoginOptions.

bbirman (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
JohnsonEricAtSalesforce: Can you use `getSupportActionBar()` instead?

JohnsonEricAtSalesforce (approved):
⭐️ Looks good from here.  Only one optional suggestion regarding replacing `setActionBar()` with `getSupportActionBar()` if the title is actually a requirement.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
wmathurin: I would have to make DevInfoActivity a AppCompatActivity first. Let me give it a go.

wmathurin: Done with https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2402/commits/c02cb39cbcec5ab1c9f00bedc7ae8e062ece5762

PR #2403, opened 2023-05-15
Using prompt=login during advanced authentication so that user must re-enter credentials

wmathurin:
Previously if a user did a logout followed by a login, they would go straight to the allow/deny screen. 

Changes were made initially in the internal fork of the Mobile SDK (see [here](https://git.soma.salesforce.com/Android/SalesforceMobileSDK-Android-Public/pull/283)). 

There is a new property shareBrowserSession (and a method setBrowserLoginEnabled) on SalesforceSDKManager to go back to the old behavior. 

That flag can also be set through the AuthConfig.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -620,7 +620,7 @@ public class SalesforceSDKManager implements LifecycleObserver {
     /**
      * Returns whether share browser session is enabled.
      *
-     * @return True - if share browser session is enabled, False - otherwise..
+     * @return True - if share browser session is enabled, False - otherwise.
      */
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -616,14 +617,24 @@ public synchronized void setUseWebServerAuthentication(boolean useWebServerAuthe
         this.useWebServerAuthentication = useWebServerAuthentication;
     }
 
+    /**
+     * Returns whether share browser session is enabled.
+     *
+     * @return True - if share browser session is enabled, False - otherwise..
```
JohnsonEricAtSalesforce, line 623: Optional - Is there an extra period at the end?

JohnsonEricAtSalesforce (approved):
⭐️ Looks good from here!

PR #2406, opened 2023-05-19
Removing external storage and soup spec in SmartStore

wmathurin:
As announced a year ago, we are removing external storage support.
Here is the doc mentioning the removal: https://developer.salesforce.com/docs/atlas.en-us.mobile_sdk.meta/mobile_sdk/offline_external_storage.htm


brandonpage (approved):

wmathurin:
Briefcase tests failed because the briefcase is setup for specific connected apps (and I just recently updated the TEST_CREDENTIALS for the project on CircleCI - but using a different connected app when logging in). 
I just got new test credentials for testandroid@gs0.mobilesdk.com for the connected app "Android Unit Tests", so now the tests should pass.

PR #2417, opened 2023-06-06
@W-13544970: [Android] [SDK Opt-In Parameter] Enable blocking of mobile app access to "salesforce integration" users in Mobile SDK

JohnsonEricAtSalesforce:
This follow-up to #2414 adds an opt-in for the Salesforce integration user authentication block which defaults to `false`.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
@@ -812,7 +812,7 @@ public class OAuthWebviewHelper implements KeyChainAliasCallback {
                     HttpAccess.DEFAULT, tr.idUrlWithInstance, tr.authToken);
 
                 // Request the authenticated user's information to determine if it is a Salesforce integration user.  This is a synchronous network request, so it must be performed here in the background stage.
-                isSalesforceIntegrationUser = fetchIsSalesforceIntegrationUser(tr);
+                isSalesforceIntegrationUser = SalesforceSDKManager.getInstance().shouldBlockSalesforceIntegrationUser() && fetchIsSalesforceIntegrationUser(tr);
             } catch (Exception e) {
                 backgroundException = e;
             }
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -674,7 +674,7 @@ protected void onPostExecute(OAuth2.TokenEndpointResponse tr) {
             final SalesforceSDKManager mgr = SalesforceSDKManager.getInstance();
 
             // Failure cases.
-            if (isSalesforceIntegrationUser) {
+            if (mgr.shouldBlockSalesforceIntegrationUser() && isSalesforceIntegrationUser) {
```
bbirman, line 677: is it possible to check `mgr.shouldBlockSalesforceIntegrationUser()` before making the extra API call to get if the user is an integration user?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -812,7 +812,7 @@ protected TokenEndpointResponse performRequest(TokenEndpointResponse tr) throws
                     HttpAccess.DEFAULT, tr.idUrlWithInstance, tr.authToken);
 
                 // Request the authenticated user's information to determine if it is a Salesforce integration user.  This is a synchronous network request, so it must be performed here in the background stage.
-                isSalesforceIntegrationUser = fetchIsSalesforceIntegrationUser(tr);
+                isSalesforceIntegrationUser = SalesforceSDKManager.getInstance().shouldBlockSalesforceIntegrationUser() && fetchIsSalesforceIntegrationUser(tr);
```
wmathurin, line 815: NIT could we rename the variable since it might not always represent whether the user is indeed an integration user.
Maybe call it `shouldBlockUser` ?
Then line 677 of `OauthWebViewHelper.java`, we could simply do `if(shouldBlockUser)`

JohnsonEricAtSalesforce: I was on the fence of doing just that earlier also.  Thanks for the encouragement to boost it into action.  How's this? 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2417/commits/5d802437ee7d965c70b8648d0a91dc60d286ebdd

wmathurin (approved):

brandonpage (approved):

bbirman (approved):

PR #2418, opened 2023-06-06
Not allowing arbitrary websites in the login web view

wmathurin:
Added helper method `fromBundleWithSafeLoginUrl` on `LoginOptions` that builds a `LoginOptions` from the provided bundle but checks the loginUrl. If it is not one of the login servers, then the selected login server is used instead.

This is to address W-11369931


JohnsonEricAtSalesforce (approved):
⭐️

PR #2421, opened 2023-06-08
Configurable authentication flow

wmathurin:
SalesforceSDKManager has a new flag: useHybridAuthentication (defaulting to true) that causes us to use:
- when true: hybrid_token (authorization user agent flow), hybrid_auth_code (code exchange during web server flow) and hybrid_refresh (refresh flow)
- when false: token (authorization user agent flow), authorization_code (code exchange) and refresh_token (refresh flow)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -653,6 +655,23 @@ public synchronized void setBrowserLoginEnabled(boolean browserLoginEnabled, boo
         }
     }
 
+    /**
+     * Returns whether hybrid authentication flow should be used
+     *
+     * @return True - if hybrid authentication flow should be used, False - otherwise.
+     */
+    public boolean shouldUseHybridAuthentication() {
```
bbirman, line 663: I like the name-- I'll match

bbirman (approved):

PR #2422, opened 2023-06-12
Addressing deprecations during build

wmathurin:
Moved package from AndroidManifest.xml to build.gradle to address
> package="com.salesforce.androidsdk" found in source AndroidManifest.xml: /home/circleci/SalesforceMobileSDK-Package/tmp20230608T035347.445/androidnativekotlin/mobile_sdk/SalesforceMobileSDK-Android/libs/SalesforceSDK/AndroidManifest.xml.
Setting the namespace via a source AndroidManifest.xml's package attribute is deprecated.
Please instead set the namespace (or testNamespace) in the module's build.gradle file, as described here: https://developer.android.com/studio/build/configure-app-module#set-namespace

Also addressed
> Addressing "Software Components will not be created automatically for Maven publishing from Android Gradle Plugin 8.0"


JohnsonEricAtSalesforce (approved):
⭐️

PR #2423, opened 2023-06-12
@W-12727938: Crash after login if first backgrounded. Android Debug build only.

JohnsonEricAtSalesforce:
🚀 _Ready for review._ ⭐️

This is an investigation into the stack track in [W-12727938](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001NgIPYYA3/view) and related the related [Github Issue](https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/317).  What I've seen consistently in my attempts to replicate the test is _slightly_ different.  My React Native app allows log in and out perfectly so long as the app is never paused while the login activity is displayed.  A pause can be triggered by returning to the Android launcher, for instance.

If the app is paused while the login activity is displayed, the `MainActivity` instance is destroyed which unregisters the broadcast intent receiver that login completion depends on.  On login, the app simply finishes with login activity finish and there's nothing to receive the broadcast intent and start the next activity.  The user simply drops to the launcher.  Maybe a white screen could happen if a custom app had logic wired in the various `onUserSwitched` methods but was missing the actual authentication event like this?

It seems we'll need a system level hint to keep the activity where the broadcast intent receivers are registered from being destroyed while the user is working with login activity, even on app pause.

I've used `startActivityForResult` in place of broadcast intents quite a bit in the past, which does accomplish this.  Here's one change in `ClientManager` that resolved the occurrence of the issue I was experiencing.

According to our discussion from draft status, this is a short-term solution for 11.0.  A more comprehensive solution will be crafted in a future release using the team's ideas.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -113,7 +113,7 @@ public class ClientManager {
             if (options != null) {
                 i.putExtras(options);
             }
-            activityContext.startActivity(i);
+            activityContext.startActivityForResult(i, 0);
         }
 
         // Account found
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -113,6 +113,22 @@ public class ClientManager {
             if (options != null) {
                 i.putExtras(options);
             }
+            /*
+             * Special Note: `LoginActivity` does not actually return a result.
+             * However, it does start broadcast intents that need to be received
+             * by the starting activity.  Since login activity is started in a
+             * new task, the starting activity would become available to be
+             * destroyed which unregisters its broadcast intent receivers.
+             *
+             * Using `startActivityForResult` starts a new task with the
+             * starting activity as the "base" intent with login activity as its
+             * "visible" sub-activity.  This keeps the starting activity from
+             * being eagerly destroyed and sets it as the activity to be started
+             * if the user returns the this task after it may have been fully
+             * destroyed due to memory pressure.
+             *
+             * TODO: This short term solution will be replaced in a future release.
+             */
             activityContext.startActivityForResult(i, 0);
         }
 


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -113,7 +113,7 @@ public void getRestClient(Activity activityContext, RestClientCallback restClien
             if (options != null) {
                 i.putExtras(options);
             }
-            activityContext.startActivity(i);
+            activityContext.startActivityForResult(i, 0);
```
JohnsonEricAtSalesforce, line 116: I'm not certain we'd merge it this way, but this one line allows me to pause the app at the login activity, resume and still be able to complete login since the system keeps the activity with the broadcast intent receivers while awaiting the fictitious result.

wmathurin: I'm okay with that approach for the bug fix (maybe with a big comment to explain the reason for using startActivityForResult). We can try and figure out if there is a more elegant approach post 11.0.
@brandonpage what do you think?

JohnsonEricAtSalesforce: That's a thought, since we could do this short term and maybe a heftier effort post 11.0.  For this short-term fix, I suspect this same change would need to be applied to the other locations where `LoginActivity` is started.

JohnsonEricAtSalesforce:
By the by, I re-read (for the zillionth time) the [Tasks and the back stack](https://developer.android.com/guide/components/activities/tasks-and-back-stack) documentation to be sure I really grok the current state of what's happening here and get a solution that's targeted to Android SDK best practice.  It might take a little longer to test it out, but stay tuned and I reckon we'll have something polished and worth the wait.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
brandonpage: This seems fine for an 11.0 bugfix.  Ditto on the explanation comment, maybe even with a todo.  Now that we see the underlying problem I think any other fix would be a bit too risky so close to release.

JohnsonEricAtSalesforce: Thanks for the feedback, all.  I did a lot of reading and testing today on emulators and my devices.  Before switching to `startActivityForResult`, the bug is really easy to consistently reproduce 100% of the time.  Afterwards, it seems to consistently behave as expected.

One test I did a lot was send the app to background and run enough memory-heavy video games (my kids!) to force the app to terminate and resume from the task's base intent when using the recent tasks list.  (Note: Minecraft is the memory-use winner).  The base intent when using `startActivityForResult` is able to create a new task and activity stack that will let the user log in.

I _did_ try to find another Android API that would create the same sub-activity relationship without the "result" semantic this technically doesn't use.  I couldn't find a better match though at the activity level.  I'll add a comment to explain this and we can look for a solution to the underlying issue in the next release.

JohnsonEricAtSalesforce:
For historical purposes, I realized this was related to a task @gkotula worked on for Trust not long ago.  See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2203/files

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
JohnsonEricAtSalesforce: I should mention: I _did_ look at other parts of the app that start login activity.  This one spot seems the one we can apply this work-around to at the moment within the MSDK code.  If anyone thinks there's another, let me know how to test it.

brandonpage (approved):

wmathurin (approved):

JohnsonEricAtSalesforce:
I created [[MSDK Android] Improved Authentication Related Events And Navigation](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001TtBCxYAN/view) as a follow-up.

PR #2427, opened 2023-06-21
@W-13610752: LoginActivity UI appears different in 11.0

JohnsonEricAtSalesforce:
🥁 _Ready for review_ 🎸

This updates the `SalesforceSDK` style so that our MSDK 11 updates using `AppCompatActivity` will theme exactly as they do in 10.2.0.  This applies to both light and dark modes plus all the activities in the sample apps.  Also, this ensures the theme is the same API 24 to current.

I tested this side-by-side with 10.2.0 in MobileSyncExplorer in login activity and all the other activities in the sample app.

### DIFF
--- a/libs/SalesforceSDK/res/values/sf__styles.xml
+++ b/libs/SalesforceSDK/res/values/sf__styles.xml
@@ -2,7 +2,9 @@
 
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <style name="SalesforceSDK" parent="Theme.AppCompat.DayNight">
+    <style name="SalesforceSDK" parent="Theme.AppCompat.Light.DarkActionBar">
+        <item name="colorPrimary">@color/sf__primary_color</item>
+        <item name="titleTextColor">@color/sf__secondary_color</item>
         <item name="android:actionBarStyle">@style/SalesforceSDKActionBar</item>
         <item name="android:colorPrimary">@color/sf__primary_color</item>
         <item name="android:colorPrimaryDark">@color/sf__primary_color</item>
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/AccountSwitcherActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/AccountSwitcherActivity.java
@@ -26,11 +26,12 @@
  */
 package com.salesforce.androidsdk.ui;
 
-import android.app.Activity;
 import android.os.Bundle;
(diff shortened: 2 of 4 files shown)


wmathurin (approved):

libs/SalesforceSDK/res/values/sf__styles.xml
```diff
@@ -2,7 +2,9 @@
 
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <style name="SalesforceSDK" parent="Theme.AppCompat.DayNight">
+    <style name="SalesforceSDK" parent="Theme.AppCompat.Light.DarkActionBar">
```
JohnsonEricAtSalesforce, line 5: Now that I really looked over the 10.2 theme for the app bar, it qualifies as "dark" since it always uses white text on blue, regardless of light or dark mode.

libs/SalesforceSDK/res/values/sf__styles.xml
```diff
@@ -2,7 +2,9 @@
...
+        <item name="colorPrimary">@color/sf__primary_color</item>
```
JohnsonEricAtSalesforce, line 6: AppCompat picks up these two colors from a different attribute set, so I is necessary to have both `colorPrimary` and `android:colorPrimary`.

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/AccountSwitcherActivity.java
```diff
@@ -46,7 +47,7 @@
  *
  * @author bhariharan
  */
-public class AccountSwitcherActivity extends Activity {
+public class AccountSwitcherActivity extends AppCompatActivity {
```
JohnsonEricAtSalesforce, line 50: I was doing another test and realized `AccountSwitcherActivity` also uses the `SalesforceSDK` style for its theme.  Without being an `AppCompatActivity`, it wouldn't display the AppCompat provided action bar.

This one seems to reparent without any other issues.

JohnsonEricAtSalesforce:
We might need to be on the lookout for other activities whose action bar behavior has changed where ever the `SalesforceSDK` style is used.  I'll take another look around before merging.

JohnsonEricAtSalesforce:
For testing records, I verified `KeyValueStoreInspectorActivity` manages its theme in code, overriding the manifest, and so it is themed in 11.0 as it is in 10.2.

PR #2431, opened 2023-06-28
New tests where we query soup elements with non-ascii characters

wmathurin:
Three new tests:
- non-ascii characters in string indexed field, 
- non-ascii characters in json1 indexed field, 
- non-ascii characters in non-indexed field.


JohnsonEricAtSalesforce (approved):
⭐️

brandonpage (approved):

PR #2432, opened 2023-06-28
@W-12610994: [Android] Update reference to SHA1PRNG algorithm

JohnsonEricAtSalesforce:
🥁 _Ready For Review_ 🎸

This replaces a few explicit references to the deprecated SHA1PRNG random number generation algorithm with APIs that allow the Android system to choose the recommended algorithm.  That should have this code compliant with the current recommendation.

I did some extra verification with both fresh installs of the app an upgrades from `dev` during log on, log out and all the unit tests.

Here's a quick compilation of related research.
👉🏻 https://android-developers.googleblog.com/2013/02/using-cryptography-to-store-credentials.html
👉🏻 https://android-developers.googleblog.com/2016/06/security-crypto-provider-deprecated-in.html
👉🏻 https://developer.android.com/guide/topics/security/cryptography#kotlin

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
@@ -41,7 +41,6 @@ import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
-import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.SecureRandom;
@@ -60,7 +59,6 @@ public class Encryptor {
     private static final String AES_CBC_CIPHER = "AES/CBC/PKCS5Padding";
     private static final String AES_GCM_CIPHER = "AES/GCM/NoPadding";
     private static final String MAC_TRANSFORMATION = "HmacSHA256";
-    private static final String SHA1PRNG = "SHA1PRNG";
     private static final String RSA_PKCS1 = "RSA/ECB/PKCS1Padding";
     private static final String BOUNCY_CASTLE = "BC";
     private static final int READ_BUFFER_LENGTH = 1024;
@@ -72,7 +70,7 @@ public class Encryptor {
      * @return Initialized cipher.
      */
     public static Cipher getEncryptingCipher(String encryptionKey)
-            throws NoSuchAlgorithmException, InvalidAlgorithmParameterException, InvalidKeyException {
+            throws InvalidAlgorithmParameterException, InvalidKeyException {
         final byte[] keyBytes = Base64.decode(encryptionKey, Base64.DEFAULT);
         return getEncryptingCipher(keyBytes, generateInitVector());
     }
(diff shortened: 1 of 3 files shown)


wmathurin (commented):
Do we have to do anything with already encrypted data?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -72,7 +70,7 @@ public class Encryptor {
      * @return Initialized cipher.
      */
     public static Cipher getEncryptingCipher(String encryptionKey)
-            throws NoSuchAlgorithmException, InvalidAlgorithmParameterException, InvalidKeyException {
```
JohnsonEricAtSalesforce, line 75: I traced through the code to find some now unused `NoSuchAlgorithmException` references.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -534,8 +532,9 @@ private static byte[] decryptWithPrivateKey(PrivateKey privateKey, String data,
         return null;
     }
 
-    private static byte[] generateInitVector() throws NoSuchAlgorithmException {
-        final SecureRandom random = SecureRandom.getInstance(SHA1PRNG);
+    private static byte[] generateInitVector() {
+        // Create the system recommended secure random number generator provider algorithm.
+        final SecureRandom random = new SecureRandom();
```
JohnsonEricAtSalesforce, line 537: The documentation I referenced in the description indicates it's preferred to use this constructor which allows `SecureRandom` to choose the recommended available algorithm.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -165,13 +163,9 @@ private synchronized static String generateUniqueId(String name, int length) {
...
             try {
-
-                // Uses SecureRandom to generate an AES-256 key.
-                final SecureRandom secureRandom = SecureRandom.getInstance(SHA1PRNG);
-
-                // SecureRandom does not require seeding. It's automatically seeded from system entropy.
+                // Create the key generator with its recommended secure random number generator provider algorithm.
                 final KeyGenerator keyGenerator = KeyGenerator.getInstance(AES);
-                keyGenerator.init(length, secureRandom);
+                keyGenerator.init(length);
```
JohnsonEricAtSalesforce, line 168: For `KeyGenerator`, the documentation in the description plus the Javadoc agreed it was recommended to let it have control over creation of the random number generator, its algorithm and seeding.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
brandonpage: Could be it was needed for an older Android API level (23 or 21 perhaps?).

brandonpage (approved):
Looks Great!

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
JohnsonEricAtSalesforce: Now that the underlying methods of ours are not specifying algorithms, this exception is no longer declared by the methods we're calling.

JohnsonEricAtSalesforce:
> Do we have to do anything with already encrypted data?

Existing keys remain as-is, so existing data will be decrypted as-is also.

PR #2435, opened 2023-07-06
@W-13660464: [THGO] Android MobileSDK 11.0 crashes on app launching

JohnsonEricAtSalesforce:
This is a follow-up to https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2401, where a dependency injection circular dependency risk was inadvertently introduced with `Context` access via `SalesforceSDKManager`.

https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2401/files#r1198388762

To work-around this issue for now, an additional level of asynchronous access is reintroduced.  A to-do is included to derive a means to the application `Context` that is safe from this dependency injection circular dependency.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
@@ -26,6 +26,8 @@
  */
 package com.salesforce.androidsdk.analytics;
 
+import static java.util.concurrent.TimeUnit.HOURS;
+
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Build;
@@ -53,6 +55,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledFuture;
 
 /**
  * This class contains APIs that can be used to interact with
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -474,10 +482,28 @@ private void resetAnalyticsPolicy() {
         e.commit();
     }
 
-    private static void recreateAnalyticsPublishPeriodicWorkRequest() {
-        AnalyticsPublishingWorker.Companion.reEnqueueAnalyticsPublishPeriodicWorkRequest(
-                SalesforceSDKManager.getInstance().getAppContext(),
-                sPublishFrequencyInHours
-        );
+    private static ScheduledFuture<?> recreateAnalyticsPublishPeriodicWorkRequest() {
```
JohnsonEricAtSalesforce, line 485: Overall and since this is a temporary work-around, I accurately restored this to its 10.2 state.  I did add a little Java syntax cleanup and inspector warning resolution to the modified code.  I also added a detailed to-do as well to explain why this is here and document some suggestions for the future.

JohnsonEricAtSalesforce:
For the moment, do not merge this pull request.  I'm doing additional research that might change the recommended solution.

JohnsonEricAtSalesforce:
Based on latest findings in https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001UlVSiYAN/view and today's discussion at MSDK Daily, there's a better solution to the root cause of this crash in the app code.  We'll close this pull request and follow-up with the app in GUS.

PR #2436, opened 2023-07-07
Switch to using Restricted Context in CircleCI.

brandonpage:


.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ aliases:
     command: |
       sudo npm i npm@latest -g
       sudo npm install -g shelljs@0.8.5
-      sudo npm install -g cordova@8.1.2
+      sudo npm install -g cordova
```
wmathurin, line 51: Should we do cordova@12.0.0 to match the version we explicitly support?

brandonpage: We can.  I just figured we should always try to use the latest so if it breaks we know right away.

wmathurin: :+1:
Somehow I thought we were enforcing the exact version of cordova to use. But we only have a "minVersion" that we check for the Cordova CLI - see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Package/blob/dev/shared/constants.js#L61).

wmathurin (approved):

PR #2446, opened 2023-08-09
@W-13673997: [MSDK Android] Improve Debug Menu Access In Android Emulator

JohnsonEricAtSalesforce:
🥁 _Ready for review!_ 🎸

  This change adds a new "Show Salesforce Mobile SDK Developer Support" local notification for MSDK apps running with the debug build configuration.  The benefits hopefully include more intuitive access to the developer support dialog plus an alternative to the difficult to access shake gesture.

  The shake gesture became more difficult to access since Android Emulator, as an embedded tool window of Android Studio, generally cannot receive general keyboard shortcuts.

  For testing, here's some of the logic I thought of.  More ideas are welcome!
- The notification is only displayed for debug builds when developer support is enabled
  -  Builds other than debug will not show the notification, regardless of developer support enablement, since the manifest will not include the post notification permissions
- The notification should be displayed only while an app is authenticated and in the foreground
- The code around the shake gesture hasn't changed, for those who are familiar with it
- The app will request post notifications on login, but will considerately not re-request this permission if it is declined
  - This is a general guide for permissions management from Google

For extra credit, I also enabled dark mode support for the developer support dialog.

Give it a try and let me know how it looks.  I've testing in native and hybrid apps so far between API 24 and 33.  React Native testing is in the works.

### DIFF
--- a/libs/SalesforceSDK/res/values/sf__strings.xml
+++ b/libs/SalesforceSDK/res/values/sf__strings.xml
@@ -96,6 +96,10 @@
 	<string name="sf__biometric_opt_in_deny">Use Password</string>
 	<string name="sf__biometric_signout_user">Signing out %s.</string>
 
-	<!-- Dev support -->
-	<string name="sf__dev_support_title">Mobile SDK Dev Support</string>
+	<!-- Developer Support -->
+	<string name="sf__dev_support_title">Mobile SDK Developer Support</string>
+	<string name="sf__notifications_local_show_dev_support_action_title">Show Developer Support</string>
+	<string name="sf__notifications_local_show_dev_support_content">Show Salesforce Mobile SDK developer support</string>
+	<string name="sf__notifications_local_show_dev_support_text">Tap to display Salesforce Mobile SDK developer support in the active app.</string>
+	<string name="sf__notifications_local_show_dev_support_title">Salesforce Mobile Developer Support</string>
 </resources>
--- a/libs/SalesforceSDK/res/values/strings.xml
+++ b/libs/SalesforceSDK/res/values/strings.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!--
-      These aren't prefixed as they're expected to be overriden   
-      Strings of the SDK that are used by SDK-based applications live in sf__strings.xml
+      These aren't prefixed as they're expected to be overridden Strings of the
+      SDK that are used by SDK-based applications live in sf__strings.xml
     -->
     <string name="app_name">SalesforceSDK</string>
     <string name="account_type">com.salesforce.androidsdk</string>
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -26,10 +26,15 @@
  */
 package com.salesforce.androidsdk.app;
 
+import static com.salesforce.androidsdk.R.style.SalesforceSDK_AlertDialog;
+import static com.salesforce.androidsdk.R.style.SalesforceSDK_AlertDialog_Dark;
+
 import android.accounts.Account;
 import android.accounts.AccountManager;
 import android.app.Activity;
 import android.app.AlertDialog;
+import android.app.Application;
+import android.app.Application.ActivityLifecycleCallbacks;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
@@ -40,6 +45,7 @@ import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.os.AsyncTask;
 import android.os.Build;
+import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.provider.Settings;
@@ -48,6 +54,7 @@ import android.view.View;
 import android.webkit.CookieManager;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.LifecycleObserver;
 import androidx.lifecycle.OnLifecycleEvent;
@@ -71,6 +78,8 @@ import com.salesforce.androidsdk.config.AdminSettingsManager;
 import com.salesforce.androidsdk.config.BootConfig;
 import com.salesforce.androidsdk.config.LoginServerManager;
 import com.salesforce.androidsdk.config.RuntimeConfig;
+import com.salesforce.androidsdk.developer.support.activities.ShowDeveloperSupportActivity;
+import com.salesforce.androidsdk.developer.support.notifications.local.ShowDeveloperSupportNotifier;
 import com.salesforce.androidsdk.push.PushMessaging;
 import com.salesforce.androidsdk.push.PushNotificationInterface;
 import com.salesforce.androidsdk.push.PushService;
@@ -152,6 +161,19 @@ public class SalesforceSDKManager implements LifecycleObserver {
     protected Context context;
     private LoginOptions loginOptions;
     private final Class<? extends Activity> mainActivityClass;
+
+    /**
+     * Null or an authenticated Activity for private use when developer support
+     * is enabled.
+     */
+    private Activity authenticatedActivityForDeveloperSupport = null;
+
+    /**
+     * Null or the Android Activity lifecycle callbacks object registered when
+     * developer support is enabled.
+     */
+    private ActivityLifecycleCallbacks activityLifecycleCallbacksForDeveloperSupport;
+
     private Class<? extends Activity> loginActivityClass = LoginActivity.class;
     private Class<? extends AccountSwitcherActivity> switcherActivityClass = AccountSwitcherActivity.class;
     private ScreenLockManager screenLockManager;
@@ -395,6 +417,9 @@ public class SalesforceSDKManager implements LifecycleObserver {
         if (!TextUtils.isEmpty(idpAppPackageName)) {
             INSTANCE.setIDPAppPackageName(idpAppPackageName);
         }
+
+        // Set up developer support.
+        setupDeveloperSupport(context);
     }
 
     /**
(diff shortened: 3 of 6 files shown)


JohnsonEricAtSalesforce:
Note this `11.1.0` feature shouldn't merge to upstream until after `11.0.1`.

libs/SalesforceSDK/res/values/sf__strings.xml
```diff
@@ -96,6 +96,10 @@
 	<string name="sf__biometric_opt_in_deny">Use Password</string>
 	<string name="sf__biometric_signout_user">Signing out %s.</string>
 
-	<!-- Dev support -->
-	<string name="sf__dev_support_title">Mobile SDK Dev Support</string>
+	<!-- Developer Support -->
+	<string name="sf__dev_support_title">Mobile SDK Developer Support</string>
```
JohnsonEricAtSalesforce, line 100: I placed these string resources here to match the location of the one existing "Developer Support" string.  Is this a match?  The content can change as we see fit.

libs/SalesforceSDK/res/values/strings.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!--
-      These aren't prefixed as they're expected to be overriden   
-      Strings of the SDK that are used by SDK-based applications live in sf__strings.xml
+      These aren't prefixed as they're expected to be overridden Strings of the
```
JohnsonEricAtSalesforce, line 4: This was a simple typo.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -152,6 +161,19 @@ public class SalesforceSDKManager implements LifecycleObserver {
     protected Context context;
     private LoginOptions loginOptions;
     private final Class<? extends Activity> mainActivityClass;
+
+    /**
+     * Null or an authenticated Activity for private use when developer support
+     * is enabled.
+     */
+    private Activity authenticatedActivityForDeveloperSupport = null;
```
JohnsonEricAtSalesforce, line 169: This is perhaps the most interesting part of the whole effort.  Requesting permissions requires an activity.  Responding to a notification action, in newer versions of Android, requires a call to `Activity.startActivity` even though we only want to show a dialog on the existing activity - anything else doesn't close the system drawer.  All the developer support dialog management is done here, in a singleton that only has an Application for context.

To integrate this with a smaller code footprint, this Activity variable keeps a reference to the top Activity post-login.  Further on, this is set by a utility method called from an Activity Lifecycle Callback.  That provides an Activity to request permissions.  Also, a new `ShowDeveloperSupportActivity` acts as a disposable trigger from the notification action.

I had a few other ideas to integration the permissions management and notification action, but they started to present API change to apps.  The approach here became the best black box integration in my trials.  Any other ideas?  I like the way this turned out, especially that it's entirely under-the-hood to apps.  I'd be game to evolve it in a different direction also.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -868,6 +893,113 @@ public void startSwitcherActivityIfRequired() {
...
+    /**
+     * Updates developer support features to match an Android Activity lifecycle
+     * event, including updating the app's authenticated Android Activity.
+     *
+     * @param authenticatedActivity The new authenticated Android Activity
+     *                              instance, or null
+     * @param lifecycleActivity     The Android Activity instance provided by
+     *                              the Activity lifecycle event
+     */
+    private void updateDeveloperSupportForActivityLifecycle(
```
JohnsonEricAtSalesforce, line 905: The display and hide of the developer support notification and dialog are handled here, which was a nice way to extract the logic from the Activity Lifecycle Callback object.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1218,32 +1350,37 @@ public ClientManager getClientManager(String jwt, String url) {
...
-                                        (dialog, which) -> {
-                                            devActionHandlers[which].onSelected();
-                                            devActionsDialog = null;
-                                        })
-                                .setOnCancelListener(dialog -> devActionsDialog = null)
-                                .setTitle(R.string.sf__dev_support_title)
-                                .create();
-                devActionsDialog.show();
-            }
+        if (devActionsDialog != null) devActionsDialog.dismiss();
```
JohnsonEricAtSalesforce, line 1363: This is the only new line in this block, which simply guards against showing more than one dialog at a time.  The rest is just a lambda conversion and whitespace.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1465,12 +1602,26 @@ protected void onAppBackgrounded() {
         ((ScreenLockManager) getScreenLockManager()).onAppBackgrounded();
         ((BiometricAuthenticationManager) getBiometricAuthenticationManager())
                 .onAppBackgrounded();
+
+        // Hide the Salesforce Mobile SDK "Show Developer Support" notification
+        if (authenticatedActivityForDeveloperSupport != null) {
```
JohnsonEricAtSalesforce, line 1607: I piggy-backed on this existing application lifecycle event to hide the notification when the app is delivered to the background.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -1465,12 +1602,26 @@ protected void onAppBackgrounded() {
...
     }
 
     @OnLifecycleEvent(Lifecycle.Event.ON_START)
     protected void onAppForegrounded() {
         ((ScreenLockManager) getScreenLockManager()).onAppForegrounded();
         ((BiometricAuthenticationManager) getBiometricAuthenticationManager())
                 .onAppForegrounded();
+
+        // Display the Salesforce Mobile SDK "Show Developer Support" notification
+        if (getUserAccountManager().getCurrentAccount() != null && authenticatedActivityForDeveloperSupport != null) {
```
JohnsonEricAtSalesforce, line 1621: I piggy-backed on this existing application lifecycle event to show the notification when the app is delivered to the foreground.  This includes an extra check to ensure the notification isn't shown without authentication.

libs/SalesforceSDK/src/com/salesforce/androidsdk/developer/support/activities/ShowDeveloperSupportActivity.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import androidx.appcompat.app.AppCompatActivity
+
+/**
+ * An Activity that, when started, finishes immediately and shows the developer
+ * support dialog from the previous Activity.  SalesforceSDKManager implements
+ * the logic for observing the Activity lifecycle and showing the developer
+ * support dialog.
+ */
+class ShowDeveloperSupportActivity : AppCompatActivity() {
```
JohnsonEricAtSalesforce, line 11: This Activity serves as the "action" for the new notification.  I had tried the old broadcast intent trick, but then remembered how newer API levels require a call to `Activity.startActivity` from notification actions.  Finishing this activity accomplishes the same integration in the `SalesforceSDKManager` Activity Lifecycle Callback.

libs/SalesforceSDK/src/com/salesforce/androidsdk/developer/support/notifications/local/ShowDeveloperSupportNotifier.kt
```diff
@@ -0,0 +1,204 @@
...
+import com.salesforce.androidsdk.R.string.sf__notifications_local_show_dev_support_text
+import com.salesforce.androidsdk.R.string.sf__notifications_local_show_dev_support_title
+import com.salesforce.androidsdk.developer.support.activities.ShowDeveloperSupportActivity
+import com.salesforce.androidsdk.developer.support.notifications.local.ShowDeveloperSupportNotifier.Companion.NotificationId.SHOW_DEVELOPER_SUPPORT
+
+/**
+ * Provides the "Show Developer Support" local notification, including
+ * notification permissions and set up.
+ */
+internal class ShowDeveloperSupportNotifier {
```
JohnsonEricAtSalesforce, line 37: This class is a container for the functions which manage the notifications and permissions.  Most of it should seem pretty boilerplate.

libs/SalesforceSDK/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,13 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!-- Debug-Only: Post notifications permission supports the show developer support notification. -->
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
JohnsonEricAtSalesforce, line 4: Here's everything that's only for AndroidManifest.xml in debug builds.

JohnsonEricAtSalesforce (commented):
Here's a quick code walkthrough!

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
JohnsonEricAtSalesforce: I should mention, the `context` variable on `SalesforceSDKManager` gets an inspector flag by Android Studio.  The flow of Context for Activity and Application can get some future polish in the modernization epic.

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -868,6 +893,113 @@ public void startSwitcherActivityIfRequired() {
...
+    private void updateDeveloperSupportForActivityLifecycle(
+            @Nullable Activity authenticatedActivity,
+            @NonNull Activity lifecycleActivity
+    ) {
+
+        // Show the developer support dialog, if applicable.
+        if (lifecycleActivity.isDestroyed() && lifecycleActivity instanceof ShowDeveloperSupportActivity) {
+            showDevSupportDialog(authenticatedActivityForDeveloperSupport);
+            return;
+        }
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: It is an interesting part of the flow since the developer support dialog can theoretically be shown by any activity in the app - but as a dialog attached to any activity not as a new activity.  "Actions" on the notification need to start an activity to trigger the dismissal of the system views that display the notification or have a broadcast intent attached to the content intent of the notification.  The content intent recently got an inspector warning since Material design prefers an activity intent attached to an action instead of using the notification's content intent - especially a broadcast intent.  It's not a deprecation, though, so I believe it's acceptable to suppress if the design is required.

On one hand, I did like that using a dedicated activity doesn't require suppressing the content intent inspector warning.  On the other, I agree it's counterintuitive.

I filed a follow-up we can use to investigate using the content intent and see if it looks better.   Let's take a look and see which option we elect.  Thanks for the thoughts!

[W-13985181: [MSDK Android] Remove Unneeded Show Developer Support Activity
](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001YxPZ9YAN/view)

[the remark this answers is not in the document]
JohnsonEricAtSalesforce: ...And here's the [follow-up code](https://github.com/JohnsonEricAtSalesforce/SalesforceMobileSDK-Android/pull/1#pullrequestreview-1587687813)

PR #2447, opened 2023-08-10
Optionally support use custom domain button in webview.

brandonpage:
- Opt-in regex pattern to specify domains.
- If a valid url is entered via the "Use Custom Domain" button on the login page store it in the server list.
- Open custom domain urls in Chrome (advanced auth) if appropriate.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -48,6 +48,7 @@ import android.view.View;
 import android.webkit.CookieManager;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.LifecycleObserver;
 import androidx.lifecycle.OnLifecycleEvent;
@@ -100,6 +101,7 @@ import java.util.Locale;
 import java.util.SortedSet;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.regex.Pattern;
 
 /**
  * This class serves as an interface to the various
@@ -182,7 +184,7 @@ public class SalesforceSDKManager implements LifecycleObserver {
 
     private boolean useHybridAuthentication = true; // hybrid authentication flows ON by default - but app can opt out by calling setUseHybridAuthentication(false)
 
-    private boolean shouldInferCustomDomain = false; // Do not detect use of Custom Domain input from login webview but app can opt in by calling setInferCustomDomain(ture)
+    private Pattern customDomainInferencePattern;
 
     private Theme theme =  Theme.SYSTEM_DEFAULT;
     private String appName;
@@ -676,20 +678,25 @@ public class SalesforceSDKManager implements LifecycleObserver {
     }
 
     /**
-     * Returns whether the SDK should infer if the user has entered a new login server through
-     * the "Use Custom Domain" button on the login screen.
+     * Returns the pattern used to detect the use of "Use Custom Domain" input from login web view.
+     *
+     * @return pattern if set or null
      */
-    public boolean shouldInferCustomDomain() {
-        return this.shouldInferCustomDomain;
+    public synchronized Pattern getCustomDomainInferencePattern() {
+        return customDomainInferencePattern;
     }
 
     /**
-     * Sets whether the SDK should infer if the user has entered a new login server through
-     * the "Use Custom Domain" button on the login screen.
-     * @param shouldInferCustomDomain
+     *  Detect use of "Use Custom Domain" input from login web view using the given regex.
+     *  Example for a specific org:
+     *    "^https:\\/\\/mobilesdk\\.my\\.salesforce\\.com\\/\\?startURL=%2Fsetup%2Fsecur%2FRemoteAccessAuthorizationPage\\.apexp"
+     *  For any my domain:
+     *    "^https:\\/\\/[a-zA-Z0-9]+\\.my\\.salesforce\\.com/\\?startURL=%2Fsetup%2Fsecur%2FRemoteAccessAuthorizationPage\\.apexp"
+     *
+     * @param pattern regex to use when detecting use of custom domain on login
      */
-    public synchronized void setShouldInferCustomDomain(boolean shouldInferCustomDomain) {
-        this.shouldInferCustomDomain = shouldInferCustomDomain;
+    public synchronized void setCustomDomainInferencePattern(@Nullable Pattern pattern) {
+        this.customDomainInferencePattern = pattern;
     }
 
     /**
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -182,7 +182,7 @@ public class SalesforceSDKManager implements LifecycleObserver {
 
     private boolean useHybridAuthentication = true; // hybrid authentication flows ON by default - but app can opt out by calling setUseHybridAuthentication(false)
 
-    private boolean shouldInferCustomDomain = true; // Detect use of Custom Domain input from login webview - but app  can opt out by calling setInferCustomDomain(false)
+    private boolean shouldInferCustomDomain = false; // Do not detect use of Custom Domain input from login webview but app can opt in by calling setInferCustomDomain(ture)
 
     private Theme theme =  Theme.SYSTEM_DEFAULT;
     private String appName;
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -181,6 +181,9 @@ public class SalesforceSDKManager implements LifecycleObserver {
     private boolean useWebServerAuthentication = true; // web server flow ON by default - but app can opt out by calling setUseWebServerAuthentication(false)
 
     private boolean useHybridAuthentication = true; // hybrid authentication flows ON by default - but app can opt out by calling setUseHybridAuthentication(false)
+
+    private boolean shouldInferCustomDomain = true; // Detect use of Custom Domain input from login webview - but app  can opt out by calling setInferCustomDomain(false)
+
     private Theme theme =  Theme.SYSTEM_DEFAULT;
     private String appName;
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -522,8 +551,10 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
                 return true;
             }
 
-			boolean isDone = uri.toString().replace("///", "/").toLowerCase(Locale.US).startsWith(loginOptions.getOauthCallbackUrl().replace("///", "/").toLowerCase(Locale.US));
-            if (isDone) {
+            String formattedUrl = uri.toString().replace("///", "/").toLowerCase(Locale.US);
+            String callbackUrl = loginOptions.getOauthCallbackUrl().replace("///", "/").toLowerCase(Locale.US);
+			boolean authFlowFinished = formattedUrl.startsWith(callbackUrl);
+            if (authFlowFinished) {
```
brandonpage, line 557: This should be no behavior change, just making it easier to read/understand.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -541,8 +569,37 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+                    host.endsWith(".force-user-content.com") ||
+                    host.endsWith(".salesforce-experience.com") ||
+                    host.endsWith(".salesforce-scrt.com")));
+        }
+
+        private boolean isNewLoginUrl(Uri uri) {
+            if (!uri.toString().contains(getLoginUrl()) && isSalesforceUrl(uri.getHost())) {
+                String path = uri.getQuery();
+                return path != null &&
+                        path.startsWith("startURL=/setup/secur/RemoteAccessAuthorizationPage.apexp?source=");
```
brandonpage, line 599: This condition may be too specific as it runs the chance of breaking if the url changes in the future.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -522,8 +524,34 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+                    // Checks the config for the selected login server
+                    (new AuthConfigTask(this)).execute();
+                } catch (Exception e) {
+                    SalesforceSDKLogger.e(TAG, "Unable to retrieve auth config.");
+                }
+            }
+
+            String formattedUrl = uri.toString().replace("///", "/").toLowerCase(Locale.US);
+            String callbackUrl = loginOptions.getOauthCallbackUrl().replace("///", "/").toLowerCase(Locale.US);
+			boolean authFlowFinished = formattedUrl.startsWith(callbackUrl);
```
wmathurin, line 553: NIT what's up with indentation here?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -541,8 +569,37 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
                         onAuthFlowComplete(tr);
                     }
                 }
+            } else {
+                // Only allow redirect for requests with authorization headers for Salesforce urls.
+                if (request.getRequestHeaders() != null && request.getRequestHeaders().containsKey("Authorization")) {
```
wmathurin, line 574: Why do we have that check?
We don't want to follow redirects blindly if we are adding bearer tokens to the request (during rest calls) to avoid leaking them out but why does it matter during login?

wmathurin (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -541,8 +569,37 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+                    host.endsWith(".site.com") ||
+                    host.endsWith(".lightning.com") ||
+                    host.endsWith(".salesforce-sites.com") ||
+                    host.endsWith(".force-user-content.com") ||
+                    host.endsWith(".salesforce-experience.com") ||
+                    host.endsWith(".salesforce-scrt.com")));
+        }
+
+        private boolean isNewLoginUrl(Uri uri) {
+            if (!uri.toString().contains(getLoginUrl()) && isSalesforceUrl(uri.getHost())) {
```
bbirman, line 596: I only added support for *.my.salesforce.com on iOS, do we want to open it up to more?

wmathurin: @bbirman but you have that logic in SFNetwork for regular rest calls not during login, right?

bbirman: The login one just checks for `.my.salesforce.com/?startURL=/setup/secur/RemoteAccessAuthorizationPage.apexp` (draft incoming), and then the other logic to check for Salesforce URLs and conditionally redirect is just in SFNetwork like you said

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
brandonpage: Github not interpreting a tab as 4 spaces?  Looks fine in the IDE but I'll fix it in my next commit.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
brandonpage: [iOS wasn't following any authorized redirects but changed to follow them if they are SFDC urls.](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3629/files)  Just trying to reach parity.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -181,6 +181,9 @@ public class SalesforceSDKManager implements LifecycleObserver {
     private boolean useWebServerAuthentication = true; // web server flow ON by default - but app can opt out by calling setUseWebServerAuthentication(false)
 
     private boolean useHybridAuthentication = true; // hybrid authentication flows ON by default - but app can opt out by calling setUseHybridAuthentication(false)
+
+    private boolean shouldInferCustomDomain = true; // Detect use of Custom Domain input from login webview - but app  can opt out by calling setInferCustomDomain(false)
```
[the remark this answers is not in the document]
brandonpage: I don't really care either way.  Is it going to be opt in now and opt out in 12.0 or opt in forever?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
brandonpage: Good callout, it should be more narrow.  Is that the only domain though?  What about community url (`*.force.com`)?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
wmathurin: @bbirman this [iOS code](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3629/files#diff-187e3eba81f3f8f772d3d913e610b04183eef011a4fa67e390eea88262571015) in SFNetwork only runs during REST calls not during login, or am I wrong?

wmathurin: In other word, shouldn't the logic above move to RestClient.java?

bbirman: Yup on iOS it only applies for the RestClient equivalent, no changes for the web view

brandonpage: 😳. That makes a lot of sense.  I will fix this.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
bbirman: I feel like the UI nudges towards just adding a prefix to .my.salesforce.com so I've been on the fence about it, do either of you have a preference? is the startURL portion the same for communities too?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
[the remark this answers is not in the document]
bbirman: good question... maybe opt in until we get more feedback and then reevaluate?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -542,7 +570,21 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
                 }
             }
-            return isDone;
+
+            return authFlowFinished;
+        }
+
+        private boolean isNewLoginUrl(Uri uri) {
+            String host = uri.getHost();
+            String path = uri.getQuery();
```
wmathurin, line 579: We should call it query.
We should also get the path and make sure it is empty based on @bbirman findings about SSO.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -525,7 +526,10 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
             }
 
             // Check if user entered a custom domain
-            if (SalesforceSDKManager.getInstance().shouldInferCustomDomain() && isNewLoginUrl(uri)) {
+            String host = uri.getHost();
+            Pattern customDomainPattern = SalesforceSDKManager.getInstance().getCustomDomainInferencePattern();
+            if (host != null && !getLoginUrl().contains(host) && customDomainPattern != null
```
wmathurin, line 531: Comparing with [iOS](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3636/files#diff-fc38a1c173394406330d16308ea5eee16aabbbe120e879cf10650d93b4ebe613R780), should we have a flag like `domainUpdated`, or is that achieved by setting the login server which causes `getLoginUrl().contains(host)` to be true if you are back here a second time?

wmathurin (approved):

wmathurin:
Maybe we should rename the PR from "Fully support use custom domain button in webview" to "Optionally support use custom domain in webview" ;-)

PR #2448, opened 2023-08-10
Allow app to specify a url to override oauth token endpoint.

brandonpage:
This is very likely too heavy handed of an approach and may need to be refined.


wmathurin (approved):
LGTM.
Maybe a new test in https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/config/BootConfigTest.java ?

PR #2457, opened 2023-09-08
@W-14093670: [MSDK Android] Flaky MobileSync Tests, Possibly Related To `ArrayBlockingQueue`

JohnsonEricAtSalesforce:
🌈 _*Ready For Review*_ 🛸

This effort aims to stabilize the flaky results of the MobileSync instrumented tests.

For summary, there were a few issues related to concurrency in `ArrayBlockingQueue` instances accessed by multiple threads plus an exception being thrown by the before and after processing in one test which didn't need the setup and teardown code after all.  Read on for more juicy detail, if desired.

Tests that utilized `SyncUpdateCallbackQueue` with multiple `SyncState` identifiers were failing because that queue collected `SyncState` instances from concurrent processes where the order across identifiers is not consistent.  The fix in the test is to collect those instances and segregate them by `SyncState` id so that the test can assert the correct processing of each sync independently from the other.

Enough of `SyncUpdateCallbackQueue` changed in Java that it benefitted from a Kotlin conversion as well.

Next, the various `BatchSyncUpTarget` constructor tests seemed to be failing for reasons unrelated to the code in the tests themselves.  My theory was that the `@Before` and `@After` processing that was being inherited by the test class might be causing the exceptions.  So, those got moved to a stand-alone test class and now seem to pass consistently.

For extra credit, `ArrayBlockingQueue` from `java.util.concurrent` has been replaced with `Channel` from Kotlin Coroutines.


libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/util/SyncUpdateCallbackQueue.kt
```diff
@@ -0,0 +1,98 @@
...
+    private var syncStatesById: Map<Long, Channel<SyncState>>
+
+    /**
+     * Initializes a new instance with pre-allocated queues for each provided
+     * SyncState id.
+     */
+    init {
+        check(syncStateIds.isNotEmpty()) { "At least one SyncState id must be provided." }
+
+        syncStatesById = syncStateIds.associateWith { Channel(UNLIMITED) }
```
JohnsonEricAtSalesforce, line 58: For some modernization icing on the cake, this replaces java.util's `ArrayBlockingQueue` with a Kotlin Coroutines `Channel`.

JohnsonEricAtSalesforce (commented):
Here's some detailed code walkthrough notes.  🥃

wmathurin (approved):
LGTM

PR #2459, opened 2023-09-21
Modernization of MobileSync library

wmathurin:
## Steps
1. **Convert** Mobile Sync library to Kotlin with Android Studio
2. **Clean up generated code**: e.g. **get rid of optionals** wherever null are not expected (even if the java code was not explicit about it), use **Kotlin constructs** like return if / ?: / with / data classes / object classes etc.
3. Make sure **MobileSync library** / **MobileSync tests** / **sample apps** (MobileSyncExplorer, MobileSyncExplorerHybrid) and **downstream libraries** (SalesforceHybrid and SalesforceReact) **compile** - **without having to make changes outside the MobileSync library**
4. Make sure **MobileSync tests pass**, **SalesforceHybrid** and **SalesforceReact mobile sync tests pass**
5. Make sure **mobile sync sample apps work**: MobileSyncExplorer and MobileSyncExplorerHybrid
6. Get rid of **NoteSync** hybrid sample app (simple hybrid app that syncs down notes using SOAP)
7. Add **coroutine wrappers** (get inspiration from MobileSyncExplorerKotlin)
8. Make sure **mobile sync template apps work**: MobileSyncExplorerKotlin and MobileSyncExplorerReactNative

## Notes
(1) See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/619a1cbee393b2976a2979c5aed0146258a35bf1
(3) Had to do a few small changes (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/f6e77b324ecf0d9fb558c1ad1a5f0acc3a1c86f0) - also had to keep MobileSyncMananger in java (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/5758ba9c68e14e2da2e9df9c80d66022dc5ef004)
(6) See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/a51392931d887b56c8bc44fa6375e5627a6e53a9
(7) See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/c3ec299df83d6f337fd57a36ce8473e4c3677d7f and https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/bfac5b14223f4f34395ac837f6d36eb5204bc1af
(8) See https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/371


wmathurin:
@brandonpage @JohnsonEricAtSalesforce As part of the modernization, I removed the NoteSync application (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/a51392931d887b56c8bc44fa6375e5627a6e53a9).
It was a simple hybrid application that synched down notes but at the time it was written there was no REST API for notes, so instead it uses a custom sync down target that uses SOAP.

JohnsonEricAtSalesforce:
@wmathurin - When this is ready for upstream `dev`, could we do a squash since it has a higher number of in-progress commits? That would simplify our commit history in `dev` afterwards.

JohnsonEricAtSalesforce:
Also, just a note it might take a moment longer to post a good review since there's plenty of changes.  Stay tuned 📻

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.kt
```diff
@@ -83,7 +61,29 @@ protected synchronized void upgradeSObject() {
...
+         *
+         * @return Instance of this class.
+         */
+        @get:Synchronized
+        val instance: MobileSyncUpgradeManager
+            get() {
+                if (INSTANCE == null) {
+                    INSTANCE = MobileSyncUpgradeManager()
+                }
+                return INSTANCE!!
```
JohnsonEricAtSalesforce, line 86: This works, though I'm always curious when I see a force unwrap and if control flow in Kotlin.  It looks like `INSTANCE` is private, so this could use a thread-safe lazy delegate like this, if you want.

```
val instance by lazy { MobileSyncUpgradeManager() }
```

I checked to see if it compiles that way, at least in MSDK's own workspace.

JohnsonEricAtSalesforce: I believe that lets us take the get:Synchronized away as well.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/app/MobileSyncUpgradeManager.kt
```diff
@@ -83,7 +61,29 @@ protected synchronized void upgradeSObject() {
...
-        return getInstalledVersion(MOBILE_SYNC_KEY);
+    val installedSobjectVersion: String
+        get() = getInstalledVersion(MOBILE_SYNC_KEY)
+
+    companion object {
+        /**
+         * Key in shared preference file for mobile sync version.
+         */
+        private const val MOBILE_SYNC_KEY = "mobile_sync_version"
+        private const val TAG = "MobileSyncUpgradeManager"
```
JohnsonEricAtSalesforce, line 72: Is `TAG` unused?  If there are inspector suggestions available that very well checked by the tool, like an unused private member, this might a good time to sweep those away.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/AdvancedSyncUpTask.kt
```diff
@@ -0,0 +1,110 @@
...
+                batch.add(record)
+            }
+
+            // Process batch if max batch size reached or at the end of dirtyRecordIds
+            if (batch.size == maxBatchSize || i == totalSize - 1) {
+                (target as AdvancedSyncUpTarget).syncUpRecords(
+                    syncManager,
+                    batch,
+                    options.fieldlist,
+                    options.mergeMode!!,
```
JohnsonEricAtSalesforce, line 78: I wasn't sure if this force-unwrap could be avoided.  It looks like this is a guard when deserializing `SyncOptions` from JSON in the event merge mode isn't specified.  Is there a reasonable default value? If it must be specified, could this operation or the creation of the sync skip this operation and log a warning?

wmathurin: I've been getting rid of force unwrap as much as possible. I'll look at this one.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/CollectionSyncUpTarget.kt
```diff
@@ -0,0 +1,139 @@
...
+        }
+        val recordIdToRemoteModDate = fetchLastModifiedDates(syncManager, nonLocallyCreatedRecords)
+        for (record in nonLocallyCreatedRecords) {
+            val storeId = record.getString(SmartStore.SOUP_ENTRY_ID)
+            val localModDate = RecordModDate(
+                JSONObjectHelper.optString(record, modificationDateFieldName),
+                isLocallyDeleted(record)
+            )
+            val remoteModDate = recordIdToRemoteModDate[storeId]
+            storeIdToNewerThanServer[storeId] = isNewerThanServer(localModDate, remoteModDate!!)
```
JohnsonEricAtSalesforce, line 130: I also wasn't sure about a way to resolve this force unwrap, but took a look.  If the remote modified date is somehow missing, is there a default action rather than let this throw? For instance, does that record actually need an update to cleanse the missing date value, so that value in the map should be true?  Just thinking of ideas since the force unwrap is the same as throwing an exception on null.  Perhaps, if this is a real guard, could this logic throw a more detailed exception than the default unwrap provided by Kotlin?

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/AdvancedSyncUpTask.kt
JohnsonEricAtSalesforce: Thanks.  If there's a logic alternative, great.  Perhaps if not, we move this to a guard with a clearer exception detail than the default provided by the force unwrap.  That could look something like `val mergeMode = options.mergeMode ?: throw IllegalArgumentException("Insert developer friendly message here.")`

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfig.kt
```diff
@@ -0,0 +1,158 @@
...
+import com.salesforce.androidsdk.util.ResourceReaderHelper
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Class encapsulating syncs definition.
+ *
+ * Config expected in a resource or assets file in JSON with the following:
+ * {
```
JohnsonEricAtSalesforce, line 48: Optional: I was just curious but Kdoc supports ```preformatted code``` using backticks just like markdown does.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfig.kt
```diff
@@ -0,0 +1,158 @@
...
+        } catch (e: JSONException) {
+            MobileSyncLogger.e(TAG, "Unhandled exception parsing json", e)
+        }
+    }
+
+    /**
+     * Return true if syncs are defined in config
+     * @return
+     */
+    fun hasSyncs(): Boolean {
```
JohnsonEricAtSalesforce, line 104: Optional: This could be a statement instead of a block.
```
fun hasSyncs() = syncConfigs?.let { it.length() > 0 } == true
```

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfig.kt
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Create the syncs from the config in the given store
+     * NB: only feedback is through the logs - the config is static so getting it right is something the developer should do while writing the app
+     * @param store
+     */
+    fun createSyncs(store: SmartStore) {
+        val syncConfigs = syncConfigs ?: return
+        val syncManager: SyncManager = SyncManager.getInstance(null, null, store)
```
JohnsonEricAtSalesforce, line 115: Optional: After the Kotlin conversion, a lot of explicit types can be removed, such as `SyncManager` here.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfig.kt
```diff
@@ -0,0 +1,158 @@
...
+                    val options: SyncOptions = SyncOptions.fromJSON(optionsJson)
+                    val soupName = syncConfig.getString(StoreConfig.SOUP_NAME)
+                    MobileSyncLogger.d(TAG, "Creating sync:$syncName")
+                    when (syncType) {
+                        SyncState.Type.syncDown -> {
+                            val target = SyncDownTarget.fromJSON(targetJson)
+                            syncManager.createSyncDown(target, options, soupName, syncName)
+                        }
+
+                        SyncState.Type.syncUp -> {
```
JohnsonEricAtSalesforce, line 138: Optional: Static imports for `syncUp` and `syncDown` could aid readability.  Are they part of our public API or could they be renamed with the conventional leading uppercase?

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/config/SyncsConfig.kt
wmathurin: Right now we only run / publish javadoc back to [github.io](https://forcedotcom.github.io/SalesforceMobileSDK-Android/index.html). But we should think about using kdoc and updating inlined documentation in the code accordingly. Maybe a new story / epic ?

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/AdvancedSyncUpTask.kt
```diff
@@ -0,0 +1,110 @@
...
+class AdvancedSyncUpTask(syncManager: SyncManager, sync: SyncState, callback: SyncUpdateCallback?) :
+    SyncUpTask(syncManager, sync, callback) {
+    @Throws(JSONException::class, IOException::class)
+    override fun syncUp(
+        sync: SyncState,
+        callback: SyncUpdateCallback?,
+        dirtyRecordIds: List<String>
+    ) {
+        val soupName = sync.soupName
+        val target = sync.target as SyncUpTarget
```
JohnsonEricAtSalesforce, line 52: Optional: The Kotlin converter does leave unchecked casts such as this one, which I lean towards using an optional cast on with `as?`.  Looking at this specific line, all the code downstream seems to accept `null`.  Food for thought, but no changes required.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/AdvancedSyncUpTask.kt
```diff
@@ -0,0 +1,110 @@
...
+        val options = sync.options
+        val totalSize = dirtyRecordIds.size
+        val maxBatchSize = (target as AdvancedSyncUpTarget).maxBatchSize
+        val batch: MutableList<JSONObject> = ArrayList()
```
JohnsonEricAtSalesforce, line 56: Optional: A lot of specific collection types can switch to Kotlin's collections, such as  `mutableListOf<JSONObject>()` in this case.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/AdvancedSyncUpTask.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        // Figuring out what records need to be synced up based on merge mode and last mod date on server
+        val recordIdToShouldSyncUp = shouldSyncUpRecords(syncManager, target, dirtyRecords, options)
+
+        // Syncing up records
+        for (i in 0 until totalSize) {
+            checkIfStopRequested()
+            val record = dirtyRecords[i]
+            val recordId = record.getString(SmartStore.SOUP_ENTRY_ID)
+            if (java.lang.Boolean.TRUE == recordIdToShouldSyncUp[recordId]) {
```
JohnsonEricAtSalesforce, line 68: This could just be `true`, I suspect.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/CleanSyncGhostsTask.kt
```diff
@@ -24,69 +24,73 @@
...
+        status: SyncState.Status,
+        progress: Int,
+        callback: SyncUpdateCallback?
+    ) {
         // Not a true sync
         // Leaving sync state alone
     }
 
-    @Override
-    protected void runSync() throws Exception {
```
JohnsonEricAtSalesforce, line 59: Is it significant that this is no longer open to subclass override?

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/LayoutSyncManager.kt
```diff
@@ -0,0 +1,335 @@
...
+         */
+        fun onSyncComplete(
+            objectAPIName: String?, formFactor: String?, layoutType: String?,
+            mode: String?, recordTypeId: String?, layout: Layout?
+        )
+    }
+
+    companion object {
+        const val SOUP_NAME = "sfdcLayouts"
+        const val QUERY = "SELECT {" + SOUP_NAME + ":_soup} FROM {" + SOUP_NAME +
```
JohnsonEricAtSalesforce, line 239: Optional: This could use a Kotlin string template.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/LayoutSyncManager.kt
```diff
@@ -0,0 +1,335 @@
...
+                "} WHERE {" + SOUP_NAME + ":" + Constants.ID + "} = '%s-%s-%s-%s-%s'"
+        private const val TAG = "LayoutSyncManager"
+        private val INDEX_SPECS = arrayOf(
+            IndexSpec(Constants.ID, SmartStore.Type.json1)
+        )
+        private val INSTANCES: MutableMap<String, LayoutSyncManager> = HashMap()
```
JohnsonEricAtSalesforce, line 245: No Change Required: I was curious to double check: Kotlin's `MutableMap` isn't thread safe like our old friend `HashMap`.  So this might not be a candidate for `mutableMapOf`.  So this probably needs to stay as-is.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/MetadataSyncManager.kt
```diff
@@ -0,0 +1,259 @@
...
+         * Callback triggered when metadata sync completes.
+         *
+         * @param metadata Metadata.
+         */
+        fun onSyncComplete(metadata: Metadata?)
+    }
+
+    companion object {
+        const val SOUP_NAME = "sfdcMetadata"
+        const val QUERY = "SELECT {" + SOUP_NAME + ":_soup} FROM {" + SOUP_NAME +
```
JohnsonEricAtSalesforce, line 164: Optional: Here's another string template conversion that'd be a quick win.  It might be worth looking across the whole module for them.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/MetadataSyncManager.kt
```diff
@@ -0,0 +1,259 @@
...
+            smartStore: SmartStore?
+        ): MetadataSyncManager {
+            val user = account ?: MobileSyncSDKManager.getInstance().userAccountManager.cachedCurrentUser
+            val store = smartStore ?: MobileSyncSDKManager.getInstance().getSmartStore(user, communityId)
+            val syncManager: SyncManager =
+                SyncManager.getInstance(user, communityId, store)
+            val uniqueId = ((if (user != null) user.userId else "") + ":"
+                    + store.database.path)
+            var instance = INSTANCES[uniqueId]
+            if (instance == null) {
```
JohnsonEricAtSalesforce, line 225: Optional: This could Elvis with the `instance` initialization above.  A quick regex search across the module reviewing if.*==null can find these.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/CleanSyncGhostsTask.kt
JohnsonEricAtSalesforce: Great question about changes to API or behavior that isn't part of MSDK's documented expectation. Thinking of other libraries, like Nimbus, there has been a significant impact to internal consumers when subtle things like this change. Many consumers also use the library in ways that weren't intended, but were heavily committed to it.

The compatibility and release question wasn't fully resolved with Nimbus. From personal experience, we mitigated in several ways that worked in the long run.

One, we did on a case by case basis either restore a changed API or require a consumer to adopt it. That required a lot of post-release discussion after Nimbus 5 for Android.

Two, there was a case where we patched the previous release of Nimbus 4 with a fix required by a consumer where their release was significantly impacted. The consumer later adopted the latest version.

So, I believe changing undocumented API or behavior that is internal and not part of MSDK's acceptance criteria in 11.1 is acceptable so we can move the library forward. We might just need to be willing to patch an 11.0.2 in a worst case scenario or a quick 11.1.1 to provide support for consumers. As you've been doing, we hold changing documented, public API until 12.x. Think that works?

Sorry, long answer based on Nimbus experiences 🏁

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncDownTarget.kt
```diff
@@ -0,0 +1,363 @@
...
+    @Throws(IOException::class, JSONException::class)
+    protected fun getChildrenRemoteIdsWithSoql(
+        syncManager: SyncManager,
+        soqlForChildrenRemoteIds: String
+    ): Set<String> {
+
+        // Makes network request and parses the response.
+        var records = startFetch(syncManager, soqlForChildrenRemoteIds)
+        val remoteChildrenIds: MutableSet<String> = HashSet(parseChildrenIdsFromResponse(records))
+        while (true) {
```
JohnsonEricAtSalesforce, line 190: @wmathurin - Is this intended to be `while (true)`?

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SoqlSyncDownTarget.kt
```diff
@@ -0,0 +1,256 @@
...
+    @Throws(IOException::class, JSONException::class)
+    protected fun getRemoteIdsWithSoql(
+        syncManager: SyncManager,
+        soqlForRemoteIds: String
+    ): Set<String> {
+        val remoteIds: MutableSet<String> = HashSet()
+
+        // Makes network request and parses the response.
+        remoteIds.addAll(parseIdsFromResponse(startFetch(syncManager, soqlForRemoteIds)))
+        while (true) {
```
JohnsonEricAtSalesforce, line 199: @wmathurin - Is this intended to be `while (true)`? The previous code had a check on `records != null`.

JohnsonEricAtSalesforce (changes requested):
@wmathurin - I'm still reviewing other changes, but I didn't realize Github wouldn't let me continue adding single comments for you to work on while I had started a review with more actionable items - My bad.  I didn't want you to have to wait for everything.  Here's what I have up to the moment with a few actionable items among optional discussion.

I tagged you on a couple actionable questions where the some while flows switched to a fixed `true` condition.  I wasn't sure that was intentional or the converter at work.

Thanks for the tremendous effort on this one.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncDownTarget.kt
```diff
@@ -0,0 +1,363 @@
...
+            // So we target parent-children 'goups' where the parent changed
+            // And we only download the changed children
+            childrenWhere.append(
+                buildModificationDateFilter(
+                    childrenInfo.modificationDateFieldName,
+                    maxTimeStamp
+                )
+            )
+            parentWhere.append(buildModificationDateFilter(modificationDateFieldName, maxTimeStamp))
+                .append(if (TextUtils.isEmpty(parentSoqlFilter)) "" else " and ")
```
JohnsonEricAtSalesforce, line 241: Optional: Another quick Kotlin update can be moving away from `TextUtils` in a lot of cases.  This could be `parentSoqlFilter.isNullOrEmpty()`.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/SyncManager.kt
```diff
@@ -0,0 +1,850 @@
...
+import com.salesforce.androidsdk.rest.ApiVersionStrings
+import com.salesforce.androidsdk.rest.RestClient
+import com.salesforce.androidsdk.rest.RestRequest
+import com.salesforce.androidsdk.rest.RestResponse
+import com.salesforce.androidsdk.smartstore.store.SmartStore
+import kotlinx.coroutines.NonCancellable
+import kotlinx.coroutines.withContext
+import org.json.JSONException
+import java.io.IOException
+import java.util.concurrent.Executors
```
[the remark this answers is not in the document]
wmathurin: I thought about that, but we would have to be sure it can still be called from Java.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/SyncDownTask.kt
```diff
@@ -24,91 +24,88 @@
...
-        if (mergeMode == SyncState.MergeMode.LEAVE_IF_CHANGED) {
-            idsToSkip = target.getIdsToSkip(syncManager, soupName);
+        var idsToSkip: Set<String>? = null
+        if (mergeMode == MergeMode.LEAVE_IF_CHANGED) {
+            idsToSkip = target.getIdsToSkip(syncManager, soupName)
         }
-
-        while (records != null) {
-            checkIfStopRequested();
+        while (true) {
```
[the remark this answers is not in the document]
wmathurin: Original code was doing:
```java
while (records != null) {
    (...)
    // Fetch next records, if any.
    records = target.continueFetch(syncManager);
}
```

Now the code does
```kotlin
while (true) {
    val recordsIn = records ?: break
    // Fetch next records, if any.
    records = target.continueFetch(syncManager)
}
```

But I guess the following is probably easier to read
```kotlin
while (records != null) {
    val recordsIn = records
    // Fetch next records, if any.
    records = target.continueFetch(syncManager)
}
```

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncDownTarget.kt
wmathurin: There is a break line 193 when records is null. I'll rewrite in a more readable way.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/CleanSyncGhostsTask.kt
wmathurin: 👍

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncUpTarget.kt
```diff
@@ -0,0 +1,606 @@
...
+        refIdToServerId: Map<String, String>,
+        response: RecordResponse?
+    ): Boolean {
+        var needReRun = false
+        val soupName = childrenInfo.soupName
+        val lastError = response?.errorJson?.toString()
+
+        // Delete case
+        if (isLocallyDeleted(record)) {
+            if (isLocallyCreated(record) // we didn't go to the sever
```
JohnsonEricAtSalesforce, line 322: Optional: `sever` 👉🏻 `server`

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncDownTarget.kt
wmathurin: Addressed with https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/4a71a36046f63c109ee9f836ec01999a4f543c86

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SoqlSyncDownTarget.kt
wmathurin: startFetch always return something not null while continue fetch can return null.
The full loop does:
``` kotlin
        while (true) {
            syncManager.checkAcceptingSyncs()

            // Fetch next records, if any.
            val records = continueFetch(syncManager) ?: break
            remoteIds.addAll(parseIdsFromResponse(records))
        }
```
Another Elvis-break - maybe not the most readable code?

wmathurin: Before it did
```java
while (records != null) {
            syncManager.checkAcceptingSyncs();

            // Fetch next records, if any.
            records = continueFetch(syncManager);
            remoteIds.addAll(parseIdsFromResponse(records));
        }
```

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncDownTarget.kt
```diff
@@ -0,0 +1,375 @@
...
+        return getLatestModificationTimeStamp(records, modificationDateFieldName)
+    }
+
+    /**
+     * When sync down fetches records from older to newer, the maxTimeStamp for the sync
+     * can be updated throughout the sync, and as a result running a paused (or killed) sync
+     * does not refetch all records.
+     * @return true if sync down is sorted by latest modification time stamp
+     */
+    open val isSyncDownSortedByLatestModification: Boolean
```
JohnsonEricAtSalesforce, line 247: Optional: It doesn't look like this needs a get accessor since it could simply `= false`.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SoqlSyncDownTarget.kt
wmathurin: I'll rewrite for clarity.

wmathurin: See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/985768f21d26146683e7f8efb342680e6b604c6d

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncUpTarget.kt
```diff
@@ -0,0 +1,573 @@
...
+
+    // Last sync error
+    protected var lastError: String? = null
+    /**
+     * Construct SyncUpTarget
+     */
+    /**
+     * Construct SyncUpTarget
+     */
+    /**
```
JohnsonEricAtSalesforce, line 89: Optional: It looks like there's two extra comment blocks here.

wmathurin: Java -> Kotlin conversion combined multiple method into a single method with default parameter values (and @JvmOverloads annotations). But it kept all the java doc comments ;-) Will cleanup.

wmathurin: See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/667e0fb3b229c4e525a5f6b5a96e29b8357d9ecc

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncDownTarget.kt
wmathurin: Fixed with https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/89de87833440f6973962d29e930ee88a909ec61b

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/manager/AdvancedSyncUpTask.kt
wmathurin: Good point - replaced unsafe cast:
-        val target = sync.target as SyncUpTarget
+        val target = sync.target as? SyncUpTarget ?: throw MobileSyncException("A SyncUpTarget was expected")

See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/ddf0bca32ce3dcde823c485a4753dbfe74b5adfa

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/util/SyncOptions.kt
```diff
@@ -0,0 +1,105 @@
...
+        const val FIELDLIST = "fieldlist"
+
+        /**
+         * Build SyncOptions from json
+         * @param options as json
+         * @return
+         * @throws JSONException
+         */
+        @JvmStatic
+		@Throws(JSONException::class)
```
JohnsonEricAtSalesforce, line 70: Optional: This source could use a quick run of the formatter.

wmathurin: Good idea. Done in https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2459/commits/cd6d5161f823fb43048b15898db2c95bc427d187

JohnsonEricAtSalesforce (approved):
Thanks, @wmathurin, for all the really heavy lifting here.  Reading through the results of the Kotlin converter in all-day-long detail, it seems it did a faithful job with our Java.  Plus, you've captured a lot of Kotlin-style bonus points.

This looks good-to-go, though we'll want to give everything a solid regression test and be prepared for a few subtle API or runtime behavior changes on release.

wmathurin:
Thanks @JohnsonEricAtSalesforce 
For testing I ran our various apps: 

- MobileSyncExplorer (java), 
- MobileSyncExplorerHybrid (java + cordova), 
- MobileSyncExplorerKotlin (kotlin) 

I also ran the:
- native java tests, 
- hybrid tests 
- react native tests. 

I also added some new Kotlin tests for the new corountine wrappers methods.

I did not have to change any of the "downstream" code: 
- dependent libraries SalesforceHybrid and SalesforceReact, 
- sample apps
- tests 

(Except for some imports in the kotlin test see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/f6e77b324ecf0d9fb558c1ad1a5f0acc3a1c86f0).

PR #2461, opened 2023-09-27
Prevent ANR and deprecate cause.

brandonpage:


JohnsonEricAtSalesforce (approved):
⭐️

brandonpage:
The test failure is legitimate but unrelated to the changes in this PR.  I will fix that is a separate PR after merging this.

wmathurin (commented):
Will 5s be enough? Should we provide a way for the app to choose the timeout value?

brandonpage:
@wmathurin  [5 seconds is the longest timeout we can use to prevent ANR's](https://developer.android.com/topic/performance/vitals/anr#:~:text=An%20ANR%20is%20triggered%20for,onCreate()%20and%20Service).

PR #2463, opened 2023-09-28
Support Predictive System Back.

brandonpage:
I tried to implement this about a year ago when it was first announced.  At that time, users could opt-in via a setting in the developer options of their API 33 (Android 13) device and enable it with a manifest option to test their apps, but the feature wouldn't be forced on for all apps until API 34 (Android 14).  I was able to work around the issue it caused for our Screen Lock (a.k.a passcode) feature but not the lack of support from Cordova and React Native.  Because in app back navigation completely broke for those skews we deferred support until API 34.  

Since then, Google has changed the strategy somewhat.  As of the last Android 14 beta the visual impact of the feature still cannot be seen without the user opting in through developer options (maybe it will be in the final release or pixel exclusive for a year? who knows).  **However**, even without the visual goodness the side effects of breaking functionality still remain if enabled!  Google has also carried the API 33 opt-in manifest setting forward and make the feature opt-in (instead of forced on) in API 34.  In addition, the feature can now be toggled on or off per activity which unblocks the issues we faced last year.  

TL;DR:  Google made it opt-in and configurable per activity.  I fixed the issues it caused for Screen Lock + Biometric Auth and disabled it for Cordova + React Native apps.

Testing Matrix:
| | API 24-32 | API 33 | API 34 |
| :--- | :---: | :---: | :---: |
| Native |  :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Screen Lock | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Biometric Unlock | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Cordova | :white_check_mark: | :x: * | :white_check_mark: |
| React Native | :white_check_mark: | :x: * |  |

* It appears Google has not fixed the issue with API 33 that we experienced last year (in app back navigation is broken).  I have added an example of setting the manifest opt-in value conditionally based on API level in MobileSyncExplorerHybrid.

### DIFF
--- a/libs/SalesforceHybrid/AndroidManifest.xml
+++ b/libs/SalesforceHybrid/AndroidManifest.xml
@@ -4,5 +4,8 @@
     android:versionCode="82"
     android:versionName="11.1.0.dev">
 
-    <application />
+    <application>
+        <activity android:name="com.salesforce.androidsdk.phonegap.ui.SalesforceDroidGapActivity"
+            android:enableOnBackInvokedCallback="false" />
+    </application>
 </manifest>
--- a/libs/SalesforceReact/AndroidManifest.xml
+++ b/libs/SalesforceReact/AndroidManifest.xml
@@ -3,6 +3,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     android:versionCode="82"
     android:versionName="11.1.0.dev">
-
-    <application />
+    
+    <application>
+        <activity android:name=".ui.SalesforceReactActivity"
+            android:enableOnBackInvokedCallback="false" />
+    </application>
 </manifest>
--- a/native/NativeSampleApps/MobileSyncExplorer/AndroidManifest.xml
+++ b/native/NativeSampleApps/MobileSyncExplorer/AndroidManifest.xml
@@ -8,7 +8,8 @@
 	<application android:icon="@drawable/sf__icon"
 	    android:label="@string/app_name"
 		android:name="com.salesforce.samples.mobilesyncexplorer.MobileSyncExplorerApp"
-		android:manageSpaceActivity="com.salesforce.androidsdk.ui.ManageSpaceActivity">
+		android:manageSpaceActivity="com.salesforce.androidsdk.ui.ManageSpaceActivity"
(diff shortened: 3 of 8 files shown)


native/NativeSampleApps/MobileSyncExplorer/AndroidManifest.xml
```diff
@@ -8,7 +8,8 @@
 	<application android:icon="@drawable/sf__icon"
 	    android:label="@string/app_name"
 		android:name="com.salesforce.samples.mobilesyncexplorer.MobileSyncExplorerApp"
-		android:manageSpaceActivity="com.salesforce.androidsdk.ui.ManageSpaceActivity">
+		android:manageSpaceActivity="com.salesforce.androidsdk.ui.ManageSpaceActivity"
+		android:enableOnBackInvokedCallback="true">
```
brandonpage, line 12: I figured leave one app opted-in would be good for testing.

libs/SalesforceHybrid/AndroidManifest.xml
```diff
@@ -4,5 +4,8 @@
     android:versionCode="82"
     android:versionName="11.1.0.dev">
 
-    <application />
+    <application>
+        <activity android:name="com.salesforce.androidsdk.phonegap.ui.SalesforceDroidGapActivity"
+            android:enableOnBackInvokedCallback="false" />
```
brandonpage, line 9: I haven't see any word on them supporting this.

libs/SalesforceReact/AndroidManifest.xml
```diff
@@ -3,6 +3,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     android:versionCode="82"
     android:versionName="11.1.0.dev">
-
-    <application />
+    
+    <application>
+        <activity android:name=".ui.SalesforceReactActivity"
+            android:enableOnBackInvokedCallback="false" />
```
brandonpage, line 9: [RN will eventually support this but it isn't in yet](https://github.com/facebook/react-native/issues/34529).  We can remove this when we update to whatever version that may be.

wmathurin (approved):

PR #2470, opened 2023-10-10
@W-14246983: [Android][Hybrid] Hybrid Templates Use Maven Central MSDK Dependencies By Default

JohnsonEricAtSalesforce:
🥁 _Ready For Review_ 🎸

  This is a companion to https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/pull/626

  The update here allows the MSDK Gradle workspace to build even when the React Native local repository is not present.  That is the case when MSDK is a composite build with the hybrid template apps.  A warning is printed to indicate this is the case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ executors:
     docker:
       - image: *cimg
     environment:
-      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
+      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError" -Dorg.gradle.daemon=false'
 
 jobs:
   static-analysis:
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,15 +1,23 @@
+val nodeModulesMsdk = File("${rootProject.projectDir}/libs/SalesforceReact/node_modules/react-native/android") // For stand-alone MSDK builds.
+val nodeModulesTemplate = File("${rootProject.projectDir}/../../node_modules/react-native/android") // For template app builds.
+val hasNodeModules = nodeModulesMsdk.exists() || nodeModulesTemplate.exists()
+
 include("libs:SalesforceAnalytics")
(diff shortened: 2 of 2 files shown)


settings.gradle.kts
```diff
@@ -1,15 +1,23 @@
...
 include("libs:SmartStore")
 include("libs:MobileSync")
 include("libs:SalesforceHybrid")
-include("libs:SalesforceReact")
-include("native:NativeSampleApps:RestExplorer")
-include("native:NativeSampleApps:MobileSyncExplorer")
+if (hasNodeModules) {
+    include("libs:SalesforceReact")
+} else {
+    logger.warn("Skipping `SalesforceReact` module since local node repository is missing.  Has `yarn install` been run in the directory containing `package.json`?")
```
JohnsonEricAtSalesforce, line 13: @wmathurin - Since we don't have `node_modules` in the case of the hybrid templates, this allows the MSDK workspace to guard for that and build the other libraries.  A warning is printed.

I'd been considering this anyways as a developer experience enhancement.  This lets the build run, plus gives the developer a solid clue that they need to populate the local repositories.  That's better than the default "unable to resolve artifacts" error message, I suspect.

JohnsonEricAtSalesforce: I could update the message to refer to running the `install.sh`, since that has the `yarn install` that would have been missed.

JohnsonEricAtSalesforce: That will be a big improvement for our side of the build 💫

wmathurin (approved):

.circleci/config.yml
```diff
@@ -67,7 +67,7 @@ executors:
     docker:
       - image: *cimg
     environment:
-      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
+      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError" -Dorg.gradle.daemon=false'
```
brandonpage, line 70: This might be allocating too much, but we can change it in the future if it ends up being a problem.

brandonpage (approved):

PR #2474, opened 2023-10-16
Add ability to attach OkHttp Interceptors by consuming App

mahendrn:
Keep an reference to the builder locally so interceptors added by consuming Apps are retained during the creation of RestClient.


brandonpage (approved):
LGTM.  We will merge this after 11.1 goes live.

PR #2481, opened 2023-10-27
Configurable custom tab browser

wmathurin:
Through SalesforceSDKManager, app can specify an alternate browser or indicate it wants the default to be used.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
@@ -735,8 +735,9 @@ public class SalesforceSDKManager implements LifecycleObserver {
 
     /**
      * Return the custom tab browser to use during advanced authentication
-     * NB: the default browser on the device is used is null is specified or if the specified browser
-     *     is not installed
+     * NB: the default browser on the device is used
+     * - if null is specified
+     * - or if the specified browser is not installed
      * @return custom tab browser
      */
     public synchronized String getCustomTabBrowser() {
@@ -745,8 +746,9 @@ public class SalesforceSDKManager implements LifecycleObserver {
 
     /**
      * Set the custom tab browser to use during advanced authentication
-     * NB: the default browser on the device is used is null is specified or if the specified browser
-     *     is not installed
+     * NB: the default browser on the device is used
+     * - if null is specified
+     * - or if the specified browser is not installed
      * @param customTabBrowser pass null to use the default browser on the device
      */
     public synchronized void setCustomTabBrowser(String customTabBrowser) {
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -732,6 +733,26 @@ public synchronized void setCustomDomainInferencePattern(@Nullable Pattern patte
         this.customDomainInferencePattern = pattern;
     }
 
+    /**
+     * Return the custom tab browser to use during advanced authentication
+     * NB: the default browser on the device is used is null is specified or if the specified browser
```
brandonpage, line 738: Should be `if` instead of `is` in "the default browser on the device is used **_if_** null".

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
```diff
@@ -732,6 +733,26 @@ public synchronized void setCustomDomainInferencePattern(@Nullable Pattern patte
...
+     *     is not installed
+     * @return custom tab browser
+     */
+    public synchronized String getCustomTabBrowser() {
+        return customTabBrowser;
+    }
+
+    /**
+     * Set the custom tab browser to use during advanced authentication
+     * NB: the default browser on the device is used is null is specified or if the specified browser
```
brandonpage, line 748: Same `if` instead of `is`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.java
brandonpage: What happens if null is set and the default browser does not support custom tabs?

wmathurin: What happens today if you don't have chrome installed and the default browser does not support custom tab?

brandonpage: If you have Adv Auth properly setup with the intent filter in the app manifest it does not use the default browser.  It forces Chrome or fails with a toast message telling you to install chrome.

wmathurin: Now with the code change, it will only fail and show a toast if the requested browser is not found.

If you did not request a specific browser (by setting customTabBrowser to null), no error is thrown because doesBrowserExist() returns false (see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2481/files#diff-33ae398946ac646b4e530934b50bfef16db75f0d5b0ce6be67d86c29cae155fcR418)) and no package is set on the intent, causing the default browser to launch and load the login url.

If the default browser does not support custom tab, the flow will fail. Do we have an example of browser that doesn't support custom tab?

brandonpage: I think all of the major browsers support Custom Tabs so maybe it's not a big deal.  However, I think it would be a little better to get the default browsers package name and [check if it is in the list of apps that support Custom Tabs installed on the device](https://developer.chrome.com/docs/android/custom-tabs/howto-custom-tab-check/).

brandonpage (approved):

PR #2486, opened 2023-11-08
Update Firebase Messaging to 23.3.1.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
```diff
@@ -109,41 +120,19 @@ object PushMessaging {
...
-         * isn't thrown, this means that the initialization has already been completed.
-         */
-        try {
-            FirebaseApp.getInstance(appName)
-        } catch (e: IllegalStateException) {
-            SalesforceSDKLogger.i(TAG, "Firebase hasn't been initialized yet", e)
-            FirebaseApp.initializeApp(context, firebaseOptions, appName)
-        }
-    }
-
```
brandonpage, line 146: None of the private functions in this class were removed, just moved below the public functions.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushNotificationInterface.kt
```diff
@@ -42,4 +45,14 @@ interface PushNotificationInterface {
...
+
+    /**
+     * This method is optional, you may simply return null.
+     *
+     * Use this method to provide the correct FirebaseMessaging instance if the one
+     * you intend to use is not the default.
+     *
+     * @return the FirebaseMessaging instance to use.
+     */
+    fun supplyFirebaseMessaging(): FirebaseMessaging?
```
brandonpage, line 57: I believe this is still necessary for some apps due issues in Firebase.  I did remove the method that would allow customers to supply their own `FirebaseApp` as that should not be necessary.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
```diff
@@ -960,15 +960,11 @@ protected void addAccount(final UserAccount account) {
...
+    	 * Registers for push notifications, if setup by the app.
     	 * This step needs to happen after the account has been added by client
     	 * manager, so that the push service has all the account info it needs.
     	 */
-        final Context appContext = SalesforceSDKManager.getInstance().getAppContext();
-        final String pushNotificationId = BootConfig.getBootConfig(appContext).getPushNotificationClientId();
-        if (!TextUtils.isEmpty(pushNotificationId)) {
-            PushMessaging.register(appContext, account);
-        }
+        PushMessaging.register(SalesforceSDKManager.getInstance().getAppContext(), account);
```
brandonpage, line 967: Is there any reason we can think of to pass context in here instead of just having it in the object?  It is passed to just about ever function in `PushMessaging` and that just feels ugly/unnecessary.

wmathurin: Not really.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.java
brandonpage: Static references to Android Context are a memory leak 😞

PR #2487, opened 2023-11-10
Moving to SQLCipher 4.5.5

wmathurin:
Notable changes:
- packages have moved from `net.sqlcipher.database` to `net.zetetic.database.sqlcipher`
- library is loaded with a different call: `System.loadLibrary("sqlcipher")` instead of `SQLiteDatabase.loadLibs(context)`
- DBOpenHelper needs the encryption key at creation time (previously the key was provided when opening the database): a few public APIs had to be changed to accommodate the new behavior

To do:
- fix a few test failures (the explain plan is a bit different in some cases)
- **relax concurrency restriction in SmartStore** (<- that's the big one that will give us performance improvement but require significant testing)

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/AlterSoupLongOperation.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/AlterSoupLongOperation.java
@@ -32,7 +32,7 @@ import android.text.TextUtils;
 import com.salesforce.androidsdk.smartstore.store.SmartStore.SmartStoreException;
 import com.salesforce.androidsdk.smartstore.util.SmartStoreLogger;
 
-import net.sqlcipher.database.SQLiteDatabase;
+import net.zetetic.database.sqlcipher.SQLiteDatabase;
 
 import org.json.JSONException;
 import org.json.JSONObject;
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
@@ -34,9 +34,11 @@ import com.salesforce.androidsdk.analytics.EventBuilderHelper;
 import com.salesforce.androidsdk.smartstore.util.SmartStoreLogger;
 import com.salesforce.androidsdk.util.ManagedFilesHelper;
 
-import net.sqlcipher.database.SQLiteDatabase;
-import net.sqlcipher.database.SQLiteDatabaseHook;
-import net.sqlcipher.database.SQLiteOpenHelper;
+import net.zetetic.database.DatabaseErrorHandler;
+import net.zetetic.database.sqlcipher.SQLiteConnection;
+import net.zetetic.database.sqlcipher.SQLiteDatabase;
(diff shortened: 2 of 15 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -103,43 +105,49 @@ public static synchronized List<String> getGlobalDatabasePrefixList(Context ctx,
...
+	 * @param encryptionKey Password for database.
+	 * @param ctx           Context.
+	 * @param account       User account.
 	 * @return DBOpenHelper instance.
 	 */
-	public static synchronized DBOpenHelper getOpenHelper(Context ctx,
-			UserAccount account) {
-		return getOpenHelper(ctx, account, null);
+	public static synchronized DBOpenHelper getOpenHelper(String encryptionKey, Context ctx,
+														  UserAccount account) {
```
brandonpage, line 114: I assume this is just github weirdness?

brandonpage: Also a side note:  If all the public API's are static synchronized this class could become a standard Kotlin object (in a future modernization PR).

brandonpage (approved):
LGTM so far.  I am excited to see the relaxed concurrency restrictions!

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/AlterSoupLongOperation.java
```diff
@@ -220,7 +220,11 @@ private void alterSoupInternal(AlterSoupStep toStep)  {
 				reIndexSoup();
 			if (toStep == AlterSoupStep.RE_INDEX_SOUP) break;
 		case RE_INDEX_SOUP:
-			dropOldTable();		
+			dropOldTable();
+			// After upgrading to 4.5.5
+			// Functional indices are only picked after one query is run against the database
+			// See @W-14397675
+			store.getLongOperations();
 			if (toStep == AlterSoupStep.DROP_OLD_TABLE) break;
```
wmathurin, line 228: 3 tests were failing:

- testUpsertAgainstChangedSoup
- testDeleteAgainstChangedSoup
- testExactQueryAgainstChangedSoup

In those tests, we alter a soup (change its indexes) then run a query with explain query plan to make sure the new index is being used.

The test fails when the new index is a json functional index (e.g. json_extract (soup, '$.key')).

1) I verified that the indexes are correctly created (with additional logging and by querying the sqlite_master table).
2) I verified that json functional indices are being leveraged if setup initially (by doing explain query plan during other tests and with MobileSyncExplorer).
3) If I add a call to checkDatabaseIndexes() in the test after the alterSoup() call, then it works. All that checkDatabaseIndexes() does is calling dbOpenHelper.getWritableDatabase() followed by a query against the sqlite_master table to get the DDL statements that created the indices on the table. But debugging it looks like both the alterSoup and the checkDatabaseIndexes are going through the same SQLDatabase object.
4) If I simply run any query after alterSoup() e.g. store.getLongOperations(), then it works.

The "fix" for now is to run a query at the end of the alter soup operation.
Not sure why this is needed to "settle" the database.

PR #2488, opened 2023-11-14
Run static analysis per library instead of for the project.

brandonpage:
This should be more efficient and hopefully git rid of our OOM testing issues.


brandonpage:
I will remove the expected `static-analysis` task from being required after this PR is merged.

wmathurin (approved):

PR #2516, opened 2024-02-22
Using new cipher for RSA (but trying with old cipher in case of failure)

wmathurin:
- encrypt and decrypt method use new cipher but try old cipher in case of failure
- key generation code had to be tweaked (to support different digest and encryption padding)
- had issue with default provider so using AndroidKeyStoreBCWorkaround for new cipher
- Added test

### DIFF
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
@@ -60,6 +60,13 @@ public class Encryptor {
     private static final String AES_GCM_CIPHER = "AES/GCM/NoPadding";
     private static final String MAC_TRANSFORMATION = "HmacSHA256";
     private static final String RSA_PKCS1 = "RSA/ECB/PKCS1Padding";
+
+    // Cipher in use until 248 (API 59)
+    public static final String LEGACY_NOTIFICATION_CIPHER = RSA_PKCS1;
+
+    // Cipher in use starting 250 (API 60)
+    public static final String NOTIFICATION_CIPHER = "RSA/ECB/OAEPWithSHA-256AndMGF1Padding";
+
     private static final String BOUNCY_CASTLE = "BC";
     private static final int READ_BUFFER_LENGTH = 1024;
 
@@ -503,7 +510,7 @@ public class Encryptor {
         return output;
     }
 
-    private static byte[] encryptWithPublicKey(PublicKey publicKey, String data, String cipher) {
+    public static byte[] encryptWithPublicKey(PublicKey publicKey, String data, String cipher) {
         if (publicKey == null || TextUtils.isEmpty(data)) {
             return null;
         }
@@ -517,7 +524,7 @@ public class Encryptor {
         return null;
     }
 
-    private static byte[] decryptWithPrivateKey(PrivateKey privateKey, String data, String cipher) {
+    public static byte[] decryptWithPrivateKey(PrivateKey privateKey, String data, String cipher) {
         if (privateKey == null || TextUtils.isEmpty(data)) {
             return null;
         }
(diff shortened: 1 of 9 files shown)


brandonpage (approved):
LGTM!

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/security/KeyStoreWrapperTest.java
```diff
@@ -139,4 +154,25 @@ public void testDecryptDataEncryptedWithLegacyRSACipher() {
...
     }
+
+    /**
+     * Upgrade test: client upgraded and server upgraded
+     * 1. Client generated key pair before upgrading to 11.1.1
+     * 2. Server is upgraded and encrypts push notification with stored key using new RSA cipher mode
+     * 3. Client tries to decrypt push notification
+     */
+    @Test
+    public void testDecryptDataEncryptedWithNewCipherForKeyCreatedBeforeChange() {
```
wmathurin, line 165: That scenario fails. Because when generating a key, you have to specify what encryption paddings / digest etc can be used with a given key. So a key generated pre 11.1.1 won't be allowed to decrypt using a different cipher mode even though the server will probably successfully encrypt with it using a different cipher mode.

brandonpage (approved):

wmathurin:
Need more work (upgrade step) because of scenario 3 in https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2516/commits/f02a21e4f82f6d19bbf23c24d3976cb11f5554dc

wmathurin:
Testing upgrade scenario end to end (old client logins and registers, client upgrades, server sends notification).
The upgrade happens when SalesforceSDKManager initializes and therefore before the application had a chance to specify the push notification receiver. As a result, the upgrade code does not do anything since it does not see push as being setup.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
```diff
@@ -48,7 +49,7 @@ import com.salesforce.androidsdk.util.SalesforceSDKLogger
  * retrieval of push notification registration information from a
  * private shared preference file.
  */
-object PushMessaging {
+public object PushMessaging {
```
brandonpage, line 52: Why does this need to be public?  Looks like it can be accessed from the upgrade manager without this change.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/KeyStoreWrapper.java
```diff
@@ -211,6 +226,41 @@ private PrivateKey getPrivateKey(String algorithm, String name, int length) {
...
+                 * TODO: Remove this check once minVersion > 28.
+                 */
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+
+                    /*
+                     * Disabling StrongBox based on Google's recommendation - it's not a good
+                     * fit for this use case, since the key will need to be retrieved multiple
+                     * times. Besides, StrongBox Keymaster is available only on a few devices,
+                     * such as the Pixel 3 and Pixel 3 XL at this time.
+                     */
```
brandonpage, line 251: Unrelated, but we should re-evaluate this at some point since it's been more than a few years.

brandonpage (approved):

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/security/Encryptor.java
```diff
@@ -422,7 +439,11 @@ public static String decryptWithRSA(PrivateKey privateKey, String data) {
      * @return Decrypted data.
      */
     public static byte[] decryptWithRSABytes(PrivateKey privateKey, String data) {
-        return decryptWithPrivateKey(privateKey, data, RSA_PKCS1);
+        byte[] result =  decryptWithPrivateKey(privateKey, data, RSA_ECB_OAEP, /* logErrorOnFailure */ false);
```
wmathurin, line 442: App upgrading to 11.1.1 will fail to decrypt keys with OAEP and it's okay, they should succed to decrypt them with RSA_PKCS1. Only logging a warning if OAEP fails. An error will be logged if we completely failed to decrypt the data.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
wmathurin: Actually it was already public (that's the default in Kotlin) and adding public was redundant (Android Studio actually complained). But I guess we should have made that class internal.

brandonpage: Yeah, I had originally written internal but I couldn't remember if there was a reason it was public.  I think it's just always been public so they kt conversion didn't change that in a minor release.

PR #2517, opened 2024-02-27
Native login

brandonpage:
### This PR includes:
- suspended `login(username: String, password: String)` function
- `getFallbackWebAuthenticationIntent` function
- `shouldShowBackButton` property/function
- Functions to enable Biometric Authentication via Native Login.
- Added `nativeLogin` property to UserAccount.
- Fix for truly unauthenticated rest requests (necessary for Bio Auth with Native Login).

### Usage: 
SalesforceSDKManager's `initNative` now takes an additional optional native login activity class.

App calls `useNativeLogin` in it's main activity :
```
SalesforceSDKManager.getInstance().useNativeLogin(clientId, redirectUri, loginUrl);
```

In their own UI the App calls:
```
val result = SalesforceSDKManager.getInstance().nativeLoginManager.login("username", "password")
```

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
@@ -59,6 +59,8 @@ public class UserAccountBuilder {
     private String contentDomain;
     private String contentSid;
     private String csrfToken;
+
+    private Boolean nativeLogin = false;
     private Map<String, String> additionalOauthValues;
 
     /**
@@ -339,6 +341,17 @@ public class UserAccountBuilder {
     }
 
     /**
+     * Sets if the user authenticated with native login.
+     *
+     * @param nativeLogin if the user authenticated with native login.
+     * @return Instance of this class.
+     */
+    public UserAccountBuilder nativeLogin(Boolean nativeLogin) {
+        this.nativeLogin = nativeLogin;
+        return this;
+    }
+
+    /**
      * Sets additional OAuth values.
      *
      * @param additionalOauthValues Additional OAuth values.
@@ -358,6 +371,6 @@ public class UserAccountBuilder {
         return new UserAccount(authToken, refreshToken, loginServer, idUrl, instanceServer, orgId,
                 userId, username, accountName, communityId, communityUrl, firstName, lastName,
                 displayName, email, photoUrl, thumbnailUrl, additionalOauthValues, lightningDomain,
-                lightningSid, vfDomain, vfSid, contentDomain, contentSid, csrfToken);
+                lightningSid, vfDomain, vfSid, contentDomain, contentSid, csrfToken, nativeLogin);
     }
 }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
@@ -343,11 +343,11 @@ public class UserAccountBuilder {
     /**
      * Sets if the user authenticated with native login.
      *
-     * @param nativeLogin if the user authenticated with native login.
+     * @param isNativeLogin if the user authenticated with native login.
      * @return Instance of this class.
      */
-    public UserAccountBuilder nativeLogin(Boolean nativeLogin) {
-        this.nativeLogin = nativeLogin;
+    public UserAccountBuilder nativeLogin(Boolean isNativeLogin) {
+        this.nativeLogin = isNativeLogin;
         return this;
     }
 
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -160,6 +160,12 @@ open class SalesforceSDKManager protected constructor(
     internal val nativeLoginActivity: Class<out Activity>? = null,
 ) : LifecycleObserver {
 
+    constructor(
+        context: Context,
+        mainActivity: Class<out Activity>,
+        loginActivity: Class<out Activity>? = null
+    ) : this(context, mainActivity, loginActivity, nativeLoginActivity = null)
+
     /** The Android context */
     val appContext: Context = context
 
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -1472,13 +1472,15 @@ open class SalesforceSDKManager protected constructor(
         private fun init(
             context: Context,
             mainActivity: Class<out Activity>,
-            loginActivity: Class<out Activity>
+            loginActivity: Class<out Activity>? = null,
+            nativeLoginActivity: Class<out Activity>? = null,
         ) {
             if (INSTANCE == null) {
                 INSTANCE = SalesforceSDKManager(
                     context,
                     mainActivity,
-                    loginActivity
+                    loginActivity,
+                    nativeLoginActivity,
                 )
             }
             initInternal(context)
@@ -1548,7 +1550,28 @@ open class SalesforceSDKManager protected constructor(
         ) = init(
             context,
             mainActivity,
-            loginActivity
+            loginActivity,
+        )
+
+        /**
+         * Initializes required components. Native apps must call one overload
+         * of this method before using the Salesforce Mobile SDK.
+         *
+         * @param context The Android context
+         * @param mainActivity The app's main activity class
+         * @param loginActivity The app's login activity
+         * @param nativeLoginActivity The app's native login activity
+         */
+        fun initNative(
+            context: Context,
+            mainActivity: Class<out Activity>,
+            loginActivity: Class<out Activity>? = null,
+            nativeLoginActivity: Class<out Activity>,
+        ) = init(
+            context,
+            mainActivity,
+            loginActivity,
+            nativeLoginActivity,
         )
 
         /**
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -157,7 +157,7 @@ open class SalesforceSDKManager protected constructor(
     protected val context: Context,
     mainActivity: Class<out Activity>,
     private val loginActivity: Class<out Activity>? = null,
-    private val nativeLoginActivity: Class<out Activity>? = null,
+    internal val nativeLoginActivity: Class<out Activity>? = null,
 ) : LifecycleObserver {
 
     /** The Android context */
@@ -197,8 +197,12 @@ open class SalesforceSDKManager protected constructor(
      */
     val loginActivityClass: Class<out Activity>
         get() {
-            // TODO: check fallback to webview here
-            return nativeLoginActivity ?: loginActivity ?: LoginActivity::class.java
+            return nativeLoginActivity ?: webviewLoginActivityClass
+        }
+
+    internal val webviewLoginActivityClass: Class<out Activity>
+        get() {
+            return loginActivity ?: LoginActivity::class.java
         }
 
     /** The class for the account switcher activity */
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -88,7 +88,7 @@ import com.salesforce.androidsdk.auth.OAuth2.revokeRefreshToken
 import com.salesforce.androidsdk.auth.idp.SPConfig
 import com.salesforce.androidsdk.auth.idp.interfaces.IDPManager
 import com.salesforce.androidsdk.auth.idp.interfaces.SPManager
-import com.salesforce.androidsdk.auth.nativeLogin.NativeLoginManager as NativeLoginManagerInterface
+import com.salesforce.androidsdk.auth.nativeLoginInterface.NativeLoginManager as NativeLoginManagerInterface
 import com.salesforce.androidsdk.auth.NativeLoginManager
 import com.salesforce.androidsdk.config.AdminPermsManager
 import com.salesforce.androidsdk.config.AdminSettingsManager
@@ -229,7 +229,7 @@ open class SalesforceSDKManager protected constructor(
     }
 
     /** The Salesforce SDK manager's native login manager */
-    var nativeLoginManager: NativeLoginManagerInterface? = null
+    lateinit var nativeLoginManager: NativeLoginManagerInterface
         private set
 
     /** Optionally enables features for automated testing.
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -197,19 +197,13 @@ open class SalesforceSDKManager protected constructor(
      */
     private var showDeveloperSupportBroadcastIntentReceiver: BroadcastReceiver? = null
 
+    internal val webviewLoginActivityClass: Class<out Activity> = loginActivity ?: LoginActivity::class.java
+
     /**
      * The class of the activity used to perform the login process and create
      * the account.
      */
-    val loginActivityClass: Class<out Activity>
-        get() {
-            return nativeLoginActivity ?: webviewLoginActivityClass
-        }
-
-    internal val webviewLoginActivityClass: Class<out Activity>
-        get() {
-            return loginActivity ?: LoginActivity::class.java
-        }
+    val loginActivityClass: Class<out Activity> = nativeLoginActivity ?: webviewLoginActivityClass
 
     /** The class for the account switcher activity */
     var accountSwitcherActivityClass = AccountSwitcherActivity::class.java
@@ -548,7 +542,7 @@ open class SalesforceSDKManager protected constructor(
     var shouldBlockSalesforceIntegrationUser = false
 
     /**
-     * Creates a NativeLoginManager instance that allows the app to use it's
+     * Creates a NativeLoginManager instance that allows the app to use its
      * own native UI for authentication.
      *
      * @param consumerKey The Connected App consumer key.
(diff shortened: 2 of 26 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -88,6 +88,8 @@ import com.salesforce.androidsdk.auth.OAuth2.revokeRefreshToken
 import com.salesforce.androidsdk.auth.idp.SPConfig
 import com.salesforce.androidsdk.auth.idp.interfaces.IDPManager
 import com.salesforce.androidsdk.auth.idp.interfaces.SPManager
+import com.salesforce.androidsdk.auth.nativeLoginInterface.NativeLoginManager as NativeLoginManagerInterface
```
brandonpage, line 91: This little trick makes the code super readable.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -154,9 +156,16 @@ open class SalesforceSDKManager protected constructor(
...
+    private val loginActivity: Class<out Activity>? = null,
+    internal val nativeLoginActivity: Class<out Activity>? = null,
 ) : LifecycleObserver {
 
+    constructor(
+        context: Context,
+        mainActivity: Class<out Activity>,
+        loginActivity: Class<out Activity>? = null
+    ) : this(context, mainActivity, loginActivity, nativeLoginActivity = null)
+
```
brandonpage, line 168: Adding the new param with a default value makes this a non-breaking change in Kotlin, but not Java unfortunately. The new constructor covers that case.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/nativeLoginInterface/NativeLoginManager.kt
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * If the native login view should show a back button.
+     */
+    val shouldShowBackButton: Boolean
+
+    /**
+     * The username of the locked account or null.  Can be used to pre-populate the username field
+     * or in a message telling the user which account biometric will unlock.
+     */
+    val biometricAuthenticationUsername: String?
```
brandonpage, line 51: Just a reminder:  Unlike `let` in Swift, `val` in Kotlin is **not immutable**.  `val` is simply _readonly_.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/nativeLoginInterface/NativeLoginManager.kt
```diff
@@ -0,0 +1,66 @@
...
+    InvalidPassword,      // Password does not meet the weakest Salesforce criteria.
+    InvalidCredentials,   // Username/password combination is incorrect.
+    UnknownError,
+    Success,
+}
+
+/**
+ *  Manage native login flow.
+ */
+interface NativeLoginManager {
```
brandonpage, line 40: This class is missing `cancelAuthentication` and `biometricAuthenticationSuccess` compared to it's iOS counterpart because they just aren't necessary on Android.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -531,6 +547,20 @@ open class SalesforceSDKManager protected constructor(
...
+    /**
+     * Creates a NativeLoginManager instance that allows the app to use it's
+     * own native UI for authentication.
+     *
+     * @param consumerKey The Connected App consumer key.
+     * @param callbackUrl The Connected App redirect URI.
+     * @param communityUrl The login url for native login.
+     * @return The Native Login Manager.
+     */
+    fun useNativeLogin(consumerKey: String, callbackUrl: String, communityUrl: String): NativeLoginManagerInterface {
```
brandonpage, line 559: As I mentioned above since we aliased "com.salesforce.androidsdk.auth.**_nativeLoginInterface_**.NativeLoginManager" to `NativeLoginManagerInterface` we can simply use that instead of having to write the _entire path_ for it OR the implementation class of the same name.  And the best part is the consumer still sees the return type as the interface `NativeLoginManager`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/nativeLoginInterface/NativeLoginManager.kt
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Use this function to get the intent for fallback web based authentication.
+     *
+     * In your native login activity start this intent in such a way as to [get a result callback](https://developer.android.com/training/basics/intents/result).
+     * A result code of `Activity.RESULT_OK` means the user has successfully authenticated via the webview and you
+     * can finish your activity.
+     *
+     * @return the intent to be started for fallback web based authentication.
+     */
+    fun getFallbackWebAuthenticationIntent() : Intent
```
brandonpage, line 73: I really wanted to handle this for the customer but without an activity (or even if they give me a reference to their activity) it just isn't possible.  Good news is the code the customer will have to implement pretty simple.  Here is an example of the new (`startActivityForResult` is deprecated) way to do it:

```
private val fallbackResult = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
    if (result.resultCode == Activity.RESULT_OK) {
        finish()
    }
}
```

And in their button they simply call `fallbackResult.launch(nativeLoginManager.getFallbackWebAuthenticationIntent())`

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -389,7 +389,12 @@ public static final IdServiceResponse callIdentityService(HttpAccess httpAccesso
      * @param authToken Access token.
      */
     public static final Request.Builder addAuthorizationHeader(Request.Builder builder, String authToken) {
-        return builder.header(AUTHORIZATION, BEARER + authToken);
+        if (builder.getHeaders$okhttp().get(AUTHORIZATION) == null) {
+            return builder.header(AUTHORIZATION, BEARER + authToken);
+        } else {
+            // Don't override header for Native Login authentication.
+            return builder;
+        }
```
brandonpage, line 397: This feels a bit dirty if you consider that we probably just shouldn't add the interceptor that inserts this authorization header to unauthenticated requests, but attempting that gets pretty complicated.  The client builder is associated with the current account, not the rest client being used.  So even if we are using an unauthenticated rest client, the rest client of the current user is till attached.  

The only scenario in which this matter is Biometric Auth with Native Login.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
```diff
@@ -338,6 +340,17 @@ public UserAccountBuilder csrfToken(String csrfToken) {
         return this;
     }
 
+    /**
+     * Sets if the user authenticated with native login.
+     *
+     * @param nativeLogin if the user authenticated with native login.
+     * @return Instance of this class.
+     */
+    public UserAccountBuilder nativeLogin(Boolean nativeLogin) {
```
wmathurin, line 349: Wouldn't it be better to call it setNativeLogin??

brandonpage: I don't have a problem renaming it, but since it's a builder class all of the setters are named this way.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -135,7 +137,7 @@ import java.util.concurrent.ConcurrentSkipListSet
 import java.util.regex.Pattern
 import com.salesforce.androidsdk.auth.idp.IDPManager as DefaultIDPManager
 import com.salesforce.androidsdk.auth.idp.SPManager as DefaultSPManager
-import com.salesforce.androidsdk.security.interfaces.BiometricAuthenticationManager as DefaultBiometricAuthenticationManager
+import com.salesforce.androidsdk.security.interfaces.BiometricAuthenticationManager as BiometricAuthenticationManagerInterface
```
JohnsonEricAtSalesforce, line 140: I was just noticing this interface and internal implementation looks just like NativeLoginManager where we have the interface an an explicitly `internal` implementation.  `InternalBiometricAuthenticationManager`?

Of course, there's lot of ways to name interface/protocols and implementations 🤯

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -192,7 +201,15 @@ open class SalesforceSDKManager protected constructor(
      * The class of the activity used to perform the login process and create
      * the account.
      */
-    var loginActivityClass: Class<out Activity> = LoginActivity::class.java
+    val loginActivityClass: Class<out Activity>
+        get() {
```
JohnsonEricAtSalesforce, line 205: Both of these could drop their braces, return and `=` a single expression, for optional Kotlin style points.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -221,6 +238,10 @@ open class SalesforceSDKManager protected constructor(
         LoginServerManager(appContext)
     }
 
+    /** The Salesforce SDK manager's native login manager */
+    lateinit var nativeLoginManager: NativeLoginManagerInterface
```
JohnsonEricAtSalesforce, line 242: Wouldn't this be safer as an optional rather than a `lateinit`? Since this is public to apps, they could get themselves into a crash if they don't remember to check if this is initialized.  `lateinit` is neat to have, but I've seen enough random crashes leak through that I almost never use it even privately anymore.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -531,6 +547,20 @@ open class SalesforceSDKManager protected constructor(
      */
     var shouldBlockSalesforceIntegrationUser = false
 
+    /**
+     * Creates a NativeLoginManager instance that allows the app to use it's
```
JohnsonEricAtSalesforce, line 551: Simply typo: This should be possessive `its` instead of a contraction for `it is`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -0,0 +1,271 @@
...
+        val authRequestBody = createRequestBody(
+            RESPONSE_TYPE to CODE_CREDENTIALS,
+            CLIENT_ID to clientId,
+            REDIRECT_URI to redirectUri,
+            CODE_CHALLENGE to codeChallenge,
+        )
+        val authRequest = RestRequest(
+            RestRequest.RestMethod.POST,
+            RestRequest.RestEndpoint.LOGIN,
+            loginUrl + OAUTH_AUTH_PATH, // Full path for unauthenticated request
```
JohnsonEricAtSalesforce, line 117: Could this use a Kotlin string template? This should be equivalent, but perhaps it'd be nice just for consistency.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -0,0 +1,271 @@
...
+    internal fun isValidPassword(password: String): Boolean {
+        val containsNumber = password.contains("[0-9]".toRegex())
+        val containsLetter = password.contains("[A-Za-z]".toRegex())
+
+        return containsNumber && containsLetter && password.length >= MIN_PASSWORD_LENGTH
+                && password.toByteArray().size <= MAX_PASSWORD_LENGTH_BYTES
+    }
+
+    private suspend fun suspendFinishAuthFlow(tokenResponse: RestResponse): NativeLoginResult {
+        val appContext = SalesforceSDKManager.getInstance().appContext
```
JohnsonEricAtSalesforce, line 189: I believe we have to eliminate this Context reference in `SalesforceSDKManager` someday.  Should we have native login manager receive it another way now to make that easier?  Perhaps we could have the app provide it in the initialize native login method?  I might be thinking to far ahead on this one.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -0,0 +1,271 @@
...
+                            Intent(appContext, SalesforceSDKManager.getInstance().mainActivityClass).apply {
+                                setPackage(appContext.packageName)
+                                flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                            }
+                        )
+                    }
+
+                    /**
+                     * We need to wait until the main activity is launched here before
+                     * returning success back to the apps Native Login Activity.  If we
```
JohnsonEricAtSalesforce, line 213: `apps` should probably be the possessive `app's`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    private suspend fun suspendFinishAuthFlow(tokenResponse: RestResponse): NativeLoginResult {
+        val appContext = SalesforceSDKManager.getInstance().appContext
+        val loginOptions = LoginOptions(loginUrl, redirectUri, clientId, emptyArray<String>())
+        val tokenEndpointResponse = OAuth2.TokenEndpointResponse(tokenResponse.rawResponse)
+        tokenResponse.consumeQuietly()
+
+        return suspendCoroutine { continuation ->
+            OAuthWebviewHelper(appContext, object : OAuthWebviewHelper.OAuthWebviewHelperEvents {
```
JohnsonEricAtSalesforce, line 195: I'm sure this is food for though much later, but it seems unwieldy to have to use `OAuthWebviewHelper` in this context.  I see how it works in the context of the codebase it is today, so no action required.  I have a work item in the backlog where it could be neat to see how we could separate `OAuthWebviewHelper` into smaller units of decoupled logic and then this could use just the one it needs without the two skipped implementations.

Just something to ponder for the future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -0,0 +1,271 @@
...
+                                flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                            }
+                        )
+                    }
+
+                    /**
+                     * We need to wait until the main activity is launched here before
+                     * returning success back to the apps Native Login Activity.  If we
+                     * don't when they call finish() the activity will just be restarted
+                     * and will hand around forever behind the main activity.
```
JohnsonEricAtSalesforce, line 215: Should `hand` be `hang`?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Re-reading my own notes thus far, they're all optional except this one.  I do feel we'd save some apps an otherwise obvious crash or two by making this an optional.

JohnsonEricAtSalesforce (changes requested):
Looking great here, overall.  Everything else I noted is optional review, but there's just the one follow-up regarding the `lateinit` on our public Native Login Manager property we should review before merging.

https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2517#discussion_r1520475464

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
brandonpage: You are right on the package name:
```
Names of packages are always lowercase and do not use underscores (org.example.project). Using multi-word names is generally discouraged, but if you do need to use multiple words, you can either just concatenate them together or use the camel case (org.example.myProject).
```
`.native.login.interfaces` is a great suggestion, I see you did that with local notifications.  I'll follow that pattern here too.

Personally I hate class names with `Default` (when there is only one impl) or `Internal` in them because they are just generic junk added on to make the name unique. Might as well by `NativeLoginManager1` for all the value they add lol.  

The reason I like this naming convention is because it is the clearest naming for us **and** the consumer simultaneously.  `auth.native.login.interface.NativeLoginManager` is an interface but to the consumer it is also the implementation.  When we return an instance of that interface to them they get a concrete implementation.  But internally it's just an interface to us, so being able to refer to it as `NativeLoginManagerInterface` internally and have `auth.NativeLoginManager` keep it's name is the best of both worlds imo.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
brandonpage: Perhaps.  That seems like a setup issue that customers wouldn't actually see in the wild (similar to a comment I made on your PR earlier).  I feel like customers are either going to use the feature and call `useNativeLogin` in the app's `onCreate` or they aren't going to touch `nativeLoginManager`.  Dealing with optionals is pretty trivial in Kotlin but a bit more annoying in Java.  🤷 

Thoughts on this @wmathurin ?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
brandonpage: 🤷 Could be, but I that `loginUrl + OAUTH_AUTH_PATH` is more readable than `"$loginUrl$OAUTH_AUTH_PATH"`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
brandonpage: Yeah, I agree.  `OAuthWebviewHelper` is designed around the webview lifecycle and the hooks into it are engrained all over.  I knew it would be too much work for this story/PR to try to separate it into smaller pieces but it would be really cool if we could do that in the future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.java
wmathurin: That's fine keeping it the same. May have the param name change to isNativeLogin??

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
wmathurin: Optionals are annoying but compile time failure is still preferable to a run time one (even if unlikely).

JohnsonEricAtSalesforce (approved):
Thanks for the updates!

wmathurin (approved):

PR #2518, opened 2024-02-27
Fix login.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1660,7 +1660,7 @@ open class SalesforceSDKManager protected constructor(
      */
     fun fetchAuthenticationConfiguration(
         completion: (() -> Unit)? = null
-    ) = CoroutineScope(Main).launch {
+    ) = CoroutineScope(Default).launch {
```
brandonpage, line 1663: This fixes the native browser not opening when it should for Adv Auth.

wmathurin (approved):

PR #2523, opened 2024-03-13
IdServiceResponse now pulls same fields than iOS

wmathurin:
Matching the [fields pulled on iOS](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.h).
NB: There are a few more not pulled on either platform.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -109,7 +109,6 @@ public class OAuth2 {
     private static final String PHOTOS = "photos";
     private static final String PICTURE = "picture";
     private static final String THUMBNAIL = "thumbnail";
-
     private static final String AUTHORIZATION_CODE = "authorization_code";
     private static final String HYBRID_AUTH_CODE = "hybrid_auth_code";
     private static final String CODE = "code";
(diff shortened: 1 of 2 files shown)


brandonpage (approved):
LGTM.  But these new values aren't actually exposed on the user account/stored/used for anything yet, right?

wmathurin:
@brandonpage You're right. It is not a super useful change.

wmathurin:
@bbirman are we storing those values with the user?

brandonpage:
> @bbirman are we storing those values with the user?

I believe `IdData` is [stored on the user](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/f28a6d32b287a679aba30f2a4f8974e20284d9e2/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccount.m#L204) on iOS.

bbirman:
> @bbirman are we storing those values with the user?

Yup, it looks like it's on SFUserAccount, stored with UserAccount.plist

wmathurin:
Thanks so I guess I should send another PR to do something similar for Android.

brandonpage:
> Thanks so I guess I should send another PR to do something similar for Android.

It doesn't necessarily have to be done now.  I took a quick stab at rewriting `UserAccount.java` in Kotlin (simply using properties makes that 800 line file less than 400) while working on Native Login but there is a ton of ambiguous nullability that would require careful upgrade testing.  If no one is asking for them could add these new fields as part of a modernization story there to avoid duplicating a lot of upgrade testing.

wmathurin:
@brandonpage I might just send the locale through for now since that was the original ask of the bug/story.

wmathurin:
@brandonpage just exposing language and locale to user account: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2525
That was painful!!

brandonpage:
> @brandonpage just exposing language and locale to user account: #2525 That was painful!!

I tried to warn you lol.  I just made a note to bring this up in planning.

PR #2533, opened 2024-03-22
@W-14584405: [Android] OTP implementation

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

This is ready for final review now that it has been tested fully in the updated template app.  See https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/400

Thanks! 🤘🏻

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -448,7 +448,7 @@ internal class NativeLoginManager(
         // TODO: Inquire regarding modern JSON serialization in MSDK. ECJ20240317
         fun toJson() = JSONObject().apply {
             put("recaptcha", recaptcha)
-            put("recaptchaevent", recaptchaevent)
+            put("recaptchaevent", recaptchaevent?.toJson())
             put("username", username)
             put("verificationmethod", verificationMethod)
         }
@@ -473,7 +473,14 @@ internal class NativeLoginManager(
         val siteKey: String,
         val expectedAction: String = "login",
         val projectId: String
-    )
+    ) {
+        fun toJson() = JSONObject().apply {
+            put("token", token)
+            put("siteKey", siteKey)
+            put("expectedAction", expectedAction)
+            put("projectId", projectId)
+        }
+    }
 
     // endregion
     // region Private Implementation
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -26,10 +26,15 @@
  */
 package com.salesforce.androidsdk.auth
 
-import android.accounts.AccountManager
+import android.accounts.AccountManager.KEY_INTENT
 import android.content.Intent
+import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
+import android.content.Intent.FLAG_ACTIVITY_SINGLE_TOP
 import android.os.Bundle
-import android.util.Base64
+import android.util.Base64.NO_PADDING
+import android.util.Base64.NO_WRAP
+import android.util.Base64.URL_SAFE
+import android.util.Base64.encodeToString
 import com.salesforce.androidsdk.accounts.UserAccount
 import com.salesforce.androidsdk.app.SalesforceSDKManager
 import com.salesforce.androidsdk.auth.OAuth2.AUTHORIZATION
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -250,22 +265,305 @@ internal class NativeLoginManager(
...
+     * @return The native login result
+     */
+    private suspend fun submitAccessTokenRequest(
+        authorizationResponse: RestResponse,
+        codeVerifier: String
+    ): NativeLoginResult {
+        if (authorizationResponse.isSuccess) {
+            val code = authorizationResponse.asJSONObject().get(CODE).toString()
+            val authEndpoint = authorizationResponse.asJSONObject().get(SFDC_COMMUNITY_URL).toString()
+            val tokenRequestBody = createRequestBody(
```
wmathurin, line 513: Doesn't the Oauth2 exchangeCode method do that?

wmathurin: Shouldn't we use HYBRID_AUTH_CODE for the grant type when hybrid auth flow is enabled? See https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java#L320

That causes the token end point to return a bunch of cookies that some app (e.g. SApp) can use when "hydrating" their web view.

brandonpage: That one is on me.  I missed that in my username/password PR.  

And I believe we aren't using OAuth2's `exchangeCode` because we need to handle/suspend the TokenEndpointResponse (which involves making the user info request) to return success or failure at the right time.  If we don't the app will call finish on the native login view too early and it will just get recreated and sit in the background forever (since we won't have launched he main activity yet).

brandonpage: iOS also does not use hybrid_auth_code.  😬

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -250,22 +265,312 @@ internal class NativeLoginManager(
...
+     * "email" or "sms"
+     */
+    private data class OtpRequestBody(
+        val recaptcha: String?,
+        val recaptchaevent: OtpRequestBodyReCaptchaEvent?,
+        val username: String,
+        val verificationMethod: String
+    ) {
+        // TODO: Inquire regarding modern JSON serialization in MSDK. ECJ20240317
+        fun toJson() = JSONObject().apply {
```
JohnsonEricAtSalesforce, line 449: @brandonpage or @wmathurin - I took a casual glance around the rest of MSDK and didn't see any newer way of working with JSON beyond json.org.  Is that true? It would be trivial to update this to Kotlin Serialization.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
JohnsonEricAtSalesforce: @wmathurin and @brandonpage - Do we need a change here? I read Brandon's note above as "works as expected", but wasn't sure.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -250,22 +265,312 @@ internal class NativeLoginManager(
...
+        // React to the OTP response.
+        return when (otpResponse.isSuccess) {
+            true -> {
+                runCatching {
+                    // Decode the OTP response to obtain the OTP identifier.
+                    otpResponse.consumeQuietly()
+                    OtpRequestResult(Success, otpResponse.asJSONObject().get("identifier").toString())
+                }.getOrElse { e ->
+                    SalesforceSDKLogger.e(TAG, "Cannot JSON decode OTP response body due to a decoding error with message '${e.message}'.")
+                    otpResponse.consumeQuietly()
```
brandonpage, line 361: NIT:  Do we need to consume the response here?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
brandonpage: I wrote a crossplatform WI for this.

brandonpage (approved):

PR #2534, opened 2024-03-25
Add interceptor to integration user info check to make authenticated redirects possible.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -1100,18 +1104,43 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
     private fun fetchIsSalesforceIntegrationUser(
         tokenEndpointResponse: TokenEndpointResponse?
     ): Boolean {
-        val url = "$loginUrl/services/oauth2/userinfo"
-        val builder: Builder = Builder().url(url).get()
+        val baseUrl = tokenEndpointResponse?.instanceUrl ?: loginUrl
```
brandonpage, line 1107: I changed this to `instanceUrl` to match iOS, but with the fix in the rest of this PR it isn't strictly necessary.

bbirman (approved):

wmathurin (approved):

PR #2535, opened 2024-03-29
Push notification encryption (follow-up)

wmathurin:
- cipher modes are in an enum
- methods to encrypt/decrypt with RSA expect a cipher mode
- there is a new method (`Encryptor.decryptWithRSAMultiCipherNodes`) called by `PushNotificationDecryptor` which tries the new RSA cipher mode and then the legacy one
- `SalesforceKeyGenerator` is using the same cipher mode as before (otherwise the encryption key is not readable after upgrade)
- during push device registration with Salesforce, we add the `CipherName` parameter if the app is configured to talk to API 60 (or greater)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
@@ -83,7 +83,7 @@ public class SalesforceKeyGenerator {
      * @return Unique ID.
      */
     public static String getUniqueId(String name, int length) {
-        return generateUniqueId(name, length);
+        return generateUniqueIdIfNoneStored(name, length);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -153,33 +153,45 @@ private synchronized static String generateEncryptionKey(String name) {
...
-                // Generates a random UUID 128-bit key instead.
-                uniqueId = UUID.randomUUID().toString();
+            uniqueId = Encryptor.decryptWithRSA(privateKey, id, Encryptor.CipherMode.RSA_PKCS1);
+            if (uniqueId == null) {
+                SalesforceSDKLogger.e(TAG, "Unable to decrpyt unique identifier stored");
             }
+        }
+
+        // Other create a new one and store it
+        if (uniqueId == null) {
```
wmathurin, line 170: Previously if we failed to decrypt the stored unique id, we would return null which would cause the app to crash. Now it will recreate a unique id and log the problem encountered.

NB: That situation should not happen. I ran into it when I changed the cipher mode for all RSA calls. With this PR, only push notification encryption will use the new cipher mode. If/when we decide to change the cipher mode for the SalesforceKeyGenerator call, we will need to add a upgrade step.

brandonpage (approved):

PR #2537, opened 2024-04-01
@W-14831222: [Android] NullPointerException: userAccount must not be null on app foreground with biometrics

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

  This guards for a reported null pointer crash in the biometric authentication manager.  Though, it does not solve for why an Android `Account` exists that a `UserAccount` cannot be derived from.  Potentially, more research would be needed in  `UserAccountManager.buildUserAccount(Account)` to identify the gap or to gather metrics related to the actual `Account` data at runtime.  The immediate goal is to reduce the impact of the observed crashes and improve the user experience. Here's a follow-up work item for that effort: [W-15381271: -MSDK Android- Android Account Is Sometimes Null In BiometricAuthenticationManager shouldLock
](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001nRqC3YAK/view)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -36,6 +36,8 @@ import android.content.SharedPreferences.Editor;
 import android.os.Bundle;
 import android.text.TextUtils;
 
+import androidx.annotation.Nullable;
+
 import com.salesforce.androidsdk.app.Features;
 import com.salesforce.androidsdk.app.SalesforceSDKManager;
 import com.salesforce.androidsdk.auth.AuthenticatorService;
@@ -413,7 +415,7 @@ public class UserAccountManager {
 	 * @param account Account object.
 	 * @return UserAccount object.
 	 */
-	public UserAccount buildUserAccount(Account account) {
+	public @Nullable UserAccount buildUserAccount(Account account) {
 		if (account == null) {
 			return null;
 		}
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -413,7 +415,7 @@ public void signoutUser(UserAccount userAccount, Activity frontActivity, boolean
 	 * @param account Account object.
 	 * @return UserAccount object.
 	 */
-	public UserAccount buildUserAccount(Account account) {
+	public @Nullable UserAccount buildUserAccount(Account account) {
```
JohnsonEricAtSalesforce, line 418: The nullable annotation should have been on this method all along since it has two explicit paths to null.

In the case of this crash, the `Account` is not null or the calling method would have simply returned false with its existing guard.  So, the second null return in this method is:

```
if (authToken == null || instanceServer == null || userId == null || orgId == null) {
  return null;
}
```

That's been in place since about 2014, according to the annotated diff.  There's not much more detail as to why this was an expected condition, though.  At the least, now the calling code recognizes there is a nullable return and won't crash.

The follow-up work item in GUS would be a suitable time to investigate that logic in the wild, hopefully.  If anyone has a better suggestion to audit this logic do let me know.

libs/SalesforceSDK/src/com/salesforce/androidsdk/security/BiometricAuthenticationManager.kt
```diff
@@ -49,13 +49,17 @@ internal class BiometricAuthenticationManager: AppLockManager(
         get() { return SalesforceSDKManager.getInstance().userAccountManager.currentUser }
 
     override fun shouldLock(): Boolean {
+        val userAccountManager = SalesforceSDKManager.getInstance().userAccountManager
+
         val elapsedTime = System.currentTimeMillis() - lastBackgroundTimestamp
-        val account = SalesforceSDKManager.getInstance().userAccountManager.currentAccount ?: return false
-        val userAccount = SalesforceSDKManager.getInstance().userAccountManager.buildUserAccount(account)
+        val account = userAccountManager.currentAccount ?: return false
+        val userAccount = userAccountManager.buildUserAccount(account) ?: return false
```
JohnsonEricAtSalesforce, line 56: The new guard on this line is the crux of avoiding the crash.

brandonpage (approved):

wmathurin (approved):

PR #2543, opened 2024-04-05
Remove shouldReloadPage

brandonpage:
This singular variable has always controlled too many distinct scenarios:
- If the webview or custom tab should reload when the app returns from the background.
- If the webview or custom tab should reload when we transition from the ServerPicker to LoginActivity.
- If `Intent.FLAG_ACTIVITY_NO_HISTORY` should be set on the custom tab activity.

This leads to bad scenarios like: you need to _not_ reload on background for MFA so you sacrifice the correct server being loaded from the picker when you switch.  <- That sucks.

**I believe all of these scenarios only have one real correct choice and it is to **not reload** ever, unless the server changes.  If a reload is needed for some extreme case we cannot foresee the manual option remains.**

[This is the PR that originally added shouldReloadPage](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2033), which set it to true by default.  This was done to make it possible to transition between webview and custom tab, however that also requires the no history flag which breaks MFA.  To work around this I have decided to **relaunch** the LoginActivity instead of reloading the webview (which feels a little bad) to support this transition.  This is necessary since there is no way to programmatically close a custom tab 🙄.  The only time we need to reload the webview is after we fetch the auth config because we don't know if login should be in a custom tab or not until then.  

P.S: An added benefit is this removes a reason for apps to subclass `LoginActivity` and override the default `OAuthWebviewHelper`.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -106,7 +106,7 @@ import com.salesforce.androidsdk.rest.ClientManager.LoginOptions.fromBundleWithS
 import com.salesforce.androidsdk.security.BiometricAuthenticationManager
 import com.salesforce.androidsdk.security.BiometricAuthenticationManager.Companion.SHOW_BIOMETRIC
 import com.salesforce.androidsdk.ui.OAuthWebviewHelper.OAuthWebviewHelperEvents
-import com.salesforce.androidsdk.ui.ServerPickerActivity.CHANGE_SERVER_INTENT
+import com.salesforce.androidsdk.util.AuthConfigUtil.AUTH_CONFIG_COMPLETE_INTENT_ACTION
 import com.salesforce.androidsdk.util.EventsObservable
 import com.salesforce.androidsdk.util.EventsObservable.EventType.AuthWebViewCreateComplete
 import com.salesforce.androidsdk.util.EventsObservable.EventType.LoginActivityCreateComplete
@@ -134,7 +134,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
 
     private var webviewHelper: OAuthWebviewHelper? = null
 
-    private var changeServerReceiver: ChangeServerReceiver? = null
+    private var authConfigReceiver: AuthConfigReceiver? = null
 
     private var receiverRegistered = false
 
@@ -217,11 +217,11 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         )
         certAuthOrLogin()
         if (!receiverRegistered) {
-            changeServerReceiver = ChangeServerReceiver().also { changeServerReceiver ->
+            authConfigReceiver = AuthConfigReceiver().also { changeServerReceiver ->
                 registerReceiver(
                     this,
                     changeServerReceiver,
-                    IntentFilter(CHANGE_SERVER_INTENT),
+                    IntentFilter(AUTH_CONFIG_COMPLETE_INTENT_ACTION),
                     ContextCompat.RECEIVER_NOT_EXPORTED
                 )
             }
@@ -242,7 +242,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
 
     override fun onDestroy() {
         if (receiverRegistered) {
-            unregisterReceiver(changeServerReceiver)
+            unregisterReceiver(authConfigReceiver)
             receiverRegistered = false
         }
         super.onDestroy()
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -259,9 +259,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
 
         // Reload the login page for every new intent to ensure the correct login server is selected
         webviewHelper?.run {
-            if (shouldReloadPage) {
-                loadLoginPage()
-            }
+            loadLoginPage()
```
brandonpage, line 262: Alway reload when return from the ServerPicker.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -347,15 +345,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-        if (wasBackgrounded) {
-            webviewHelper?.run {
-                if (shouldReloadPage) {
-                    clearView()
-                    loadLoginPage()
-                }
-            }
-            wasBackgrounded = false
-        }
+        wasBackgrounded = false
```
brandonpage, line 348: Never reload when returning from the background.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -466,15 +453,6 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
...
         }
 
-        /*
-         * Prevent Chrome custom tab from staying in the activity history stack.
-         * This flag ensures that the Chrome custom tab is dismissed once the
-         * login process is complete
-         */
-        if (shouldReloadPage) {
-            customTabsIntent.intent.setFlags(Intent.FLAG_ACTIVITY_NO_HISTORY)
-        }
```
brandonpage, line 476: This is no longer needed since we switched to `singleTop`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
```diff
@@ -297,11 +299,27 @@ public void rebuildDisplay() {
...
+        }
+        intent.putExtras(options);
+
+        /*
+         * The activity needs to be created (instead of just reloading the webview)
+         * to support the transition from custom tab to standard webview login.
+         *
+         * The only other way to do this is with the NO_HISTORY flag on the custom tab, however
+         * this will cause it to always reload on background -- breaking MFA.
+         */
```
brandonpage, line 321: This comment explains most of the changes in the PR.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -254,14 +254,13 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         // If the intent is a callback from Chrome, process it and do nothing else
         if (isChromeCallback(intent)) {
             completeAuthFlow(intent)
+            webviewHelper?.clearView()
```
brandonpage, line 257: This was a small ask from Salesforce App.  Because we make the identity call after auth is complete there is a gap of time where the custom tab has been closed but we don't dismiss the login activity yet.  This change means the user sees a white screen instead of a login screen.  Perhpas the spinner should be shown here a well?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
```diff
@@ -297,11 +299,28 @@ public void rebuildDisplay() {
...
+
+        /*
+         * The activity needs to be recreated (instead of just reloading the webview)
+         * to support the transition from custom tab to standard webview login.  If we do this
+         * with the CLEAR_TOP flag we ensure unnecessary Custom Tabs are destroyed.
+         *
+         * The only other way to do this is with the NO_HISTORY flag on the custom tab, however
+         * this will cause it to always reload on background -- breaking MFA.
+         */
+        ctx.startActivity(intent);
```
brandonpage, line 323: This comment explains most of the PR.

JohnsonEricAtSalesforce (approved):
🎉 Thanks for taking the time to thoroughly research this and get some much needed cleanup done.  Looks great! 

⭐️⭐️⭐️⭐️⭐️

PR #2545, opened 2024-04-10
Delay push notification unregistration until device has network connectivity.

brandonpage:


JohnsonEricAtSalesforce (approved):
🤘🏻This highlights the earlier benefit of switching to the new Work Manager in modernization efforts.

brandonpage:
The failing tests are reproducible on dev. We can investigate and fix them before this or next week during testing.

wmathurin:
Will there be a problem if the revoke token call gets to the server before we de-register?

PR #2546, opened 2024-04-10
@W-14893505: [ASA Program] [#SA-01][Android] App Enables Dangerous File Access via setAllowFileAccessFromFileURLs

JohnsonEricAtSalesforce:
🎸 _Ready For Review!_ 🥁

  This removes file URL access from the login web view to comply with the security goal of this work item.  This was discussed among the team today and there doesn't seem to be a known reason for why this is enabled.  The login use cases I tried today all seem to function just as well without this option.  If anyone has login edge cases that might have been affected, do let me know and I'll investigate.  

  Thanks!


brandonpage (approved):
👍 Apparently this setting was deprecated with API 30 for being insecure.  

I don't think it will happen, but if someone gives us a good use case for it we can implement [androidx.webkit.WebViewAssetLoader](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader) instead.

brandonpage (approved):

PR #2548, opened 2024-04-11
Fix server picker issues with native login fallback to web based auth.

brandonpage:
Because of the new `CLEAR_TOP` flag we should explicitly use the Webview Login Activity Class when switching servers to avoid accidentally dismissing the webview activity.

I was tempted to convert `ServerPickerActivity` to Kotlin instead of making `webviewLoginActivityClass` public as it isn't much work.  However, customers can extend `ServerPickerActivity` and since a lot of it is Java public or protected we would need to carefully consider the external interface before and after.  I think it makes more sense to evaluate what the interface should be (which has not been done for may years), make deprecations if necessary and convert it in 13.0.


JohnsonEricAtSalesforce (approved):
🤘🏻

PR #2550, opened 2024-04-17
Fix "App was built for an older version of Android" issue for Sample Apps.

brandonpage:


JohnsonEricAtSalesforce (approved):
Thanks, @brandonpage! 🤘🏻

PR #2551, opened 2024-04-18
@W-15547113: [MSDK Android] Crash When Returning To Login View From Server Picker View

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

  This resolves what looks like an easy-to-fix crash when returning to login from the server picker.  The crash is due to `startActivity` being called on a `Context` reference that isn't always an `Activity` instance.  Since the method is already `ServerPickerActivity`, calling `startActivity` on `this` seems like it'd do just what the exception is asking for.

  @brandonpage - Does this align with the comment in this method?


brandonpage (approved):

brandonpage:
I must have been asleep that day since I just copied over the startActivity code from `BiometricAuthenticationManager`, which isn't an Activity but uses `Intent.FLAG_ACTIVITY_NEW_TASK` to avoid this crash.  I removed the new task flag because this is an activity... but forgot to use it's own context 🤦.

PR #2552, opened 2024-04-18
@W-15548296: [MSDK Android] NetworkOnMainThreadException In App Created From Hybrid Local Template

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

  This resolves a network on main thread exception that I discovered in the hybrid template apps.

  This is interesting in that the call to `getMyDomainAuthConfig` is happening in a different location than it does in `SalesforceSDKManager.fetchAuthenticationConfiguration` with very subtly different logic.  The key difference is that the hybrid app doesn't call `setBrowserLoginEnabled` when the authorization config is obtained.  @brandonpage - Any chance you know why that is? It looks intentional downstream in `isSamlLoginRedirect`.

  Perhaps in the future the switch around `setBrowserLoginEnabled` could be added to the `SalesforceSdkManager` version so redundancy is reduced.

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.kt
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.kt
@@ -26,7 +26,6 @@
  */
 package com.salesforce.androidsdk.phonegap.ui
 
-import android.net.Uri.parse
 import android.os.Bundle
 import android.view.KeyEvent
 import android.webkit.URLUtil.isHttpsUrl
@@ -57,7 +56,11 @@ import com.salesforce.androidsdk.util.AuthConfigUtil.MyDomainAuthConfig
 import com.salesforce.androidsdk.util.AuthConfigUtil.getMyDomainAuthConfig
 import com.salesforce.androidsdk.util.EventsObservable
 import com.salesforce.androidsdk.util.EventsObservable.EventType.GapWebViewCreateComplete
+import com.salesforce.androidsdk.util.SalesforceSDKLogger
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers.Default
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeout
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import org.apache.cordova.CallbackContext
 import org.apache.cordova.CordovaActivity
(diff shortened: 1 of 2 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.kt
```diff
@@ -581,21 +584,26 @@ open class SalesforceDroidGapActivity : CordovaActivity(), SalesforceActivityInt
...
-            ?.url
-            ?.trim { it <= ' ' } ?: return
-
-        if (loginServer == PRODUCTION_LOGIN_URL || loginServer == SANDBOX_LOGIN_URL || !isHttpsUrl(loginServer) || parse(loginServer) == null
-        ) {
-            return
-        }
-
-        authConfig = getMyDomainAuthConfig(loginServer)
+        CoroutineScope(Default).launch {
```
JohnsonEricAtSalesforce, line 587: This just catches the call to `getMyDomainAuthConfig` here up with the concurrency and logging code that's in the `SalesforceSdkManager` version.  It _is_ redundant, but does avoid any more intrusive change to the pending 12.0 release.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -564,6 +564,7 @@ open class SalesforceSDKManager protected constructor(
      * license
      * @return The Native Login Manager.
      */
+    @Suppress("unused")
```
JohnsonEricAtSalesforce, line 567: Unrelated, but this resolves an inspector warning.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1399,12 +1400,12 @@ open class SalesforceSDKManager protected constructor(
      * Determines whether the device has a compact screen.
      * Taken directly from https://developer.android.com/guide/topics/large-screens/large-screen-cookbook#kotlin
      */
-    fun compactScreen(activity: Activity) : Boolean {
+    fun compactScreen(activity: Activity): Boolean {
```
JohnsonEricAtSalesforce, line 1403: Also a rider to the bill, however the rest of this source is formatted to Android Studio defaults already.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1443,6 +1444,7 @@ open class SalesforceSDKManager protected constructor(
          * TODO: Remove the suppress lint annotation once the Android context is no longer retained.
          */
         @JvmField
+        @SuppressLint("StaticFieldLeak")
```
JohnsonEricAtSalesforce, line 1447: I'd thought this suppress was in place quite some time ago, since there's a TODO already around the bigger issue present here.

brandonpage (approved):
I didn't realize this existed and am not sure why it needs to.  Seems like it should just use the similar (probably identical at one point) code in the `SalesforceSDK` library.  Let's make a WI for that for a future release.

PR #2563, opened 2024-05-03
Move to sqlcipher 4.5.7

wmathurin:


brandonpage (approved):

.circleci/gitChangedLibs.rb
```diff
@@ -19,9 +19,10 @@
 
 for prfile in prfiles
   path = prfile["filename"]
-  for lib in libsTopoSorted
-    if path.include? lib
-      libsModified.add(lib)
+  libsTopoSorted.each do |lib|
```
wmathurin, line 22: @brandonpage what do you think?
We would run tests in dependent libraries too.

brandonpage: My ruby is a little rusty but conceptually I like it.  👍

brandonpage (approved):

wmathurin:
Only Salesforce React tests fail - but for memory issues unrelated to SQLCipher change.

PR #2567, opened 2024-05-09
@W-15061223: [Android] Add additional native login feature: forgot password

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

Here MSDK acquires the headless forgot password flow from the Salesforce Identity API.

https://help.salesforce.com/s/articleView?id=sf.remoteaccess_headless_forgot_password_flow.htm&type=5

This follows the style we used for login and password-less login in 12.0.  Most of the new code should look quite familiar with the subtle exception of some new utility methods to keep reused logic from becoming redundant.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -641,7 +641,7 @@ internal class NativeLoginManager(
     )
 
     /**
-     * Generates either the reCAPTCHA token parameter for non- enterprise
+     * Generates either the reCAPTCHA token parameter for non-enterprise
      * reCAPTCHA configurations or the reCAPTCHA event parameter for enterprise
      * reCAPTCHA configurations.
      * @param reCaptchaToken A reCAPTCHA token provided by the reCAPTCHA SDK
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -286,7 +286,7 @@ internal class NativeLoginManager(
                 username = trimmedUsername
             ).toJson()
         }.onFailure { e ->
-            SalesforceSDKLogger.e(TAG, "Cannot JSON encode OTP request body due to an encoding error with message '${e.message}'.", e)
+            SalesforceSDKLogger.e(TAG, "Cannot JSON encode start password reset request body due to an encoding error with message '${e.message}'.", e)
             return UnknownError
         }.getOrNull()
 
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -376,7 +376,7 @@ internal class NativeLoginManager(
     // endregion
     // region Salesforce Identity API Headless, Password-Less Login Via One-Time-Passcode
 
-    override suspend fun startPasswordLessAuthorization(
+    override suspend fun submitOtpRequest(
         username: String,
         reCaptchaToken: String,
         otpVerificationMethod: OtpVerificationMethod
@@ -395,7 +395,7 @@ internal class NativeLoginManager(
         )
         // Generate the start password-less login request body.
         val startPasswordLessLoginRequestBodyString = runCatching {
-            StartPasswordLessLoginRequestBody(
+            OtpRequestBody(
                 recaptcha = reCaptchaParameterGenerationResult.nonEnterpriseReCaptchaToken,
                 recaptchaevent = reCaptchaParameterGenerationResult.enterpriseReCaptchaEvent,
                 username = trimmedUsername,
@@ -439,21 +439,7 @@ internal class NativeLoginManager(
         }
     }
 
-    @Deprecated(
-        "The replacement function provided in 12.1.0 uses a more consistent naming convention.",
-        ReplaceWith("submitStartPasswordLessAuthorizationRequest(username, reCaptchaToken, otpVerificationMethod)")
-    )
-    override suspend fun submitOtpRequest(
-        username: String,
-        reCaptchaToken: String,
-        otpVerificationMethod: OtpVerificationMethod
-    ): OtpRequestResult = startPasswordLessAuthorization(
-        username,
-        reCaptchaToken,
-        otpVerificationMethod
-    )
-
-    override suspend fun completePasswordLessAuthorization(
+    override suspend fun submitPasswordlessAuthorizationRequest(
         otp: String,
         otpIdentifier: String,
         otpVerificationMethod: OtpVerificationMethod
@@ -512,20 +498,6 @@ internal class NativeLoginManager(
         )
     }
 
-    @Deprecated(
-        "The replacement function provided in 12.1.0 uses a more consistent naming convention.",
-        ReplaceWith("submitCompletePasswordLessAuthorizationRequest(otp, otpIdentifier, otpVerificationMethod)")
-    )
-    override suspend fun submitPasswordlessAuthorizationRequest(
-        otp: String,
-        otpIdentifier: String,
-        otpVerificationMethod: OtpVerificationMethod
-    ) = completePasswordLessAuthorization(
-        otp,
-        otpIdentifier,
-        otpVerificationMethod
-    )
-
     /**
      * A data class for the start password reset request body.
      * @param recaptcha The reCAPTCHA token provided by the reCAPTCHA Android
@@ -578,7 +550,7 @@ internal class NativeLoginManager(
      * @param verificationMethod The OTP verification code's delivery method in
      * "email" or "sms"
      */
-    private data class StartPasswordLessLoginRequestBody(
+    private data class OtpRequestBody(
         val recaptcha: String?,
         val recaptchaevent: ReCaptchaEventRequestParameter?,
         val username: String,
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -46,6 +46,7 @@ import com.salesforce.androidsdk.auth.OAuth2.CODE_VERIFIER
 import com.salesforce.androidsdk.auth.OAuth2.GRANT_TYPE
 import com.salesforce.androidsdk.auth.OAuth2.HYBRID_AUTH_CODE
 import com.salesforce.androidsdk.auth.OAuth2.OAUTH_AUTH_PATH
+import com.salesforce.androidsdk.auth.OAuth2.OAUTH_ENDPOINT_HEADLESS_FORGOT_PASSWORD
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -123,13 +124,8 @@ internal class NativeLoginManager(
...
         val trimmedPassword = password.trim()
 
-        if (!isValidUsername(trimmedUsername)) {
-            return InvalidUsername
-        }
-
-        if (!isValidPassword(trimmedPassword)) {
-            return InvalidPassword
-        }
+        trimmedUsername.mapInvalidUsernameToResult()?.let { return it }
```
JohnsonEricAtSalesforce, line 127: These parameter validations are used in a lot of places now, so I made them into a one-liner.  The original method is still used and tested, so this is just optional syntactic sugar.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
         return requestBodyString.toRequestBody(mediaType)
     }
 
-    // region Headless, Password-Less Login Via One-Time-Passcode
+    // region Salesforce Identity API Headless Forgot Password Flow
```
JohnsonEricAtSalesforce, line 269: I made a few comment and naming updates to really call any open-source reader's attention to the correct location in our official Salesforce Identity API docs.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
...
 
-    override suspend fun submitOtpRequest(
```
JohnsonEricAtSalesforce, line 275: Github's diff engine misaligned this part of the diff.  The logic of `submitOtpRequest` is simpler further in the source.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
...
+    override suspend fun startPasswordReset(
```
JohnsonEricAtSalesforce, line 271: Now that I'm getting used to these two-step identity flows, I decided on a really clear "start" and "complete" naming prefix.  That seems to make it really obvious which method goes where in the app's control flow.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
...
-        if (!isValidUsername(username.trim())) {
-            return OtpRequestResult(InvalidUsername)
-        }
-        val reCaptchaSiteKeyId = when {
-            reCaptchaSiteKeyId == null -> {
-                SalesforceSDKLogger.e(TAG, "A reCAPTCHA site key wasn't and must be provided when using enterprise reCAPATCHA.")
-                return OtpRequestResult(UnknownError)
+        val trimmedUsername = username.trim()
+        trimmedUsername.mapInvalidUsernameToResult()?.let { return it }
+        val reCaptchaParameterGenerationResult = generateReCaptchaParameters(reCaptchaToken)
```
JohnsonEricAtSalesforce, line 279: The logic of validating and choosing between the Salesforce Identity API's reCAPTCHA parameters is quite a bit of code I didn't want to duplicate.  So, here's a dandy method to handle that where ever we need to.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
...
+                UnknownError
+            }
         }
-        val googleCloudProjectId = when {
-            googleCloudProjectId == null -> {
-                SalesforceSDKLogger.e(TAG, "A Google Cloud project id wasn't and must be provided when using enterprise reCAPATCHA.")
-                return OtpRequestResult(nativeLoginResult = UnknownError)
+    }
+
+    override suspend fun completePasswordReset(
```
JohnsonEricAtSalesforce, line 322: Here's the `completePasswordReset` method to match the `startPasswordReset` method.  Everything here follows the patterns we used in 12.0 for other native login methods.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
...
-        val reCaptchaEventParameter = when {
-            isReCaptchaEnterprise -> OtpRequestBodyReCaptchaEvent(
-                token = reCaptchaToken,
-                siteKey = reCaptchaSiteKeyId,
-                projectId = googleCloudProjectId
-            )
+    // endregion
+    // region Salesforce Identity API Headless, Password-Less Login Via One-Time-Passcode
+
+    override suspend fun startPasswordLessAuthorization(
```
JohnsonEricAtSalesforce, line 379: Here's a question: In 12.0 I rather naively named this `submitOtpRequest`.  Now I realize that's very ambiguous in the larger scope of multiple Salesforce Identity flows that will request and use an OTP.  Also, it doesn't help the app developer know which comes first in the start and completion of this two-step flow.

Without breaking the API in 12.1, I deprecated the former method and made it a proxy to call the newer method.  The app code is _much_ more intuitive afterwards.  Any objections to that in general or in a point release?  It might be nice to do this before adoption takes off.

I could roll this back if we feel it's a blocker.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -430,23 +512,78 @@ internal class NativeLoginManager(
...
+     * SDK.  This is not used with reCAPTCHA Enterprise
+     * @param recaptchaevent The reCAPTCHA parameters for use with reCAPTCHA
+     * Enterprise
+     * @param username A valid Salesforce username.  Note that email may be used
+     * for community users
      * @param verificationMethod The OTP verification code's delivery method in
      * "email" or "sms"
      */
-    private data class OtpRequestBody(
+    private data class StartPasswordLessLoginRequestBody(
```
JohnsonEricAtSalesforce, line 581: In favor of renaming the password-less login methods and data classes, there's not a generic "OTP Request" in the API.  Each endpoint and step in the flow has a distinct object graph.

JohnsonEricAtSalesforce (commented):
Here's a quick self-review with introductory comments to kick off the review.  Thanks for reading.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
brandonpage: As I said in my other comment I am not a huge fan or renaming immediately.  If we are going to rename I just think we should consider multiple options and potentially have a more cohesive naming scheme for all the interface APIs.  We should also consider that we could wait an rename these in the future if/when this functionality expands beyond communities -- which will require us to at least update if not recreate the API.

brandonpage (changes requested):
The code LGTM, but I am not yet sold on the API renaming.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
JohnsonEricAtSalesforce: I simply reverted this in the interest of expediency 🏁

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/interfaces/NativeLoginManager.kt
```diff
@@ -72,12 +72,55 @@ interface NativeLoginManager {
...
-     * password-less login flow. This fulfills step three of the headless
-     * password-less login flow.
+     * Submits a request to start a password reset to the Salesforce Identity
+     * API headless forgot password flow.  This fulfills step one of the
+     * headless forgot password flow.
+     *
+     * See https://help.salesforce.com/s/articleView?id=sf.remoteaccess_headless_forgot_password_flow.htm&type=5
+     *
+     * @param username A valid Salesforce username.  Note that email may be used
+     * for community users
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: It was worded something like this in the Identity API docs, but your suggestion makes perfect sense to me 👍🏻

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/interfaces/NativeLoginManager.kt
```diff
@@ -88,27 +131,49 @@ interface NativeLoginManager {
...
+        username: String,
+        reCaptchaToken: String,
+        otpVerificationMethod: OtpVerificationMethod
+    ): OtpRequestResult
+
+    @Deprecated(
+        "Use the replacement function provided in 12.1.0 for more consistent naming conventions.",
+        ReplaceWith("startPasswordLessAuthorization(username, reCaptchaToken, otpVerificationMethod)")
+    )
     suspend fun submitOtpRequest(
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I reverted this one to move our cycle time along 🚀

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -270,108 +266,194 @@ internal class NativeLoginManager(
...
+        }.onFailure { e ->
+            SalesforceSDKLogger.e(TAG, "Cannot JSON encode start password reset request body due to an encoding error with message '${e.message}'.", e)
+            return UnknownError
+        }.getOrNull()
+
+        // Create the start password reset request.
+        val startPasswordResetRequest = RestRequest(
+            POST,
+            LOGIN,
+            "$loginUrl$OAUTH_ENDPOINT_HEADLESS_FORGOT_PASSWORD",
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: Either way, we ought to be consistent in how strings are assembled.  I lean towards Kotlin's templates since that's what the IDE inspector it guiding the community at large to.  I'll quickly switch if needed here to move this through, though.

wmathurin:
For the API rename, maybe introduce the new one, keep the old one (have it call the new one) and mark the old one as deprecated `@Deprecated("Use the ... method - will be removed in 13.0", replaceWith = ReplaceWith("..."))`.
The template can use the new one.

JohnsonEricAtSalesforce:
> For the API rename, maybe introduce the new one, keep the old one (have it call the new one) and mark the old one as deprecated `@Deprecated("Use the ... method - will be removed in 13.0", replaceWith = ReplaceWith("..."))`. The template can use the new one.

I already rolled it back for now, since it was extra credit polish to the developer experience.  We can revisit another day.

wmathurin (approved):

PR #2574, opened 2024-06-13
Add Logout Reason to Revoke API call.

brandonpage:
I have elected not to add support for passing the logout reason from hybrid or React Native apps because the audience for this change is internal apps and none of them use those libraries anymore.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -606,6 +606,9 @@ public class ClientManager {
                         	if (Looper.myLooper() == null) {
                                 Looper.prepare();
                         	}
+                            // Note: As of writing (2024) this call will never succeed because revoke API is an
+                            // authenticated endpoint.  However, there is no harm in attempting and the debug logs
+                            // produced may help developers better understand the state of their app.
                             SalesforceSDKManager.getInstance()
                                     .logout(null, null, false, OAuth2.LogoutReason.TOKEN_EXPIRED);
                         }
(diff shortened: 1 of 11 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -606,7 +606,8 @@ public String getNewAuthToken() {
                         	if (Looper.myLooper() == null) {
                                 Looper.prepare();
                         	}
-                            SalesforceSDKManager.getInstance().logout(null, false);
+                            SalesforceSDKManager.getInstance()
+                                    .logout(null, null, false, OAuth2.LogoutReason.TOKEN_EXPIRED);
```
wmathurin, line 610: When the refresh token has expired, there won't be a (successful) client call to the server to revoke the token and therefore the logout reason will never be sent to the server. But the server should have created a log line when the refresh token expires.

TLDR: TOKEN_EXPIRED reason can never be logged by the client only by the server

wmathurin: That being said, it's fine to use that reason here (even though it won't leave the client).

brandonpage: Great callout.  I think I will leave it as there are no negative side affects but I will add a comment noting this.

bbirman: I'll update on iOS too

wmathurin (approved):

brandonpage:
I have fixed he (unrelated) broken `SalesforceSDK` tests in #2575.

PR #2576, opened 2024-06-14
Add LogoutReason to logout events.

brandonpage:


wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/LogoutCompleteReceiver.kt
```diff
@@ -24,29 +24,31 @@
...
+            onLogoutComplete(LogoutReason.valueOf(reason.uppercase(Locale.ROOT)))
         }
     }
 
-    protected abstract void onLogoutComplete();
+    /** @deprecated Will be removed in 13.0.  Use the new [onLogoutComplete] that provides a [LogoutReason] instead. */
+    protected abstract fun onLogoutComplete()
+
+    // TODO: Make this abstract in 13.0 to replace the above onLogoutComplete()
+    protected open fun onLogoutComplete(reason: LogoutReason) { }
```
brandonpage, line 53: Reasoning:  making a function abstract makes it **required** for implementers of the class.  This is technically a breaking change.  So my thought process here is to make the new function concrete so that it is optional, for now.  

If we made the old function concrete it would spare teams from having to implement it [and the deprecation warning) after moving to the new function.  This doesn't seem necessary but let me know your opinions.

wmathurin (approved):

PR #2577, opened 2024-06-19
Add CodeCov.

brandonpage:


brandonpage:
The cleanest way to test this (without redoing everything on my fork) would probably be to merge this and send another test PR so that the CI actually runs.  If anyone feels strongly about testing this fully before merging let me know, but I will note that the settings used in `codecov.yml` are the same as they were before we removed CodeCov in 9.1.0.

wmathurin (approved):

PR #2585, opened 2024-06-25
Convert Firebase Test Lab code coverage for CodeCov.

brandonpage:
Apparently [CodeCov no longer supports .ec files](https://docs.codecov.com/docs/supported-report-formats), which is the only thing we can generate in Firebase Test Cloud.  So I wrote a gradle task to convert the unsupported format to XML CodeCov can process (and html we can read) with JoCoCo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,6 +171,10 @@ jobs:
               exit 1
             fi
           when: always
+      - run: 
+          name: Convert Code Coverage
+          command: ./gradlew libs:<< parameters.lib >>:convertCodeCoverage
+          when: always
       - when:
           condition: << parameters.pr >>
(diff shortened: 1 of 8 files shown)


brandonpage:
[Coverage exists!](https://output.circle-artifacts.com/output/job/d8b0159b-0d95-4636-b8bd-891591d83d18/artifacts/0/libs/SalesforceHybrid/build/reports/jacoco/convertedCodeCoverage/html/com.salesforce.androidsdk.phonegap.plugin/index.html)

This PR needs to be merged so we can get a baseline via a nightly build on dev, then the nicer CodeCov site will show results and compare subsequent PRs to that.

wmathurin (approved):

PR #2586, opened 2024-06-26
ReactNativeMap.isNull throws if the provided key is not in the map

wmathurin:
It was not always like that - but that change of behavior means some of the bridge method might not work if an optional argument is not provided. I noticed the issue because testRemoveFromSoup started failing.


bbirman (approved):

wmathurin:
SalesforceReact tests all passing (that's the only change in this PR).

PR #2587, opened 2024-06-26
@W-15993541/W-15993636: [Android] Publish AILTN logs on background/Register push notifications on foreground

JohnsonEricAtSalesforce:
🎸 _Ready For Review!_ 🥁

Note: This also includes [W-15993636](https://gus.lightning.force.com/a07EE00001uKF3pYAG) since the two topics look so similar in the code and test well together.

This changes MSDK to avoid resuming the app for background tasks that could incur additional authorization refresh events.  The default behavior is now to perform those tasks when the app resumes or pauses (foreground or background).  Configuration options are provided so app can opt-in to the old behavior, if desired.

I was able to test the apps to verify the new AILTN publishing in each configuration.  Also, I was able to test push notifications registration according to each configuration.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
@@ -27,14 +27,17 @@
 package com.salesforce.androidsdk.analytics
 
 import android.content.Context
+import androidx.work.Constraints
 import androidx.work.ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE
 import androidx.work.ListenableWorker.Result.success
-import androidx.work.PeriodicWorkRequest.Builder
+import androidx.work.NetworkType.CONNECTED
+import androidx.work.OneTimeWorkRequest
+import androidx.work.PeriodicWorkRequest
 import androidx.work.WorkManager.getInstance
 import androidx.work.Worker
 import androidx.work.WorkerParameters
 import com.salesforce.androidsdk.accounts.UserAccountManager
-import com.salesforce.androidsdk.analytics.AnalyticsPublishingWorker.Companion.reEnqueueAnalyticsPublishPeriodicWorkRequest
+import com.salesforce.androidsdk.analytics.AnalyticsPublishingWorker.Companion.enqueueAnalyticsPublishWorkRequest
 import java.util.concurrent.TimeUnit.HOURS
 
 /**
@@ -42,11 +45,17 @@ import java.util.concurrent.TimeUnit.HOURS
  * This class is intended to be instantiated by the background tasks work
  * manager.
  *
- * Use [reEnqueueAnalyticsPublishPeriodicWorkRequest] to enqueue a analytics
- * publish periodic work request.  Previous requests will be cancelled.
+ * [enqueueAnalyticsPublishWorkRequest] is used internally by the Salesforce
+ * Mobile SDK to enqueue analytics publish work requests to match the current
+ * analytics publishing configuration. Only one request may be active at a time.
+ * Only the Salesforce Mobile SDK should call this method as it is not intended
+ * for public use.
  *
  * @param context The Android context provided by the work manager
  * @param workerParams The worker parameters provided by the work manager
+ * @see [SalesforceAnalyticsManager.setPublishOnceTimeOnAppBackgroundEnabled]
+ * @see [SalesforceAnalyticsManager.setPublishPeriodicallyOnFrequencyEnabled]
+ * @see [SalesforceAnalyticsManager.setPublishPeriodicallyFrequencyHours]
  * @see <a href='https://developer.android.com/guide/background'>Android
  * Background Tasks</a>
  */
@@ -73,36 +82,67 @@ internal class AnalyticsPublishingWorker(
 
     companion object {
 
-        private const val publishAnalyticsPeriodicWorkName = "SalesforceAnalyticsPublishingPeriodicWork"
+        /**
+         * The Android background tasks name of the publish analytics work
+         * request.
+         * Note: This name is also used for one-time work, yet maintains this
+         * value for backwards compatibility and to ensure both periodic and one
+         * time work are mutually exclusive.
+         */
+        private const val PUBLISH_ANALYTICS_WORK_NAME = "SalesforceAnalyticsPublishingPeriodicWork"
 
         /**
-         * Enqueues a persistent background tasks periodic work request to
-         * publish stored analytics for the current user at the provided
-         * interval.
+         * Enqueues a persistent background tasks work request to publish stored
+         * analytics for the current user.
          *
          * If a work request is already queued, it will be cancelled before
          * the replacement is enqueued.
          *
+         * The publish hours interval parameter determines between background
+         * periodic publishing and one-time publishing.  Note, background
+         * periodic publishing starts or resumes the host app and may incur
+         * licensing costs if authorization token refresh is required.
+         *
+         * Only the Salesforce Mobile SDK should call this method as it is not
+         * intended for public use.
+         *
          * @param context The Android context
-         * @param publishHoursInterval The interval at which to publish in hours
+         * @param periodicBackgroundPublishingHoursInterval The interval for
+         * periodic background publishing in hours or null to publish one time
+         * only
          * @return UUID The worker's unique id, which may be used for
          * cancellation
          */
-        fun reEnqueueAnalyticsPublishPeriodicWorkRequest(
+        fun enqueueAnalyticsPublishWorkRequest(
             context: Context,
-            publishHoursInterval: Long
-        ) = Builder(
-            AnalyticsPublishingWorker::class.java,
-            publishHoursInterval,
-            HOURS
-        ).build().also { publishAnalyticsPeriodicWorkRequest ->
-            runCatching {
-                getInstance(context)
-            }.getOrNull()?.enqueueUniquePeriodicWork(
-                publishAnalyticsPeriodicWorkName,
-                CANCEL_AND_REENQUEUE,
-                publishAnalyticsPeriodicWorkRequest
-            )
-        }.id
+            periodicBackgroundPublishingHoursInterval: Long? = null
+        ) = when (periodicBackgroundPublishingHoursInterval) {
+
+            null -> OneTimeWorkRequest.Builder(
+                AnalyticsPublishingWorker::class.java
+            ).setConstraints(
+                Constraints.Builder().setRequiredNetworkType(CONNECTED).build()
+            ).build().also { publishAnalyticsOneTimeWorkRequest ->
+                runCatching {
+                    getInstance(context)
+                }.getOrNull()?.enqueue(publishAnalyticsOneTimeWorkRequest)
+            }.id
+
+            else -> PeriodicWorkRequest.Builder(
+                AnalyticsPublishingWorker::class.java,
+                periodicBackgroundPublishingHoursInterval,
+                HOURS
+            ).setConstraints(
+                Constraints.Builder().setRequiredNetworkType(CONNECTED).build()
+            ).build().also { publishAnalyticsPeriodicWorkRequest ->
+                runCatching {
+                    getInstance(context)
+                }.getOrNull()?.enqueueUniquePeriodicWork(
+                    PUBLISH_ANALYTICS_WORK_NAME,
+                    CANCEL_AND_REENQUEUE,
+                    publishAnalyticsPeriodicWorkRequest
+                )
+            }.id
+        }
     }
 }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
@@ -188,7 +188,10 @@ public class SalesforceAnalyticsManager {
      *
      * @param periodicBackgroundPublishingHoursInterval The interval for
      *                                                  periodic background
-     *                                                  publishing in hours
+     *                                                  publishing in hours. It
+     *                                                  is recommended to keep
+     *                                                  this value under seven
+     *                                                  days
      * @see #setPublishPeriodicallyOnFrequencyEnabled(boolean)
      */
     public static synchronized void setPublishPeriodicallyFrequencyHours(
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
@@ -70,8 +70,10 @@ public class SalesforceAnalyticsManager {
     private static final String UNAUTH_INSTANCE_KEY = "_no_user";
 
     private static Map<String, SalesforceAnalyticsManager> INSTANCES;
-    private static boolean sPublishHandlerActive;
-    private static int sPublishFrequencyInHours = DEFAULT_PUBLISH_FREQUENCY_IN_HOURS;
+    private static boolean isPublishWorkRequestEnqueued;
+    private static boolean isPublishOneTimeOnAppBackgroundEnabled = true;
+    private static boolean isPublishPeriodicallyOnFrequencyEnabled = false;
+    private static int publishPeriodicallyFrequencyHours = DEFAULT_PUBLISH_FREQUENCY_IN_HOURS;
     private static int sEventPublishBatchSize = DEFAULT_BATCH_SIZE;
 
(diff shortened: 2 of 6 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -182,13 +184,77 @@ public static synchronized void reset(UserAccount account, String communityId) {
...
+        setPublishPeriodicallyOnFrequencyEnabled(true);
+    }
+
+    /**
+     * Sets the interval for periodic background publishing in hours.
+     *
+     * @deprecated Planned for removal 13.0.
+     * Use {@link #setPublishPeriodicallyFrequencyHours(int)} )}.
+     */
+    @Deprecated()
```
JohnsonEricAtSalesforce, line 207: Without refactoring to make the names consistent, I though the new code wouldn't be very intuitive.  So, here are some deprecations that I believe will make the implementation more friendly while keeping compatibility with 12.0.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -898,7 +901,12 @@ open class SalesforceSDKManager protected constructor(
         frontActivity: Activity?,
         showLoginPage: Boolean = true,
     ) {
-        logout(account, frontActivity, showLoginPage)
+        logout(
+            account = account,
+            frontActivity = frontActivity,
+            showLoginPage = showLoginPage,
+            reason = LogoutReason.UNKNOWN
```
JohnsonEricAtSalesforce, line 908: @brandonpage - This resolves the stack overflow on logout we discussed earlier.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1465,6 +1481,17 @@ open class SalesforceSDKManager protected constructor(
         screenLockManager?.onAppForegrounded()
         (biometricAuthenticationManager as? BiometricAuthenticationManager)?.onAppForegrounded()
 
+        // Review push-notifications registration for the current user, if enabled.
+        userAccountManager.currentUser?.let { userAccount ->
```
JohnsonEricAtSalesforce, line 1485: This is MSDK's internal call for one-time-on-foreground push notifications registration.  It's wrapped with a check for the current user since I believe that's the key condition.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
```diff
@@ -184,7 +184,6 @@ object PushMessaging {
             val packageInfo = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
                 context.packageManager.getPackageInfo(context.packageName, PackageManager.PackageInfoFlags.of(0))
             } else {
-                @Suppress("DEPRECATION")
```
JohnsonEricAtSalesforce, line 187: This deprecation was unneeded.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -127,18 +130,17 @@ open class PushService {
         }
 
         /*
-         * Although each user account should be registered with the SFDC API
```
JohnsonEricAtSalesforce, line 130: For my own satisfaction, it's cool to finally understand this requirement.

JohnsonEricAtSalesforce (commented):
Here are some self-review notes to start the conversation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -437,37 +462,64 @@ open class PushService {
...
+         * @param action The push notifications registration action
+         * @param delayDays For registration actions, the interval in days
+         * between periodic registration or null for immediate one-time
+         * registration
          */
         internal fun enqueuePushNotificationsRegistrationWork(
             userAccount: UserAccount?,
             action: PushNotificationsRegistrationAction,
-            delayMilliseconds: Long?
+            delayDays: Long?
```
brandonpage, line 484: Days or hours?

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -73,36 +82,67 @@ internal class AnalyticsPublishingWorker(
...
-        ).build().also { publishAnalyticsPeriodicWorkRequest ->
-            runCatching {
-                getInstance(context)
-            }.getOrNull()?.enqueueUniquePeriodicWork(
-                publishAnalyticsPeriodicWorkName,
-                CANCEL_AND_REENQUEUE,
-                publishAnalyticsPeriodicWorkRequest
-            )
-        }.id
+            periodicBackgroundPublishingHoursInterval: Long? = null
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: Well, that's a great question.  It turns out that's just a carry through of the type Android Background Tasks expects for its repeat interval parameter.  I could make our parameter something else and cast it to long when passed to the work request builder, but I'm not sure we'd be adding value in that.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -73,36 +82,67 @@ internal class AnalyticsPublishingWorker(
...
+                AnalyticsPublishingWorker::class.java
+            ).setConstraints(
+                Constraints.Builder().setRequiredNetworkType(CONNECTED).build()
+            ).build().also { publishAnalyticsOneTimeWorkRequest ->
+                runCatching {
+                    getInstance(context)
+                }.getOrNull()?.enqueue(publishAnalyticsOneTimeWorkRequest)
+            }.id
+
+            else -> PeriodicWorkRequest.Builder(
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I actually did test that.  Android Background Tasks provides a reasonable default in `MIN_PERIODIC_INTERVAL_MILLIS` which is currently set to fifteen minutes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -102,7 +104,7 @@ public static synchronized SalesforceAnalyticsManager getInstance(UserAccount ac
     /**
      * Returns the instance of this class associated with this user and community.
      *
-     * @param account User account.
+     * @param account     User account.
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I occasionally try re-formatting existing sources to Android Studio's current defaults.  If the changes are minimal, I've been game to include a little cleanup.  I could undo this if we want.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
JohnsonEricAtSalesforce: I went for days here since this value is private to our SDK and always set to six days.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -182,13 +184,77 @@ public static synchronized void reset(UserAccount account, String communityId) {
...
+     * Sets the interval for periodic background publishing in hours.
+     *
+     * @param periodicBackgroundPublishingHoursInterval The interval for
+     *                                                  periodic background
+     *                                                  publishing in hours
+     * @see #setPublishPeriodicallyOnFrequencyEnabled(boolean)
+     */
+    public static synchronized void setPublishPeriodicallyFrequencyHours(
+            int periodicBackgroundPublishingHoursInterval
+    ) {
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: That's a neat suggestion.  One one hand, Android Background Tasks will automatically switch to a reasonable minimum of fifteen minutes and provides that for us.  Do we have an upper bound we'd want to enforce and what would it be?  I could get that in.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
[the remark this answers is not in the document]
brandonpage: Fair enough.  As long as we are validating the input Long vs Int vs whatever doesn't really matter here if it eventually needs to be an Long.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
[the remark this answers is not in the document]
brandonpage: Nice.  I assume the same min value is used for negative input?

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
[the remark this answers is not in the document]
brandonpage: Our original 6 day value might be a reasonable upper limit since it is not common knowledge that unused push tokens are removed after 7 days.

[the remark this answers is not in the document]
brandonpage: @JohnsonEricAtSalesforce  If we don't put an upper limit we should mention in the method description that unused registrations are removed after 7 days so the customer is aware of the risks of using a larger value.

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -27,26 +27,35 @@
...
- * publish periodic work request.  Previous requests will be cancelled.
+ * [enqueueAnalyticsPublishWorkRequest] is used internally by the Salesforce
+ * Mobile SDK to enqueue analytics publish work requests to match the current
+ * analytics publishing configuration. Only one request may be active at a time.
+ * Only the Salesforce Mobile SDK should call this method as it is not intended
+ * for public use.
  *
  * @param context The Android context provided by the work manager
  * @param workerParams The worker parameters provided by the work manager
+ * @see [SalesforceAnalyticsManager.setPublishOnceTimeOnAppBackgroundEnabled]
```
wmathurin, line 56: NIT should the name be `setPublishOnceOnAppBackgroundEnabled` or `setPublishOneTimeOnAppBackgroundEnabled` instead?

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
```diff
@@ -70,8 +70,10 @@ public class SalesforceAnalyticsManager {
     private static final String UNAUTH_INSTANCE_KEY = "_no_user";
 
     private static Map<String, SalesforceAnalyticsManager> INSTANCES;
-    private static boolean sPublishHandlerActive;
-    private static int sPublishFrequencyInHours = DEFAULT_PUBLISH_FREQUENCY_IN_HOURS;
+    private static boolean isPublishWorkRequestEnqueued;
+    private static boolean isPublishOneTimeOnAppBackgroundEnabled = true;
```
wmathurin, line 74: Do we want to support isPublishOneTimeOnAppBackgroundEnabled and isPublishPeriodicallyOnFrequencyEnabled both flags being true (or false). Should we instead have an enum with three values no-publishing, publish-on-background, publish-periodically?

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -73,36 +82,67 @@ internal class AnalyticsPublishingWorker(
...
-            }.getOrNull()?.enqueueUniquePeriodicWork(
-                publishAnalyticsPeriodicWorkName,
-                CANCEL_AND_REENQUEUE,
-                publishAnalyticsPeriodicWorkRequest
-            )
-        }.id
+            periodicBackgroundPublishingHoursInterval: Long? = null
+        ) = when (periodicBackgroundPublishingHoursInterval) {
+
+            null -> OneTimeWorkRequest.Builder(
```
wmathurin, line 121: Should we do that if isPublishOneTimeOnAppBackgroundEnabled is true?

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
```diff
@@ -27,26 +27,35 @@
...
+ * Mobile SDK to enqueue analytics publish work requests to match the current
+ * analytics publishing configuration. Only one request may be active at a time.
+ * Only the Salesforce Mobile SDK should call this method as it is not intended
+ * for public use.
  *
  * @param context The Android context provided by the work manager
  * @param workerParams The worker parameters provided by the work manager
+ * @see [SalesforceAnalyticsManager.setPublishOnceTimeOnAppBackgroundEnabled]
+ * @see [SalesforceAnalyticsManager.setPublishPeriodicallyOnFrequencyEnabled]
+ * @see [SalesforceAnalyticsManager.setPublishPeriodicallyFrequencyHours]
```
wmathurin, line 58: I wonder if we should support three modes (and therefore have three methods):
- no publishing: turnOffPublishing()
- publish on background: publishOnceOnAppBackground()
- publish periodically with period x: publishPeriodically(frequencyHours)

Right now it looks like you could publish on background and periodically, or you could publish periodically without specifying a frequency.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -413,8 +414,32 @@ open class PushService {
...
-        private const val MILLISECONDS_IN_SIX_DAYS = 518400000L
+        /**
+         * Specifies if push notification (re-)registration should occur one
+         * time when the app is brought to the foreground.
+         *
+         * This is mutually exclusive with
+         * [isPushNotificationsRegisterPeriodicallyEnabled] and is the
+         * default option of the two.
+         */
+        var isPushNotificationsRegistrationOneTimeOnAppForegroundEnabled = true
```
wmathurin, line 425: Same comment as above. Instead of two separate flags (which gives us 4 possible states), could we have an enum with 3 values: no-re-registration, re-registration on foreground, periodic re-registration.
Again we could offer three methods to pick the "mode".

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -437,37 +462,64 @@ open class PushService {
...
             val workData = Data.Builder()
                 .putString("USER_ACCOUNT", userAccountJson)
                 .putString("ACTION", action.name)
                 .build()
-            val workRequest: OneTimeWorkRequest =
-                OneTimeWorkRequest.Builder(PushNotificationsRegistrationChangeWorker::class.java)
-                    .setInputData(workData)
-                    .setInitialDelay(delayMilliseconds ?: 0, TimeUnit.MILLISECONDS)
+
+            when (delayDays) {
```
wmathurin, line 496: Instead of testing on the delayDays, we would use the enum mode value.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/SalesforceAnalyticsManager.java
brandonpage: I like this idea.

JohnsonEricAtSalesforce: I _also_ like that idea. I'll get a follow-up work item open to review feedback from @wmathurin also.

libs/SalesforceSDK/src/com/salesforce/androidsdk/analytics/AnalyticsPublishingWorker.kt
JohnsonEricAtSalesforce: Today's "Eagle Eyes" award goes to @wmathurin - Thanks!

PR #2588, opened 2024-06-26
Moving to SQLCipher 4.6.0

wmathurin:


wmathurin:
All tests passing (at least locally) except for some BriefcaseSyncDown tests - those failures are unrelated to the SQLCipher upgrade and will be investigated separately.

PR #2593, opened 2024-07-17
Fix auth config fetch.

brandonpage:
Yesterday on an [iOS PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3741) I proudly proclaimed that we didn't need the same fix on Android because it was implemented years ago.  Which _was_ true.  However, in linking the Android code as proof I noticed that it had been broken slightly by a Kotlin conversion; this PR rectifies that.


wmathurin (approved):

JohnsonEricAtSalesforce (approved):
💫

PR #2594, opened 2024-07-22
@W-16171409: [Android] Add QR Code Login Support in MSDK

JohnsonEricAtSalesforce:
🎸 _Ready For Review!_ 🥁

  This adds support for log in via log in QR codes generated with the UI Bridge API to MSDK.  This completes a very detailed spike previously completed by @wmathurin and directly inherits from that codebase.

  Before reviewing, I highly recommend reading this document and references in detail:
  👉🏻 https://salesforce.quip.com/JXmvAwirhR3V

  There's actually very little if any logic change from the spike code.  What I did try to apply was my novice's eye to the topic so we can name and document the code such that it'll be a bit more approachable for those uninitiated to the UI Bridge API, how that is delivered via the QR code and the many touch points that has in MSDK's log in logic.  I did the same in the companion pull request that introduces commented out content to our template guiding a developer through enabling this in a new app.

  The companion template update is in https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/417

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -117,6 +117,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.launch
 import org.json.JSONObject
+import java.net.URLDecoder
 import com.salesforce.androidsdk.R.layout.sf__login as sf__login_layout
 import com.salesforce.androidsdk.R.menu.sf__login as sf__login_menu
 
@@ -591,9 +592,14 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
     }
 
     private fun extractBridgeJsonParam(qrContents: String): String? {
-        val regex = """\?bridgeJson=(\{.*\})""".toRegex()
-        val matchResult = regex.find(qrContents)
-        return matchResult?.groups?.get(1)?.value
+        // Coming from intent (external QR reader) we have ?bridgeJson={...}
+        // Coming from embedded QR reader we have ?bridgeJson=%7B...%7D
+        val regexExternal = """\?bridgeJson=(\{.*\})""".toRegex()
+        val regexInternal = """\?bridgeJson=(%7B.*%7D)""".toRegex()
+        return regexExternal.find(qrContents)?.groups?.get(1)?.value
+            ?: regexInternal.find(qrContents)?.groups?.get(1)?.value?.let {
+                URLDecoder.decode(it, "UTF-8")
+            }
     }
 
     private fun parseBridgeInfo(jsonString: String): BridgeInfo {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -605,7 +605,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
     private fun parseBridgeInfo(jsonString: String): BridgeInfo {
         val jsonObject = JSONObject(jsonString)
         val frontdoorBridgeUrl = jsonObject.getString(FRONTDOOR_BRIDGE_URL_KEY)
-        val pkceCodeVerifier = jsonObject.getString(PKCE_CODE_VERIFIER_KEY)
+        val pkceCodeVerifier = jsonObject.optString(PKCE_CODE_VERIFIER_KEY, "")
         return BridgeInfo(frontdoorBridgeUrl, pkceCodeVerifier)
     }
 
@@ -613,6 +613,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
      * Login with qr contents
      * @param qrContents a url or url query containing a bridgeJson parameter
      *      bridgeJson parameter should contain a url-encoded JSON with two values: frontdoor_bridge_url and pkce_code_verifier
+     *      if not pkce_code_verifier is specified we use user agent flow
      * @return true if a frontdoor bridge url could be extracted from qrContents
      */
     fun loginFromQR(qrContents: String?): Boolean {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -581,7 +581,12 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
                 if (result.contents != null) {
                     val qrContents = result.contents
                     Log.i(TAG, "qrContents-->" + qrContents)
-                    parseQRContents(qrContents)
+                    parseQRContents(qrContents)?.let {
+                        webviewHelper?.loginWithFrontdoorBridgeUrl(
+                            it.frontdoorBridgeUrl,
+                            it.pkceCodeVerifier
+                        )
+                    }
                 }
                 else {
                     // TBD
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -121,7 +121,6 @@ import java.net.URLDecoder
 import com.salesforce.androidsdk.R.layout.sf__login as sf__login_layout
 import com.salesforce.androidsdk.R.menu.sf__login as sf__login_menu
 
-
 /**
  * Login activity authenticates a user. Authorization happens inside a web view.
  * Once an authorization code is obtained, it is exchanged for access and
@@ -187,7 +186,6 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
 
         // Setup content view
         setContentView(sf__login_layout)
-
         if (salesforceSDKManager.isIDPLoginFlowEnabled) {
             findViewById<Button>(sf__idp_login_button).visibility = VISIBLE
         }
@@ -365,9 +363,9 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         wasBackgrounded = false
     }
 
-    public override fun onSaveInstanceState(bundle: Bundle) {
-        super.onSaveInstanceState(bundle)
-        webviewHelper?.saveState(bundle)
+    public override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        webviewHelper?.saveState(outState)
     }
 
     override fun onKeyDown(keyCode: Int, event: KeyEvent) =
@@ -569,7 +567,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
 
     private fun isQRLogin(intent: Intent): Boolean {
         return SalesforceSDKManager.getInstance().isQRLoginFlowEnabled
-            && intent.data?.path?.contains(LOGIN_QR_PATH) == true
+                && intent.data?.path?.contains(LOGIN_QR_PATH) == true
     }
 
     private data class BridgeInfo(
@@ -755,4 +753,3 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         const val PKCE_CODE_VERIFIER_KEY = "pkce_code_verifier"
     }
 }
-
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -797,19 +797,19 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         // region QR Code Login Via UI Bridge API Constants
 
         /** The QR code login intent path */
-        const val LOGIN_QR_PATH = "/login/qr"
+        private const val LOGIN_QR_PATH = "/login/qr"
 
         /** The login QR code's UI Bridge API parameter's JSON frontdoor bridge URL key */
-        const val FRONTDOOR_BRIDGE_URL_KEY = "frontdoor_bridge_url"
+        private const val FRONTDOOR_BRIDGE_URL_KEY = "frontdoor_bridge_url"
 
         /** The login QR code's UI Bridge API parameter's JSON PKCE code verifier key */
-        const val PKCE_CODE_VERIFIER_KEY = "pkce_code_verifier"
+        private const val PKCE_CODE_VERIFIER_KEY = "pkce_code_verifier"
 
         /** A regular expression to extract the UI Bridge API parameter JSON from the intent's login QR code content */
-        val qrCodeBridgeJsonRegexExternal by lazy { """\?bridgeJson=(\{.*\})""".toRegex() }
+        private val qrCodeBridgeJsonRegexExternal by lazy { """\?bridgeJson=(\{.*\})""".toRegex() }
 
         /** A regular expression to extract the UI Bridge API parameter JSON from the app's login QR code content */
-        val qrCodeBridgeJsonRegexInternal by lazy { """\?bridgeJson=(%7B.*%7D)""".toRegex() }
+        private val qrCodeBridgeJsonRegexInternal by lazy { """\?bridgeJson=(%7B.*%7D)""".toRegex() }
 
         // endregion
     }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -169,7 +169,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         salesforceSDKManager.setViewNavigationVisibility(this)
 
         val loginOptions = if (isQRLogin) {
-            // Get app login options
+            // Get the app's login options
             salesforceSDKManager.loginOptions
         } else {
             // Get login options from the intent's extras
@@ -226,7 +226,6 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
             this
         )
 
-        // QR login
         if (isQRLogin) {
             loginFromQR("?" + intent.data?.query)
         } else {
@@ -565,33 +564,26 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         }
     }
 
-    private fun isQRLogin(intent: Intent): Boolean {
-        return SalesforceSDKManager.getInstance().isQRLoginFlowEnabled
-                && intent.data?.path?.contains(LOGIN_QR_PATH) == true
-    }
+    private fun isQRLogin(intent: Intent) = SalesforceSDKManager.getInstance().isQRLoginFlowEnabled
+            && intent.data?.path?.contains(LOGIN_QR_PATH) == true
 
-    private data class BridgeInfo(
+    private data class UiBridgeInfo(
         val frontdoorBridgeUrl: String,
         val pkceCodeVerifier: String
     )
 
-    private fun parseQRContents(qrContents: String?): BridgeInfo? {
-        if (qrContents != null) {
-            d(TAG, "qrContents-->${qrContents}")
-            val bridgeJsonString = extractBridgeJsonParam(qrContents)
-            d(TAG, "bridgeJsonString-->${bridgeJsonString}")
-            if (bridgeJsonString != null) {
-                val bridgeInfo = parseBridgeInfo(bridgeJsonString)
-                d(TAG, "bridgeInfo-->${bridgeInfo}")
-                return bridgeInfo
+    private fun parseQRContents(qrContents: String?) =
+        qrContents?.let { qrContentsUnwrapped ->
+            extractBridgeJsonParam(qrContentsUnwrapped)?.let { bridgeJsonStringUnwrapped ->
+                parseBridgeInfo(bridgeJsonStringUnwrapped)
             }
         }
-        return null
-    }
 
     private fun extractBridgeJsonParam(qrContents: String): String? {
-        // Coming from intent (external QR reader) we have ?bridgeJson={...}
-        // Coming from embedded QR reader we have ?bridgeJson=%7B...%7D
+        /*
+         * Coming from intent (external QR reader) expect ?bridgeJson={...}
+         * Coming from embedded QR reader expect ?bridgeJson=%7B...%7D
+         */
         val regexExternal = """\?bridgeJson=(\{.*\})""".toRegex()
         val regexInternal = """\?bridgeJson=(%7B.*%7D)""".toRegex()
         return regexExternal.find(qrContents)?.groups?.get(1)?.value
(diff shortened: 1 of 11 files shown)


JohnsonEricAtSalesforce:
It's worth noting I did intend to include commit's from the spike performed by @wmathurin.  I rebased his branch on `dev` before adding my own commits.  Plus, a squash will make this very tidy on `dev` when we're done while preserving his history here.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -345,9 +361,9 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
         wasBackgrounded = false
     }
 
-    public override fun onSaveInstanceState(bundle: Bundle) {
-        super.onSaveInstanceState(bundle)
-        webviewHelper?.saveState(bundle)
+    public override fun onSaveInstanceState(outState: Bundle) {
```
JohnsonEricAtSalesforce, line 364: This was an unrelated yet low-risk inspector cleanup.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -654,9 +670,144 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
+     * The UI Bridge API JSON parameter should contain URL-encoded JSON with two
+     * values:
+     * - frontdoor_bridge_url
+     * - pkce_code_verifier
+     *
+     * If pkce_code_verifier is not specified then the user agent flow is used
+     * @return Boolean true if a log in attempt is possible using the provided QR
+     * code content, false otherwise
+     */
+    fun loginFromQrCode(
```
JohnsonEricAtSalesforce, line 688: All the new methods here should look _very_ familiar to @wmathurin.  I did apply quite a bit of comment and naming thought though to reflect my own journey in understanding how it all worked plus see if I could make the semantic match through all the reference documentation.  There's some Kotlin style added as well for extra credit.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -1399,6 +1399,21 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
         }
     }
 
+    /**
+     * Automatically log in using the provided UI Bridge API parameters.
+     * @param frontdoorBridgeUrl The UI Bridge API front door bridge API
+     * @param pkceCodeVerifier The PKCE code verifier
+     */
+    internal fun loginWithFrontdoorBridgeUrl(
```
JohnsonEricAtSalesforce, line 1407: @wmathurin - I made this `internal`.  Was there any reason it wasn't before?

wmathurin: The packing of front door bridge URL and PKCE in a QR code will be up to the apex or java page.
Hopefully it will use the attributes names we recommend, if it does not, an app could parse them out itself and call this method directly?
I guess if we make the method internal, it will force the server page to use the attributes names in the JSON that we expect.

JohnsonEricAtSalesforce: That's very helpful, since I was also starting to wonder where a custom APEX with different structure would tie in.  I'll make this public again!

wmathurin (approved):

JohnsonEricAtSalesforce:
Hold merging this to `dev` and the `12.1.0` codebase until we confirm the iOS version will also make it in time for the release.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: @JohnsonEricAtSalesforce @wmathurin Is there any kind of profile limitation to this login option? If not should we consider moving this to a standalone utility class that could be used by LoginActivity and native login consumers?  Perhaps native login support is a new story to itself, just trying to be forward thinking so we don't have to deprecate and move APIs around if possible.

wmathurin: The bridge url is supposed to be loaded in a web view to get the tokens or the code - it is not a headless API. So it's another form of web view based login.

brandonpage: Yes, but despite it being a web based login it is possible to do it without the customer ever seeing a webview.  You can put a button on a native login screen that has the same code as the `QrCodeEnabledLoginActivity` from the template app and if the user is pre authorized (a requirement for headless flow) they don't have to accept connected app policy.  This is obviously work well beyond the scope of this PR and would require a headless webview but it sounds possible, right?

wmathurin: It's like the web view to get a code for the SP in the IDP-based login. You have to show it because user interaction might be needed. 
But you are right that you could launch it from the native login screen.

wmathurin: Yeah, if the user is preauthorized, we could hide the web view. I guess we could let the mobile app know whether to show the web view or not by adding a field in the JSON we encode in the QR URL.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -661,7 +661,7 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
                     )
 
                     else -> when {
-                        instance.useWebServerAuthentication -> onWebServerFlowComplete(params["code"])
+                        codeVerifier?.isNotEmpty() == true -> onWebServerFlowComplete(params["code"])
```
brandonpage, line 664: This is a bug and needs to be reverted as it will break User Agent flow login.  `codeVerifier` is always **created** no matter which flow we are using.

brandonpage (changes requested):
Where the QR Code Login work should live isn't a blocker, but the User Agent Flow issue I mentioned is.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
wmathurin: You are right, I did not try user agent flow when doing the spike.
But we can't just revert.
With QR login, we will decide what flow to use based on what came through the JSON encoded in the QR URL.
If the JSON contains a code verifier, we use it and we should to a web server flow. And we should set a new flag saying that web server flow should be used.
That flag should be set to instance.useWebServerAuthentication by default.
We should use the new flag here.

JohnsonEricAtSalesforce: Thanks, @brandonpage!  I'll take a look.

brandonpage: Okay.  So if I understand correctly the QR URL can/should override the app's setting for `useWebServerAuthentication`?  Sounds like something that should be documented. 

Also is User Agent Flow supported?  The code above appears to require PKCE.

wmathurin: We support both, if the QR encodes a code verifier we expect to the bridge URL to do a web server flow and the redirect to get us a code. If the QR comes with an empty/null code verifier, then the code verifier ends up not being defined and the code above will flow the user agent branch. The bridge URL should do a user agent flow and the redirect get us the refresh tokens.

During the spike, I played with both, but the web server flow case meant a longer URL encoded in the QR which made it harder to read with the reader. I could not get it to work completely. See https://salesforce.quip.com/JXmvAwirhR3V#temp:C:IIO55768e89acdc49948a3be45bf

wmathurin: @JohnsonEricAtSalesforce Before closing the story, we need to make sure user agent and web server flow work with and without QR. During the spike, I did not test them all or get them all working.

JohnsonEricAtSalesforce: @wmathurin & @brandonpage - I broke out [W-16336715: [MSDK Android] Correct QR Code Log In When Using Web Server](https://gus.lightning.force.com/a07EE00001xT1B3YAK) to track this in detail.  I'll commit that here before we merge for 12.1 or whichever release QR code login lands in soon.

I also broke out [W-16336722](https://gus.lightning.force.com/a07EE00001xTbt7YAC) to track the other bug in Quip regarding the main activity not being displayed.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -660,9 +667,16 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
...
                     )
 
-                    else -> when {
-                        instance.useWebServerAuthentication -> onWebServerFlowComplete(params["code"])
-                        else -> onAuthFlowComplete(TokenEndpointResponse(params))
+                    else -> {
+                        // Determine if presence of override parameters require the user agent flow.
+                        val overrideWithUserAgentFlow = isUsingFrontDoorBridgeUrl && frontDoorBridgeCodeVerifier == null
+                        when {
+                            instance.useWebServerAuthentication && !overrideWithUserAgentFlow ->
```
JohnsonEricAtSalesforce, line 674: 💤 [No after hours reply needed] @brandonpage & @wmathurin, this is my current thinking for storing the front door bridge URL and its optional accompanying code verifier as dedicated properties so we may then use them to determine the authentication flow between web server and user agent.  That's for this thread from earlier:

https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2594#discussion_r1688807301

This looks a lot like what I have in place for iOS in SFOAuthCoordinator.

Just having an outside force overriding the app's chosen auth flow is complicated, so I want to internalize that from the app and try to keep our code simple.  I was hoping just storing those and then clearly determining this one local variable would keep that readable.  I'm open the suggestions, though.

On the iOS side, SFOAuthCoordinator is the WKNavigationDelegate.  It would possibly be a lot easier to maintain if that was a separate class chosen just when a front door URL is in place.  I'm thinking "one class per job" rather than a single class with an overly complicated control flow.  I _didn't_ do that yet, but perhaps when we convert that to Swift and here we eliminate OAuthWebViewHelper we could really simplify the code for this complicated set of logic flow.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -164,14 +169,19 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
 
         // Protect against screenshots
         window.setFlags(FLAG_SECURE, FLAG_SECURE)
 
-        // Fetch authentication configuration if required
-        salesforceSDKManager.fetchAuthenticationConfiguration()
+        // Fetch authentication configuration except for QR code login.
+        if (!isDeepLinkedQrCodeLogin) {
+            salesforceSDKManager.fetchAuthenticationConfiguration()
+        }
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: @wmathurin - Do you recall the logic on this one?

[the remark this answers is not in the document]
wmathurin: I took it out during the spike because it uses the selected login server which might not match the one specified in the QR url. To keep it, we have to refactor fetchAuthenticationConfiguration to optionally accept a loginServer and during a QR login, we should pass the loginServer across.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -148,16 +148,23 @@ import java.util.function.Consumer
...
     "This class will no longer be public starting in Mobile SDK 13.0.",
     level = DeprecationLevel.WARNING,
 )
 open class OAuthWebviewHelper : KeyChainAliasCallback {
 
+    /** The default, locally generated code verifier */
     private var codeVerifier: String? = null
 
+    /** For Salesforce Identity UI Bridge API support, an overriding front door bridge URL to use in place of the default initial URL */
+    var isUsingFrontDoorBridgeUrl = false
```
brandonpage, line 163: NIT: The comment description makes this sound like it is a URL, not a boolean.  And does this or `frontDoorBridgeCodeVerifier` need to be public?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -660,9 +667,16 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
                         null
                     )
 
-                    else -> when {
-                        instance.useWebServerAuthentication -> onWebServerFlowComplete(params["code"])
-                        else -> onAuthFlowComplete(TokenEndpointResponse(params))
+                    else -> {
+                        // Determine if presence of override parameters require the user agent flow.
+                        val overrideWithUserAgentFlow = isUsingFrontDoorBridgeUrl && frontDoorBridgeCodeVerifier == null
```
brandonpage, line 672: When would this occur?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: With this as it is, are we skipping a task somehow during a deep linked QR Code Log In?  Also, I usually thought of a "deep linked" QR code log in as coming from an external QR code reader but the change in server would apply to internally read QR codes as well, I reckon.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -738,7 +752,7 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
                 create(loginOptions.loginUrl),
                 loginOptions.oauthClientId,
                 code,
-                codeVerifier,
+                frontDoorBridgeCodeVerifier ?: codeVerifier,
```
brandonpage, line 755: This just feels dangerous.  What if you login with qrCode (so `frontDoorBridgeCodeVerifier` is set), we logout and try a standard login?  Is `OAuthWebviewHelper` (and thus `frontDoorBridgeCodeVerifier`) still in memory?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -654,9 +670,147 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
+        const val FRONTDOOR_BRIDGE_URL_KEY = "frontdoor_bridge_url"
+
+        /** The login QR code's UI Bridge API parameter's JSON PKCE code verifier key */
+        const val PKCE_CODE_VERIFIER_KEY = "pkce_code_verifier"
+
+        /** A regular expression to extract the UI Bridge API parameter JSON from the intent's login QR code content */
+        val qrCodeBridgeJsonRegexExternal by lazy { """\?bridgeJson=(\{.*\})""".toRegex() }
+
+        /** A regular expression to extract the UI Bridge API parameter JSON from the app's login QR code content */
+        val qrCodeBridgeJsonRegexInternal by lazy { """\?bridgeJson=(%7B.*%7D)""".toRegex() }
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: Indeed not.  See 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2594/commits/42f440f32e23e362a10f5eaf0ab9144f8c517f2b

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
JohnsonEricAtSalesforce: Good cleanup for both points.  👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2594/commits/74d1572e8da7bdaa6be98dc4eba9217f43531739 I feel like Android Studio has stopped suggesting access modifiers for some reason.  It used to be very aggressive about suggesting private or reductions in access when applicable.  🤨

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
JohnsonEricAtSalesforce: This occurs when the QR code provides a front door bridge URL but _not_ a code verifier, which implicitly sends us to user agent flow.  That's my understanding after chatting with @wmathurin.  Possibly one of most interesting points in all of this is ensuring this dovetails well with out existing preferences between user agent and web server flow.

I've tried quite a few combinations to test it out, such as changing the app's `useWebServerAuthentication` setting plus using QR codes of both preferences.  It matches my understanding of it thus far, but I love more input if there's any to have.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
JohnsonEricAtSalesforce: I was really curious about state here also.  My tests showed the login state and all the objects I expected being recreated each log in attempt.  That would match my expectation since a log out really should set the app back to a clean slate.

Additionally, I also just tested using a switch user/add user scenario while another user is logged in.  I got a new OAuthWebViewHelper instance, so everything is back to a clean log in state.

Do we have any scenarios where we actually would want to retain these objects?  I suspect that's a "no".

JohnsonEricAtSalesforce: Here's a thought that I was already considering: I feel we need handling when, somehow, a QR Code Log In fails. The user could theoretically fall back to username and password with the same web view.

In `OAuthWebViewHelper.onAuthFlowError`, we could null these two parameters.

Would it make sense to do the same in `onAuthFlowComplete`?  I see that is called _after_ `onWebServerFlowComplete` and `doCodeExchangeEndpoint`.

With both in place, that would help us reset state for error flows (where I think we need it) and on success flows (where we're discarding the `OAuthWebViewHelper` anyways).

I'm trying to do some destructive testing with failing QR Codes 💣

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I was trying to make sure I understood this one.  With the code as we have it, one can only log in via a QR Code/Front Door Bridge from the connected app configured in `bootconfig.xml`.  @wmathurin, isn't that a constraint and if so we wouldn't need this check.

Otherwise, do have a need to support QR codes that would log the user into something they couldn't log into via the web view? Sorry, I'm just making sure I know the scope.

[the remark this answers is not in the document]
wmathurin: In the spike, I was extracting the login server from the QR url - see https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/7468f20a6a052767da718ec2c1a1c05f8422c332.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
brandonpage: Yeah, if QR Login somehow fails was the other scenario I was thinking of.  Might be good to reset it in `onAuthFlowError`.

brandonpage (approved):
LGTM.  Though I think I would still like to see us extracting the login server from the QR Code URL [like Wolf mentioned here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2594#discussion_r1783751322).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: @wmathurin - I can see the current code you and I have here, since this includes your spike commits, does the URL extraction from the QR code and sets login URL on login options.  The trouble is that fails to log in.  Say, if I set the app's default login URL to "Production"/`https://login.salesforce.com` and then use a QR code from `https://mobilesdkatsdb6--c.test1.vf.pc-rnd.force.com/apex/qrcodelogin`.  The web view displays "Webpage Not Available"/"The webpage at mobilesdk://android/pn/tester?code=... could not be loaded because: net:ERR_UNKNOWN_URL_SCHEME"

If I recall the code, we'd need to also have the connected app's callback URL so the auth flow would be able to handle the authorization response page.  Do we want to do that and is this actually a realistic scenario?

Just for fun, I did do a little switch users magic and it _is_ possible to sign in to the app across two different login URLs, but I wasn't sure we actually support that.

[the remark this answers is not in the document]
JohnsonEricAtSalesforce: Also, as the code is now this boolean is set from `isDeepLinkedQrCodeLogin`.  That method probably needs to change in the follow-up pull request for [W-16336722](https://gus.lightning.force.com/a07EE00001xTbt7YAC).  How the QR Code payload is structured is theoretically app to the APEX/VisualForce Page set up by the MSDK user, so that method should probably move to the app/template code since so the app can interpret the deep link intent and pass explicit options to MSDK.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
JohnsonEricAtSalesforce: @brandonpage - Take a look at 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2594/commits/a2a35754b194e30abaccb51ac1373cafbd706e47 - That adds a new reset method and calls to it from three points in the logic where I reckon we need to get back to the initial state.  Of course, I'm not aware of a way to re-use the object after a successful authentication transaction.  That one may not be needed.

JohnsonEricAtSalesforce: The failure case is the most notable addition.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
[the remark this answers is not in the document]
wmathurin: An app can disallow users from adding login servers by hiding the button to pick servers.
However when the app does not hide the button, users are allowed to add login servers.

When allowed, if QR points to a login server they don't have or haven't selected, I don't think it would be a good user experience to require the user to first add the login server then scan the QR code. It kind of defeats the point of having QR as an easier way to log in.

When the button is hidden, if we extract the login server from the QR URL like I did in the spike, it would allow the user to change login server even though the app thought it was prohibiting it.

The correct behavior should probably be:
When doing a login from QR URL, extract the login server, check if the login server is part of the list of login servers or if adding login servers is allowed. If that is true, proceed with login (including the fetching of the auth config).
If the login server is not part of the list of login servers and adding login servers is disallowed, fail with a message to the server indicating the problem ("QR uses a login server that is not configured in mobile application").

If that's too much for 12.2, we could create a new story for it and simply document in 12.2 what's not supported.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
wmathurin: I guess we could instead fail if the configured auth flow does not match the auth flow the QR URL implies. And show a message to the user e.g. "QR URL is for a user-agent flow but mobile application is configured for web server flow".

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I broke this out into [W-16893221](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000022ReOAYA0/view).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
JohnsonEricAtSalesforce: Actually, this switch to the flow in the QR code is working for both Android and iOS.  It's a one-time override, since a success or failure will reset everything back to the flow set by the app.

PR #2604, opened 2024-08-23
Fix issue that can cause the identity service call to fail on login.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -153,6 +153,13 @@ public synchronized static void clearCaches(UserAccount userAccount) {
 		String orgId = userAccount != null ? userAccount.getOrgId() : null;
 		String userId = userAccount != null ? userAccount.getUserId() : null;
 		String cacheKey = computeCacheKey(orgId, userId);
+		OAuthRefreshInterceptor interceptor = OAUTH_REFRESH_INTERCEPTORS.get(cacheKey);
+		if (interceptor != null) {
+			// There is no way to remove an interceptor from the OkHttp builder so get rid of the
+			// access token to prevent it from being used by the OAuthRefreshInterceptor in memory.
+			interceptor.setAuthToken(null);
+		}
+
```
wmathurin, line 162: Maybe we need to revisit https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2474/files which introduced the issue. 

HttpAccess.getOkhttpClientBuilder() used to create a new OkHttpClient.Builder every time it is called, and then we had caches (for OkHttpClient's, OkHttpClient.Builder and OauthRefreshInterceptor) and clearCaches() would make sure to clean them on logout.

Once we started caching OkHttpClient.Builder, we are no longer starting from a clean slate, the code above address the issue but will be confusing to read (we tweak an interceptor then we drop it from the cache a few lines later).

Also, what is happening currently when you have multiple users logged in. Aren't they all using the same OkHttpClient.Builder? Does this builder end up with multiple interceptors because of line 217 below?

wmathurin: Also that follow up change (https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2517/commits/97bf92abdc81870d7fb0841440a3369288e98989) would not have been needed if HttpAccess.getOkHttpClientBuilder() was still returning a new builder each time.

wmathurin: I think we should go back to returning a new builder in HttpAcess.getOkHttpClientBuilder().
But maybe we can offer a way to customize those builders?

wmathurin: Maybe something like: https://github.com/wmathurin/SalesforceMobileSDK-Android/commit/dd354c7a051069a5470f0d835dabf65823b83017

brandonpage: @wmathurin Perhaps all we really need to do is start with a clean state earlier in the process?  The reason this only affects the identity service call is because that is made before we create the new rest client.  That is why we add the auth header (which is overridden in the interceptor -- the bug) to that call specifically.  

I believe users should have their own `OkHttpClient.Builder` no matter what because we have a RestClient per user.  Again, we haven't made the RestClient for the new user when he identity call is made.

RE: Customizing the builder:  I don't think customizing the builder was the original problem, it was having those customizations wiped away whenever we get the RestClient.  

The safest and easiest thing to do for the patch is to revert the original change.  We should keep that as an option in case we can't find a clean solution to resolve all of the side affects it causes.

wmathurin: Because we keep using the same builder and more and more interceptors can possibly be added, it feels broken as it is.

Also with the HttpAccess changes the "**If none cached, create new one**" below is not true.

So if we don't want to add the OkHttpClientBuilderCustomizer I suggested [here](https://github.com/wmathurin/SalesforceMobileSDK-Android/commit/dd354c7a051069a5470f0d835dabf65823b83017), we should at least revert the changes in HttpAccess so that getOkHttpClientBuilder create a fresh/new builder each time.

```java
    private synchronized void setOkHttpClientBuilder() {
                final String cacheKey = getCacheKey();
                OkHttpClient.Builder okHttpClientBuilder = OK_CLIENT_BUILDERS.get(cacheKey);

                // If none cached, create new one
                if (okHttpClientBuilder == null) {
                        if (cacheKey.equals("unauthenticated")) {
                                okHttpClientBuilder = httpAccessor.getUnauthenticatedOkHttpBuilder();
                        } else {
                                okHttpClientBuilder = httpAccessor.getOkHttpClientBuilder()
                                                .addInterceptor(getOAuthRefreshInterceptor());
                        }

                        OK_CLIENT_BUILDERS.put(getCacheKey(), okHttpClientBuilder);
                }
                this.okHttpClientBuilder = okHttpClientBuilder;
        }
```

brandonpage: Yeah, `HttpAccess` being a singleton really complicates this and might lead to multiuser issues. 

The safest and easiest thing to do for the patch is to revert the original change.  We should keep that as an option in case we can't find a clean solution to resolve all of the side affects it causes.

PR #2607, opened 2024-08-26
Revert retain client builder

brandonpage:


JohnsonEricAtSalesforce (approved):
🤘🏻

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/HttpAccess.java
```diff
@@ -102,24 +100,26 @@ public HttpAccess(Context app, String userAgent) {
      * and user agent interceptor for an authenticated client.
      */
     public OkHttpClient.Builder getOkHttpClientBuilder() {
```
wmathurin, line 102: NIT Should we make the "createNewClientBuilder" public and deprecate the other two (getOkHttpClientBuilder() and getUnauthenticatedOkHttpBuilder()). It better reflects the fact it create a new builder every time it is called.

wmathurin (approved):

PR #2611, opened 2024-09-06
Developement

Ritesh175862:
IDEA url updated

### DIFF
--- a/hybrid/HybridSampleApps/AccountEditor/AndroidManifest.xml
+++ b/hybrid/HybridSampleApps/AccountEditor/AndroidManifest.xml
@@ -35,7 +35,7 @@
             'scheme', 'host' and 'path' with their corresponding values from your connected app.
 
             For example, if the callback URL of your connected app is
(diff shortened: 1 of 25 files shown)


brandonpage:
@Ritesh175862 I think you pushed to the main repo instead of your fork.

hybrid/HybridSampleApps/AccountEditor/AndroidManifest.xml
```diff
@@ -35,7 +35,7 @@
             'scheme', 'host' and 'path' with their corresponding values from your connected app.
 
             For example, if the callback URL of your connected app is
-            "testsfdc:///mobilesdk/detect/oauth/done",
+            "https://login.salesforce.com/services/oauth2/success",
```
brandonpage, line 38: Just for reference I would not advise using a URL (especially this URL) as a the callback URI.  If you need to use Advanced (browser) Auth the callback is used to deep link back into your app and Google requires proof of ownership for URLs.  There is no way to prove you own `https://login.salesforce.com`.

PR #2484, opened 2023-11-01
Change launch mode of LoginActivity from singleInstance to singleTop.

brandonpage:

PR #2613, opened 2024-09-11
Fix advanced auth login issue after switching servers and other issues.

brandonpage:
Issues covered here:
- Failed PKCE Login due to the code verifier being replaced mid flow.  
  - Fixed by now allowing the webview to reload in LoginActivity `onNewIntent` while we are using advanced auth.
- If you switch servers the LoginActivity can appear again instead of the app after a successful login.
  - Partly fixed by the above restriction of reloading the webview (which checks the well known config and launches the login view again when it is done).
  - Partly fixed by switching the LoginActivity from `SingleTop` to `SingleTask`.  I really didn't want to do this because it is **technically a breaking chance** for consuming apps, but the LoginActivity randomly restarts after finish.  I think this is an Android bug we can't work around.  
- Advanced Auth login using the custom domain webview button (requires opt-in pattern) was broken.
  - Somehow for this flow it was possible for a few methods (`loadingLoginPage` and `showError`) to no longer be running on the correct thread so I ensured they run on main. 
- I also cleaned up the presentation of Advanced Auth a little by blanking out the webview when we launch a custom tab.  This means if the user backs our of the custom tab they see a blank screen instead of the webview still loaded with the _previous_ login server.  They will also see this blank screen briefly as we finish PKCE.

### DIFF
--- a/hybrid/HybridSampleApps/AccountEditor/AndroidManifest.xml
+++ b/hybrid/HybridSampleApps/AccountEditor/AndroidManifest.xml
@@ -46,7 +46,7 @@
         <!--
         <activity android:name="com.salesforce.androidsdk.ui.LoginActivity"
             android:theme="@style/SalesforceSDK"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
             android:exported="true">
 
             <intent-filter>
--- a/hybrid/HybridSampleApps/MobileSyncExplorerHybrid/AndroidManifest.xml
+++ b/hybrid/HybridSampleApps/MobileSyncExplorerHybrid/AndroidManifest.xml
@@ -47,7 +47,7 @@
         <!--
         <activity android:name="com.salesforce.androidsdk.ui.LoginActivity"
             android:theme="@style/SalesforceSDK"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
             android:exported="true">
 
             <intent-filter>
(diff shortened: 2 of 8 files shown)


wmathurin:
So we are reverting https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2484. Do you remember why we did https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2484?

brandonpage:
> So we are reverting #2484. Do you remember why we did #2484?

It's more tweaking that reverting.  We made that change because `SingleInstance` made the custom tab a new "task", which meant it was it's own entry in the back stack and app switcher menu.  The only way to avoid that is to use the no history flag, but that breaks 2FA.   IIRC there are actually a bunch of other weird side affects to `SingleInstance`, I don't know why we were ever using.  Maybe it was to work around a _chrome_ custom tab issue that have long since been fixed when the feature was brought to the OS level?

When I chose to move to `singleTop` I had also considered using `singleTask`.  I settled on `singleTop` because the documentation describes it as being the closest thing to the default (non specified) launch mode.  From my testing `singleTask` behaves in exactly the same way in all scenarios for our use case with the exception that the issue of the login activity randomly starting again after login does not happen.  I will admit it is very strange I didn't run into this when we made  #2484, but it also seems to only happen on certain devices.  As far as I can tell it appears to be unrelated to any code change we have made since (besides changing `launchMode`).  

I would honestly love to keep it at `singleTop` and avoid the breaking change but I've spent about a week trying to do that.  I have repro steps and a lot of info I could share on this.  I'm happy to hold the PR if anyone on the team wants to try to reproduce and find the root cause (but I suspect it is simply an Android OS bug).

wmathurin:
I read the old PR too quickly - we were not using singleTask before - so it's not a revert.
We have gone from singleInstance to singleTop to singleTask.
Thanks for the explanation.

wmathurin (approved):

PR #2614, opened 2024-09-13
Make SalesforceSDKManager's ScreenLockManager getter public.

brandonpage:
I appears we simply missed this when we converted `SalesforceSDKManager` to Koltin in 12.0.


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -1032,7 +1033,7 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
             if (id?.screenLockTimeout?.compareTo(0) == 1) {
                 SalesforceSDKManager.getInstance().registerUsedAppFeature(FEATURE_SCREEN_LOCK)
                 val timeoutInMills = (id?.screenLockTimeout ?: 0) * 1000 * 60
-                instance.screenLockManager?.storeMobilePolicy(
+                (instance.screenLockManager as ScreenLockManager?)?.storeMobilePolicy(
```
brandonpage, line 1036: Off topic, but I really hate `instance` here.  If there is anything worse than the Singleton Pattern it is making references to Singletons 😞

wmathurin (approved):

PR #2615, opened 2024-09-13
Fix push deregister error.

brandonpage:
In #2545 I made a change to delay push deregister until the device had network connection so that the operation would succeed (eventually) even if the user chose to logout while offline.  At the time of testing this worked correctly in all cases but since a change has been made that causes an error on deregister (likely the recent bugfix #2607, which retained the tokens after logout).  The error doesn't actually cause the unregister operation to fail and is invisible to the end user, but an internal team ran into issues with automated testing because of it.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
@@ -160,10 +160,9 @@ object PushMessaging {
                 FirebaseInstallations.getInstance(firebaseApp).delete()
                 firebaseApp.delete()
             }
+
             unregisterSFDCPush(context, account)
         }
-
-        unregisterSFDCPush(context, account)
     }
 
     /**
@@ -470,14 +469,7 @@ object PushMessaging {
         account: UserAccount?,
         action: PushNotificationsRegistrationAction,
     ) {
-        if (account == null) {
-            enqueuePushNotificationsRegistrationWork(
-                userAccount = null,
-                action = action,
-                pushNotificationsRegistrationType = ReRegistrationOnAppForeground,
-                delayDays = null
-            )
-        } else if (isRegistered(context, account)) {
+        if (account == null || isRegistered(context, account)) {
             enqueuePushNotificationsRegistrationWork(
                 userAccount = account,
                 action = action,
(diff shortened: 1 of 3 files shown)


wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
```diff
@@ -160,10 +160,9 @@ object PushMessaging {
                 FirebaseInstallations.getInstance(firebaseApp).delete()
                 firebaseApp.delete()
             }
+
             unregisterSFDCPush(context, account)
         }
-
-        unregisterSFDCPush(context, account)
```
brandonpage, line 166: This seems like an oversight.  There is no reason for `unregisterSFDCPush` to be called twice here (once even if it wasn't registered 👀).

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushMessaging.kt
```diff
@@ -470,14 +469,7 @@ object PushMessaging {
...
     ) {
-        if (account == null) {
-            enqueuePushNotificationsRegistrationWork(
-                userAccount = null,
-                action = action,
-                pushNotificationsRegistrationType = ReRegistrationOnAppForeground,
-                delayDays = null
-            )
-        } else if (isRegistered(context, account)) {
+        if (account == null || isRegistered(context, account)) {
```
brandonpage, line 472: This is just code simplification, no logic change.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -477,53 +482,55 @@ open class PushService {
...
-                    workerClass = PushNotificationsRegistrationChangeWorker::class.java
-                ).setInputData(workData)
-                    .setConstraints(constraints)
-                    .build().also { workRequest ->
-                        workManager.enqueueUniqueWork(
-                            PUSH_NOTIFICATIONS_REGISTRATION_WORK_NAME,
-                            REPLACE,
-                            workRequest
-                        )
+            if (action == Register) {
```
brandonpage, line 485: This is actually the main issue.  

The unified diff looks weird due to indentation but the change here is simply wrapping this `when (pushNotificationsRegistrationType)` in an `if (action == Register)` so we don't enqueue a **deregister** job here with the pushNotifications**Registration**Type and another (expected) deregister job below.

JohnsonEricAtSalesforce (approved):
🤘🏻

PR #2616, opened 2024-09-19
Saving more fields from token end point in user account - simplified data passing auth / user account / account manager

wmathurin:
## New token end point fields captured 

Capturing a few extra fields from token end point (cookie client src, cookie sid client and sid cookie name). 
Saving them to the user account.
_We will need them when we change `SalesforceDroidGapActivity` from using front-door to using sessions returned by hybrid login and refresh flows._

### Class diagram highlighting new or modified methods/fields

```mermaid
classDiagram
    class UserAccount {
        +cookieClientSrc
        +cookieSidClient
        +sidCookieName
    }
    
    class UserAccountBuilder {
        +cookieClientSrc
        +cookieSidClient
        +sidCookieName
        +allowUnset: Boolean
        +populateFromIdServiceResponse()
        +populateFromTokenEndpointResponse()
        +populateFromUserAccount()
    }
    
    class UserAccountManager {
        -buildAuthBundle()
        +createAccount()
        +updateAccount()
    }
    
    class OAuthWebviewHelper {
    }

    class BaseFinishAuthFlowTask {
        +onPostExecute()
    }
    
    class ClientManager {
        +createNewAccount()
        +peekRestClient()
    }

    class AccMgrAuthTokenProvider {
        +getNewAuthToken()
        -refreshStaleToken()
    }
    
    class AuthenticationService {
        +getAuthToken()
    }
    
    UserAccount <-- UserAccountBuilder : builds
    UserAccountManager --> UserAccount : manages
    OAuthWebviewHelper --> BaseFinishAuthFlowTask: has
    BaseFinishAuthFlowTask --> UserAccountBuilder : uses
    ClientManager --> UserAccountManager : uses
    ClientManager --> AccMgrAuthTokenProvider: has
    AccMgrAuthTokenProvider --> UserAccountBuilder: uses
    ClientManager --> UserAccountBuilder : uses
    OAuthWebviewHelper --> UserAccountManager : uses
    AuthenticationService --> UserAccountManager : uses
    AuthenticationService --> UserAccountBuilder : uses
```

### `UserAccount`
- New fields mentioned above

#### New method: `populateFromIdServiceResponse` 
- called from `onPostExecute` of `BaseFinishAuthFlowTask` in `OAuthWebviewHelper`  at the end of login flow
#### New method: `populateFromTokenEndpointResponse` 
- called from `onPostExecute` of `BaseFinishAuthFlowTask` in `OAuthWebviewHelper` at the end of login flow
- called from `refreshStaleToken` of `ClientManager`'s `AccMgrAuthTokenProvider` (called when `RestClient` gets a 401)
- called from `getAuthToken` of `AuthenticationService` 
#### New method: `populateFromUserAccount`
- it allows a builder to start with existing values that are then updated by calling `populateFromTokenEndpointResponse` with the response returned by a call to refresh end point
- called from `refreshStaleToken` of `ClientManager`'s `AccMgrAuthTokenProvider` (called when `RestClient` gets a 401)
- called from `getAuthToken` of `AuthenticationService `
#### New field/setter `allowUnset`
- it defaults to true
- when false, passing null to a setter does not unset a value previously set and passing a map to `additionAuthValues` causes the new map and old map to be merged
- it is used doing refresh to merge values returned by refresh with existing values to mirror the old code (where everything was done "manually" in multiple places)

#### New method: `buildAuthBundle`
- creates a bundle from a `UserAccount` for authenticator service
- the bundle uses keys understood by authenticator service
- the bundle encrypt most values
#### New method: `createAccount`
- called from `ClientManager`'s `createNewAccount` (which is now a one liner)
- more readable than the old `createNewAccount` thanks to `buildAuthBundle`
#### New method: `updateAccount`
- called from `refreshStaleToken `of `ClientManager`'s `AccMgrAuthTokenProvider` (called when `RestClient` gets a 401)
- called from `getAuthToken` of `AuthenticationService` 

### `OAuthWebviewHelper`
- simplified `onPostExecute`
- got rid of `AccountOptions` inner class - `accountOptions` is simply the `userAccount.toBundle()`

### `ClientManager`
- simplified `createNewAccount` - code moved to `UserAccountManager`
- simplified `peekRestClient` - calling `UserAccountManager`'s `buildUserAccount` to get all the values from account manager
- simplified `getNewAuthToken` - like before it calls `refreshStaleToken`, but that method now returns a `UserAccount` instead of a bundle with encrypted values
- simplified `refreshStaleToken` - uses `UserAccountBuilder`'s new methods `populateFromUserAccount`, `allowUnset` and `populateFromTokenEndpointResponse`

### Added / updated tests 
- `UserAccountTest`
- `ClientManagerTest`
- `UserAccountManagerTest`
- a few others (leveraging `UserAccountManager.createAccount` instead of having to set up a `ClientManager` and calling `createNewAccount()`


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -192,15 +206,21 @@ public class UserAccount {
...
 	/**
-	 * Parameterized constructor.
+	 * Construct UserAccount from JSON.
 	 *
 	 * @param object JSON object.
+	 * @param applicationName the application name.
+	 * @param additionalOauthKeys addtional auth keys.
 	 */
-	public UserAccount(JSONObject object) {
+	UserAccount(JSONObject object, String applicationName, List<String> additionalOauthKeys) {
```
wmathurin, line 223: This constructor does not call methods SalesforceSDKManager (to get application name of additional oauth keys) - which makes testing easier.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -232,17 +251,29 @@ public UserAccount(JSONObject object) {
...
+	}
+
+	/**
+	 * Construct UserAccount from Bundle.
 	 *
 	 * @param bundle Bundle.
+	 * @param additionalOauthKeys additional oauth keys.
 	 */
-	public UserAccount(Bundle bundle) {
+	UserAccount(Bundle bundle, List<String> additionalOauthKeys) {
```
wmathurin, line 276: Same idea as above - a constructor that does not call SalesforceSDKManager.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -86,7 +87,11 @@ public class UserAccount {
 	public static final String NATIVE_LOGIN = "native_login";
 	public static final String LANGUAGE = "language";
 	public static final String LOCALE = "locale";
-
+	public static final String COOKIE_CLIENT_SRC = "cookie-clientSrc";
```
wmathurin, line 90: The new fields!

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -752,9 +831,10 @@ public int hashCode() {
 	/**
 	 * Returns a JSON representation of this instance.
 	 *
+	 * @param additionalOauthKeys: values to pull from the additionalOauthValues
 	 * @return JSONObject instance.
 	 */
-	public JSONObject toJson() {
+	JSONObject toJson(List<String> additionalOauthKeys) {
```
wmathurin, line 837: The public toJson takes no argument and calls this one with the additionalOauthKeys returned by SalesforceSDKManager. Having this method allows for simpler tests.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.kt
```diff
@@ -0,0 +1,529 @@
...
+    private var clientId: String? = null
+    private var additionalOauthValues: Map<String, String>? = null
+    private var allowUnset = true
+
+    /**
+     * Set fields from token end point response
+     * @param tr token endpoint response
+     * @return Instance of this class
+     */
+    fun populateFromTokenEndpointResponse(tr: TokenEndpointResponse?): UserAccountBuilder {
```
wmathurin, line 75: Key method to get values from token end point response into a user account object.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.kt
```diff
@@ -0,0 +1,529 @@
...
+            .cookieSidClient(tr.cookieSidClient)
+            .sidCookieName(tr.sidCookieName)
+    }
+
+    /**
+     * Set fields from id service response
+     * @param id id service response
+     * @return Instance of this class
+     */
+    fun populateFromIdServiceResponse(id: IdServiceResponse?): UserAccountBuilder {
```
wmathurin, line 106: Key method to get values from id service response into a user account object.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.kt
```diff
@@ -0,0 +1,529 @@
...
+            .language(id.language)
+            .locale(id.locale)
+    }
+
+    /**
+     * Set fields from an existing UserAccount
+     * @param userAccount user account.
+     * @return Instance of this class
+     */
+    fun populateFromUserAccount(userAccount: UserAccount): UserAccountBuilder {
```
wmathurin, line 127: Used during refresh to apply new values onto existing objects (to use populateXXX, we need a UserAccountBuilder with the values of an existing UserAccount).

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.kt
```diff
@@ -0,0 +1,529 @@
...
+    }
+
+    /**
+     * Sets additional OAuth values.
+     *
+     * @param additionalOauthValues Additional OAuth values.
+     * @return Instance of this class.
+     */
+    fun additionalOauthValues(additionalOauthValues: Map<String, String>?): UserAccountBuilder {
+        if (!allowUnset) {
```
wmathurin, line 494: The merge logic is important - and there is a new test to make sure it is right.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1791,7 +1791,7 @@ open class SalesforceSDKManager protected constructor(
         fun encrypt(
             data: String?,
             key: String?
-        ): String? = Encryptor.encrypt(data, key)
+        ): String? = if (data == null || key == null) null else Encryptor.encrypt(data, key)
```
wmathurin, line 1794: The null check here saves us many null checks in the calling code

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1809,7 +1809,7 @@ open class SalesforceSDKManager protected constructor(
         fun decrypt(
             data: String?,
             key: String?
-        ): String? = Encryptor.decrypt(data, key)
+        ): String? = if (data == null || key == null) null else Encryptor.decrypt(data, key)
```
wmathurin, line 1812: The null check here saves us many null checks in the calling code

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -638,7 +641,7 @@ private void populateFromJSON(JSONObject parsedResponse) {
                 firstName = parsedResponse.getString(FIRST_NAME);
                 lastName = parsedResponse.getString(LAST_NAME);
                 displayName = parsedResponse.getString(DISPLAY_NAME);
-                final JSONObject photos = parsedResponse.getJSONObject(PHOTOS);
+                final JSONObject photos = parsedResponse.optJSONObject(PHOTOS);
```
wmathurin, line 644: Old code - if it can get null, we need to to optJSONObject not getJSONObject (which throws).
Because of the catch below, the code will continue with all the fields below the throwing line not set :-(

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -649,7 +652,7 @@ private void populateFromJSON(JSONObject parsedResponse) {
                 userId = parsedResponse.getString(USER_ID);
                 orgId = parsedResponse.getString(ORG_ID);
                 nickname = parsedResponse.getString(NICKNAME);
-                final JSONObject urls = parsedResponse.getJSONObject(URLS);
+                final JSONObject urls = parsedResponse.optJSONObject(URLS);
```
wmathurin, line 655: Old code - if it can get null, we need to to optJSONObject not getJSONObject (which throws).
Because of the catch below, the code will continue with all the fields below the throwing line not set :-(

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -768,6 +771,9 @@ public static class TokenEndpointResponse {
         public String contentDomain;
         public String contentSid;
         public String csrfToken;
+        public String cookieClientSrc;
```
wmathurin, line 774: New fields!

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -395,6 +395,8 @@ public RestResponse sendSync(RestRequest restRequest, Interceptor... interceptor
 
     /**
      * All immutable information for an authenticated client (e.g. username, org ID, etc.).
+     *
+     * TODO revisit this class - some of information is NOT immutable e.g. sids
```
wmathurin, line 399: The description for the class mentions "immutable information" like username and org id.
But it also olds SIDs which can change during a refresh :-(

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -167,30 +164,164 @@ public void testCreateAccountFromBundle() {
...
-                String.format("%s (%s) (%s)", TEST_USERNAME, TEST_INSTANCE_URL, SalesforceSDKManager.getInstance().getApplicationName()),
-                account.getAccountName());
+        checkTestAccount(account);
+    }
+
+    /**
+     * Tests populating account from token end point response and id response
+     */
+    @Test
+    public void testPopulateFromTokenEndpointAndIdService() throws JSONException {
```
wmathurin, line 174: Test for new methods populateFromTokenEndpoint and populateFromIdServiceResponse

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -167,30 +164,164 @@ public void testCreateAccountFromBundle() {
...
+                .clientId(TEST_CLIENT_ID)
+                .build();
+        checkTestAccount(account);
+    }
+
+    /**
+     * Tests that allowUnset behaves as expected
+     */
+    @Test
+    public void testAllowUnset() {
```
wmathurin, line 192: Making sure the allowUnset works as desired

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -167,30 +164,164 @@ public void testCreateAccountFromBundle() {
...
+                .loginServer(null)
+                .loginServer("login-server-3")
+                .build().getLoginServer());
+    }
+
+    /**
+     * Tests that allowUnset behaves as expected
+     */
+    @Test
+    public void testAllowUnsetForAdditionalOauthValues() {
```
wmathurin, line 248: Making sure the allowUnset works as desired for additional oauth values

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -217,6 +347,17 @@ private JSONObject createTestAccountJSON() throws JSONException{
         object.put(UserAccount.LOCALE, TEST_LOCALE);
         object.put(UserAccount.PHOTO_URL, TEST_PHOTO_URL);
         object.put(UserAccount.THUMBNAIL_URL, TEST_THUMBNAIL_URL);
+        object.put(UserAccount.LIGHTNING_DOMAIN, TEST_LIGHTNING_DOMAIN);
```
wmathurin, line 350: Test should include all the fields in case of a copy / paste error when assigning fields from map values etc.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/util/BundleTestHelper.kt
```diff
@@ -0,0 +1,49 @@
+package com.salesforce.androidsdk.util
+
+import android.os.Bundle
+import org.junit.Assert
+
+class BundleTestHelper {
```
wmathurin, line 6: Helper class - we have a similar class for comparing JSONObjects already

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/util/BundleTestHelper.kt
```diff
@@ -0,0 +1,49 @@
...
+            }
+
+            // Fail if one is null and the other is not
+            if (expected == null || actual == null) {
+                Assert.fail("$message: One of the bundles is null.")
+                return
+            }
+
+            // Check if both bundles contain the same keys
+            for (key in expected.keySet()) {
```
wmathurin, line 22: We could just compare the sizes and Assert.fail but then we would not know what fields are missing.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/OAuthWebviewHelper.kt
```diff
@@ -1156,44 +1096,8 @@ open class OAuthWebviewHelper : KeyChainAliasCallback {
...
-            accountOptions?.vfDomain,
-            accountOptions?.vfSid,
-            accountOptions?.contentDomain,
-            accountOptions?.contentSid,
-            accountOptions?.csrfToken,
-            accountOptions?.nativeLogin,
-            accountOptions?.language,
-            accountOptions?.locale
-        )
+        val extras = clientManager.createNewAccount(account)
```
wmathurin, line 1100: Old code creates new account from accountOptions instead of the account passed to the method.
Not sure why since the only caller to that method is the one creating accountOptions from the account.
Why do we have accountOptions??

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountManagerTest.java
```diff
@@ -121,175 +78,126 @@ public void tearDown() throws Exception {
...
-        		TEST_AUTH_TOKEN, TEST_INSTANCE_URL,
-        		TEST_LOGIN_URL, TEST_IDENTITY_URL,
-        		TEST_CLIENT_ID, TEST_ORG_ID_2,
-        		TEST_USER_ID_2, null, null, null,
-                null, null, null, null, null,
-                null, null, null, null, null,
-                null, null, null, false,
-                TEST_LANGUAGE, TEST_LOCALE);
+    private void checkSameUserAccount(UserAccount expected, UserAccount actual) {
+        BundleTestHelper.checkSameBundle("Not the expected user account", expected.toBundle(), actual.toBundle());
```
wmathurin, line 201: We are comparing every fields (previously we were doing a Assert.equals but the UserAccount's equals method only looks at userId and orgId).

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -247,57 +381,213 @@ private Bundle createTestAccountBundle() {
...
     /**
-     * Check for equality of two bundles.
-     *
-     * @param one the first bundle
-     * @param two the second bundle
-     * @return true if the keys/values match
-     *         false otherwise
+     * Create test account
+     */
+    public static UserAccount createTestAccount() {
```
wmathurin, line 403: Static method to create a test account (`UserAccount` object not the account manager account) that this test suite and others all use (e.g. `UserAccountManagerTest` and `ClientManagerTest`).

Also creating the account manager account is a lot simpler now. 
No need to set up `LoginOptions` / `ClientManager` and call a method that takes many args, simply call `UserAccountManager.createAccount(userAccount)`.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/security/BiometricAuthenticationManagerTest.kt
```diff
@@ -60,19 +58,7 @@ class BiometricAuthenticationManagerTest {
...
-        )
-        val clientManager = ClientManager(ctx, ClientManagerTest.TEST_ACCOUNT_TYPE, loginOptions, true)
-        clientManager.createNewAccount(userAccount.accountName, userAccount.username, userAccount.refreshToken, userAccount.authToken,
-            userAccount.instanceServer, userAccount.loginServer, userAccount.idUrl, userAccount.userId, userAccount.orgId,
-            userAccount.userId, userAccount.communityId, userAccount.communityId, userAccount.firstName, userAccount.lastName,
-            userAccount.displayName, userAccount.email, userAccount.photoUrl, userAccount.thumbnailUrl, userAccount.additionalOauthValues,
-            userAccount.lightningDomain, userAccount.lightningSid, userAccount.vfDomain, userAccount.vfSid, userAccount.contentDomain,
-            userAccount.contentSid, userAccount.csrfToken, userAccount.nativeLogin, userAccount.language, userAccount.locale
-        )
+        UserAccountManager.getInstance().createAccount(userAccount);
```
wmathurin, line 61: Simpler! No need for LoginOptions and ClientManager.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/security/ScreenLockManagerTest.kt
```diff
@@ -184,18 +183,7 @@ class ScreenLockManagerTest {
...
-        )
-        val clientManager = ClientManager(ctx, ClientManagerTest.TEST_ACCOUNT_TYPE, loginOptions, true)
-        clientManager.createNewAccount(userAccount.accountName, userAccount.username, userAccount.refreshToken, userAccount.authToken,
-            userAccount.instanceServer, userAccount.loginServer, userAccount.idUrl, userAccount.userId, userAccount.orgId,
-            userAccount.userId, userAccount.communityId, userAccount.communityId, userAccount.firstName, userAccount.lastName,
-            userAccount.displayName, userAccount.email, userAccount.photoUrl, userAccount.thumbnailUrl, userAccount.additionalOauthValues,
-            userAccount.lightningDomain, userAccount.lightningSid, userAccount.vfDomain, userAccount.vfSid, userAccount.contentDomain,
-            userAccount.contentSid, userAccount.csrfToken, userAccount.nativeLogin, userAccount.language, userAccount.locale
-        )
+        UserAccountManager.getInstance().createAccount(userAccount);
```
wmathurin, line 186: Simpler! No need for LoginOptions and ClientManager.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/SalesforceActivity.java
```diff
@@ -57,7 +57,7 @@ public void onResume() {
 
     @Override
     public void onPause() {
-        super.onPause();
+		super.onPause();
```
brandonpage, line 60: NT:  Is this an intended change?  (hard to tell what spacing actually is on github)

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.kt
```diff
@@ -0,0 +1,529 @@
...
+ */
+package com.salesforce.androidsdk.accounts
+
+import com.salesforce.androidsdk.auth.OAuth2.IdServiceResponse
+import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse
+
+/**
+ * This class helps build a UserAccount object.
+ */
+class UserAccountBuilder private constructor() {
```
brandonpage, line 35: I don't think it applies/matters for this class but just as a reminder when we convert from Java to Kotlin we have to be cognizant that by default Java is overrideable and Kotlin is not.  I keep a [list of things internal apps are extending in these tests](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/PublicOverridesTest.kt).

brandonpage (approved):
Looks great.  Lots of thorough cleanup here!

wmathurin:
@brandonpage I merged too quickly, I'll address your comments in the next PR

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountBuilder.kt
```diff
@@ -0,0 +1,529 @@
...
+    }
+
+    /**
+     * Sets the allow unset behavior
+     * When false:
+     * - passing null to a setter does not unset a value previously set
+     * - passing a map to additionAuthValues causes the new map and old map to be merged
+     * @param allowUnset new value for allowUnset
+     */
+    fun allowUnset(allowUnset: Boolean): UserAccountBuilder {
```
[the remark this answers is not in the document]
wmathurin: I believe the refresh token response should have the same fields as the login token response. But if it has a few missing fields, it should not cause them to be gone. The old code has some checks in some places, it would not blindly set some fields with the fields coming from the token response, it would check that there is actually a value.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -768,13 +772,18 @@ public static class TokenEndpointResponse {
...
 
         /**
          * Parameterized constructor built during login flow.
          *
          * @param callbackUrlParams Callback URL parameters.
+         * @param additionalOauthKeys Additional oauth keys.
          */
-        public TokenEndpointResponse(Map<String, String> callbackUrlParams) {
+        @VisibleForTesting(otherwise = VisibleForTesting.PACKAGE_PRIVATE)
+        public TokenEndpointResponse(Map<String, String> callbackUrlParams, List<String> additionalOauthKeys) {
```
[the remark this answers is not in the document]
wmathurin: Previously we had `public TokenEndpointResponse(Map<String, String> callbackUrlParams)` which get the `additionalOauthKeys` from `SalesforceSDKManager`.
Now we still have `public TokenEndpointResponse(Map<String, String> callbackUrlParams)` but there is also this one: `public TokenEndpointResponse(Map<String, String> callbackUrlParams, List<String> additionalOauthKeys)` used by the tests (so that the test does not have to set/unset `additionalOauthKeys` on SalesforceSDKManager.

brandonpage: Yeah, sorry I forgot to remove the comment after seeing the second constructor.

brandonpage:
> @brandonpage I merged too quickly, I'll address your comments in the next PR

No worries, pretty much everything was NIT or feedback for future releases.

PR #2617, opened 2024-09-19
Android 15 Support - Part 1

brandonpage:
This PR covers supporting Android 15 devices **prior to apps updating to API 35**.  I will follow up with a Part 2 that fixes issues and makes necessary improvements for apps targeting API 35.

Changes:
- Fixed endless advanced auth loop is no browser exits.
   - This isn't specific to Android 15 pre se, but a user is significantly more likely to run into this scenario in the new [Private Space](https://source.android.com/docs/security/features/private-space) profile.
-  Update [Predictive System Back](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture) behavior.  
   - "Predictable" for the OS really means that the app doesn't control it anymore.  I have basically moved all of necessary cleanup work to `onDestroy` so that it happens no matter what.
   - The only difference in behavior (besides the fancy new animations ✨) is pressing/gesturing back while locked (Passcode or Bio Auth) now does nothing instead of closing the app.  

All aspects of the SDK seem to work fine in Private Spaces and I have found no indication we have Android for Work logic that needs to be updated take this into account.  ~~The only thing left to test here is IDP within a Private Space, which I am setting up now.~~  Edit:  Working!

I tested [Support for 16 KB page size](https://developer.android.com/guide/practices/page-sizes) with somewhat confusing results on an emulator.  My concern was SQLCipher, but it was updated to support 16 KB page size in the version we are currently using (4.6.1).  Login an sync worked fine but it crashed with an error that didn't make much sense not long after.  I will test again on a physical Pixel 8 Pro after I upgrade it to Android 15.  However, this is still an experimental feature and I don't think we necessarily need to support it now.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -58,7 +58,7 @@ import android.webkit.WebView
 import android.widget.Button
 import android.widget.Toast.LENGTH_SHORT
 import android.widget.Toast.makeText
-import android.window.OnBackInvokedDispatcher.PRIORITY_DEFAULT
+import androidx.activity.addCallback
 import androidx.appcompat.app.AppCompatActivity
 import androidx.biometric.BiometricManager
 import androidx.biometric.BiometricManager.Authenticators.BIOMETRIC_STRONG
@@ -129,19 +129,12 @@ import com.salesforce.androidsdk.R.menu.sf__login as sf__login_menu
  * OAuth web view helper class.
  */
 open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
-
     private var wasBackgrounded = false
-
     private var webviewHelper: OAuthWebviewHelper? = null
-
     private var authConfigReceiver: AuthConfigReceiver? = null
-
     private var receiverRegistered = false
-
     private var accountAuthenticatorResponse: AccountAuthenticatorResponse? = null
-
     private var accountAuthenticatorResult: Bundle? = null
-
     private var biometricAuthenticationButton: Button? = null
 
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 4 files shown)


wmathurin (approved):

brandonpage:
IDP flows work on Android 15 and even (when both apps are within a) Private Space.  I did notice an error when doing the flows inside a Private Space:
```
 E  Failed to retrieve application info (Ask Gemini)
    android.content.pm.PackageManager$NameNotFoundException: Package com.salesforce.samples.restexplorer not found for user 10
    	at android.content.pm.LauncherApps.getApplicationInfo(LauncherApps.java:1315)
    	at com.google.android.gms.internal.wear_companion.zzfzj.zzc(com.google.android.libraries.wear.companion:wear-companion-android@@2.29.0:13)
    	at com.google.android.gms.internal.wear_companion.zzevw.zza(com.google.android.libraries.wear.companion:wear-companion-android@@2.29.0:255)
    	at com.google.android.gms.internal.wear_companion.zzepo.zzz(com.google.android.libraries.wear.companion:wear-companion-android@@2.29.0:61)
    	at com.google.android.gms.internal.wear_companion.zzepo.zzB(com.google.android.libraries.wear.companion:wear-companion-android@@2.29.0:74)
    	at com.google.android.gms.internal.wear_companion.zzepo.zzg(com.google.android.libraries.wear.companion:wear-companion-android@@2.29.0:27)
    	at com.google.android.gms.internal.wear_companion.zzeox.run(com.google.android.libraries.wear.companion:wear-companion-android@@2.29.0:9)
    	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
    	at java.lang.Thread.run(Thread.java:1012)
```

Maybe this is because the Pixel Watch App is installed in (and can't be removed from ?!?) the Private Space?  Figured it was worth mentioning.

PR #2621, opened 2024-09-25
Android 15 support

brandonpage:
- Bumped compileSdk and targetSdk of sample apps to 35
- Fixed an issue related to new intent rules in Android 15 that broke IDP initiated IDP flow.  
- Enabled [Edge to Edge](https://developer.android.com/develop/ui/views/layout/edge-to-edge) UI for `ScreenLockActivity` and fixed the status bar icon color in dark mode.
- Opted out of Edge to Edge for most Mobile SDK activities (screenshot and more information below -- **please read**).
- Opted out of Edge to Edge for Hybrid Apps.

IMG_0002

The left is what the login screen looks like by default with Edge to Edge if the App targets API 35.  The opt out flag is not advertised at all by Google, was introduced as **deprecated** and has code comments clearly stating this should only be used as a temporary solution.  The reason I chose to do this for MSDK 12.2 is because with our XML based UIs poses some issues.  For example, I can adjust the elements on the login screen to be in the correct place but the status bar above our blue url bar will remain transparent.  The API used to color is is _removed_ in API 35.  There are hacks to work around it like adding another UI element there but since customers expect to be able to theme that like they do today it complicates things significantly.  Considering how simple our screens are I think the best thing to do going forward is to rewrite them with Jetpack Compose in 13.0.  This would include Login, Server Picker, User Switcher, and Dev Info.  

I believe we will need to use the workaround for Hybrid apps until Cordova updates to support API 35.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath("com.android.tools.build:gradle:8.5.0")
+        classpath("com.android.tools.build:gradle:8.6.1")
         classpath("io.github.gradle-nexus:publish-plugin:1.1.0")
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.24")
         classpath("org.jacoco:org.jacoco.core:0.8.12")
--- a/hybrid/HybridSampleApps/MobileSyncExplorerHybrid/AndroidManifest.xml
+++ b/hybrid/HybridSampleApps/MobileSyncExplorerHybrid/AndroidManifest.xml
@@ -16,7 +16,7 @@
         android:name="com.salesforce.androidsdk.phonegap.app.HybridApp"
         android:manageSpaceActivity="com.salesforce.androidsdk.ui.ManageSpaceActivity"
 		android:icon="@drawable/sf__icon"
-        android:enableOnBackInvokedCallback="@bool/isGraterThan33"
+        android:enableOnBackInvokedCallback="@bool/isGreaterThan33"
         tools:ignore="UnusedAttribute">
 
         <!--  Main activity -->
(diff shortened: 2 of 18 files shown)


wmathurin (commented):
Don't we bump up the versions in x.0 version and only certify it runs on the latest version in the y.2 version (assuming y is x-1) ??

hybrid/HybridSampleApps/MobileSyncExplorerHybrid/AndroidManifest.xml
```diff
@@ -15,11 +16,11 @@
         android:name="com.salesforce.androidsdk.phonegap.app.HybridApp"
         android:manageSpaceActivity="com.salesforce.androidsdk.ui.ManageSpaceActivity"
 		android:icon="@drawable/sf__icon"
-        android:enableOnBackInvokedCallback="@bool/isGraterThan33">
+        android:enableOnBackInvokedCallback="@bool/isGraterThan33"
```
wmathurin, line 19: I realize it was there before - but what is that?

brandonpage:
> Don't we bump up the versions in x.0 version and only certify it runs on the latest version in the y.2 version (assuming y is x-1) ??

I only bumped the the version for the sample apps.  Our libraries still target API 34 (Android 14).

brandonpage:
CI is failing because it does not recognize `windowOptOutEdgeToEdgeEnforcement` due to it being added in API 35.  This make me realize I should gate this with a `values-v35` folder so apps that haven't updated don't see this error.  I'll test and update the PR.

build.gradle.kts
```diff
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath("com.android.tools.build:gradle:8.5.0")
+        classpath("com.android.tools.build:gradle:8.6.1")
```
wmathurin, line 12: Are gradle upgrades source of friction for apps team adoption of MSDK?

wmathurin (approved):

build.gradle.kts
brandonpage: Yeah, if they consume via source.  I noticed a warning telling me I should use 8.6 if I target API 35 though so I updated it.

brandonpage:
I will send a new PR taking a different approach, by not opting out of Edge to Edge.

PR #2624, opened 2024-09-28
Android 15 support 

brandonpage:
- Bumped compileSdk and targetSdk of sample apps to 35 and Gradle to 8.6.1 to match.
- Fixed an issue related to new intent rules in Android 15 that broke IDP initiated IDP flow.
- Enabled [Edge to Edge](https://developer.android.com/develop/ui/views/layout/edge-to-edge) UI for ScreenLockActivity and fixed the status bar icon color in dark mode.
- Conditionally (if app targets API 35+) enabled Edge to Edge for all other screens.  Fixed layouts taking the entire screen and showing behind status and navigation bars.  Fixed status and navigation bar colors for all screens.

### DIFF
--- a/libs/SalesforceSDK/res/drawable/sf__fix_status_bar.xml
+++ b/libs/SalesforceSDK/res/drawable/sf__fix_status_bar.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:id="@+id/sf__status_bar_background" android:height="100dp" >
+        <shape android:shape="rectangle">
+            <size android:height="100dp" android:width="100dp"/>
+            <solid android:color="@color/sf__primary_color"/>
+        </shape>
+    </item>
+
+    <item android:id="@+id/sf__nav_bar_background" android:height="100dp">
+        <shape android:shape="rectangle">
+            <size android:height="100dp" android:width="100dp"/>
+            <solid android:color="@android:color/transparent"/>
+        </shape>
+    </item>
+</layer-list>
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPManager.kt
@@ -119,7 +119,6 @@ internal class IDPManager(
                     spConfig,
                     codeChallenge
                 ) { result ->
-                    authCodeActivity.finish()
                     onResult(result)
                 }
             }
@@ -214,6 +213,7 @@ internal class IDPManager(
                 addCategory(Intent.CATEGORY_DEFAULT)
             }
             startActivity(activeFlow.context, launchIntent)
+            activeFlow.authCodeActivity?.finish()
         } else {
             activeFlow.onStatusUpdate(Status.ERROR_RECEIVED_FROM_SP)
         }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/EdgeToEdgeUtil.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/EdgeToEdgeUtil.kt
@@ -1,0 +1,36 @@
+package com.salesforce.androidsdk.ui
+
+import android.graphics.drawable.LayerDrawable
+import android.os.Build.VERSION_CODES.UPSIDE_DOWN_CAKE
+import android.view.View
+import androidx.activity.enableEdgeToEdge
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.content.res.ResourcesCompat
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.updatePadding
+import com.salesforce.androidsdk.R.drawable.sf__fix_status_bar
+import com.salesforce.androidsdk.R.id.sf__status_bar_background
+
+// TODO: Remove this in 13.0 after rewriting screens in compose.
+internal fun AppCompatActivity.fixEdgeToEdge(view: View) {
+    if (application.applicationInfo.targetSdkVersion > UPSIDE_DOWN_CAKE) {
+        enableEdgeToEdge()
+        ViewCompat.setOnApplyWindowInsetsListener(view) { listenerView, windowInsets ->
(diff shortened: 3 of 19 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/EdgeToEdgeUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+            // Fix transparent status bar not matching action bar
+            val background = ResourcesCompat.getDrawable(resources, sf__fix_status_bar, null)
+            val statusBarFiller = (background as LayerDrawable).findDrawableByLayerId(sf__status_bar_background)
+            statusBarFiller.setBounds(0, 0, insets.right, insets.bottom)
+            view.setBackgroundResource(sf__fix_status_bar)
+
+            WindowInsetsCompat.CONSUMED
+        }
+    }
+}
```
brandonpage, line 36: This is where all of UX magic happens.

libs/SalesforceSDK/res/drawable/sf__fix_status_bar.xml
```diff
@@ -0,0 +1,16 @@
...
+        </shape>
+    </item>
+
+    <item android:id="@+id/sf__nav_bar_background" android:height="100dp">
+        <shape android:shape="rectangle">
+            <size android:height="100dp" android:width="100dp"/>
+            <solid android:color="@android:color/transparent"/>
+        </shape>
+    </item>
+</layer-list>
```
brandonpage, line 16: This is sort of a hack to fill in the status and navigation bars with the appropriate color since they are now forced transparent and completely unsettable in Edge to Edge.  Note that I am changing the height of the top rectangle to match the device status bar height in code!

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPManager.kt
```diff
@@ -214,6 +213,7 @@ internal class IDPManager(
                 addCategory(Intent.CATEGORY_DEFAULT)
             }
             startActivity(activeFlow.context, launchIntent)
+            activeFlow.authCodeActivity?.finish()
```
brandonpage, line 216: The IDP fix.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/EdgeToEdgeUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+        ViewCompat.setOnApplyWindowInsetsListener(view) { listenerView, windowInsets ->
+            val insets = windowInsets.getInsets(
+                WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.displayCutout()
+            )
+
+            // Fix screens being drawn behind status and navigation bars
+            listenerView.updatePadding(insets.left, insets.top, insets.right, insets.bottom)
+
+            // Fix transparent status bar not matching action bar
+            val background = ResourcesCompat.getDrawable(resources, sf__fix_status_bar, null)
```
wmathurin, line 28: Is there a chance background could be null? Should we check for null? E.g. if they subclass our activity and use a different layout.

brandonpage: "background" is the new [sf__fix_status_bar.xml](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2624/files/5a859840bbab9d2ccd2f7a5f632982a9d9cc2162#diff-d3e42c86dfab7e3f197037eb5cd45c220cce7298278a935c00277d3a038010e5) drawable added in this PR.  I set it as the background of the activity later with `view.setBackgroundResource(sf__fix_status_bar)`.  

The drawable is split in half, the top being `sf__primary_color` and the bottom is transparent.  I am getting a reference to it here so I can set the height of the top to the height of the status bar.  This mimics the old look.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/EdgeToEdgeUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.core.content.res.ResourcesCompat
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.updatePadding
+import com.salesforce.androidsdk.R.drawable.sf__fix_status_bar
+import com.salesforce.androidsdk.R.id.sf__status_bar_background
+
+// TODO: Remove this in 13.0 after rewriting screens in compose.
+internal fun AppCompatActivity.fixEdgeToEdge(view: View) {
+    if (application.applicationInfo.targetSdkVersion > UPSIDE_DOWN_CAKE) {
```
brandonpage, line 17: I should actually check this _and_ `SDK_INT` so we don't do anything special on pre-Android 15 devices.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/EdgeToEdgeUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+            )
+
+            // Fix screens being drawn behind status and navigation bars
+            listenerView.updatePadding(insets.left, insets.top, insets.right, insets.bottom)
+
+            // Fix transparent status bar not matching action bar
+            val background = ResourcesCompat.getDrawable(resources, sf__fix_status_bar, null)
+            val statusBarFiller = (background as LayerDrawable).findDrawableByLayerId(sf__status_bar_background)
+            statusBarFiller.setBounds(0, 0, insets.right, insets.bottom)
+            view.setBackgroundResource(sf__fix_status_bar)
```
wmathurin, line 31: I think `setBackgroundResource` does a `getDrawable` followed by `setBackground` so we could just do:
`view.setBackground(background)`

Also we might only want to set the background once:
```
if (view.background != background) {
    view.setBackground(background)
}
```

brandonpage: I tried the if but every time the breakpoint was hit `view.background` did not equal `background` and I did not seeing it being called multiple times.   I am updating this code though some of it is actually unnecessary.

wmathurin (approved):

PR #2625, opened 2024-10-03
@W-16336722: [MSDK Android] QR Code Log In Using External QR Code Reader Does Not Display Main Activity

JohnsonEricAtSalesforce:
🤘🏻 _*Ready For Review!*_ 🎸

  This resolves an issue where scanning the login QR code from the Android QR scanner tile would log in, but then finish without starting the main activity.  The root cause and the fix details are in my walk through notes.

  Also, I significantly cleaned up the public and private API MSDK exposes for QR Code Login and UI Bridge front door URL login.  The source commentary should also be much more helpful now.

  This is a companion to https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/426


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -215,6 +215,18 @@ open class SalesforceSDKManager protected constructor(
      */
     val loginActivityClass: Class<out Activity> = nativeLoginActivity ?: webviewLoginActivityClass
 
+    /**
+     * For Salesforce Identity API UI Bridge support, the overriding front door bridge URL to
+     * use in place of the default initial login URL
+     */
+    var frontDoorBridgeUrl: String? = null
```
JohnsonEricAtSalesforce, line 222: When the app receives a deep-linked login QR code from the external QR code reader, that Intent has to be received by the app's main activity such that the login broadcast intent receivers will be registered on a cold start of the app.  In that event, the main application will be about to trigger login since the call to fetch the REST client will automatically display login as a side effect and return null.  These two new variables our on SDK manager singleton give the client manager and login activity the hint that there is a pending front door URL to use.

Overall, I feel we'd be better if the existing SDK design gave us another means of doing this.  But, there are a couple patterns in place we'd have to resolve first for trust work:
- Either move the broadcast intent receivers to the application class rather than a specific activity or publish login events on another bus
- Make the display of the login activity pass through the app rather than be a side effect of getting the REST client
- Remove the singleton pattern of the SDK manager

I've been thinking about a more future-forward means of passing these two data points between the deep-link activity (main application) and the login activity in the context of MSDK's current implementation.  I haven't gotten a good idea yet.  Shall we keep that for a to-do when we clean up the trust work above or is there a cleaner way we could use now?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -145,8 +145,17 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
         val salesforceSDKManager = SalesforceSDKManager.getInstance()
 
-        // Determine if the activity was created from a deep link intent with QR code log in via UI bridge API parameters.
-        val isDeepLinkedQrCodeLogin = isDeepLinkedQrCodeLogin(intent)
+        /*
+         * For Salesforce Identity API UI Bridge support, the overriding front door bridge URL to
+         * use in place of the default initial login URL plus the optional web server flow code
+         * verifier accompanying the front door bridge URL
+         */
+        val frontDoorBridgeUrl = salesforceSDKManager.frontDoorBridgeUrl
```
JohnsonEricAtSalesforce, line 153: Here's the use of the new SDK manager properties.  I also remove the QR code semantic here since maybe this code could be receiving a front door URL from another means.  Note how both properties are unset after use.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -163,17 +172,21 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
 
         // Protect against screenshots
         window.setFlags(FLAG_SECURE, FLAG_SECURE)
 
-        // Fetch authentication configuration except for QR code login.
-        if (!isDeepLinkedQrCodeLogin) {
+        /*
+         * Fetch authentication configuration except when using Salesforce Identity API UI Bridge, since the front door URL may be for another login server.
+         *
+         * Support for adding new login servers from the front door bridge URL is not yet implemented.  Unknown login servers will fail login.
```
JohnsonEricAtSalesforce, line 187: I documented our pending implementation of dynamically added login servers here.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -675,144 +693,193 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
 
     /**
-     * Automatically log in with a UI Bridge API login QR code.
-     * @param loginQrCodeContent The login QR code content.  This should be
-     * either a URL or URL query containing the UI Bridge API JSON parameter.
-     * The UI Bridge API JSON parameter should contain URL-encoded JSON with two
-     * values:
-     * - frontdoor_bridge_url
-     * - pkce_code_verifier
+     * Automatically log in with a UI Bridge API front door bridge URL and PKCE code verifier.
```
JohnsonEricAtSalesforce, line 696: The remainder of the diff largely reflects my attempt at comprehensively documenting all the properties and methods login activity provides for use in the "reference" implementation of QR code login.  I attempted to name and document everything so we have a clean, understandable and customizable public API.  It's easier to read in Android Studio than git.

@wmathurin - I moved everything back to login activity (from our conversation at DSU) so all apps following our reference implementation's APEX and Visualforce Pages can simply call these reusable methods.  Apps that wish to customize will also have properties they can tweak or they can replace the entire implementation if they're responsible for obtaining the front door URL and code verifier.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -675,144 +693,193 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-    /**
-     * Automatically log in with a UI Bridge API front door bridge URL and PKCE
-     * code verifier.
      * @param frontdoorBridgeUrl The UI Bridge API front door bridge URL
-     * @param pkceCodeVerifier The PKCE code verifier
+     * @param pkceCodeVerifier The optional PKCE code verifier, which is not required for User Agent
+     * Authorization Flow but is required for Web Server Authorization Flow
      */
     @Suppress("MemberVisibilityCanBePrivate")
     fun loginWithFrontdoorBridgeUrl(
```
JohnsonEricAtSalesforce, line 710: This is our universal entry point to MSDK's front door bridge URL login logic, even if the app chooses not to use our reference implementation's APEX, Visualforce Page and the associated login QR code URL/JSON format.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -675,144 +693,193 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-    private fun uiBridgeApiParametersFromUiBridgeApiJson(
-        uiBridgeApiParameterJsonString: String
-    ) = JSONObject(uiBridgeApiParameterJsonString).let { uiBridgeApiParameterJson ->
-        UiBridgeApiParameters(
-            uiBridgeApiParameterJson.getString(FRONTDOOR_BRIDGE_URL_KEY),
-            when (uiBridgeApiParameterJson.has(PKCE_CODE_VERIFIER_KEY)) {
-                true -> uiBridgeApiParameterJson.optString(PKCE_CODE_VERIFIER_KEY)
-                else -> null
-            }
+    fun loginWithFrontdoorBridgeUrlFromQrCode(
```
JohnsonEricAtSalesforce, line 729: This is our entry point to MSDK's front door bridge URL login logic when the app does choose to use our reference implementation's APEX, Visualforce Page and the associated login QR code URL/JSON format.   The app just passes the URL from the QR code to MSDK and magic ensues.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -675,144 +693,193 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
+         *
+         * Apps may customize this so long as it matches the server-side Apex class or other code
+         * generating the QR code.
+         *
+         * Apps need not use the QR code login URL structure provided in this companion object if
+         * they wish to entirely customize the QR code login URL format and implement a custom
+         * parsing scheme.
+         */
+        @Suppress("MemberVisibilityCanBePrivate")
+        var qrCodeLoginUrlPath = "/login/qr"
```
JohnsonEricAtSalesforce, line 767: Instead of private constants, I made these companion properties variables so the app could make partial adjustments to the formatting of the QR code login URL and JSON while still reusing the logic.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -675,144 +693,193 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-        /** The login QR code's UI Bridge API parameter's JSON PKCE code verifier key */
-        private const val PKCE_CODE_VERIFIER_KEY = "pkce_code_verifier"
+        // endregion
+        // region QR Code Login Via Salesforce Identity API UI Bridge Private Implementation
+
+        /**
+         * For QR code login URLs, a regular expression to extract the Salesforce Identity API UI
+         * Bridge parameter JSON string.
+         */
+        private val qrCodeLoginJsonRegex by lazy {
```
JohnsonEricAtSalesforce, line 846: @wmathurin - I couldn't reproduce the "external" regular expression you documented in the spike.  My internal and external QR code URLs are identical on a Google Pixel 9 Pro.

JohnsonEricAtSalesforce (commented):
Here's my self review with a detailed code walk-through.  Thanks for reading and sharing ideas.

libs/MobileSync/build.gradle.kts
```diff
@@ -99,5 +99,5 @@ android {
 }
 
 kotlin {
-    jvmToolchain(17)
+    jvmToolchain(21)
```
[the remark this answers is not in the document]
brandonpage: Uhhhh, this seems like a chance we probably want to save for 13.0.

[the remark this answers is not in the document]
JohnsonEricAtSalesforce: @brandonpage - Could you see where your Java 17 was being resolved from when you updated to Ladybug? If we don't want this update now, then possibly I just need to go through the process to acquire Java 17 locally.

This is what got me:
```
/Applications/Android\ Studio.app/Contents/jbr/Contents/Home/bin/java -version
openjdk version "21.0.3" 2024-04-16
OpenJDK Runtime Environment (build 21.0.3+-79915915-b509.11)
OpenJDK 64-Bit Server VM (build 21.0.3+-79915915-b509.11, mixed mode)
```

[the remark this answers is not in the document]
brandonpage: I think mine is set from a Java install I did, not Android studio.  If I run `java --version` from anywhere I get:

```
openjdk 17.0.8.1 2023-08-24
OpenJDK Runtime Environment Homebrew (build 17.0.8.1+0)
OpenJDK 64-Bit Server VM Homebrew (build 17.0.8.1+0, mixed mode, sharing)
```

Android Studio settings look like this:
Screenshot 2024-10-07 at 9 11 31 AM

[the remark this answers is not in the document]
brandonpage: @JohnsonEricAtSalesforce I got a strange error when I tried to use `JAVA_HOME` from the above list and the IDE suggested `jbr-21`.  Gradle sync/compile/run works with that on the latest Ladybug feature drop so I really don't think we need to update the toolchain value.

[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I was able to download JBR 17 via Android Studio and install it in the recommended location of `~/Library/Java`.  It works, so I rolled this back.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
wmathurin: +1 having a TODO for future cleanup (or maybe a story on GUS).

wmathurin (approved):

brandonpage (approved):

PR #2628, opened 2024-10-09
@W-16814802: [iOS] Add Deep Link From External QR Code Reader To iOS Native Swift Template QR Code Login (Parity For Android)

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!*_ 🥁

  When preparing the documentation for QR code log in on iOS, I noticed Android's implementation was more probably more different than required.  So, this brings a large portion of the documentation for the two in line.

  Also, Android still had one more bug where the deep link intent to main activity could be triggered again when returning to the activity.   That is fixed now.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -727,7 +727,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
      * [LoginActivity.loginWithFrontdoorBridgeUrl].
      *
      * @param qrCodeLoginUrl The QR code login URL
-     * @return Boolean true if a log in attempt is possible using the provided QR code log in URL,
+     * @return Boolean true if a log in attempt is possible using the provided QR code login URL,
      * false otherwise
      */
     fun loginWithFrontdoorBridgeUrlFromQrCode(
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -152,14 +152,13 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
          * frontdoor bridge URL.
          */
         val isUsingFrontDoorBridge = isFrontdoorBridgeUrlIntent(intent) || isQrCodeLoginUrlIntent(intent)
-        val uiBridgeApiParameters = when (isQrCodeLoginUrlIntent(intent)) {
-            true -> uiBridgeApiParametersFromQrCodeLoginUrl(intent.data?.toString())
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -146,16 +146,21 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-        val frontDoorBridgeUrl = salesforceSDKManager.frontDoorBridgeUrl
-        val frontDoorBridgeCodeVerifier = salesforceSDKManager.frontDoorBridgeCodeVerifier
-        val isUsingFrontDoorBridge = frontDoorBridgeUrl != null
-        // Reset the the Salesforce SDK manager's UI bridge support.
-        salesforceSDKManager.frontDoorBridgeUrl = null
-        salesforceSDKManager.frontDoorBridgeCodeVerifier = null
+        val isUsingFrontDoorBridge = isFrontdoorBridgeUrlIntent(intent) || isQrCodeLoginUrlIntent(intent)
+        val uiBridgeApiParameters = when (isQrCodeLoginUrlIntent(intent)) {
+            true -> uiBridgeApiParametersFromQrCodeLoginUrl(intent.data?.toString())
+            else -> intent.getStringExtra(EXTRA_KEY_FRONTDOOR_BRIDGE_URL)?.let { frontdoorBridgeUrl ->
```
brandonpage, line 157: NIT:  [Kotlin style guide says to prefer if to when for binary conditions](https://kotlinlang.org/docs/coding-conventions.html#if-versus-when).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/SalesforceActivity.java
```diff
@@ -38,6 +38,13 @@ public abstract class SalesforceActivity extends AppCompatActivity implements Sa
 
 	private final SalesforceActivityDelegate delegate;
 
+	/**
+	 * An option to build the REST client automatically on activity resume, which has the side
+	 * effect of displaying the default login activity when no user is logged in.
+	 */
+	@Deprecated
+	protected boolean buildRestClientOnResume = true;
```
brandonpage, line 46: What exactly is this for?  The host app to set in their activity if they are using QR Code Login?  It does not appear to be used anywhere.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: I gotcha in 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2628/commits/e2f40be4bff9149bfb5be9bde9694868cfa7d287

JohnsonEricAtSalesforce (commented):
Oops - My self-review notes on this one were still pending.  It's better late than never 🙃

wmathurin (approved):
LGTM -  we can always refine/revisit in 13.0

PR #2630, opened 2024-10-21
Test and Test App Cleanup

brandonpage:
- Fixed Server Picker UI tests.
- Downgraded hybrid test apps to API 34 to avoid Edge to Edge issues.
- Fixed Edge to Edge issues on native test apps.
- Fixed a couple AndroidManifest issues with test apps.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
@@ -300,7 +300,9 @@ public class ServerPickerActivity extends AppCompatActivity implements
          * The only other way to do this is with the NO_HISTORY flag on the custom tab, however
          * this will cause it to always reload on background -- breaking MFA.
          */
-        startActivity(intent);
+        if (!SalesforceSDKManager.getInstance().getIsTestRun()) {
+            startActivity(intent);
+        }
         finish();
     }
 }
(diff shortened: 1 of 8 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ServerPickerActivity.java
```diff
@@ -300,7 +300,9 @@ public void onAuthConfigFetched() {
          * The only other way to do this is with the NO_HISTORY flag on the custom tab, however
          * this will cause it to always reload on background -- breaking MFA.
          */
-        startActivity(intent);
+        if (!SalesforceSDKManager.getInstance().getIsTestRun()) {
+            startActivity(intent);
+        }
```
brandonpage, line 305: I really don't love this, but it seems to be the only way to get the tests to run consistently.

native/NativeSampleApps/test/RestExplorerTest/src/com/salesforce/samples/restexplorer/ServerPickerActivityTest.java
```diff
@@ -26,7 +26,6 @@
  */
 package com.salesforce.samples.restexplorer;
 
-import static androidx.test.InstrumentationRegistry.getInstrumentation;
```
brandonpage, line 29: Every change in this class is either fixing a warning or changing `Thread.sleep` to `device.waitForIdle` (which speeds things up).

wmathurin (approved):

PR #2631, opened 2024-11-01
@W-17092709: [MSDK Android] Cannot Log In To AccountEditor Hybrid Sample App

JohnsonEricAtSalesforce:
🥁 _*Ready For Review*_ 🤘🏻

  This corrects a potential crash where a call the load the SQLCipher library was outside of the critical section to synchronize multithreaded access to it.  The trick was that only the call to `AtomicBoolean.compareAndSet(...)` was thread safe.  Once it completed and `loadLibrary` was working other threads could continue.

  There was a way to do this with `AtomicBoolean` which I tried in the first commit.  Then, I though more about the `Lock` object I often use with Kotlin Concurrent's `withLock` extension.  There _is_ a way to do that in Java so we move one step towards Kotlin conversion here.  Also, we don't seem to need the boolean as `loadLibrary` gracefully completes in milliseconds once the library is already loaded.

  I'd previously observed that changes to the dependency structure could affect this from compile time.  Also, running on a device compared to an emulator was a notable change.  Those _remain_ true observations, but only because it triggers the race condition.

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
@@ -65,7 +65,13 @@ public class DBOpenHelper extends SQLiteOpenHelper {
 	public static final String DATABASES = "databases";
 	private String dbName;
 
-	private static AtomicBoolean libsLoaded = new AtomicBoolean(false);
+	/**
+	 * Indicates if the SQLCipher native library is loaded. This is used as an object monitor for
+	 * concurrency critical sections and as an atomic variable.
+	 * <p>
+	 * Replace this use of java.uti.concurrent with Kotlin Concurrent's Lock.withLock
+	 */
+	private static final AtomicBoolean isSqlCipherLibraryLoaded = new AtomicBoolean(false);
 
 	/*
 	 * Cache for the helper instances
@@ -179,7 +185,7 @@ public class DBOpenHelper extends SQLiteOpenHelper {
                 SmartStoreLogger.e(TAG, "Error occurred while creating JSON", e);
 			}
 			EventBuilderHelper.createAndStoreEvent(eventName, account, TAG, storeAttributes);
-			loadLibsIfNeeded(ctx);
+			loadNativeLibrariesIfNeeded();
 			helper = new DBOpenHelper(encryptionKey, ctx, fullDBName);
 			openHelpers.put(fullDBName, helper);
 		}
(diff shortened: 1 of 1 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java
```diff
@@ -191,9 +198,12 @@ protected DBOpenHelper(String encryptionKey, Context context, String dbName) {
 		this.dbName = dbName;
 	}
 
-	protected static void loadLibsIfNeeded(Context context) {
-		if (libsLoaded.compareAndSet(false, true)) {
-			 System.loadLibrary("sqlcipher");
+	protected static void loadNativeLibrariesIfNeeded() {
```
wmathurin, line 201: I can see how it avoids concurrent calls but how does it avoids loading the library more than once.

Should you do something like:
```java
isSqlCipherLibraryLoaded.lock();
try {
  if (!libsLoaded.compareAndSet(false, true)) {
      System.loadLibrary("sqlcipher");
  }   
} finally {
  isSqlCipherLibraryLoaded.unlock();
}
```		

Also wouldn't it be simpler to do
```java
synchronized(this.class) {
   ...
}
```java

Comment (JohnsonEricAtSalesforce) [reply libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java:201]: According to my research and the AI summary on Google, `loadLibrary` has built in protections against duplication.  So, our goal here becomes making sure we don't let dependent threads through while one is still running `loadLibrary`, which was the case before.

I also profiled `loadLibrary` and it runs hundreds of times faster on subsequent calls for the same library, so there's no performance impact and that also hints that it is a fast-return on those additional calls.  It runs in less than 10 ms., so we can enjoy reducing our code footprint for readability knowing this.  We can also move away from using both `synchronized` and the atomic values.

Good question.  I believe we've got that covered.

Comment (wmathurin) [reply libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java:201]: Cool.

Don't you think the following would be more readable:
```java
synchronized(DBOpenHelper.class) { 
  System.loadLibrary("sqlcipher");
}
```

JohnsonEricAtSalesforce: Thanks!  Synchronizing on a monitor object or class would work and provide similar readability.  `Lock` provides some benefits beyond creating a critical section just with the synchronization keyword.

https://developer.android.com/reference/kotlin/java/util/concurrent/locks/Lock

One reason I read to at least create a dedicated monitor topic for each concurrency subject is so that different topics don't compete for the overall class/instance monitor.  That can pose future refactoring needs if there is more than one concurrency topic per instance/class.  A `Lock` instance provides benefits beyond a simple monitor object, also, which also is really cool when we migrate this Java to Kotlin 🚀

wmathurin (approved):

PR #2610, opened 2024-09-05
Using Single Access API to get front door URL

wmathurin:
- Added `RestRequest.getRequestForSingleAccess` (with new tests in `RestClientTest` and `RestRequestTest`) which calls `/services/oauth2/singleaccess` to generate a frontdoor URL to bridge into UI sessions.
- Added `shouldRefresh` method in `RestClient.OAuthRefreshInterceptor`: it now refreshes when getting either a 401 or a 403 with reason `Bad_OAuth_Token` while going to `/services/oauth2` (as long as the biometric is not enabled and locked).
- Using `getRequestForSingleAccess` in `IDPAuthCodeHelper` instead of building 
a frontdoor URL directly - tested the IDP/SP flows manually (including the case where the access token has expired).

NB: Left `SalesforceDroidGapActivity.getFronDoorUrl` - we will address it in a separate PR by using the cookies returned by the hybrid authentication flow instead of going through front door.

PR #2632, opened 2024-11-01
Support for JWT-based access token

wmathurin:
## Overview
Connected app / external client app can be configured to issue JWT-based access token (for more information see [here](https://help.salesforce.com/s/articleView?id=sf.connected_app_enable_jwt.htm&type=5)).

The mobile application does not need to know if it is using a JWT-based access token or an opaque token **in most cases**.
Login and refresh will just work as before.

However, front door URL cannot be built directly using the JWT-based access token. Instead the [single access API](https://help.salesforce.com/s/articleView?id=sf.frontdoor_singleaccess.htm&type=5) should be called to generate front door URL.

We already made changes in the Mobile SDK to be JWT-based access token ready:
- We introduced a REST wrapper for single access API and leveraged it in the IDP-SP flow (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2610)
- We moved away from front-door and hydrated the web view session in hybrid remote apps using the cookies returned by the oauth hybrid flow (https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2618)

## Changes in this PR
- We are saving the token format and parent sid returned by the token end point during login and refresh in the user account.
- We added a class `JwtAccessToken` to allow apps to inspect the JWT encoded in the JWT-based access token.
- We modified `SalesforceWebViewCookieManager` i.e. the class responsible for hydrating the web view sessions in hybrid remote apps when JWT-based access token are in use.

## Testing
- We added tests to check that token format and parent sid are properly captured and saved.
- We added tests for `JwtAccessToken`.
- We manually verified that login/refresh works in native applications and hybrid remote applications.
- We manually verified that the IDP-SP flows work when JWT-based access token are being used.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/JwtAccessToken.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/JwtAccessToken.kt
@@ -1,0 +1,63 @@
+package com.salesforce.androidsdk.auth
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import java.util.Base64
+
+@Serializable
+data class JwtHeader(
+    @SerialName("alg") val algorithn: String? = null,
+    @SerialName("typ") val type: String? = null,
+    @SerialName("kid") val keyId: String? = null,
+    @SerialName("tty") val tokenType: String? = null,
+    @SerialName("tnk") val tenantKey: String? = null,
+    @SerialName("ver") val version: String? = null,
+)
+
+@Serializable
+data class JwtPayload(
+    @SerialName("aud") val audience: List<String>? = null,
+    @SerialName("exp") val expirationTime: Int? = null,
+    @SerialName("iss") val issuer: String? = null,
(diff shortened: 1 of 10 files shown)


wmathurin:
NB: I decided not to change the transparent refresh logic. I'm not sure that the added complexity of preemptively refreshing before sending a request if using an expired JWT-based access is worth doing. Thoughts?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/JwtAccessToken.kt
```diff
@@ -0,0 +1,63 @@
...
+    @SerialName("nbf") val notBeforeTime: Int? = null,
+    @SerialName("sub") val subject: String? = null,
+    @SerialName("scp") val scopes: String? = null,
+    @SerialName("client_id") val clientId: String? = null,
+)
+
+/**
+ * Data class for decoding JWT-based access token
+ */
+data class JwtAccessToken(
```
brandonpage, line 32: 🚀  Love it.

brandonpage (approved):

PR #2633, opened 2024-11-04
Added Null check to prevent crash for TrailHead Go Android

umashankar-sf:
[com.salesforce.trailheadgo.app_issue_fa4327a6a3154091d49279fa5eba233c_crash_session_6722C99202280001286FF83BCF5223CF_DNE_0_v2_stacktrace.txt](https://github.com/user-attachments/files/17622028/com.salesforce.trailheadgo.app_issue_fa4327a6a3154091d49279fa5eba233c_crash_session_6722C99202280001286FF83BCF5223CF_DNE_0_v2_stacktrace.txt)

https://console.firebase.google.com/project/trailhead-go/crashlytics/app/android:com.salesforce.trailheadgo.app/issues/3acb1308797bc3f5179c50ae5d9d0335?time=last-ninety-days&types=crash&versions=4.1.1%20(4)%20(4010101)&sessionEventKey=672B3B16020000015F77FD9E92962BED_2012546185538799571

https://console.firebase.google.com/project/trailhead-go/crashlytics/app/android:com.salesforce.trailheadgo.app/issues/fa4327a6a3154091d49279fa5eba233c?time=last-ninety-days&types=crash&versions=4.1.1%20(4)%20(4010101)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/AppLockManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/security/AppLockManager.kt
@@ -63,9 +63,9 @@ internal abstract class AppLockManager(
         EventsObservable.get().notifyEvent(EventsObservable.EventType.AppUnlocked)
     }
 
-    fun getAccountPrefs(account: UserAccount): SharedPreferences {
+    fun getAccountPrefs(account: UserAccount?): SharedPreferences {
         val ctx = SalesforceSDKManager.getInstance().appContext
-        return ctx.getSharedPreferences(policyKey + account.userLevelFilenameSuffix, Context.MODE_PRIVATE)
+        return ctx.getSharedPreferences(policyKey + account?.userLevelFilenameSuffix, Context.MODE_PRIVATE)
     }
 
     fun getGlobalPrefs(): SharedPreferences {
@@ -73,7 +73,7 @@ internal abstract class AppLockManager(
         return ctx.getSharedPreferences(policyKey, Context.MODE_PRIVATE)
     }
 
-    fun getPolicy(account: UserAccount): Policy {
+    fun getPolicy(account: UserAccount?): Policy {
         val accountPolicy = getAccountPrefs(account)
         return accountPolicy.getBoolean(enabledKey, false) to accountPolicy.getInt(timeoutKey, 0)
     }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -655,7 +655,6 @@ public class ClientManager {
         /**
          * Build a LoginOptions from the given bundle
          * @param options - bundle
-         * @return
          */
         public static LoginOptions fromBundle(Bundle options) {
             if (options == null) {
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
@@ -658,6 +658,9 @@ public class ClientManager {
          * @return
          */
         public static LoginOptions fromBundle(Bundle options) {
+            if (options == null) {
+                return new LoginOptions(null, null, null, null, null, null); // To prevent crash if null Bundle is passed
+            }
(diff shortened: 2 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/AppLockManager.kt
```diff
@@ -63,17 +63,17 @@ internal abstract class AppLockManager(
...
+        return ctx.getSharedPreferences(policyKey + account?.userLevelFilenameSuffix, Context.MODE_PRIVATE)
     }
 
     fun getGlobalPrefs(): SharedPreferences {
         val ctx = SalesforceSDKManager.getInstance().appContext
         return ctx.getSharedPreferences(policyKey, Context.MODE_PRIVATE)
     }
 
-    fun getPolicy(account: UserAccount): Policy {
+    fun getPolicy(account: UserAccount?): Policy {
```
brandonpage, line 76: This is incorrect.  There are two places where this is API is used and both check the account is not null first.

brandonpage (changes requested):
I don't understand what scenario this change is for.  When would the `options` Bundle ever be null?  

If you are launching the LoginActivity yourself (which I do not recommend) you need to supply the correct values so we can use them [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/a0e26fc4a3b340b6e72825abb205945f851053f7/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt#L182).

umashankar-sf:
@brandonpage Please find the logs and screenshot from Firebase, It is crashing a lot for TrailHead GO app.

**Biometric crash Screenshot and Logs**
image

[biometric error.txt](https://github.com/user-attachments/files/18002826/biometric.error.txt)

**Bundle null Crash**
image
[bundle null.txt](https://github.com/user-attachments/files/18002829/bundle.null.txt)

Let's connect if you need more details

umashankar-sf:
@brandonpage  [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/a0e26fc4a3b340b6e72825abb205945f851053f7/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt#L182) the code is just passing `intent.extras` we are not checking null value.
`extras` is a nullable property. So this can be `null`. This is where crash is happening according to the Logs.
 In this PR I just added null check.

@zzou @jaewoo-lee 
```

val loginOptions = when {
            isUsingFrontDoorBridge -> salesforceSDKManager.loginOptions
            else -> fromBundleWithSafeLoginUrl(intent.extras)
        }

```

   ```
 /**
     * Retrieves a map of extended data from the intent.
     *
     * @return the map of all extras previously added with putExtra(),
     * or null if none have been added.
     */
    public @Nullable Bundle getExtras() {
        return (mExtras != null)
                ? new Bundle(mExtras)
                : null;
    }
```

brandonpage:
> @brandonpage Please find the logs and screenshot from Firebase, It is crashing a lot for TrailHead GO app.
> 
> **Biometric crash Screenshot and Logs** image
> 
> [biometric error.txt](https://github.com/user-attachments/files/18002826/biometric.error.txt)


As I mentioned on Slack, is this not fixed by the`?` [null check here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/7de5ffc1392590cc60fce2cc4b8b6019ba292a40/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt#L1430) introduced in Mobile SDK 12?

brandonpage:
> @brandonpage [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/a0e26fc4a3b340b6e72825abb205945f851053f7/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt#L182) the code is just passing `intent.extras` we are not checking null value. `extras` is a nullable property. So this can be `null`. This is where crash is happening according to the Logs. In this PR I just added null check.

Sure, `intent.extras` is a nullable property but it should never be null on our or any subclass of our LoginActivity.  Everywhere that the login activity is called from uses `intent.extras`  to store our `LoginOptions` object.  This object is passed into `OAuthWebviewHelper` and is necessary for proper authentication and later token managment.  

We need to find out how the your app is getting into this state and fix the root cause, not add a null check that will let the app continue in a broken state.

PR #2635, opened 2024-11-07
@W-16580498: [Android] Hook to propagate MSDK logs to other systems

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!*_ 🥁

  This adds a new concept of "log receivers" to the SDK's existing logging infrastructure, plus a "factory" which can be registered by the app to create and register these receivers to named components in the logging flow.  The factory object could be a bridge to the app's dependency injection framework, such as Hilt.

  I tried this out in the template apps and the overall pattern gave the app what I expected.  I marked this ready for review with the anticipation we might have to discuss the naming convention a bit.  I thought over names such as "adapter", "destination", "target" or "handler" before settling on "receiver". I though this was a good semantic since the object only has to "receive" the log entry - What happens after that is entirely up to the implementation and our interface names don't provide implementation bias.

  I added some logic to ensure the app can also choose their own handling for our notion of component names.  In my template app, I tried a few patterns such a using a single logger that didn't use the provided component name or concatenated it to the message by mapping instances of receiver objects to component name.  We can add some sample code for this into the MSDK docs.

  As always, thanks for reading!

### DIFF
--- a/libs/SalesforceAnalytics/build.gradle.kts
+++ b/libs/SalesforceAnalytics/build.gradle.kts
@@ -77,6 +77,10 @@ android {
         buildConfig = true
     }
 
+    kotlin {
+        jvmToolchain(17)
+    }
+
     val convertCodeCoverage: TaskProvider<JacocoReport> = tasks.register<JacocoReport>("convertedCodeCoverage") {
         group = "Coverage"
         description = "Convert coverage.ec from Firebase Test Lab to XML that is usable by CodeCov."
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-present, salesforce.com, inc.
+ * Copyright (c) 2024-present, salesforce.com, inc.
  * All rights reserved.
  * Redistribution and use of this software in source and binary forms, with or
  * without modification, are permitted provided that the following conditions
--- a/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
+++ b/libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2017-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of salesforce.com, inc. nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission of salesforce.com, inc.
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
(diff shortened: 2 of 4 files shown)


libs/SalesforceAnalytics/build.gradle.kts
```diff
@@ -77,6 +77,10 @@ android {
         buildConfig = true
     }
 
+    kotlin {
+        jvmToolchain(17)
```
JohnsonEricAtSalesforce, line 81: This is the first Kotlin in this module, so this matches our other Kotlin-enabled modules.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+interface SalesforceLogReceiver {
+
+    /**
+     * Receives a log entry.
+     * @param level The log entry's level
+     * @param tag The log entry's tag
+     * @param message The log entry's message
+     */
+    fun receive(
```
JohnsonEricAtSalesforce, line 43: Kotlin interfaces cannot provide JVM overloads, alas.  Is there any new way around that I've missed over the years?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogger.java
```diff
@@ -71,31 +77,65 @@ public enum Level {
         DEBUG(2),
         VERBOSE(1);
 
-        private Integer severity;
+        private final Integer severity;
```
JohnsonEricAtSalesforce, line 80: All the inspector warnings in this source were trivial and low risk to resolve, so I included that in the spirt of good housekeeping.  Would it be a good time for a Kotlin conversion on this?

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogger.java
```diff
@@ -52,6 +53,11 @@
  */
 public class SalesforceLogger {
 
+    /**
+     * The factory for Salesforce log receivers.  Defaults to null
+     */
+    public static SalesforceLogReceiverFactory logReceiverFactory;
```
wmathurin, line 59: Should it be a public static member or should we have a static method to provide a factory (and keep the member private)?

JohnsonEricAtSalesforce:
@wmathurin - Take a look at https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2635/commits/42de28d146acafeead83ccde45a12e174c3139d9 . Thanks!

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
brandonpage: AFAIK this is the best you can do today.  [The issue is still open](https://youtrack.jetbrains.com/issue/KT-36102).

wmathurin (approved):

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogReceiver.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+package com.salesforce.androidsdk.analytics.logger
+
+import com.salesforce.androidsdk.analytics.logger.SalesforceLogger.Level
+
+/**
+ * Any object that can be registered with Salesforce logger and receive issued
+ * log entries.
+ */
+interface SalesforceLogReceiver {
```
brandonpage, line 35: I might have gone with `SalesforceLogCollector` but Receiver is equally good IMO.

libs/SalesforceAnalytics/src/com/salesforce/androidsdk/analytics/logger/SalesforceLogger.java
```diff
@@ -148,8 +212,8 @@ private boolean isDebugMode() {
                     }
                 }
             }
-        } catch (PackageManager.NameNotFoundException e) {
-            debugMode = true;
+        } catch (NameNotFoundException e) {
+            /* Intentionally blank */
```
brandonpage, line 216: This seems like a really strange assumption we had here.  Without looking too deeply, good change 👍

brandonpage (approved):

brandonpage:
> There's a lot of ways the app could use this new logging feature, but for a quick example I used a couple of anonymous adapters and ensured I could work with both level and component name. Obviously, this is a pretty trivial example and wouldn't add value in reality, but it shows where the app can inject any and all the logic they wish. Here's a screenshot!
> 
> Screenshot 2024-11-08 at 11 39 13

This could also be implemented with a single code block (like if interfaces supported JVM overloads) by assigning one to the other:

```
override fun receive(level: Level, tag: String, message: String) = receive(level, tag, message, throwable = null)
override fun receive(level: Level, tag: String, message: String, throwable: Throwable?) {
   // single impl
}
```

PR #2639, opened 2024-11-19
@W-17267009: [MSDK Android] Add Log Receiver Example Code To A Sample App

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!*_ 🤘🏻

  This follow-up to #2635 adds sample code to the REST Explorer sample app.  Also, it includes enough Kotlin conversion so the new sample code isn't in Java 😍


native/NativeSampleApps/RestExplorer/src/com/salesforce/samples/restexplorer/RestExplorerApp.kt
```diff
@@ -0,0 +1,206 @@
+/*
+ * Copyright (c) 2011-present, salesforce.com, inc.
```
wmathurin, line 2: NIT should the range be 2024-present since it's a new file?

wmathurin (approved):

native/NativeSampleApps/RestExplorer/src/com/salesforce/samples/restexplorer/RestExplorerApp.kt
JohnsonEricAtSalesforce: That's a neat question since this is a rename due to the Kotlin conversion.  Logically, it's still the same RestExplorerApp class.  From a historical perspective, I do kind of like the hint that this has git history going back before the rename, which is something I've used quite a bit in researching Kotlin logic in its Java form to learn requirements.

native/NativeSampleApps/RestExplorer/src/com/salesforce/samples/restexplorer/RestExplorerApp.kt
```diff
@@ -0,0 +1,206 @@
...
+     *
+     * @param context The application context
+     * @param mainActivity The activity to be launched after the login flow
+     * @param loginActivity The login activity
+     */
+    private constructor(
+        context: Context,
+        mainActivity: Class<out Activity?>,
+        loginActivity: Class<out Activity?>
+    ) : SalesforceSDKManager(
```
brandonpage, line 148: Unrelated to this PR entirely, but is this really a good example of how we want apps to use the Mobile SDK in 2024?  @wmathurin @JohnsonEricAtSalesforce

JohnsonEricAtSalesforce: That's a great topic!  I would suspect the answer the that is "No".  There are quite a few aspects of the "SDK Manager" that are following older patterns and I reckon this is one.  We should bring this to planning and create a modernization story (or entire epic) for what we want to do with that class.

Specifically in this context and thinking as an app developer, it would be great to not be subclassing a library-provided class or using it in a singleton context.

Let's gather our notes for planning since this will be a topic bigger than a github thread could handle.

PR #2643, opened 2024-12-08
Protect against null intent extras.

brandonpage:
`intent.extras` should never be null, we set it everywhere we instantiate our LoginActivity or the class the app provides.  However, according to #2633 it is somehow possible but the root cause is not yet known.  This is a defensive solution to the crash that does not pass a null `LoginOptions` bundle to `OAuthWebviewHelper`.


JohnsonEricAtSalesforce (approved):
⭐️

brandonpage:
This potential issue is no longer possible with #2653.

PR #2644, opened 2024-12-19
@W-16589746: [Android] REST wrappers for select SFAP APIs

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!*_ 🥁

  Now we have a new REST client wrapper with convenience request and response types for four of the `sfap_api` endpoints.  These are all Kotlin-first and also use Kotlin Serialization instead of JSON.org.  Because of that, they do look a little different than some of our Java-first/JSON.org REST wrappers.  The goal is to provide a comparable level of convenience to our library consumers.

  https://developer.salesforce.com/docs/einstein/genai/guide/access-models-api-with-rest.html#step-3-use-models-rest-api

  The usage looks like this in the template apps.  I also tested calling all of this from Java to ensure the developer experience is similar.

```
    // region `sfap_api` Testing

    /** The generation id from the `sfap_api` `generations` endpoint */
    private var generationId: String? = null

    /**
     * Fetches generated text from the `sfap_api` "generations" endpoint.
     * @return Job The Kotlin Coroutines Job running the fetch
     */
    private fun generateText() = CoroutineScope(Default).launch {

        runCatching {
            val generationsResponseBody = SfapApiClient(
                apiHostName = "dev.api.salesforce.com",
                modelName = "sfdc_ai__DefaultGPT35Turbo",
                restClient = client ?: return@launch
            ).fetchGeneratedText(
                "Tell me a story about an action movie hero with really, really cool hair."
            )

            generationId = generationsResponseBody.generation?.id
            Log.i("SFAP_API-TESTS", generationsResponseBody.generation?.generatedText ?: "")
            Log.i("SFAP_API-TESTS", generationsResponseBody.sourceJson ?: "")

            // Test submitting feedback for the generated text.
            submitFeedback()
        }.onFailure { throwable ->
            Log.i("App", "Cannot fetch generated text due to an error with message '${throwable.message}'.")
        }
    }

    /**
     * Fetches generated chat responses from the `sfap_api` "generations"
     * endpoint.
     * @return Job The Kotlin Coroutines Job running the fetch
     */
    private fun generateChat() = CoroutineScope(Default).launch {

        runCatching {
            val chatGenerationsResponseBody = SfapApiClient(
                apiHostName = "dev.api.salesforce.com",
                modelName = "sfdc_ai__DefaultGPT35Turbo",
                restClient = client ?: return@launch
            ).fetchGeneratedChat(
                requestBody = MyChatGenerationsRequestBody(
                    messagesTransient = arrayOf(
                        Message(
                            role = "user",
                            content = "Can you help me clean my kitchen?"
                        )
                    ),
                    localizationTransient = Localization(
                        defaultLocale = "en_US",
                        inputLocales = arrayOf(
                            Locale(
                                locale = "en_US",
                                probability = 0.1
                            )
                        ),
                        expectedLocales = arrayOf("en_US")
                    ),
                    /* This is an example of how to provide app-specific values for the `tags` parameter */
                    tags = Tags("This property was specified by the app rather than the Salesforce Mobile SDK.")
                )
            )
            chatGenerationsResponseBody.generationDetails?.generations?.forEach { generation ->
                Log.i("SFAP_API-TESTS", generation.content ?: return@forEach)
            }
            Log.i("SFAP_API-TESTS", chatGenerationsResponseBody.sourceJson ?: "")
        }.onFailure { throwable ->
            Log.i("App", "Cannot fetch generated chat responses due to an error with message '${throwable.message}'.")
        }
    }

    /**
     * Submits feedback for previously generated text from the `sfap_api`
     * endpoints to the `sfap_api` `feedback` endpoint.
     * @return Job The Kotlin Coroutines Job running the submission
     */
    private fun submitFeedback() = CoroutineScope(Default).launch {

        runCatching {
            val feedbackResponseBody = SfapApiClient(
                apiHostName = "dev.api.salesforce.com",
                restClient = client ?: return@launch
            ).submitGeneratedTextFeedback(
                requestBody = SfapApiFeedbackRequestBody(
                    feedback = "GOOD",
                    generationId = generationId
                )
            )

            Log.i("SFAP_API-TESTS", feedbackResponseBody.message ?: "")
            Log.i("SFAP_API-TESTS", feedbackResponseBody.sourceJson ?: "")
        }.onFailure { throwable ->
            Log.i("App", "Cannot submit feedback due to an error with message '${throwable.message}'.")
        }
    }

    /**
     * Fetches generated embeddings from the `sfap_api` "embeddings" endpoint.
     * @return Job The Kotlin Coroutines Job running the fetch
     */
    private fun generateEmbeddings() = CoroutineScope(Default).launch {

        runCatching {
            val embeddingsResponseBody = SfapApiClient(
                apiHostName = "dev.api.salesforce.com",
                modelName = "sfdc_ai__DefaultOpenAITextEmbeddingAda_002",
                restClient = client ?: return@launch
            ).fetchGeneratedEmbeddings(
                requestBody = SfapApiEmbeddingsRequestBody(
                    input = arrayOf("Please clean up after yourself")
                )
            )
            embeddingsResponseBody.embeddings?.forEach { embedding ->
                Log.i("SFAP_API-TESTS", "Fetched generated embedding '${(embedding.embedding ?: return@forEach)}'.")
            }
            Log.i("SFAP_API-TESTS", embeddingsResponseBody.sourceJson ?: "")
        }.onFailure { throwable ->
            Log.i("App", "Cannot fetch generated embeddings due to an error with message '${throwable.message}'.")
        }
    }

    // endregion
}

@Serializable
data class MyChatGenerationsRequestBody(
    private val messagesTransient: Array<Message>,
    private val localizationTransient: Localization,
    private val tags: Tags
) : SfapApiChatGenerationsRequestBody(
    messagesTransient,
    localizationTransient
) {

    @OptIn(InternalSerializationApi::class)
    override fun toJson(): String {
        return Json.encodeToString(MyChatGenerationsRequestBody::class.serializer(), this)
    }

    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (javaClass != other?.javaClass) return false
        if (!super.equals(other)) return false

        other as MyChatGenerationsRequestBody

        return tags == other.tags
    }

    override fun hashCode(): Int {
        var result = super.hashCode()
        result = 31 * result + tags.hashCode()
        return result
    }

    @Serializable
    data class Tags(
        @SerialName("custom_property")
        val customProperty: String
    )
}
```


JohnsonEricAtSalesforce:
I set this to "draft" status so we can incorporate some early feedback before anyone commits their time to a really detailed review.  I'm expecting we'll have a few bigger feedback changes on this one.

wmathurin:
Are you saying having a connected app that issues JWT-based access token with sfap_api scope is not sufficient to access those APIs??

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -90,21 +90,23 @@
 public class OAuth2 {
 
     private static final String ACCESS_TOKEN = "access_token";
-    protected static final String CLIENT_ID = "client_id";
+    protected static final String CLIENT_ID = "client_id"; // OAuth 2.0 token endpoint request body parameter names
+    protected static final String CLIENT_SECRET = "client_secret"; // OAuth 2.0 token endpoint request body parameter names
```
wmathurin, line 94: We should not ship client secret in the app binary.

JohnsonEricAtSalesforce:
> Are you saying having a connected app that issues JWT-based access token with sfap_api scope is not sufficient to access those APIs??

According to the doc at https://developer.salesforce.com/docs/einstein/genai/guide/access-models-api-with-rest.html, a CA _should_ work.  I had some trouble though, so I followed the ECA doc since that's what the doc specifically documented.

We might have to try a CA again and see if it was just me mistranslating the ECA guide to the CA configuration.

wmathurin:
There should not be any change in the auth code as long as the CA/ECA is issuing JWT-based token with the appropriate scope. That was the point of the whole org JWT support work.

wmathurin:
I think in the doc you linked above, the "salesforce app" refers to a server-side app not a mobile app.

JohnsonEricAtSalesforce:
> There should not be any change in the auth code as long as the CA/ECA is issuing JWT-based token with the appropriate scope. That was the point of the whole org JWT support work.

@wmathurin - Thanks.  I looked through your last couple pull requests another time and suddenly a whole bunch of things make a lot more sense 💡A JWT "based" token works just fine once I tweaked a few settings in my ECA.

I'll have an update in soon.  Thanks for the direction.

P.S., thanks for adding Kotlin Serialization! ❤️

JohnsonEricAtSalesforce:
@wmathurin - https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/53e552669dc43d598b8fca4d2085308d3332a85f is for you.  It switches to your JWT-based access token.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
wmathurin: @JohnsonEricAtSalesforce can line 94 go?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
```diff
@@ -0,0 +1,85 @@
...
+     * @return The endpoint's response
+     */
+    @Throws(
+        IOException::class,
+        OAuth2InvalidTokenException::class,
+        SfapApiException::class
+    )
+    fun fetch(
+        prompt: String,
+        accessToken: String
```
wmathurin, line 38: Why don't we use a RestRequest and the RestClient for those calls? That way the application would no need to handle the access token. Also if the token expires, refresh would be handled automatically.

JohnsonEricAtSalesforce (commented):
I'd forgotten to submit my self-review notes last night and it seems Github was absorbing my comment replies until then.  So, better late than never.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
JohnsonEricAtSalesforce: @wmathurin - Take a look at https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/8218fea8fc27c6ae0159366b54c64bc2275fa2be for the updated REST client.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
```diff
@@ -0,0 +1,79 @@
...
+        SfapApiException::class
+    )
+    fun fetch(
+        prompt: String
+    ): SfapApiGenerationsResponseBody {
+
+        // Request the generated text.
+        val restRequest = RestRequest(
+            POST,
+            "https://$apiHostName/einstein/platform/v1/models/$modelName/generations",
```
JohnsonEricAtSalesforce, line 44: @wmathurin - How's this look for building the request? Do we have any logic anywhere that already encapsulates the `api.salesforce.com` host along with its "dev" and "prod" concepts? Is there any thought that this code should accept an arbitrary sfap_api host?

wmathurin: Maybe it's one of the URLs returned in the [identity service response](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java#L573)

But maybe not - not seeing in the [doc](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_using_openid.htm&type=5)

Ideally we should get it from the server during login. Short of that, we could ask the app to pass it. We shouldn't hardcode it in the SDK.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
```diff
@@ -0,0 +1,79 @@
...
+ * Provides generated text from the `sfap_api` "generations" endpoint.
+ *
+ * See https://developer.salesforce.com/docs/einstein/genai/guide/access-models-api-with-rest.html
+ *
+ * @param apiHostName The Salesforce `sfap_api` hostname
+ * @param modelName The model name to request generations from.  For possible
+ * values, see https://developer.salesforce.com/docs/einstein/genai/guide/api-names.html
+ * @param restClient The REST client to use
+ */
+class SfapApiGenerations(
```
wmathurin, line 22: I wonder if we should have request/response helper objects, but let the app do the sending and decide if it wants the raw response or the parsed response.

We did something like that for [composite requests](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/CompositeRequest.java).

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
```diff
@@ -0,0 +1,79 @@
+package com.salesforce.androidsdk.auth
```
wmathurin, line 1: I should probably go to the com.salesforce.androidsdk.rest folder. We have a few request/response classes there already.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
JohnsonEricAtSalesforce: How about https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/071abae988069daf93d547c7719171aaa186a7c4 so we don't have that value and the app provides it?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
JohnsonEricAtSalesforce: We're not using JSON.org here and Kotlin Serialization took its place.  I think that's a win, so perhaps something like https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/90c9f2eb48086bf1cd60dfd2ba11381785f052a6 gives the app a convenience to generate JSON from the request and also to retrieve the endpoint's source JSON from the response.  How's that look?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/SfapApiGenerations.kt
JohnsonEricAtSalesforce: Nice.  Check that out in 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/2960b1a6f2093267ebaff6be974664fed83d0bde

JohnsonEricAtSalesforce:
@wmathurin - Here's the second of our four endpoints.  https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/262c9ec79389f44b2dd33a61e7dfff8ab56cc5bd

This also does a lot to ensure the pair of response and request model object are consistent.  Note: There _are_ subtle difference between them, so I didn't feel comfortable reusing the models between endpoints.  There might be some opportunity for that after I can analyze the remaining two endpoints.

wmathurin (commented):
Maybe we could offer ways for the app:
- to build the `RestRequest` from a `SfapApiChatGenerationsRequestBody`
- build a `SfapApiChatGenerationsResponseBody` from the `RestResponse`
And let the app use its `RestClient` to send it to the server.
Some apps might want to save unparsed responses, so access to the the raw response could be useful.

You could still keep the `SfapApiChatGenerations` class but it deviates a bit from the other requests and responses types we offer helper classes for.

JohnsonEricAtSalesforce:
I just found how the doc details enumeration values for some of the requests and responses, so I'll come back and make enums for those.

JohnsonEricAtSalesforce:
> Maybe we could offer ways for the app:
> 
> * to build the `RestRequest` from a `SfapApiChatGenerationsRequestBody`
> * build a `SfapApiChatGenerationsResponseBody` from the `RestResponse`
>   And let the app use its `RestClient` to send it to the server.
>   Some apps might want to save unparsed responses, so access to the the raw response could be useful.
> 
> You could still keep the `SfapApiChatGenerations` class but it deviates a bit from the other requests and responses types we offer helper classes for.

Thanks.  I'll do another compare and contrast with the other types.  We'll have a couple differences just in the migration from JSON.org to Kotlin Serialization for these.  We might want a Trust story to migrate those off JSON.org in a future release.

JohnsonEricAtSalesforce:
@wmathurin - Here's the third of four: `feedback`: 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/0dd03052551f85c41ba5f9a28e9ab9092c77a553

wmathurin (commented):
It's a bit odd to have a method called fetch on SfapApiFeedback to send a feedback etc.

Do you think we should have a single SfapApi class that offers the following methods:
- getChatGenerations
- sendFeedback
- getGenerations
- and future ones?
There could be some re-used code for setting headers on RestRequest.

JohnsonEricAtSalesforce:
> It's a bit odd to have a method called fetch on SfapApiFeedback to send a feedback etc.
> 
> Do you think we should have a single SfapApi class that offers the following methods:
> 
> * getChatGenerations
> * sendFeedback
> * getGenerations
> * and future ones?
>   There could be some re-used code for setting headers on RestRequest.

I was also just starting to think that would be a good idea also.  Stay tuned!

JohnsonEricAtSalesforce:
> > It's a bit odd to have a method called fetch on SfapApiFeedback to send a feedback etc.
> > Do you think we should have a single SfapApi class that offers the following methods:
> > 
> > * getChatGenerations
> > * sendFeedback
> > * getGenerations
> > * and future ones?
> >   There could be some re-used code for setting headers on RestRequest.
> 
> I was also just starting to think that would be a good idea also. Stay tuned!

@wmathurin - I think you'll like https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/cf677c80ca4ac41aa0cd79d1b853beb9a00f7190 for the initial consolidation.

wmathurin (commented):
LGTM - maybe just update the PR's description (it's describing stuff that is no longer in there).

JohnsonEricAtSalesforce:
> LGTM - maybe just update the PR's description (it's describing stuff that is no longer in there).

Thanks!  I still have a couple more tasks before I even move this out of "draft" status.  The latest commit really adds detail to the error responses.  I'm also going to turn on `ignoreUnknownKeys` in all the JSON objects plus see about a few enums if they make sense for some of the property values.

JohnsonEricAtSalesforce:
@wmathurin - Thanks all the draft review.  That really helped get this on track ⭐️ I believe we're ready for a final walk through, which I'm also doing once more.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsRequestBody.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Serializable
+    data class Locale(
+        val locale: String,
+        val probability: Double
+    )
+
+    @Serializable
+    class Tags
```
JohnsonEricAtSalesforce, line 114: @wmathurin - The documentation doesn't have any content for this `Tags` object.  Is that a concern? Or, since I left unknown keys on for requests is it up to the app to subclass this type and add some arbitrary content to it?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsResponseBody.kt
```diff
@@ -0,0 +1,184 @@
...
+ * Models a `sfap_api` "chat-generations" endpoint response.
+ */
+@Serializable
+data class SfapApiChatGenerationsResponseBody(
+    val id: String? = null,
+    val generationDetails: GenerationDetails? = null
+) {
+
+    /** The original JSON used to initialize this response body */
+    var sourceJson: String? = null
```
JohnsonEricAtSalesforce, line 43: @wmathurin - Did I answer your earlier question by calling attention to the raw text response exposed in each new response model like this?  It's also exposed on the error response model.  We don't actually have a "middle" object like we did with JSON.org now that we're using Kotlin serialization to skip that step.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsRequestBody.kt
wmathurin: The doc is indeed very vague

> Entries used by the Models API for non-generative purposes and entries used by the client for free-form data." 

See https://developer.salesforce.com/docs/einstein/genai/references/models-api?meta=type%3ATags

Assuming it's a map of String to String, what would the code look like if you try to provide them?

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsRequestBody.kt
JohnsonEricAtSalesforce: I tried a couple options this morning, but I couldn't get Kotlin Serialization to give me the developer experience I was thinking of.  🤔. I'll try a couple other options after the holiday and see what I can whip up.

JohnsonEricAtSalesforce:
@wmathurin - Though this was already approved, I _did_ really want to answer our question about the "abstract" `tags` parameter.  Since the doc doesn't specify its structure, we need a way for the app to do so.  I did a bunch of tests and reading.  It looks like the best option is for our MSDK request body type to omit that entirely and open the type to subclassing.  The subclass can take the reins then.  I added some documentation and updated the usage example in this pull request's description.

Take a quick look at https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2644/commits/ae1c8d210e1a56a68a32fc6b5e3d103707e91e55 and confirm that's a winner.

wmathurin (approved):

PR #2647, opened 2024-12-27
@W-17291582: [Android] [Instrumentation] Add QR Code feature logging to track adoption

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

This one follows along with recent changes to add a feature code for Native Login in #2637.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -718,10 +718,7 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
     fun loginWithFrontdoorBridgeUrl(
         frontdoorBridgeUrl: String,
         pkceCodeVerifier: String?
-    ) {
-        SalesforceSDKManager.getInstance().registerUsedAppFeature(FEATURE_QR_CODE_LOGIN)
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -696,39 +697,50 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
+     * required for User Agent
      * Authorization Flow but is required for Web Server Authorization Flow
      */
     @Suppress("MemberVisibilityCanBePrivate")
     fun loginWithFrontdoorBridgeUrl(
         frontdoorBridgeUrl: String,
         pkceCodeVerifier: String?
-    ) = webviewHelper?.loginWithFrontdoorBridgeUrl(frontdoorBridgeUrl, pkceCodeVerifier)
+    ) {
+        SalesforceSDKManager.getInstance().registerUsedAppFeature(FEATURE_QR_CODE_LOGIN)
```
brandonpage, line 722: I think we should move this up one call to `loginWithFrontdoorBridgeUrlFromQrCode` so we don't miss-report QR usage if  `loginWithFrontdoorBridgeUrl` is used for another flow.

JohnsonEricAtSalesforce: Good idea.  Check out today's commit 🎉

brandonpage (approved):

wmathurin (commented):
👍

PR #2651, opened 2025-01-16
@W-17615983: [Android] REST wrappers for select SFAP APIs - iOS Consistency Updates

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  While I was working on the iOS counterpart, I noticed a handful of items for Android that would be great to have.  So, here's a quick pull request to bring the two in sync.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsResponseBody.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsResponseBody.kt
@@ -157,7 +157,7 @@ data class SfapApiChatGenerationsResponseBody(
             @Serializable
             data class PromptTokensDetails(
                 @SerialName("cached_tokens") val cachedTokens: Int = 0,
-                @SerialName("audio_tokens") val audiTokens: Int = 0
+                @SerialName("audio_tokens") val audioTokens: Int = 0
             )
         }
     }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiClient.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiClient.kt
@@ -60,15 +60,19 @@ class SfapApiClient(
      * @param requestBody The `embeddings` request body
      * @return The endpoint's response
      */
+    @Suppress("unused")
     @Throws(SfapApiException::class)
     fun fetchGeneratedEmbeddings(
         requestBody: SfapApiEmbeddingsRequestBody
     ): SfapApiEmbeddingsResponseBody {
 
+        // Guards.
+        val modelNameUnwrapped = modelName ?: throw SfapApiException("Cannot fetch generated embeddings without specifying a model name.")
+
         // Submit the request.
         val restRequest = RestRequest(
             POST,
-            "https://$apiHostName/einstein/platform/v1/models/$modelName/embeddings",
+            "https://$apiHostName/einstein/platform/v1/models/$modelNameUnwrapped/embeddings",
             requestBody
                 .toJson()
                 .toRequestBody(CONTENT_TYPE_HEADER_VALUE_APPLICATION_JSON.toMediaTypeOrNull()),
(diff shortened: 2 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiChatGenerationsResponseBody.kt
```diff
@@ -157,7 +157,7 @@ data class SfapApiChatGenerationsResponseBody(
             @Serializable
             data class PromptTokensDetails(
                 @SerialName("cached_tokens") val cachedTokens: Int = 0,
-                @SerialName("audio_tokens") val audiTokens: Int = 0
```
JohnsonEricAtSalesforce, line 160: This was a typo, so we'll need this fixed before 13.0.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiClient.kt
```diff
@@ -60,15 +60,19 @@ class SfapApiClient(
      * @param requestBody The `embeddings` request body
      * @return The endpoint's response
      */
+    @Suppress("unused")
```
JohnsonEricAtSalesforce, line 63: A few unused suppressions clean up our warnings.  These don't appear when one's testing with an app that uses this code in a composite build workspace.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiClient.kt
```diff
@@ -60,15 +60,19 @@ class SfapApiClient(
...
     @Throws(SfapApiException::class)
     fun fetchGeneratedEmbeddings(
         requestBody: SfapApiEmbeddingsRequestBody
     ): SfapApiEmbeddingsResponseBody {
 
+        // Guards.
+        val modelNameUnwrapped = modelName ?: throw SfapApiException("Cannot fetch generated embeddings without specifying a model name.")
```
JohnsonEricAtSalesforce, line 70: This guard from the original pull request needed to be applied in additional locations.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiFeedbackRequestBody.kt
```diff
@@ -32,11 +32,16 @@ import kotlinx.serialization.json.Json
...
+ * `appFeedback`, subclass and introduce a new parameter of any object type
+ * named `appFeedback`.  Also, override the `toJson` method to use the subclass
+ * serializer instead of the the serializer provided by this class.
  */
 @Serializable
-data class SfapApiFeedbackRequestBody(
+@Suppress("unused")
+open class SfapApiFeedbackRequestBody(
     val id: String? = null,
-    val appFeedback: AppFeedback? = null,
```
JohnsonEricAtSalesforce, line 39: This was a potential bug.  Like the `Tags` object on another `sfap_api` request body, `appFeedback` doesn't have a documented structure.  It's an open object for the app to supply.  So, we remove it here and open the class with some documentation on how to provide that just as we did for `Tags`.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiClient.kt
brandonpage: If 3 out of the 4 functions in this class require `modelName` why is it nullable at all?  Feedback doesn't need it, but you aren't going to submit feedback without having run the model, right?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiFeedbackRequestBody.kt
```diff
@@ -32,11 +32,16 @@ import kotlinx.serialization.json.Json
 /**
  * Models a `sfap_api` `feedback` endpoint request.
  * See https://developer.salesforce.com/docs/einstein/genai/references/models-api?meta=generateText
+ *
```
brandonpage, line 35: The doc link above this line is for `generateText`, not `submitFeedback`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiClient.kt
JohnsonEricAtSalesforce: Good topic.  This method is also for a network endpoint where we've usually pondered how much to mock out the network integration and remote systems.  Maybe there could be some value in that just to provide some coverage of the request preparation and response handling based on known data at the time.  I was wondering about that also since so much of the error response handling is common between the four endpoints.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/SfapApiFeedbackRequestBody.kt
JohnsonEricAtSalesforce: Nice catch ✨

PR #2653, opened 2025-01-18
Modern Login Architecture

brandonpage:
This is the first of a few work items needed to fully complete this effort.  **Most notably: final UI/UX is not in place.**  The included UI is functional but only refined enough to prove the architecture works in all required aspects.     

# Design
Beyond migrating to Jetpack Compose the primary design goal was to modernize and simplify how the SDK (and app) interacts with the UI.  To do this I completely restructured the code to a new MVVM architecture.  This also meant completely removing webview login's use of `OAuthWebViewHelper` and `LoginOptions`.  

However, our existing login business logic works very well.  It has been tested in every imaginable way and harden over many years of development.  With that in mind I attempted to treat this as a **strict refactor** with regards to that code.  So while it may look like `OAuthWebViewHelper` was replaced with a bunch of new code; every necessary piece of it has actually been very carefully distributed to other classes without modification.  These classes were chosen or created to separate the (far too many) concerns the helper had:
- `LoginActivity` is now responsible for creating the webview and related clients, the Custom Tab, and cert based auth.
- `LoginViewModel` handles suppling the UI with all data, and delegates the successful or unsuccessful completion of auth.
- `AuthenticationUtilities` handles the actual successful completion of auth in a generic way that can be used outside of the webview context.

[](https://mermaid.live/edit#pako:eNqNVk1z2zYQ_SsY5lCm-jj06EM6smRPNWPHmcpOZjo6ZAWsSdQgoAIgZTdj__YsSEoCKSqJDuLHPuwu3r7F8lvCjcDkIuEKnFtIyCwUa83op41H9mgsuzGZ1Cu0Fdpb0JChXScOFXKPYoX0lu3Asa3FSprSqRcG9CAroOXR0mnj9T4np6XmXhrtmNTM59KxOjrLoUK2QdQMBGzJO_OGlQ4Jg6xZfoxb-6xAlZi-n66Txtz4OU2YfWvs4Te6LT1sFN7IChfg4S2Cv_UCNKte17pHyayklLSXHMI-HrySXqJj68ToYLpWZjc3xVYh4Wl7If_Pf9Sr2DLa5g7Ih85aauag6XXYsGCbF_aRfFcthQy0YMvFp0AIbx0T0c06i4Uh4F1w_wU3nyXu_kK1DWWKWTmTc8TM22n2qTdPqP9Gt6Vy4ZhNp9P3xwWTR_Q8X7oVKHTEC8el9kgSqiM4tGkMBiFmnJtS-xSaa2xVJpudA5wWoGYl7PSW5KtYLMhQN6ak80gCC1ieg85op8TdqTJ-c31NNZFUQD5YNeiJCtKLF-oDpTcFBIIV1abcCuoAx3aSip6hD8waK_9vuLFqSLPHDXX0ikKCN_Yg2JW3pJphrf54xX5XEbYxURaSxLEUkeXSGIWkSUd2hU2Qeelok_ewmVEHV9K_RPi5Ucb-STRvL8HWDydhyCq9wnt89kcbaYoS844VxiJ7WDJeB2mpCuZIJgNMpq5OLRLTyKIyIOpm2GEswhGnLVFy5om03zHU1HyhYl1bo70wxl5aKTIMAbqqf7cppdor9SMUnQDUQxS2IavTSF0fb8LMqdBXzySp7MQ40Ic9hEWHvs500cm03zCRuPYVi7X1rie7an8XQaJjZV6LhO3ipwhJqHluTYFHXPwckd3623v6xTLENeBoQ2J3tt5Bx3Rk78paY1MM_2NWXxboON0Ow1cl5-hcSiewnQ2dUCDqaJ-oZEt9aIV031XjVrn07gZoxuUHUfbq8TWU1bgwhY6N_zUuzM_r0q0pdNqxH05qTWfUcQx0ytmJO3K5KZW4I_US7YH1G9o1dW4aUhjTrPmvRBfTMjI1H9dSS5e3qNKqmDia0jIchjMtrp69pVwvgT9llggW9TmR_gsVOG7l1tOoKdXJud-jYzL5MPRVcg46PPki-IHG6fRDz8EQ6vfJ5EwJz6J_XIFknBRoC5CCvsTqaqwTSrnAdXJBtwLsU5gWr4QLQ2b1only4W2J44RIzPLk4hEUjeekGTrtZ9wesgX9jzGHR5oNNBpu28--cHn9DszGX0c)

## Additional Benefits 

The use or subclass of the new `LoginViewModel` allows an app to customize the UI in just about every way _without_ overriding LoginActivity.  Previously, any functions an app may have been overriding in `OAuthWebViewHelper` wouldn't take affect unless unless that extension class was supplied to `LoginActivity`... in another subclass.  These functions (such as webview or WebViewClient) can now just be overridden in `LoginActivity` itself.

The design adheres to MVVM best practices by using the `LoginViewModel` as an intermediary for all data required by the UI.  The view model is always retrieved from the OS via factory and never passed.  `LoginViewModel` does not reference Login or any other activity (as it outlives their lifecycles).  The activity or compose UI may call a function or modify data in the view model, but all `LiveData` updates flow down to observers.  

Because Compostables are _functions_ new UI tests can be written without the need for an activity context.  I have removed `ServerPickerActivityTest` from `RestExplorer` (where it needed to live because of the activity context) but the same cases can be covered in unit tests within the library once the UI is finished.  The same goes for the previously untested `LoginActivity` UI.

`LoginViewModel` exposes a `clientId` that can be set in a subclass, allowing an app to support multiple ECAs simultaneously.  

`LoginViewModel` has a new `singleServerCustomTabActivity` option.   This will let apps force the use of a custom tab with a single server url without the awkward blank activity in between.  

# What's In Progress
Only testing!  While all of the basic things have been thoroughly tested, certain things like QR Code Login and IDP _should_ just work but I haven't gotten to them yet.  My test plan doc will be kept up to date and this section will be updated as well when everything is completed.


libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -32,6 +32,7 @@
             android:theme="@style/SalesforceSDK"
             android:launchMode="singleTask"
             android:windowSoftInputMode="adjustResize"
+            android:configChanges="orientation|screenLayout|uiMode|screenSize|smallestScreenSize"
```
brandonpage, line 35: This allows for rotation, foldable changes, multitasking, etc without the webview reloading.

libs/SalesforceSDK/build.gradle.kts
```diff
@@ -13,30 +13,50 @@ plugins {
...
+    androidTestImplementation("androidx.arch.core:core-testing:2.2.0")
+    androidTestImplementation("androidx.compose.ui:ui-test-junit4:$composeVersion")
 }
 
 android {
     namespace = "com.salesforce.androidsdk"
     testNamespace = "com.salesforce.androidsdk.tests"
 
-    compileSdk = 34
+    compileSdk = 35
```
brandonpage, line 59: This was needed for tests but it caused an issue.  With `35` the symlink to `test_credentials.json` no longer works.  I fixed this by adding `assets.directories.add("../../shared/test")` below.  If anyone has a better solution let me know.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -209,13 +210,19 @@ open class SalesforceSDKManager protected constructor(
      */
     private var showDeveloperSupportBroadcastIntentReceiver: BroadcastReceiver? = null
 
-    val webviewLoginActivityClass: Class<out Activity> = loginActivity ?: LoginActivity::class.java
+    val webViewLoginActivityClass: Class<out Activity> = loginActivity ?: LoginActivity::class.java
```
brandonpage, line 213: I noticed we (inconsistently) use `webview` a lot but the official APIs use `webView` so this seemed like a good time to fix it everywhere.

We also use `frontdoor` instead of `frontDoor` but I'll leave that alone for now 😉.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -597,19 +604,17 @@ open class SalesforceSDKManager protected constructor(
...
+     * Optionally enables browser based login instead of web view login.
      *
      * @param browserLoginEnabled True if Chrome should be used for login; false
      * otherwise
      * @param shareBrowserSessionEnabled True enables browser session sharing;
      * false otherwise
      */
     @Synchronized
-    fun setBrowserLoginEnabled(
+    internal fun setBrowserLoginEnabled(
```
brandonpage, line 615: This is an intentionally breaking change.  I don't think we should have allowed this to begin with.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -0,0 +1,313 @@
...
+ * Called when any (User Agent flow, Web Server after PKCE, Native Login, IDP, ect) authentication
+ * method has received a [TokenEndpointResponse] from the server.
+ *
+ * This function:
+ *  * Blocks integration and managed app users.
+ *  * Retrieves the user's identity.
+ *  * Creates an Account.
+ *  * Checks for any CA/ECA settings such as Screen Lock or Biometric Authentication.
+ */
+internal fun onAuthFlowComplete(
```
brandonpage, line 75: This is basically that clean "Auth v2" interface I have been wanting.  

I left NativeLogin and IDP alone for now but we should migrate them from `OAuthWebviewHelper` to this utility so the former can be completely removed.  All necessary changes would be on internal code so no rush to do it in 13.0.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -52,6 +52,9 @@
  * @author bhariharan
  */
 public class LoginServerManager {
+	// LiveData representation of the users current selected server.
+	public MutableLiveData<LoginServer> selectedServer
+			= new MutableLiveData<>(new LoginServer("Production", PRODUCTION_LOGIN_URL, false));
```
brandonpage, line 57: If it ain't broke... wrap it in a new modern interface?  In all seriousness, I didn't feel like any changes to this class were actually necessary other than exposing a way to observe changes to the currently selected server.   

The only compelling reason I can think of to convert this class to Koltin right now is that `selectedServer` could be made internal.   Right now it could actually be set by the app, which isn't ideal.  

I might be able to restore the removed `private LoginServer selectedServer;` and use it as a backing value for `derivedStateOf` but I am not sure if that will work outside of a composable context.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -168,142 +210,114 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-        if (!isUsingFrontDoorBridge) {
-            salesforceSDKManager.fetchAuthenticationConfiguration()
-        }
-
-        // Setup content view
-        setContentView(sf__login_layout)
-        if (salesforceSDKManager.isIDPLoginFlowEnabled) {
-            findViewById<Button>(sf__idp_login_button).visibility = VISIBLE
+        if (!SalesforceSDKManager.getInstance().isDebugBuild) {
+            window.setFlags(FLAG_SECURE, FLAG_SECURE)
```
brandonpage, line 215: I see most of our internal apps removing the flag for debug builds and we have gotten this request many time externally.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -168,142 +210,114 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
-        if (!SalesforceSDKManager.getInstance().isBrowserLoginEnabled) {
-            // Reload the login page to ensure the correct login server is displayed in the webview.
-            webviewHelper?.run {
-                loadLoginPage()
-            }
-        }
+    private fun clearWebviewAndShowServerPicker() {
+        viewModel.loginUrl.value = ABOUT_BLANK
+        viewModel.showServerPicker.value = true
     }
```
brandonpage, line 318: This seemed like a better experience when backing out of a custom tab than seeing the same url loaded in the webview or a blank white activity.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -320,44 +334,120 @@ open class LoginActivity : AppCompatActivity(), OAuthWebviewHelperEvents {
...
+     * structure so long as this entry point is used to log in with the front
+     * door URL and optional PKCE code verifier.
+     *
+     * @param frontdoorBridgeUrl The UI Bridge API front door bridge URL
+     * @param pkceCodeVerifier The optional PKCE code verifier, which is not
+     * required for User Agent
+     * Authorization Flow but is required for Web Server Authorization Flow
+     */
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun loginWithFrontdoorBridgeUrl(
```
brandonpage, line 378: Thoughts on moving this and `loginWithFrontdoorBridgeUrlFromQrCode` to `LoginViewModel`?  I haven't had a chance to play around with it yet and see what makes sense.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -0,0 +1,287 @@
...
+    internal var showServerPicker = mutableStateOf(false)
+    internal val defaultTitleText: String
+        get() = if (loginUrl.value == "about:blank") "" else selectedServer.value ?: ""
+
+    // Public Overrideable LiveData
+    open var showTopBar = true
+    open var topBarColor: Color? = null
+    open var titleText: String? = null
+    open var titleComposable: (@Composable () -> Unit)? = null
+    open var loading = mutableStateOf(false)
```
brandonpage, line 78: This section will be expanded as we finish the UI.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -0,0 +1,287 @@
...
+    var additionalParameters = hashMapOf<String, String>()
+
+
+    val shouldShowBackButton = with(SalesforceSDKManager.getInstance()) {
+        !(userAccountManager.authenticatedUsers.isNullOrEmpty() || biometricAuthenticationManager?.locked ?: false)
+    }
+    val shouldShowBiometricPromptButton = with(SalesforceSDKManager.getInstance().biometricAuthenticationManager) {
+        this?.let { locked && hasBiometricOptedIn() } ?: false
+    }
+    // todo: add shouldShowIDPButton
```
brandonpage, line 105: I left several TODOs in on purpose, all related to UI code.  I am not sure which of the UI stories I will be doing vs others so it seemed better to just leave them in place so the intent was not lost.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -0,0 +1,287 @@
...
+                reloadWebview()
+            } else {
+                selectedServer.value = trimmedServer
+            }
+        }
+
+        // Update loginUrl when selectedServer updates so webview automatically reloads
+        loginUrl.addSource(selectedServer) { newServer ->
+            loginUrl.value = getAuthorizationUrl(newServer)
+        }
```
brandonpage, line 133: Goodbye `loadLoginPage` that caused us many headaches.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -209,13 +210,19 @@ open class SalesforceSDKManager protected constructor(
...
      * the account.
      */
-    val loginActivityClass: Class<out Activity> = nativeLoginActivity ?: webviewLoginActivityClass
+    val loginActivityClass: Class<out Activity> = nativeLoginActivity ?: webViewLoginActivityClass
+
+    /**
+     * ViewModel Factory the SDK will use in LoginActivity and composable functions.  Setting this will allow for
+     * visual customization without overriding LoginActivity.
+     */
+    var loginViewModelFactory = LoginViewModel.Factory
```
brandonpage, line 225: A note on this: This can _optionally_ be set but I use this everywhere because the system is subtly different than what you might expect.  The OS ensures there is only ever one instance of a view model **but inheritance counts**! 

For example:  If I create a subclass of `LoginViewModel` and set it to the viewModel val in my `LoginActivity` subclass our base `LoginViewModel` will never be instantiated.  Explicitly calling `LoginViewModel.Factory` will return the instance of the subclass of our view model.  So Apps overriding `LoginActivity` don't need to set this value. 

What the OS does is pretty elegant but not at all obvious.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+
+        // Load the webview as a composable
+        AndroidView(
+            modifier = Modifier
+                .padding(innerPadding)
+                .alpha(if (viewModel.loading.value) 0.0f else 100.0f),
+            factory = { activity.webView },
+            update = { it.loadUrl(loginUrl) },
```
brandonpage, line 177: Still a little shocked at how simple this is.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: Moving it makes sense to me.  Data and integration logic should in the view model rather than the activity, when possible.

JohnsonEricAtSalesforce (approved):
Exciting stuff!  Thanks!

wmathurin (approved):
LGTM!

PR #2654, opened 2025-01-21
New methods to get cipher provider version and FIPS status

wmathurin:
New tests

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
@@ -126,7 +126,7 @@ public class SmartStore  {
 	 * Set license key for SQLCipher
 	 * Needed when using commercial or enterprise editions of SQLCipher
 	 * Should be called before using SmartStore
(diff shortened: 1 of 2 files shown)


JohnsonEricAtSalesforce (approved):
🤘🏻

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
```diff
@@ -1647,6 +1647,24 @@ public String getSQLCipherVersion() {
...
+	public String getCipherProviderVersion() {
+		return TextUtils.join(" ", queryPragma("cipher_provider_version"));
+	}
+
+	/**
+	 * Get SQLCipher FIPS status
+	 *
+	 * @return "0" when using the community edition or the commercial edition and "1" when using the FIPS enabled enterprise edition
+	 */
+	public String getCipherFIPSStatus() {
```
brandonpage, line 1664: NIT:  Feels like this might be a bit more user friendly as a boolean instead of `"0"` or `"1"` strings.

brandonpage (approved):

PR #2656, opened 2025-01-24
Using OAEP cipher mode to encrypt unique id

wmathurin:
NB: Falling back on PKCS1 at decryption time to successfully decode unique id encoded with MSDK 12 or older

At this point PKCS1 is only used in fall back scenarios - to decrypt payloads encrypted for older version of MSDK:
- to decrypt encrypted push notifications, if server generated key was created with PKCS1 cipher mode
- to decrypt unique id, if it was encrypted with PKCS1 cipher mode


brandonpage (approved):

wmathurin:
NB: manually tested upgrade - login to app that uses old generateUniqueIdIfNoneStored and then relaunching app with the new generateUniqueIdIfNoneStored and making sure uniqueId was read and written back using new cipher mode.

PR #2659, opened 2025-01-28
@W-17410859: Rewrite ManageSpaceActivity in Jetpack Compose

JohnsonEricAtSalesforce:
🤘🏻 _*Ready For Review*_ 🎸

  This relatively simple update migrates MSDK's default "manage space activity" to Jetpack Compose.  The new layout is functionally identical to the previous one.  It also has a couple open functions to allow customization of the top bar and dialog.  Strings can be overridden as they were before.  Also, an app can replace this class entirely if they wish.

  There is a deferred work item to apply consistent theme logic to all the new Jetpack Compose activities.

  Enjoy!

Screenshot 2025-01-27 at 21 37 05


libs/SalesforceSDK/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ dependencies {
     implementation("androidx.biometric:biometric:1.2.0-alpha05")
     implementation("androidx.lifecycle:lifecycle-extensions:2.2.0")
     implementation("androidx.core:core-ktx:1.15.0")
+    implementation("androidx.activity:activity-compose:1.10.0")
```
JohnsonEricAtSalesforce, line 31: This new import gives activities the `setContent` composable and removes the need for `ComposeView`, which was intended more for legacy support in mixed layout environments.

libs/SalesforceSDK/res/values/sf__strings.xml
```diff
@@ -41,6 +41,7 @@
 	<string name="sf__server_picker_title">Change Server</string>
 
 	<!-- Manage space activity -->
+	<string name="sf__manage_space_title">Clear User Data?</string>
```
JohnsonEricAtSalesforce, line 44: The new Material dialog has a title and hero icon, which really does look nice.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -229,15 +228,12 @@ open class LoginActivity: FragmentActivity() {
...
 
         // Set content
-        setContentView(
-            ComposeView(this).apply {
-                setContent {
-                    LoginWebviewTheme {
-                        LoginView()
-                    }
-                }
+        setContent {
```
JohnsonEricAtSalesforce, line 231: Here's the simplified `setContent` composable in use.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ManageSpaceActivity.kt
```diff
@@ -0,0 +1,190 @@
...
+                    account = null,
+                    frontActivity = this@ManageSpaceActivity,
+                    showLoginPage = false,
+                    reason = USER_LOGOUT
+                )
+            },
+            titleText = if (isPreview) "Clear User Data?" else getString(sf__manage_space_title),
+            textText = if (isPreview) "Clearing user data will log the current user out. Are you sure you want to perform this action?" else getString(sf__manage_space_confirmation),
+            confirmButtonText = if (isPreview) "Yes" else getString(sf__manage_space_logout_yes),
+            dismissButtonText = if (isPreview) "No" else getString(sf__manage_space_logout_no),
```
brandonpage, line 118: What's up with the `isPreview`?  I'm not sure what `getString` you are using (I only see functions that take context?) but `stringResource` seems to work fine for me with Previews.

JohnsonEricAtSalesforce: 🤩 Now that's a nice one.  `stringResource` totally makes my day!  I'm so used to working around Compose's tight coupling with the runtime and mocking that out in @Preview I wasn't even expecting the convenience of `stringResource` 🤦🏻

JohnsonEricAtSalesforce: 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2659/commits/a8573c506b967763984c9df71840586dac343ae6

brandonpage (approved):
Looks great!

wmathurin (approved):

PR #2660, opened 2025-01-29
Move from CircleCI to Github Actions.

brandonpage:
After [extensive testing](https://github.com/brandonpage/SalesforceMobileSDK-Android/pull/29) I recreated the resulting configuration in a clean branch for this PR.  The new CI won't run on this PR because it needs to be checked into `dev` but it works very similarly to our iOS implementation and the permission check was copied over exactly.  

Example of the test results on the action summary screen:
Screenshot 2025-01-28 at 11 02 18 PM

Failures are also in-line annotations on the PR 😎


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,18 @@
+name: Nightly Tests
+
+on: 
+  workflow_dispatch:
+  schedule:
+    - cron: "0 5 * * 2,6"  # cron is UTC, this translates to 10 PM PST Mon and Fri.
```
brandonpage, line 6: We used to run different API levels on different days of the week but it's a lot more efficient to just run all of them (in parallel on Firebase Test Lab) a couple times a week.

.github/workflows/reusable-workflow.yaml
```diff
@@ -0,0 +1,127 @@
...
+        with:
+          credentials_json: '${{ secrets.GCLOUD_SERVICE_KEY }}'
+      - uses: 'google-github-actions/setup-gcloud@v2'
+        if: success() || failure()
+      - name: Run Tests
+        if: success() || failure()
+        run: |
+          # Most used according to https://gs.statcounter.com/android-version-market-share/mobile-tablet/worldwide
+          DEFAULT_API_VERSION="34" 
+          FULL_API_RANGE="28 29 30 31 32 33 34 35"
```
brandonpage, line 73: This (along with the Java and Ruby versions above) are why we _really_ need YAML Anchors.  If/when they finally add it we can move this to the top where it can be more easily found and updated when necessary.

.github/workflows/reusable-workflow.yaml
```diff
@@ -0,0 +1,127 @@
...
+          COMMAND+="--app \"native/NativeSampleApps/RestExplorer/build/outputs/apk/debug/RestExplorer-debug.apk\" "
+          COMMAND+="--test=libs/${{ inputs.lib }}/build/outputs/apk/androidTest/debug/${{ inputs.lib }}-debug-androidTest.apk "
+          COMMAND+="${DEVICES}"
+          COMMAND+="--environment-variables coverage=true,coverageFile=\"/sdcard/coverage.ec\" "
+          COMMAND+="--directories-to-pull=/sdcard  "
+          COMMAND+="--results-dir=${{ inputs.lib }}-${{github.run_number}} "
+          COMMAND+="--results-history-name=${{ inputs.lib }} "
+          COMMAND+="--timeout=20m --no-auto-google-login --no-record-video --no-performance-metrics --num-flaky-test-attempts=1"
+
+          eval "$COMMAND"
```
brandonpage, line 97: I don't love this implementation, but with the above section to add a single device for PRs and a device per API level for nightly runs I am not sure what else I can do.  It's still pretty readable so I'd rather not break it out into yet another ruby or JS script.

wmathurin (approved):
LGTM!

PR #2661, opened 2025-01-30
Fix Github Actions PR event checks.

brandonpage:
This PR adds a `is_pr` input parameter to our reusable-workflow.  The value of `github.event_name` changes from `pull_request` (or however the original flow started) to `workflow_call` in the reusale-workflow since that is technically how the _reusable-workflow was triggered_.  

CI will fail because it runs the workflow file on `dev`.  Tested [here](https://github.com/brandonpage/SalesforceMobileSDK-iOS/pull/13).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -37,7 +37,7 @@ jobs:
           # We need a sufficient depth or Danger will occasionally run into issues checking which files were modified.
           fetch-depth: 100
           # This is dangerous without the member check
-          ref: ${{ github.head_ref }}
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yaml
```diff
@@ -3,8 +3,7 @@ name: Pull Request
 on:
   # Dangerious without Member Check setup!
   pull_request_target:
-    branches:
-      - dev
+    branches: [dev, master]
```
brandonpage, line 6: CI should run on master too.  See [this discussion](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3822#discussion_r1934828572).

wmathurin (approved):

PR #2662, opened 2025-01-31
@W-17433472: Update Templates

JohnsonEricAtSalesforce:
🎸 _*Ready For Full Review!!!*_ 🥁

  This updates the Android templates where needed with MSDK support for Jetpack Compose.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.2-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
--- a/libs/MobileSync/build.gradle.kts
+++ b/libs/MobileSync/build.gradle.kts
@@ -13,10 +13,10 @@ dependencies {
     api(project(":libs:SmartStore"))
     api("androidx.appcompat:appcompat:1.7.0")
     api("androidx.appcompat:appcompat-resources:1.7.0")
-    implementation("androidx.core:core-ktx:1.13.1")
-    androidTestImplementation("androidx.test:runner:1.6.0")
-    androidTestImplementation("androidx.test:rules:1.6.0")
-    androidTestImplementation("androidx.test.ext:junit:1.2.0")
+    implementation("androidx.core:core-ktx:1.15.0")
+    androidTestImplementation("androidx.test:runner:1.6.2")
+    androidTestImplementation("androidx.test:rules:1.6.1")
+    androidTestImplementation("androidx.test.ext:junit:1.2.1")
 }
 
 android {
--- a/libs/MobileSync/build.gradle.kts
+++ b/libs/MobileSync/build.gradle.kts
@@ -1,5 +1,3 @@
-@file:Suppress("UnstableApiUsage")
-
 rootProject.ext["PUBLISH_GROUP_ID"] = "com.salesforce.mobilesdk"
 rootProject.ext["PUBLISH_VERSION"] = "13.0.0"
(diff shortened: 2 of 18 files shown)


brandonpage:
I really like the direction this is going!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.2-bin.zip
```
brandonpage, line 3: NIT:  If would be awesome if we could keep the (new) CI in sync with this.  In `.github/workflows/reusable-workflow.yaml` there is a line that specifies it: `gradle-version: 8.7` that we should bump to `8.10.2`.  

It shouldn't fail without being updated, but each run will be slower.

libs/MobileSync/build.gradle.kts
```diff
@@ -15,10 +13,10 @@ dependencies {
     api(project(":libs:SmartStore"))
     api("androidx.appcompat:appcompat:1.7.0")
     api("androidx.appcompat:appcompat-resources:1.7.0")
-    implementation("androidx.core:core-ktx:1.13.1")
-    androidTestImplementation("androidx.test:runner:1.6.0")
-    androidTestImplementation("androidx.test:rules:1.6.0")
-    androidTestImplementation("androidx.test.ext:junit:1.2.0")
+    implementation("androidx.core:core-ktx:1.15.0")
```
wmathurin, line 16: Should we pull the version number into constants like we did [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/build.gradle.kts#L16)

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -166,23 +167,35 @@ fun LoginView() {
...
+                                contentColor = colorResource(id = sf__secondary_color)
+                            )
+                        ) {
+                            Text(
+                                text = stringResource(id = sf__launch_idp),
+                                fontSize = 14.sp
+                            )
+                        }
+                    }
+                    viewModel.additionalBottomBarButtons.value.forEach { button ->
```
wmathurin, line 193: We probably should add a little todo here about adding back the bio auth login button.
Brandon had a comment about it - see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blame/7d7892fdc0d22658b618df2bf88e0f0db74d4eeb/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt#L158) - it got lost in my last PR.

libs/MobileSync/build.gradle.kts
JohnsonEricAtSalesforce: I usually don't use constants such as that since the IDE doesn't have the option to press Alt-Enter and update them.  Constants would be convenient if using a plain text editor, but actually seem to take away from the IDE functionality.  I'm flexible either way, though.

wmathurin: Interesting. Maybe we should get rid of constants then? @brandonpage

brandonpage: That is slightly annoying.  But it still tells you dependency is out of date.  Personally, I'd rather manually type the version number once alt + enter six times. 🤷

wmathurin: We could also have a [version catalog file](https://developer.android.com/build/migrate-to-catalogs) (libs.versions.toml) like they do in LSDK.

brandonpage: Plus, if we did `val thing by extra { "1.1.1" }` in SalesforceSDK you could reference it here in MobileSync.

JohnsonEricAtSalesforce: I dig the idea of switching to version catalogs.  Could we do that as a separate work item, though?

wmathurin (approved):

PR #2663, opened 2025-02-01
Fix nightly test runs.

brandonpage:
This PR changes the Nightly test structure from running one Firebase Test Lab with all supported API levels simultaneously to running them sequentially.  

If we try to run 8 instances (API 28 to 35) of our tests at once, matrixes across all of our libraries, it is simply too many simultaneous API calls to the server and the tokens gets invalidated.  This also probably would have lead to a lot of flapping tests in MobileSync when manipulating data on the server.  

Here is an example of the report from a test I ran with just 28 and 35:
Screenshot 2025-01-31 at 4 22 47 PM

In production it will show all (8 as of now) supported API levels in a single report page per library:

Screenshot 2025-01-31 at 6 37 28 PM

### DIFF
--- a/.github/workflows/reusable-workflow.yaml
+++ b/.github/workflows/reusable-workflow.yaml
@@ -75,55 +75,74 @@ jobs:
       - uses: 'google-github-actions/setup-gcloud@v2'
         if: success() || failure()
       - name: Run Tests
+        continue-on-error: true
         if: success() || failure()
         env:
           # Most used according to https://gs.statcounter.com/android-version-market-share/mobile-tablet/worldwide
-          DEFAULT_API_VERSION: "34" 
+          PR_API_VERSION: "34" 
           FULL_API_RANGE: "28 29 30 31 32 33 34 35"
           IS_PR: ${{ inputs.is_pr }}
         run: |
-          DEVICES=""
+          LEVELS_TO_TEST=$FULL_API_RANGE
+
           if $IS_PR ; then
-            DEVICES="--device model=MediumPhone.arm,version=${DEFAULT_API_VERSION},locale=en,orientation=portrait "
-          else
-            for LEVEL in $FULL_API_RANGE
+            LEVELS_TO_TEST=$PR_API_VERSION
+          fi
+
+          mkdir firebase_results
+          for LEVEL in $LEVELS_TO_TEST
             do
-                DEVICES+="--device model=MediumPhone.arm,version=${LEVEL},locale=en,orientation=portrait "
+              GCLOUD_RESULTS_DIR=${{ inputs.lib }}-api-${LEVEL}-build-${{github.run_number}}
+
+              gcloud firebase test android run \
+                --project mobile-apps-firebase-test \
+                --type instrumentation \
+                --app "native/NativeSampleApps/RestExplorer/build/outputs/apk/debug/RestExplorer-debug.apk" \
+                --test=libs/${{ inputs.lib }}/build/outputs/apk/androidTest/debug/${{ inputs.lib }}-debug-androidTest.apk \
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-workflow.yaml
```diff
@@ -75,55 +75,74 @@ jobs:
...
+                --project mobile-apps-firebase-test \
+                --type instrumentation \
+                --app "native/NativeSampleApps/RestExplorer/build/outputs/apk/debug/RestExplorer-debug.apk" \
+                --test=libs/${{ inputs.lib }}/build/outputs/apk/androidTest/debug/${{ inputs.lib }}-debug-androidTest.apk \
+                --device model=MediumPhone.arm,version=${LEVEL},locale=en,orientation=portrait \
+                --environment-variables coverage=true,coverageFile="/sdcard/coverage.ec" \
+                --directories-to-pull=/sdcard \
+                --results-dir=${GCLOUD_RESULTS_DIR} \
+                --results-history-name=${{ inputs.lib }} \
+                --timeout=20m --no-auto-google-login --no-record-video --no-performance-metrics --num-flaky-test-attempts=1 || true
```
brandonpage, line 107: The `|| true` is a little trick to ensure the test run does not return a non-zero value to the shell, since that will cause the entire step to end before we have looped through all API levels.

.github/workflows/reusable-workflow.yaml
```diff
@@ -75,55 +75,74 @@ jobs:
...
+            done
       - name: Copy Test Results
+        continue-on-error: true
```
brandonpage, line 110: I put this here and on the "Run Tests" step just to help ensure we actually loop though all API levels.

.github/workflows/reusable-workflow.yaml
```diff
@@ -75,55 +75,74 @@ jobs:
...
+
+              gsutil ls gs://test-lab-w87i9sz6q175u-kwp8ium6js0zw/${GCLOUD_RESULTS_DIR} > /dev/null 2>&1
+              if [ $? == 0 ] ; then
+                mkdir firebase_${LEVEL}
+                gsutil -m cp -r -U "`gsutil ls gs://test-lab-w87i9sz6q175u-kwp8ium6js0zw/${GCLOUD_RESULTS_DIR} | tail -1`*" ./firebase_${LEVEL}/
+                mv firebase_${LEVEL}/test_result_1.xml firebase_results/api_${LEVEL}_test_result.xml
+              else
+                if $IS_PR ; then
+                  echo "No test results found"
+                  exit 1
```
brandonpage, line 136: Error on PR to make sure expected tests were actually run.

.github/workflows/reusable-workflow.yaml
```diff
@@ -75,55 +75,74 @@ jobs:
...
+                --results-history-name=${{ inputs.lib }} \
+                --timeout=20m --no-auto-google-login --no-record-video --no-performance-metrics --num-flaky-test-attempts=1 || true
+            done
       - name: Copy Test Results
+        continue-on-error: true
         if: success() || failure()
+        env:
+          # Most used according to https://gs.statcounter.com/android-version-market-share/mobile-tablet/worldwide
+          PR_API_VERSION: "34" 
+          FULL_API_RANGE: "28 29 30 31 32 33 34 35"
```
brandonpage, line 115: As I have said on previous PR's but wow, how does Github Actions not support YAML Anchors???

I am actively watching the (4 year old!) issue requesting the feature which may actually be getting fulfilled soon, but I will specially mention this needs to be updated in multiple places on out Github Actions doc.

wmathurin (approved):

PR #2665, opened 2025-02-03
Theming for new compose classes

wmathurin:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
```diff
@@ -0,0 +1,91 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.ui.theme
+
+import android.content.Context
+import androidx.annotation.ColorInt
+import androidx.core.content.ContextCompat
+import com.salesforce.androidsdk.R
+
+object SFColors {
```
wmathurin, line 34: That way, the colors xml is still being used (and can be overriden in the client app). Or we don't want to do it this way ?? @brandonpage @JohnsonEricAtSalesforce

brandonpage: I don't mind it, but I am also not sure if it servers a purpose.  I don't think even the internal apps are using our existing colors for anything other than overriding our UI. 

If we keep the XML colors do I have to add new values there and here?

wmathurin: If I don't keep it, how would they override the colors?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SalesforceTheme.kt
```diff
@@ -0,0 +1,111 @@
...
+import androidx.compose.material3.ColorScheme
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.darkColorScheme
+import androidx.compose.material3.lightColorScheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalContext
+
+enum class SalesforceThemeType {
+    Light, Dark, DarkLogin
```
brandonpage, line 39: Does `DarkLogin` exist because we can't change the webview background color from white?  What I had been doing was hiding the WebView (via opacity) when loading so we could set the background color to whatever we want -- not sure that is relevant though.

wmathurin: I think that was the reason for DarkLogin.
Let me try with Dark and see how it looks.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
brandonpage: I was originally thinking via the view model but I don't know what is best.  My first thought is that the theme could be a data class (that could be overridden) instead of an object? 🤷

brandonpage: Or since we seem to be using a lot of Material3 components, should we be using a [MaterialTheme](https://developer.android.com/develop/ui/compose/designsystems/material3#material-theming)?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SalesforceTheme.kt
wmathurin: This is the experience using Dark (instead of DarkLogin) 

LoginWithDark

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
wmathurin: Maybe - should we provide an easy way for an application to use a different theme ?

wmathurin: Should we allow the app to provide MaterialTheme's for light and dark (through SalesforceSDKManager) and we would default to the MaterialTheme.lightColorTheme() and MaterialTheme.darkColorTheme()  ??

wmathurin:
@brandonpage @JohnsonEricAtSalesforce what do you think of this approach (https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2665/commits/5b18ebeb9c440394304bcad9331c14fdb0783bd1) for theming?

wmathurin:
Should we merge this one?
There is probably be some rework when we do the login fit and finish work e.g. where the colors should come from (properties on the LoginViewModel and/or sf__colors.xml and/or dynamic).
@JohnsonEricAtSalesforce @brandonpage

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
JohnsonEricAtSalesforce: @brandonpage & @wmathurin - Did you come up with an answer on the approach for this?  I believe we're keeping the colors in the resource XML?

JohnsonEricAtSalesforce (approved):
🤘🏻

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
wmathurin: We are keeping sf__colors.xml until at least 14.
So that's one way to override default colors.
We also added setDarkColorScheme() and setLightColorScheme() to SalesforceSDKManager - as another way to override colors.
Finally for login, the LoginViewModel exposes a few key colors as members.

As we work on login fit/finish, we might simplify things a bit??

PR #2667, opened 2025-02-04
Converting DevInfoActivity to compose

wmathurin:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.java
```diff
@@ -1,82 +0,0 @@
...
-
-    @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-        boolean isDarkTheme = SalesforceSDKManager.getInstance().isDarkTheme();
-        setTheme(isDarkTheme ? R.style.SalesforceSDK_Dark : R.style.SalesforceSDK);
-
-        // Title / layout
-        getSupportActionBar().setTitle(R.string.sf__dev_support_title);
-        setContentView(R.layout.sf__dev_info);
```
brandonpage, line 61: We should we remove this layout XML.

brandonpage (approved):

PR #2668, opened 2025-02-08
@W-17411362: LoginActivity Fit and Finish

JohnsonEricAtSalesforce:
🎸 _*Ready For Final Review!*_ 🥁

  This actually got a lot easier today since #2670 included all the bottom sheet styling for select server and add connection (which looks great. Thanks!)

  Note: This has a counterpart in the templates repository: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/444/files


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1498,18 +1501,29 @@ open class SalesforceSDKManager protected constructor(
...
     open fun setViewNavigationVisibility(activity: Activity) {
         if (!isDarkTheme || activity.javaClass.name == loginActivityClass.name) {
             /*
              * This covers the case where OS dark theme is true, but app has
              * disabled.
              */
-            activity.window.decorView.systemUiVisibility = SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR or SYSTEM_UI_FLAG_LIGHT_STATUS_BAR
+            if (SDK_INT > R) {
+                runCatching {
+                    activity.window?.insetsController?.setSystemBarsAppearance(
```
JohnsonEricAtSalesforce, line 1518: @brandonpage - Here's a modern way to accomplish what this once did with a dandy to-do so we can clean up once API 30 is our minimum.  I did some reading, and though the names have changed the overall approach is almost identical.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -194,6 +195,9 @@ open class LoginActivity: FragmentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         enableEdgeToEdge()
+        if (viewModel.dynamicBackgroundTheme.value == Dark) {
```
JohnsonEricAtSalesforce, line 198: I added this new `dynamicBackgroundTheme` state so we can know if the view model considers the background "light" or "dark".

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -316,4 +319,9 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         val title: Int,
         val onClick: () -> Unit
     )
+
+    internal enum class Theme {
```
JohnsonEricAtSalesforce, line 323: @wmathurin or @brandonpage - We haven't already made a state for this, have we?

brandonpage: https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/33f7c5cc6aff1f6baadf8d4f8a36bc18747e7a12/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt#L482

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1498,18 +1501,29 @@ open class SalesforceSDKManager protected constructor(
...
+     * @param activity The activity
      */
     open fun setViewNavigationVisibility(activity: Activity) {
         if (!isDarkTheme || activity.javaClass.name == loginActivityClass.name) {
             /*
              * This covers the case where OS dark theme is true, but app has
              * disabled.
              */
-            activity.window.decorView.systemUiVisibility = SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR or SYSTEM_UI_FLAG_LIGHT_STATUS_BAR
+            if (SDK_INT > R) {
```
wmathurin, line 1516: Should we add a TODO that says remove when min version of SDK >= API 30

wmathurin: Nevermind, I see it below.

wmathurin:
Can we get rid of [sfDarkLoginColors](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt#L175) ?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
```diff
@@ -132,6 +132,7 @@ fun sfLightColors(): ColorScheme {
         secondary = Color(SFColors.secondaryColor(context)),
         background = Color(SFColors.background(context)),
         surface = Color(SFColors.layoutBackground(context)),
+        surfaceContainer = Color(SFColors.layoutBackground(context)),
```
JohnsonEricAtSalesforce, line 135: This addition is for our `DropDownMenu` backgrounds so they match 12.x.  The default color in light theme was an odd looking pink-off-white I don't believe anyone likes 🤢

brandonpage: We can consult UX, but is there a reason not to match the picker background and use `primaryContainer`?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -123,10 +125,9 @@ fun LoginView() {
                     DropdownMenu(
                         expanded = showMenu,
                         onDismissRequest = { showMenu = false },
-                        containerColor = Color.White,
                     ) {
                         DropdownMenuItem(
-                            text = { Text("Change Server", color = Color.Gray) },
+                            text = { Text("Change Server") },
```
brandonpage, line 130: Only use localizable stings.  This is should be `stringResource(R.string.sf__pick_server)` for example.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
JohnsonEricAtSalesforce: This sets `surfaceContainer` to the same value as `primaryContainer`.  Does that accomplish what you're thinking of?  By setting `surface` here we don't have to apply any color in the layout.

Screenshot 2025-02-12 at 11 47 58

Screenshot 2025-02-12 at 11 49 33

Screenshot 2025-02-12 at 11 50 09

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
JohnsonEricAtSalesforce: Turns out there were a lot of fixed strings in the login composables.  I did a search and tried to find them all in a commit moments ago.

brandonpage:
@JohnsonEricAtSalesforce Things I was expecting to see in this PR:
- Bio Auth / IDP button(s) with Fit and Finish
  - Should we implement a QR Code button so apps don't have to do it in their code? 
  - Can we just use `additionalBottomBarButtons` for all of these?
  - What is the max number that should be shown at once?
- Fix/more consistent loading animation.  Last I checked this did not trigger every time.
- Accessibility testing.  If no work is needed that is fine.
- Compose Semantic UI Automation -- I can implement this with the customization story if you want since that one should really require it.

libs/SalesforceSDK/res/values/sf__strings.xml
```diff
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-	<!-- Errors -->
```
JohnsonEricAtSalesforce, line 4: The leading indentation for this source was pretty mixed, so since we're adding a significant number of lines I thought a gentle reformat to Android Studio's defaults would be helpful.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginServerListItem.kt
```diff
@@ -128,11 +131,12 @@ fun LoginServerListItem(
...
             horizontalArrangement = Arrangement.SpaceBetween,
             verticalAlignment = Alignment.CenterVertically,
-            modifier = Modifier.fillMaxWidth()
+            modifier = Modifier
+                .fillMaxWidth()
                 .height(intrinsicSize = IntrinsicSize.Max)
                 .onSizeChanged { size -> rowSizePixels = size }
                 .clickable(
-                    onClickLabel = stringResource(R.string.sf__server_picker_click_label),
+                    onClickLabel = stringResource(sf__server_picker_click_label),
```
JohnsonEricAtSalesforce, line 139: If we really enjoy the `R.string` prefixes instead of static imports, I could switch to that everywhere instead. Here, I just went to match what was in LoginView earlier.

JohnsonEricAtSalesforce:
> @JohnsonEricAtSalesforce Things I was expecting to see in this PR:
> 
> * Bio Auth / IDP button(s) with Fit and Finish
>   
>   * Should we implement a QR Code button so apps don't have to do it in their code?
>   * Can we just use `additionalBottomBarButtons` for all of these?
>   * What is the max number that should be shown at once?
> * Fix/more consistent loading animation.  Last I checked this did not trigger every time.
> * Accessibility testing.  If no work is needed that is fine.
> * Compose Semantic UI Automation -- I can implement this with the customization story if you want since that one should really require it.

I'm getting these entered in GUS as well.  We had some but not all of these under our general finish work item.

JohnsonEricAtSalesforce:
> * Should we implement a QR Code button so apps don't have to do it in their code?

So much of QR Code login is customizable according to the implementation we released in 12.x, including the placement and function of this button, that I feel it belongs to the app.  The real kicker in my view there is that MSDK doesn't control what that button actually does.  Our template app's example uses one version of image capture logic but each implementation can completely customize that.

It's also nice that with the `additionalBottomBarButtons` view model customization it's a really tidy one-liner to add this button there, if the app wishes to use our template's approach.

Perhaps our guiding logic should be to add MSDK-provided buttons for features where MSDK controls the logic they expose and let apps use `additionalBottomBarButtons` for anything the app provides the logic for?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginServerListItem.kt
brandonpage: That seems fine.  Removing static imports only bother me when removing the prefix makes it no longer obvious what the thing is or which base class it is from.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -244,42 +244,77 @@ internal fun PickerBottomSheet(
...
-        var mutableSelectedListItem = selectedListItem
-
-        Column(
-            modifier = Modifier
-                .animateContentSize()
-                .semantics { contentDescription = containerContentDescription }
-                .focusRequester(pickerFocus)
-                .focusable(),
+    val themeRippleConfiguration = RippleConfiguration(color = colorScheme.onSecondary)
+    CompositionLocalProvider(LocalRippleConfiguration provides themeRippleConfiguration) {
```
brandonpage, line 248: @JohnsonEricAtSalesforce I thought this was explicitly for individual components?

JohnsonEricAtSalesforce: I was thinking of the entire `PickerBottomSheet` in this case.  Where it was placed earlier it didn't apply to the broader "component" and adding it to each location seemed redundant.

brandonpage (approved):

PR #2670, opened 2025-02-10
Generic LoginServer / UserAccount Picker Bottom Sheet

brandonpage:
`PickerBottomSheet` is an extremely DRY, reusable container that displays a row list item component that corresponds with a `PickerStyle` enum value.  If we want to add a new type of Picker in the future, it should be as easy as adding a new enum value and filling in the places the compiler tells you are not exhaustive. 

The public component does a lot of setup work with SalesforceSDKManager, LoginViewModel, etc to call a secondary internal component of the same name.  This is done to facilitate previews and to make the component testable.  For example:  the internal picker takes in an `onItemSelected: (Any?, Boolean) -> Unit` parameter which can be supplied with nothing for previews and a function that tests if is actually being called in tests. 

Combining these two approaches makes the picker quite abstract but I believe though careful naming the code is still very readable (let me know if you disagree!).  

LoginServerListItem Preview Example:
Screenshot 2025-02-10 at 8 09 08 PM

UserAccountListItem Preview Example:
Screenshot 2025-02-10 at 8 09 37 PM

AddConnection Preview:
Screenshot 2025-02-10 at 9 06 34 AM

PickerBottomSheet Preview:
Screenshot 2025-02-10 at 8 10 25 PM


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -177,6 +179,14 @@ public void useSandbox() {
 	 * @param url Server URL.
 	 */
 	public void addCustomLoginServer(String name, String url) {
+		// Prevent duplicate servers.
+		for (LoginServer existingServer : getLoginServers()) {
+			if (name.equals(existingServer.name) && url.equals(existingServer.url)) {
+				setSelectedLoginServer(existingServer);
+				return;
+			}
+		}
```
brandonpage, line 188: It did not make sense to check for this in the UI code.  If we did, we would then have to surface some kind of error message.  Simply not adding the duplicate and selecting the original in the data source of truth seemed like an elegant solution.  

And I purposefully did not use our equals method because I wanted to ignore `isCustom`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -203,6 +213,39 @@ public void reset() {
 		initSharedPrefFile();
 	}
 
+	/**
+	 * Removes a login serer from the list.
+	 *
+	 * @param server the server to remove
+	 */
+	public void removeServer(LoginServer server) {
```
brandonpage, line 221: This was a bit gnarly, but necessary for the new UX.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -0,0 +1,631 @@
...
+
+            true
+        }
+    )
+    val addNewLoginServer = { name: String, url: String ->
+        loginServerManager.addCustomLoginServer(name, url)
+        viewModel.showServerPicker.value = false
+        viewModel.loading.value = true
+        viewModel.dynamicBackgroundColor.value = backgroundColor
+    }
```
brandonpage, line 193: These inline functions are the only **untested** part of the code.  My UI tests check that they are called but their actual contents can't be tested.  Maybe the Login Server ones should be added to `LoginViewModel`?  

The UserAccount ones wouldn't belong there and require reference to the activity (which should be avoided in view models IMO).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -0,0 +1,631 @@
...
+                                            }
+                                        }
+                                    }
+                                    HorizontalDivider(
+                                        thickness = STROKE_WIDTH.dp,
+                                        modifier = Modifier.padding(horizontal = PADDING_SIZE.dp),
+                                        color = colorScheme.surfaceVariant,
+                                    )
+
+                                    // Add New Connection/Account Button
```
brandonpage, line 408: Adding this to the end of the list feels a bit obtuse, but having the add button as the last list item really helps to make the UI more usable on a phone in landscape when the list is very cramped.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -0,0 +1,631 @@
...
+
+// Ensure no duplicates in the list because they are not allowed by lazy column.
+fun List<Any>.pickerDistinctBy() : List<Any> {
+    return distinctBy { listItem ->
+        when(listItem) {
+            is LoginServer -> with(listItem) { "$name$url" }
+            else -> listItem.toString()
+        }
+    }
+}
```
brandonpage, line 548: This is probably necessary/overkill since the lazy column uses the list items's toString as a unique key but I would rather just not show duplicate servers at all if they are already stored on the users device pre-13.0.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -203,6 +213,39 @@ public void reset() {
 		initSharedPrefFile();
 	}
 
+	/**
+	 * Removes a login serer from the list.
```
wmathurin, line 217: typo: ser[v]er

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/theme/SFColors.kt
```diff
@@ -90,45 +96,78 @@ object SFColors {
...
     )
 }
 
 @Composable
 fun sfDarkColors(): ColorScheme {
     val context = LocalContext.current
     return darkColorScheme(
-        primary = Color(SFColors.primaryColor(context)),
-        primaryContainer = Color(SFColors.primaryColorDark(context)),
+        primary = Color(SFColors.primaryColorDark(context)),
```
wmathurin, line 154: 👍

wmathurin (approved):
LGTM!

JohnsonEricAtSalesforce (approved):
🤘🏻

PR #2676, opened 2025-02-17
Downgrading gradle back to 8.7

wmathurin:


.github/workflows/reusable-workflow.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
         uses: android-actions/setup-android@v3
       - uses: gradle/actions/setup-gradle@v4
         with:
-          gradle-version: 8.10.2
+          gradle-version: 8.7
```
brandonpage, line 50: 🎉

brandonpage (approved):

.github/DangerFiles/TestOrchestrator.rb
```diff
@@ -9,14 +9,19 @@
...
+modified_libs = Set[]
 for file in (git.modified_files + git.added_files);
     scheme = file.split("libs/").last.split("/").first
     if LIBS.include?(scheme) 
-        modifed_libs.add(scheme)
+        modified_libs.add(scheme)
     end
 end
 
+# If modified_libs is empty, add all LIBS
```
wmathurin, line 20: @brandonpage What do you think of that change?

PR #2678, opened 2025-02-19
Remove LoginOptions and OAuthWebviewHelper.

brandonpage:
Native Login and IDP have been updated to use the new AuthentiationUtlities.


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -350,23 +349,17 @@ public void switchToUser(UserAccount user, int userSwitchType, Bundle extras) {
 	 * in ClientManager will return a RestClient instance for the new user.
 	 */
 	public void switchToNewUser() {
-		final Bundle options = SalesforceSDKManager.getInstance().getLoginOptions().asBundle();
-		switchToNewUserWithOptions(options);
+		final Bundle options = new Bundle();
```
brandonpage, line 352: This is a pattern you will see over and over but it strikes me a very odd.  The only value in`LoginOptions` that `LoginActivity` would read from a bundle is the `jwt` string 🤷.   That `jwt` value is not set on `LoginViewModel`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
```diff
@@ -122,24 +118,16 @@ public Bundle addAccount(AccountAuthenticatorResponse response,
...
-
         @Override
         public Bundle getAuthToken(AccountAuthenticatorResponse response, Account account,
                             String authTokenType, Bundle options) throws NetworkErrorException {
-
             UserAccount originalUserAccount = UserAccountManager.getInstance().buildUserAccount(account);
-            final Map<String,String> addlParamsMap = SalesforceSDKManager.getInstance().getLoginOptions().getAdditionalParameters();
+
             try {
+                final Map<String,String> addlParamsMap = originalUserAccount.getAdditionalOauthValues();
```
brandonpage, line 130: This also comes up a couple times but I think it is a bug nobody ever caught?  The additional params that were stored on `LoginOptions` were _login_ params like `login_hint=username` added to the URL for the browser to interpret.  Here we appear to be adding those params to the token refresh call instead of the additional auth params stored on the user account that we get back from the server upon login.

JohnsonEricAtSalesforce (approved):
👍🏻

PR #2680, opened 2025-02-19
W-17867189: [MSDK Android] Login View Bottom App Bar Shown Even When Empty

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
@@ -173,14 +173,17 @@ fun LoginView() {
             } else null
         }
 
-    val bottomAppBar = viewModel.bottomAppBar ?: {
-        DefaultBottomAppBar(
-            backgroundColor = viewModel.dynamicBackgroundColor,
-            button = bioAuthButton ?: idpButton ?: customButton,
-            loading = viewModel.loading.value,
-            showButton = !viewModel.authFinished.value,
-        )
-    }
+    val bottomAppBarButton = bioAuthButton ?: idpButton ?: customButton
+    val bottomAppBar = viewModel.bottomAppBar ?: bottomAppBarButton?.let { button ->
+        {
+            DefaultBottomAppBar(
+                backgroundColor = viewModel.dynamicBackgroundColor,
+                button = button,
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -507,7 +510,7 @@ private fun AppBarBackButtonPreview() {
 
 @Preview
 @Composable
-private fun AppBarBackButtonLoadinPreview() {
+private fun AppBarBackButtonLoadInPreview() {
```
JohnsonEricAtSalesforce, line 513: @brandonpage - It looked like this was the only inspector spelling suggestion in this source.  Was it supposed to be capital `I`?

brandonpage: I think we can actually remove that preview as it is identical to the one above it.  At one time I was playing around with hiding the title while loading (I missed the "g" in the name) but it was just a bad idea so I scraped it... guess I missed this in the cleanup.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -173,14 +173,17 @@ fun LoginView() {
...
-            loading = viewModel.loading.value,
-            showButton = !viewModel.authFinished.value,
-        )
-    }
+    val bottomAppBarButton = bioAuthButton ?: idpButton ?: customButton
+    val bottomAppBar = viewModel.bottomAppBar ?: bottomAppBarButton?.let { button ->
+        {
+            DefaultBottomAppBar(
+                backgroundColor = viewModel.dynamicBackgroundColor,
+                button = button,
```
brandonpage, line 181: Should this be `button = bottomAppBarButton,` ?

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
JohnsonEricAtSalesforce: Gotcha.  I removed it now.

PR #2684, opened 2025-02-27
@W-17920337: [MSDK Android] `MobileSyncTest` Test Suite Fails Due To Removal Of Login Options

JohnsonEricAtSalesforce:
🎸 *_Ready For Review_* 🥁

This gets the tests compiling again, but I do notice some test failures for both 12.x and 13.x in the suite that look like this.  If I recall, I always used to get these when testing against MobileSDK.

```
Wrong total size expected:<4> but was:<0>
```

Screenshot 2025-02-26 at 18 46 23


wmathurin (approved):

brandonpage (approved):

brandonpage:
> 🎸 __Ready For Review__ 🥁
> 
> This gets the tests compiling again, but I do notice some test failures for both 12.x and 13.x in the suite that look like this. If I recall, I always used to get these when testing against MobileSDK.
> 
> ```
> Wrong total size expected:<4> but was:<0>
> ```
> 
> Screenshot 2025-02-26 at 18 46 23

If they pass on CI, I wouldn't mess with it.  The difference in expected value is very likely due to you generating your `test_credentials.json` from a session where you logged into a different org with different data.  You should use `testandroid@gs0.mobilesdk.com`, with the proper connected app.  (I see this isn't in the tribal knowledge doc, so I will add it).

PR #2688, opened 2025-03-01
Resolving with vf domain if available and applicable

wmathurin:
Without the change there is an extra redirect happening when loading the page.


brandonpage (approved):

brandonpage:
Very NIT, but you could updated the method description above to include VF.

wmathurin:
Actually seeing the issue again. Closing.

PR #2695, opened 2025-03-19
@W-18033420: Actionable Notifications (Android) Register/De-Register Android Notification Channels For Notifications Types

JohnsonEricAtSalesforce:
🤘🏻 *_Ready For Review_* 🎸

  This relatively simple update registers Android Notification Channels for each Salesforce Notification Type on log in.  They're then removed at log out.  They new channels are also added to an Android Notification Group.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
@@ -26,6 +26,10 @@
  */
 package com.salesforce.androidsdk.push
 
+import android.app.NotificationChannel
+import android.app.NotificationChannelGroup
+import android.app.NotificationManager
+import android.app.NotificationManager.IMPORTANCE_DEFAULT
 import android.content.Intent
 import androidx.core.net.toUri
 import androidx.work.Constraints
@@ -59,6 +63,7 @@ import com.salesforce.androidsdk.push.PushService.PushNotificationReRegistration
 import com.salesforce.androidsdk.rest.ApiVersionStrings
 import com.salesforce.androidsdk.rest.ClientManager.AccMgrAuthTokenProvider
 import com.salesforce.androidsdk.rest.NotificationsApiClient
+import com.salesforce.androidsdk.rest.NotificationsTypesResponseBody
 import com.salesforce.androidsdk.rest.RestClient
 import com.salesforce.androidsdk.rest.RestClient.ClientInfo
 import com.salesforce.androidsdk.rest.RestRequest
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -237,10 +242,63 @@ open class PushService {
...
             REGISTRATION_STATUS_SUCCEEDED ->
-                fetchNotificationsTypes(userAccount ?: return)
+                registerNotificationChannels(
+                    fetchNotificationsTypes(userAccount ?: return) ?: return
+                )
 
-            UNREGISTRATION_STATUS_SUCCEEDED ->
+            UNREGISTRATION_STATUS_SUCCEEDED -> {
                 clearNotificationsTypes(userAccount ?: return)
+                removeNotificationsCategories()
```
brandonpage, line 251: Should these two lines be flipped?  If `useAccount` is null we won't remove notification categories.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -249,17 +307,23 @@ open class PushService {
...
      * @param userAccount the user account that's performing registration
      */
-    private fun fetchNotificationsTypes(userAccount: UserAccount) {
+    private fun fetchNotificationsTypes(
+        userAccount: UserAccount
+    ): NotificationsTypesResponseBody? {
         val instanceHost = userAccount.instanceServer.toUri().host
         val restClient = getRestClient(userAccount)
-        if (instanceHost != null && restClient != null) {
+        return if (instanceHost != null && restClient != null) {
```
brandonpage, line 315: NIT:  The `if/else` can be avoided with:
```
val instanceHost = userAccount.instanceServer.toUri().host ?: return null
val restClient = getRestClient(userAccount) ?: return null
```

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
```diff
@@ -237,10 +242,63 @@ open class PushService {
...
+                    NotificationChannelGroup(
+                        NOTIFICATION_CHANNEL_GROUP_SALESFORCE_ID,
+                        NOTIFICATION_CHANNEL_GROUP_SALESFORCE_NAME
+                    )
+                )
+                getNotificationChannelGroup(NOTIFICATION_CHANNEL_GROUP_SALESFORCE_ID)
+            }
+
+            notificationsTypesResponseBody.notificationTypes?.forEach { notificationType ->
+                createNotificationChannel(
```
wmathurin, line 281: I believe apps currently set up NotificationChannel themselves.
Should we continue letting them do that?

JohnsonEricAtSalesforce: I might have read wrong when I thought one of our goals was to register the channels automatically.  I really see your point though.  What if we made this automatic registration an opt-in? The app could opt-out and then use the registration types themselves to set up custom channels.

wmathurin: Opt-in might be a good idea. I don't know if it could collide with what they are doing now.

wmathurin: On iOS, categories have to be registered or the actionable notifications won't show action buttons. But on Android, I don't think channel registration is required?

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
JohnsonEricAtSalesforce: Good thinking.  If it's OK, I have that incorporated in the next pull request.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
JohnsonEricAtSalesforce: Nice.  Also, that's coming in the follow-up pull request.

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushService.kt
JohnsonEricAtSalesforce: Looking at the Android docs I believe I have our answer.
```
All posted notifications must specify a NotificationChannel ID.
```
There _are_ still SDK methods that omit channel id, however they are deprecated.

We could still go with the opt-in behavior if the app wishes to create the channels themselves.

JohnsonEricAtSalesforce: I'll merge this for now so we can get the next unit of work in review and if there is a follow-up for that channel opt-out we can file a work item for it 🌈

PR #2698, opened 2025-04-01
Moving to SQLCipher 4.7

wmathurin:


libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/SmartStoreTest.java
```diff
@@ -116,15 +116,15 @@ public void testRuntimeSettings() {
 	 */
 	@Test
 	public void testSQLCipherVersion() {
-		Assert.assertEquals("Wrong sqlcipher version", "4.6.1 community", store.getSQLCipherVersion());
+		Assert.assertEquals("Wrong sqlcipher version", "4.7.0 community", store.getSQLCipherVersion());
```
brandonpage, line 119: `sqlcipher-android:4.7.2` reports as version `4.7.0`?  Obviously it must since the test passes... just weird.

brandonpage (approved):

libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/SmartStoreTest.java
wmathurin: They did 4.7.0, 4.7.1 and 4.7.2 within a week.
Only 4.7.0 was announced on their blog.

In their newsletter they do mention that PRAGMA cipher_version will continue to report 4.7.0.

> Important SQLCipher for Android 4.7.2 Update
We are contacting you about an important update to your SQLCipher products covered under an active CipherCare support and maintenance subscription. 
>
> We have just released SQLCipher 4.7.2 for the following products:
>
> SQLCipher for Android (Commercial & Enterprise)
> SQLCipher for Android FIPS (Enterprise)
> SQLCipher for React Native (Enterprise)
> This update fixes two difficult-to-reproduce defects in SQLCipher for Android 4.7.0's new Cursor Window implementation. These could potentially cause a crash when a large query result set triggers a window resize with a fragmented heap. The issue could occur unpredictably based on a combination of schema, query result size, device memory state.
>
> We recommend upgrading to the new version if you are using the recently released 4.7.0, even if your application does not seem to be affected during testing, because the problem may only occur under limited conditions. Updating will improve stability and prevent potential crashes. Earlier versions of SQLCipher for Android below 4.7.0 are unaffected.
>
> This update only affects the SQLCipher for Android "wrapper" library packages. The updated packages still include the core SQLCipher 4.7.0 library (i.e. PRAGMA cipher_version will continue to report 4.7.0).

PR #2699, opened 2025-04-03
@W-18196428: [MSDK Android] User Identity No Longer Fetched In MSDK 13.0 Due To Network On Main

JohnsonEricAtSalesforce:
🎸 *_Ready For Review_* 🥁

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
@@ -42,9 +42,9 @@ import com.salesforce.androidsdk.analytics.SalesforceAnalyticsManager
 import com.salesforce.androidsdk.app.Features.FEATURE_BIOMETRIC_AUTH
 import com.salesforce.androidsdk.app.Features.FEATURE_SCREEN_LOCK
 import com.salesforce.androidsdk.app.SalesforceSDKManager
-import com.salesforce.androidsdk.auth.OAuth2.IdServiceResponse
 import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse
 import com.salesforce.androidsdk.auth.OAuth2.addAuthorizationHeader
+import com.salesforce.androidsdk.auth.OAuth2.callIdentityService
 import com.salesforce.androidsdk.auth.OAuth2.revokeRefreshToken
 import com.salesforce.androidsdk.config.RuntimeConfig.getRuntimeConfig
 import com.salesforce.androidsdk.push.PushMessaging.register
@@ -55,8 +55,10 @@ import com.salesforce.androidsdk.security.ScreenLockManager
 import com.salesforce.androidsdk.util.SalesforceSDKLogger.e
 import com.salesforce.androidsdk.util.SalesforceSDKLogger.w
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers.Default
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import okhttp3.HttpUrl
 import okhttp3.Interceptor
 import okhttp3.Request.Builder
@@ -83,7 +85,7 @@ private const val TAG = "AuthenticationUtilities"
  *  * Creates an Account.
  *  * Checks for any CA/ECA settings such as Screen Lock or Biometric Authentication.
  */
-internal fun onAuthFlowComplete(
+internal suspend fun onAuthFlowComplete(
     tokenResponse: TokenEndpointResponse,
     loginServer: String,
     consumerKey: String,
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -113,14 +115,17 @@ internal fun onAuthFlowComplete(
...
-    var userIdentity: IdServiceResponse? = null
-    runCatching {
-        userIdentity = OAuth2.callIdentityService(
-            HttpAccess.DEFAULT,
-            tokenResponse.idUrlWithInstance,
-            tokenResponse.authToken,
-        )
-    }
+    val userIdentity = runCatching {
+        withContext(Default) {
```
JohnsonEricAtSalesforce, line 119: The crux of the fix is here by using the default coroutine context for the call.  I also re-arranged things a bit from how it had looked in Java and previous Kotlin conversions so that the optional and smart casting is more intuitive.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -113,14 +115,17 @@ internal fun onAuthFlowComplete(
...
+            callIdentityService(
+                HttpAccess.DEFAULT,
+                tokenResponse.idUrlWithInstance,
+                tokenResponse.authToken,
+            )
+        }
+    }.onFailure { throwable ->
+        w(TAG, "Cannot fetch user identity due to an error.", throwable)
```
JohnsonEricAtSalesforce, line 127: Logging a warning seemed appropriate since this call is optional and the app can proceed with limited data.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -307,7 +312,7 @@ private fun HttpUrl.isSalesforceUrl(): Boolean {
     // List from https://help.salesforce.com/s/articleView?language=en_US&id=sf.domain_name_url_formats.htm&type=5
     val salesforceHosts = listOf(".salesforce.com", ".force.com", ".sfdcopens.com", ".site.com", ".lightning.com",
         ".salesforce-sites.com", ".force-user-content.com", ".salesforce-experience.com", ".salesforce-scrt.com")
-    return salesforceHosts.map { host.endsWith(it) }.any() { it }
+    return salesforceHosts.map { host.endsWith(it) }.any { it }
```
JohnsonEricAtSalesforce, line 315: This was the only inspector warning in this source, though unrelated to this bug.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -196,22 +197,24 @@ internal class NativeLoginManager(
         val tokenEndpointResponse = TokenEndpointResponse(tokenResponse.rawResponse)
         tokenResponse.consumeQuietly()
 
-        return withContext(Dispatchers.IO) {
```
JohnsonEricAtSalesforce, line 199: @brandonpage - Is there a specific reason this was on IO rather than Default? If I recall, doesn't IO have a lower priority though a much larger potential thread pool?  I'm used to using Default unless I'm both explicitly lowering the priority and expecting a higher quantity of jobs.  I'm game either way if we've thought it through.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -196,22 +197,24 @@ internal class NativeLoginManager(
...
-                        SalesforceSDKLogger.e(TAG, "$error: $errorDesc", e)
-                        continuation.resume(UnknownError)
-                    },
-                    onAuthFlowSuccess = { userAccount ->
-                        SalesforceSDKLogger.d(TAG, "onAuthFlowSuccess $userAccount")
-                        continuation.resume(Success)
-                    },
-                    nativeLogin = true,
-                )
+                runBlocking {
```
JohnsonEricAtSalesforce, line 202: Adding a run blocking here satisfies the suspend requirement and I saw that the callbacks require the suspend coroutine use, so it works yet felt a little code heavy.  Maybe we make this better someday when the callbacks are eliminated?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -884,11 +885,13 @@ open class LoginActivity : FragmentActivity() {
                                 )
 
                             else ->
-                                viewModel.onAuthFlowComplete(
-                                    TokenEndpointResponse(params),
-                                    ::onAuthFlowError,
-                                    ::onAuthFlowSuccess
-                                )
+                                CoroutineScope(Default).launch {
```
JohnsonEricAtSalesforce, line 888: There's no return value here, so launch is sufficient.

wmathurin:
Thanks @JohnsonEricAtSalesforce 
That sounds kind of bad, should we consider a 13.0.1 ?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
brandonpage: I was going to ask if we should use `IO` instead but read the documentation first:

> This dispatcher and its views share threads with the Default dispatcher, so using withContext(Dispatchers.IO) { ... } when already running on the Default dispatcher typically does not lead to an actual switching to another thread. In such scenarios, the underlying implementation attempts to keep the execution on the same thread on a best-effort basis.

TIL

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
brandonpage: I read through some doc and the "Implementation Note" for `IO` makes it sound like they use the same thread pool.  `IO` is supposed to be more elastic, idk if that means it has lower priority, but we don't need that so this change seems fine to me.

brandonpage (approved):

wmathurin:
@brandonpage There are quite a few fields we get from there: https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java#L571

PR #2702, opened 2025-04-11
Update CodeCov requirements.

brandonpage:
- Add `threshold: 0%` to all libraries.  This will fail a check if a PR lowers the overall coverage of a library by any amount. 
- Add new `patch` section with `target: 80%` for each library.  This will create a check per library that will ensures new code added in a PR has at least 80% coverage.  

I could have required a specific coverage amount (like 80%) per library but most are no where near where we ideally want them.  Instead, these rules should encourage us to organically work towards that goal.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,37 +6,83 @@ codecov:
 coverage:
   precision: 2
   round: down
-  range: "35...85"
+  range: "60...80"
 
   status:
+    # Overall Library Requirements
     project:
       default: false  # disable the default status that measures entire project
       SalesforceAnalytics:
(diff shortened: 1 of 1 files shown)


wmathurin (approved):

brandonpage:
I am looking into why there are no checks showing up on this PR.  No checks for patch coverage makes sense since I [didn't modify any of the files we track](https://app.codecov.io/github/forcedotcom/SalesforceMobileSDK-Android/commit/cadf7de42604f545534440857341bd350653a2e4), but I expected checks for the library thresholds.

PR #2709, opened 2025-05-05
Fix status bar icon color for all cases.

brandonpage:
This is a followup to #2703, which insured the status bar icons were always visible for any dynamic status bar color.  However, it did not take into account that the app can specify a static top app bar color.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -921,11 +921,10 @@ open class LoginActivity : FragmentActivity() {
                     ?: return@evaluateJavascript
 
                 // Ensure Status Bar Icons are readable no matter which OS theme is used.
-                val useLightIcons = when {
-                    viewModel.titleTextColor != null -> viewModel.titleTextColor!!.luminance() < 0.5
-                    viewModel.topBarColor != null -> viewModel.topBarColor!!.luminance() < 0.5
-                    else -> viewModel.dynamicBackgroundTheme.value == DARK
-                }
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -920,7 +921,11 @@ open class LoginActivity : FragmentActivity() {
                     ?: return@evaluateJavascript
 
                 // Ensure Status Bar Icons are readable no matter which OS theme is used.
-                val useLightIcons = viewModel.dynamicBackgroundTheme.value == DARK
+                val useLightIcons = when {
+                    viewModel.titleTextColor != null -> viewModel.titleTextColor!!.luminance() < 0.5
```
JohnsonEricAtSalesforce, line 925: To avoid the force unwrap here, how about setting local variables from these two values? Then, the slight chance that a change to the properties from another thread wouldn't cause an exception to be thrown.

wmathurin:
@brandonpage this should go into 13.0.1 also ?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: Realistically, I expect apps to set a static color and never touch it again or never touch it at all.  But you are right, local vals are safer so I'll do that..

brandonpage:
> @brandonpage this should go into 13.0.1 also ?

I think so.  Let me know if you disagree.

wmathurin:
Sounds good - I'll cherry pick once you merge.

wmathurin (approved):

PR #2712, opened 2025-05-08
@W-18250527: Websockets client (Android)

JohnsonEricAtSalesforce:
🥁 _*Ready For Final Review*_ 🎸

  This reviews MSDK's current capabilities for web sockets use, in particular against the Agentforce Speech Foundations API endpoints, and adds support where needed to aid apps in accessing those and similarly styled web socket endpoints.

  We can certainly discuss what I've found here and see what like and what we want to change in order to best serve MSDK's apps.  This isn't far enough along to be set in stone 🙂

  The first takeaway is that OkHttp already provides web sockets support in MSDK which isn't very far below the surface for our apps to use.  An app can create a web socket from MSDK's `RestClient.okHttpClient.newWebSocket(...)`.  That web socket also already benefits from the `OAuthRefreshInterceptor` that is applied to `okHttpClient` to automatically refresh the access token and retry when needed.

  A second finding is that `RestClient` can be instantiated with a custom `authToken` and a null `authTokenProvider` which fulfills the last part of this work item such that the app can override the current user's access token with a value the app provides.  The null `authTokenProvider` disables the refresh and retry logic.

  So, that leaves the proposed changes in a few easy layers:
  1. A simple change to make creating a new web socket from `RestClient` more intuitive: `newWebSocket(request, listener)`.
  2. Entirely optional, but I did layer a `AgentforceSpeechFoundationsApiClient` which _specifically_ handles one ASF endpoint and its concerns separately from `RestClient`.  We can keep or discard this once we review this pull request.
  3. Temporarily, I also have a [pull request](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/451) in the templates repo which shows what the end result would look like when exercising the work item's requirements in detail.  Reading or running that might make the changes really come together.


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1336,11 +1336,11 @@ open class SalesforceSDKManager protected constructor(
         }
 
     private fun accessTokenExpiration(): String {
-        val currentUSer = userAccountManager.cachedCurrentUser
+        val currentUser = userAccountManager.cachedCurrentUser
```
JohnsonEricAtSalesforce, line 1339: This is an unrelated code clean-up I'll leave in (unless it drags down my CodeCov ☺️)

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AgentforceSpeechFoundationsApiClient.kt
```diff
@@ -0,0 +1,102 @@
...
+ * - `wss://api.salesforce.com/einstein/platform/v1/models/transcribeV1/streaming-transcriptions`
+ * - https://git.soma.salesforce.com/gist/rshahaf/3ee54892dfccd17fed5da2c304b5b743
+ *
+ * TODO: Replace the documentation link with the final documentation. ECJ20250507
+ *
+ * @param apiHostName The Salesforce `sfap_api` hostname
+ * @param modelName The model name to request from.
+ * @param restClient The REST client to use
+ */
+class AgentforceSpeechFoundationsApiClient(
```
JohnsonEricAtSalesforce, line 48: Similar to `SfapApiClient` and `NotificationsApiClient`, this takes a `RestClient` and specifically makes work with this specific ASF endpoint easier.

JohnsonEricAtSalesforce: We _don't_ have to keep this class for this work item.  I did want to show it at least temporarily in this pull request so we can all see how this level of the web socket implementation could look.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AgentforceSpeechFoundationsApiClient.kt
```diff
@@ -0,0 +1,102 @@
...
+                    mapOf(
+                        "engine" to "aws",
+                        "media-encoding" to "pcm"
+                    ).toQueryParameters()
+                }"
+            )
+            .addAuthorizedAsfHeaders(restClient)
+            .build()
+
+        return restClient.newWebSocket(
```
JohnsonEricAtSalesforce, line 76: This uses the new convenience method on `RestClient`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AgentforceSpeechFoundationsApiClient.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Adds ASF sfap_api required headers, included authorization, to this
+     * request builder.
+     * @param restClient The REST client to use for authorization
+     */
+    private fun Request.Builder.addAuthorizedAsfHeaders(
+        restClient: RestClient,
+    ) = this.apply {
+        header("Authorization", "Bearer ${restClient.authToken}")
```
JohnsonEricAtSalesforce, line 98: Note this would use the custom `authToken` depending on how `restClient` was instantiated.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -340,6 +342,19 @@ public Request buildRequest(RestRequest restRequest) {
         return builder.build();
     }
 
+    /**
+     * Returns a new web socket for the provided request and web socket listener.
+     * @param request The request
+     * @param listener The web socket listener
+     * @return The web socket
+     */
+    public WebSocket newWebSocket(
```
JohnsonEricAtSalesforce, line 351: Here is the crux change, for the moment, where we can make `newWebSocket` more apparent to our `RestClient` API.

JohnsonEricAtSalesforce:
Unit tests are coming once I see the CodeCov report.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: This typo gets to live on so my CodeCov will pass, so I reverted that change.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AgentforceSpeechFoundationsApiClient.kt
JohnsonEricAtSalesforce: Since `AgentforceSpeechFoundationsApiClient` is out-of-scope for this work item, I moved it to my [templates pull request](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/451) so yinzers can see how this all works and we'll have a copy for future reference.

wmathurin (commented):
LGTM :-)

JohnsonEricAtSalesforce:
Thanks for the directional confirmation, @wmathurin.  I'll mark this one ready for final review and look at your other comments in the template pull request.

wmathurin (approved):

PR #2714, opened 2025-05-14
Beacon support

wmathurin:
- Saving beacon child consumer key and secret in user account when returned in token end point response
- Using beacon child consumer key instead of configured consumer key if available
- Updated tests
- Tested manually with beacon app setup in test environment (na54)


brandonpage (approved):
🎉

PR #2715, opened 2025-05-15
@W-17449395: [Android] Welcome login email deeplinking support in MSDK 13.1

JohnsonEricAtSalesforce:
🧱 _*Draft: Only Ready For Preview Question & Answer*_ 🚧

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -104,6 +104,7 @@ public class OAuth2 {
     private static final String BIOMETRIC_AUTHENTICATION_TIMEOUT = "BIOMETRIC_AUTHENTICATION_TIMEOUT";
     private static final int BIOMETRIC_AUTHENTICATION_DEFAULT_TIMEOUT = 15;
     private static final String HYBRID_REFRESH = "hybrid_refresh";  // Grant Type Values
+    public static final String LOGIN_HINT = "login_hint";
     private static final String REFRESH_TOKEN = "refresh_token";  // Grant Type Values
     protected static final String RESPONSE_TYPE = "response_type";
     private static final String SCOPE = "scope";
@@ -244,6 +245,7 @@ public class OAuth2 {
      * @param callbackUrl                OAuth callback URL or redirect URL.
      * @param scopes                     A list of OAuth scopes to request (e.g. {"visualforce", "api"}). If null,
      *                                   the default OAuth scope is provided.
+     * @param loginHint                  When applicable, the Salesforce Welcome Login hint
      * @param displayType                OAuth display type. If null, the default of 'touch' is used.
      * @param codeChallenge              Code challenge to use when using web server flow
      * @param addlParams                 Any additional parameters that may be added to the request.
@@ -257,6 +259,7 @@ public class OAuth2 {
             String clientId,
             String callbackUrl,
             String[] scopes,
+            String loginHint,
             String displayType,
             String codeChallenge,
             Map<String,String> addlParams) {
@@ -270,6 +273,9 @@ public class OAuth2 {
         sb.append(AND).append(CLIENT_ID).append(EQUAL).append(Uri.encode(clientId));
         if (scopes != null && scopes.length > 0) {
             sb.append(AND).append(SCOPE).append(EQUAL).append(Uri.encode(computeScopeParameter(scopes)));
+        }
+        if (loginHint != null && !loginHint.isEmpty()) {
+            sb.append(AND).append(LOGIN_HINT).append(EQUAL).append(Uri.encode(loginHint));
         }
         sb.append(AND).append(REDIRECT_URI).append(EQUAL).append(callbackUrl);
         sb.append(AND).append(DEVICE_ID).append(EQUAL).append(SalesforceSDKManager.getInstance().getDeviceId());
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
@@ -113,6 +113,7 @@ internal class IDPAuthCodeHelper private constructor(
             spConfig.oauthClientId,
             spConfig.oauthCallbackUrl,
             spConfig.oauthScopes,
+            null,
             context.getString(R.string.oauth_display_type),
             codeChallenge,
             null
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -195,6 +195,12 @@ open class LoginActivity : FragmentActivity() {
             SalesforceSDKManager.getInstance().setViewNavigationVisibility(this)
         }
 
+        // Set the Salesforce Welcome Login hint and host for the OAuth authorize URL, if applicable.
+        viewModel.loginHint = intent.getStringExtra(EXTRA_KEY_LOGIN_HINT)
+        intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let {
+            viewModel.loginUrl.value = "https://$it"
+        }
+
         /*
          * For Salesforce Identity API UI Bridge support, the overriding
          * frontdoor bridge URL to use in place of the default initial login URL
@@ -1001,6 +1007,15 @@ open class LoginActivity : FragmentActivity() {
         internal const val ABOUT_BLANK = "about:blank"
         private const val BACKGROUND_COLOR_JAVASCRIPT =
             "(function() { return window.getComputedStyle(document.body, null).getPropertyValue('background-color'); })();"
+
+        // endregion
+        // region Log In With Login Hint Public Implementation
+
+        /** Intent extra key for login hint value */
+        const val EXTRA_KEY_LOGIN_HINT = "login_hint"
+
+        /** Intent extra key for login host to use with login hint */
+        const val EXTRA_KEY_LOGIN_HOST = "login_host"
 
         // endregion
         // region QR Code Login Via Salesforce Identity API UI Bridge Public Implementation
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/LoginViewModelTest.kt
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/LoginViewModelTest.kt
@@ -156,6 +156,7 @@ class LoginViewModelTest {
         bootConfig.remoteAccessConsumerKey,
         bootConfig.oauthRedirectURI,
         bootConfig.oauthScopes,
+        null,
         SalesforceSDKManager.getInstance().appContext.getString(oauth_display_type),
         codeChallenge,
         hashMapOf<String, String>()
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -129,7 +129,14 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
     internal var dynamicHeaderTextColor =
         derivedStateOf { if (dynamicBackgroundColor.value.luminance() > 0.5) Black else White }
     internal val defaultTitleText: String
-        get() = if (loginUrl.value == ABOUT_BLANK) "" else selectedServer.value ?: ""
+        get() {
+            val loginUrl = loginUrl.value
+            return when {
+                loginUrl == ABOUT_BLANK -> ""
+                loginUrl != null -> loginUrl
+                else -> selectedServer.value ?: ""
+            }
+        }
 
     /** Additional Auth Values used for login. */
     open var additionalParameters = hashMapOf<String, String>()
@@ -162,6 +169,9 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
     // The default, locally generated code verifier
     @VisibleForTesting
     internal var codeVerifier: String? = null
+
+    /** The Salesforce Welcome Login hint parameter value for the OAuth authorize endpoint */
+    internal var loginHint: String? = null
 
     // Auth code we receive from the JWT swap for magic links.
     internal var authCodeForJwtFlow: String? = null
@@ -313,10 +323,11 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         val authorizationUrl = OAuth2.getAuthorizationUrl(
             SalesforceSDKManager.getInstance().useWebServerAuthentication,
             SalesforceSDKManager.getInstance().useHybridAuthentication,
-            URI(server),
+            loginUrl.value?.let { URI(it) } ?: URI(server),
             clientId,
             bootConfig.oauthRedirectURI,
             bootConfig.oauthScopes,
+            loginHint,
             authorizationDisplayType,
             codeChallenge,
             additionalParams
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/OAuth2Test.java
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/OAuth2Test.java
@@ -94,14 +94,14 @@ public class OAuth2Test {
 	public void testGetAuthorizationUrl() throws URISyntaxException {
 		String callbackUrl = "sfdc://callback";
 		URI authorizationUrl = OAuth2.getAuthorizationUrl(true, true, new URI(TestCredentials.LOGIN_URL),
-                TestCredentials.CLIENT_ID, callbackUrl, null, null, "some-challenge", null);
+                TestCredentials.CLIENT_ID, callbackUrl, null, null, null, "some-challenge", null);
 		URI expectedAuthorizationUrl = new URI(TestCredentials.LOGIN_URL +
                 "/services/oauth2/authorize?display=touch&response_type=code&client_id=" +
                 TestCredentials.CLIENT_ID + "&redirect_uri=" + callbackUrl + "&device_id=" +
 				SalesforceSDKManager.getInstance().getDeviceId() + "&code_challenge=some-challenge");
         Assert.assertEquals("Wrong authorization url", expectedAuthorizationUrl, authorizationUrl);
 		authorizationUrl = OAuth2.getAuthorizationUrl(true, true, new URI(TestCredentials.LOGIN_URL),
-                TestCredentials.CLIENT_ID, callbackUrl, null, "touch", "some-challenge", null);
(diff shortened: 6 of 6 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -195,6 +195,12 @@ open class LoginActivity : FragmentActivity() {
             SalesforceSDKManager.getInstance().setViewNavigationVisibility(this)
         }
 
+        // Set the Salesforce Welcome Login hint and host for the OAuth authorize URL, if applicable.
+        viewModel.loginHint = intent.getStringExtra(EXTRA_KEY_LOGIN_HINT)
+        intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let {
+            viewModel.loginUrl.value = "https://$it"
+        }
+
```
brandonpage, line 203: Just a thought:  It might end up being helpful down the line to add welcome to the server list.  I believe we record the server used to login when we create the account.

wmathurin: Also we should add welcome.salesforce.com to the default list of servers (right now we have login.salesforce.com and test.salesforce.com). That could probably go as part of this PR.

wmathurin: I meant in this file: https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/res/xml/servers.xml

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
```diff
@@ -113,6 +113,7 @@ internal class IDPAuthCodeHelper private constructor(
             spConfig.oauthClientId,
             spConfig.oauthCallbackUrl,
             spConfig.oauthScopes,
+            null,
```
JohnsonEricAtSalesforce, line 116: @wmathurin or @brandonpage - Is there any case where this call would have the login hint value?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -195,6 +195,12 @@ open class LoginActivity : FragmentActivity() {
             SalesforceSDKManager.getInstance().setViewNavigationVisibility(this)
         }
 
+        // Set the Salesforce Welcome Login hint and host for the OAuth authorize URL, if applicable.
+        viewModel.loginHint = intent.getStringExtra(EXTRA_KEY_LOGIN_HINT)
+        intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let {
+            viewModel.loginUrl.value = "https://$it"
```
JohnsonEricAtSalesforce, line 201: @Crebs - On the iOS side, does the app just provide the host, such as `welcome.salesforce.com` or does that string _also_ include the protocol?  Is it safe to code this to `https` if that's not part of the "login host" parameter value?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/LoginViewModelTest.kt
```diff
@@ -74,79 +80,105 @@ class LoginViewModelTest {
...
+        assertNotEquals(FAKE_SERVER_URL, viewModel.selectedServer.value)
 
         loginServerManager.addCustomLoginServer("fake", FAKE_SERVER_URL)
-        Assert.assertEquals(FAKE_SERVER_URL, viewModel.selectedServer.value)
+        assertEquals(FAKE_SERVER_URL, viewModel.selectedServer.value)
+    }
+
+    @Test
+    fun defaultTitleText() {
+        val regex = "https://login.salesforce.com/services/oauth2/authorize\\?display=touch&response_type=code&client_id=__CONSUMER_KEY__&scope=api%20openid%20refresh_token%20web&redirect_uri=__REDIRECT_URI__&device_id=[^=]+&code_challenge=[^=]+".toRegex()
```
JohnsonEricAtSalesforce, line 92: I made a regex out of the default value the view model already has on `dev` and abstracted values I thought would change or make the test brittle.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce, line 201: Ah, hold the press here.  I just realized I'm not using the view model properties correctly 💣 This should probably be setting `selectedServer` instead.  I'll get that fixed up!

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -271,6 +274,9 @@ public static URI getAuthorizationUrl(
         if (scopes != null && scopes.length > 0) {
             sb.append(AND).append(SCOPE).append(EQUAL).append(Uri.encode(computeScopeParameter(scopes)));
         }
+        if (loginHint != null && !loginHint.isEmpty()) {
```
wmathurin, line 277: NIT could be done with
```java
if (!TextUtils.isEmpty(loginHint)) {
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
wmathurin: Login hint won't matter here because we load the approval URL through front door using the access token of the IDP so, the page where one would enter username/password is never shown, it goes straight to the allow/deny screen if applicable or most times it goes straight to the redirect URL and the user does not need to do anything.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
JohnsonEricAtSalesforce: 🤦🏼‍♂️

wmathurin (approved):
LGTM!

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTests.kt
```diff
@@ -0,0 +1,205 @@
...
...s://mobilesdk.my.salesforce.com/services/data/v{version}/search/\",\"query\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/query/\",\"recent\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/recent/\",\"tooling_soap\":\"https://mobilesdk.my.salesforce.com/services/Soap/T/{version}/00DB0000000ToZ3\",\"tooling_rest\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/tooling/\",\"profile\":\"https://mobilesdk.my.salesforce.com/005B0000009mWwmIAE\",\"feeds\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/chatter/feeds\",\"groups\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/chatter/groups\",\"users\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/chatter/users\",\"feed_items\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/chatter/feed-items\",\"feed_elements\":\"https://mobilesdk.my.salesforce.com/services/data/v{version}/chatter/feed-elements\",\"custom_domain\":\"https://mobilesdk.my.salesforce.com\"},\"active\":true,\"user_type\":\"STANDARD\",\"language\":\"en_US\",\"locale\":\"en_US\",\"utcOffset\":-28800000,\"last_modified_date\":\"2025-02-28T18:14:06Z\"}".toByteArray()
```
wmathurin, line 38: Maybe don't put real org ids or user ids in there so that some code scan does not flag that code?

JohnsonEricAtSalesforce: Agreed.  That was on my to-do list, though I should have really done that before this commit even.

By the by, that org id appears in several places already on `dev`.  Would you want me to see if I can place dummy data in that test?

👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/f1b24c45366df2f695fbbc289df456336ba64faa/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/OAuth2Test.java#L378

👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/f1b24c45366df2f695fbbc289df456336ba64faa/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/OAuth2Test.java#L435

wmathurin: Maybe they only scan for credentials since those ids have not be flagged. For peace of mind and unless you really need org/user ids, I would still edit them out.

JohnsonEricAtSalesforce: I went crazy and really cleaned that up in https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2715/commits/541f8788177c183581c4378f1bccd438040c01bf

JohnsonEricAtSalesforce:
Sorry to add a functional commit after an approval, but I was self-reviewing and realized three things that should be better about this pull request:
1. The tests on `SalesforceSDKManager` weren't testing the correct outcome, which should be the application of `SalesforceSDKManager` properties and not the discarded authentication configuration object or the seemingly unneeded callback (Does anyone know why the callback exists in `fetchAuthenticationConfiguration`?)
2. The test suite really needed some cleanup to match our style in other tests such as `LoginViewModelTest`.
3. Finally, some of the tests didn't make sense were we don't realistically expect a thrown exception or where a null value doesn't make sense.  I'll add some on-line comments so we can look at those.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -339,11 +342,13 @@ open class SalesforceSDKManager protected constructor(
      */
     @set:Synchronized
     open var isBrowserLoginEnabled = false
-        protected set
+        @VisibleForTesting(otherwise = PROTECTED)
```
JohnsonEricAtSalesforce, line 345: I opened this up for testing on `fetchAuthenticationConfiguration`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -339,11 +342,13 @@ open class SalesforceSDKManager protected constructor(
...
+        set
 
     /** Optionally enables browser session sharing */
     var isShareBrowserSessionEnabled = false
-        private set
+        @VisibleForTesting
```
JohnsonEricAtSalesforce, line 350: I opened this up for testing on `fetchAuthenticationConfiguration`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1900,36 +1905,36 @@ open class SalesforceSDKManager protected constructor(
...
      *
+     * @param httpAccess The HTTP access to use for API integration.  Defaults
+     * to null to use the default HTTP access.  This parameter is intended for
+     * testing purposes only and should not be used in release builds.
      * @param completion An optional function to invoke at the end of the action
      */
-    fun fetchAuthenticationConfiguration(
-        completion: (() -> Unit)? = null
+    internal fun fetchAuthenticationConfiguration(
+        httpAccess: HttpAccess? = null,
```
JohnsonEricAtSalesforce, line 1914: I didn't call it out explicitly earlier, but there is this new and optional parameter so we can pass a mock for testing.  That's really similar to where we've been adding mock REST client parameters for push notifications lately.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1900,36 +1906,36 @@ open class SalesforceSDKManager protected constructor(
...
-                        shareBrowserSessionEnabled = authConfig?.isShareBrowserSessionEnabled ?: false
-                    )
-                }
+            getMyDomainAuthConfig(httpAccess, loginServer).let { authConfig ->
+                setBrowserLoginEnabled(
+                    browserLoginEnabled = authConfig?.isBrowserLoginEnabled ?: false,
+                    shareBrowserSessionEnabled = authConfig?.isShareBrowserSessionEnabled ?: false
+                )
             }
-        }.onFailure { e ->
```
JohnsonEricAtSalesforce, line 1929: Maybe this is a style topic to discuss.  There's no way to reasonably mock or even expect an exception here once we switch to `withTimeoutOrNull`.  We _can_ test with a mock that `getMyDomainAuthConfig` catches and handles its exception from the `OkHttpClient`.  So, I was considering that, in general, error logic which is "unreachable" can be removed.

If there were top-level app code calling this that really needed a guard on truly unexpected exceptions, perhaps that belongs in the app code and not throughout this library where it isn't testable?  Thoughts?  I know some would place a try-catch/runCatching around the body of every method and others never would.  What's a suitable middle ground or approach for this specific method?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1900,36 +1906,36 @@ open class SalesforceSDKManager protected constructor(
...
-            e(TAG, "Exception occurred while fetching authentication configuration", e)
         }
 
-        completion?.invoke()
+        completion.invoke()
```
JohnsonEricAtSalesforce, line 1939: I'm not sure why the completion parameter exists, but since this is an internal method would it be OK to make it required and skip a test on its nullability?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1900,36 +1906,36 @@ open class SalesforceSDKManager protected constructor(
...
      * Fetches the authentication configuration, if required.
      *
+     * @param httpAccess The HTTP access to use for API integration.  Defaults
+     * to null to use the default HTTP access.  This parameter is intended for
+     * testing purposes only and should not be used in release builds.
      * @param completion An optional function to invoke at the end of the action
      */
-    fun fetchAuthenticationConfiguration(
-        completion: (() -> Unit)? = null
+    internal fun fetchAuthenticationConfiguration(
```
JohnsonEricAtSalesforce, line 1914: Can this be an internal method? I couldn't find references to it in our public docs.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -154,10 +155,7 @@ public LoginServer getSelectedLoginServer() {
 	 *
 	 * @param server LoginServer instance.
 	 */
-	public void setSelectedLoginServer(LoginServer server) {
-		if (server == null) {
-			return;
-		}
+	public void setSelectedLoginServer(@NonNull LoginServer server) {
```
JohnsonEricAtSalesforce, line 158: Would it be acceptable to make this a required value and skip any testing around its nullability?

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -457,7 +455,7 @@ public static class LoginServer {
 		 * @param url Server URL.
 		 * @param isCustom True - if custom URL, False - otherwise.
 		 */
-		public LoginServer(String name, String url, boolean isCustom) {
+		public LoginServer(@NonNull String name, @NonNull String url, boolean isCustom) {
```
JohnsonEricAtSalesforce, line 458: Would it be acceptable to make these required values and skip any testing around their nullability?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTests.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    private val responseBody = mockk<ResponseBody>().apply {
```
JohnsonEricAtSalesforce, line 33: I moved all the mocks into members since earlier commits.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTests.kt
```diff
@@ -0,0 +1,237 @@
...
+    private val okHttpClient = mockk<OkHttpClient>().apply {
+        every { newCall(any()) } returns this@SalesforceSDKManagerTests.call
+    }
+
+    private val httpAccess = mockk<HttpAccess>().apply {
+        every { getOkHttpClient() } returns this@SalesforceSDKManagerTests.okHttpClient
+    }
+
+    @Test
+    fun salesforceSdkManager_Updates_onFetchAuthenticationConfigurationForMyDomainLoginServer() {
```
JohnsonEricAtSalesforce, line 57: I also separated one big test into smaller tests following our naming conventions from `LoginViewModelTest`.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTests.kt
```diff
@@ -0,0 +1,237 @@
...
+
+        runBlocking {
+            SalesforceSDKManager.getInstance().fetchAuthenticationConfiguration(
+                httpAccess = httpAccess,
+            ) {
+                /* Completion Does Not Require Verification */
+            }.join()
+        }
+
+        assertFalse(SalesforceSDKManager.getInstance().isBrowserLoginEnabled)
```
JohnsonEricAtSalesforce, line 78: The tests really follow these two properties now since I believe that's the actual outcome of `fetchAuthenticationConfiguration`.

JohnsonEricAtSalesforce (commented):
Here are a few additional comments to guide us through today's changes.

JohnsonEricAtSalesforce:
It's worth noting I'm still watching the test run in CI and CodeCov to see what the improvements look like.  I'll check back once I review that.

JohnsonEricAtSalesforce:
> It's worth noting I'm still watching the test run in CI and CodeCov to see what the improvements look like. I'll check back once I review that.

Though it was a bit tedious, I got the CodeCov reasonably maxed out (`91.89%`!) and really vetted the tests made sense along with a few mild improvements to the code in SalesforceSDKManager.  😅. I also preemptively fixed a bug I'd introduced in the login view title in some cases.

No more changes coming, so if anyone wants a final chance to comment here you go.  Thanks for your patience!

PR #2731, opened 2025-06-26
@W-18777804: [qr code scan][android] we should prevent a successful login of client A by scanning QR code of client B

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This adds a simple check into the `onCreate` of `LoginActivity` so that a UI Bridge API front door URL with the consumer key (client id) of another CA/ECA will not attempt log in.  The mismatching front door URL will be discarded, a toast shown and the app's current set login server will be used.  This is just as if the incorrect QR code hadn't been scanned, so the user is still in a valid state.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -208,7 +208,6 @@ open class LoginActivity : FragmentActivity() {
          * plus the optional web server flow code verifier accompanying the
          * frontdoor bridge URL.
          */
-        viewModel.isUsingFrontDoorBridge = isFrontdoorBridgeUrlIntent(intent) || isQrCodeLoginUrlIntent(intent)
         val uiBridgeApiParameters = if (isQrCodeLoginUrlIntent(intent)) {
             uiBridgeApiParametersFromQrCodeLoginUrl(intent.data?.toString())
         } else intent.getStringExtra(EXTRA_KEY_FRONTDOOR_BRIDGE_URL)?.let { frontdoorBridgeUrl ->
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -208,7 +208,6 @@ open class LoginActivity : FragmentActivity() {
          * plus the optional web server flow code verifier accompanying the
          * frontdoor bridge URL.
          */
-        viewModel.isUsingFrontDoorBridge = isFrontdoorBridgeUrlIntent(intent) || isQrCodeLoginUrlIntent(intent)
```
JohnsonEricAtSalesforce, line 211: This trivially moved a few lines down where there's more data to use.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -218,6 +217,29 @@ open class LoginActivity : FragmentActivity() {
             )
         }
 
+        /**
+         *  The Salesforce Connected App or External Client App consumer key
+         *  from the Salesforce Identity API UI Bridge front door URL.  This
+         *  is sometimes known as "client id" or "remote access consumer
+         *  key".
+         */
+        val uiBridgeApiParametersConsumerKey = uiBridgeApiParameters?.frontdoorBridgeUrl?.toUri()?.getQueryParameter("startURL")?.toUri()?.getQueryParameter("client_id")
```
JohnsonEricAtSalesforce, line 227: @wmathurin - This is the consumer key extraction I believe we were chatting about one day, maybe at Daily?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -218,6 +217,29 @@ open class LoginActivity : FragmentActivity() {
...
+
+        // Choose front door bridge use by verifying intent data and such that only front door bridge URLs with matching consumer keys are used.
+        val uiBridgeApiParametersFrontDoorBridgeUrlMatchesConsumerKey = uiBridgeApiParametersConsumerKey == viewModel.bootConfig.remoteAccessConsumerKey
+        viewModel.isUsingFrontDoorBridge = (isFrontdoorBridgeUrlIntent(intent) || isQrCodeLoginUrlIntent(intent)) && uiBridgeApiParametersFrontDoorBridgeUrlMatchesConsumerKey
```
JohnsonEricAtSalesforce: The new check here is on `uiBridgeApiParametersFrontDoorBridgeUrlMatchesConsumerKey`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -218,6 +217,29 @@ open class LoginActivity : FragmentActivity() {
...
+
+        // Alert the user if the front door bridge URL is not for this app and was discarded.
+        if (viewModel.bootConfig.remoteAccessConsumerKey != null && !uiBridgeApiParametersFrontDoorBridgeUrlMatchesConsumerKey) {
+            runOnUiThread {
+                makeText(
+                    this,
+                    "Cannot use another app's login QR Code.  Please log in to this app.",
```
JohnsonEricAtSalesforce: How's this for a default toast text? We can open this up in the view model later if desired.

wmathurin: Can we pull the error from a resource file so it can be localized or customized by apps?

JohnsonEricAtSalesforce: @brandonpage - Should we use a string resource or a view model property?  How does that fit with the other view model customizables you made?  Since this isn't a dynamic value at times, like `titleText`, does it make more sense as a traditional string resource?

PR #2733, opened 2025-06-27
Revert 13.0 Advanced Auth behavior when useWebServerAuthentication is false.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -144,6 +144,12 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
     protected open val authorizationDisplayType =
         SalesforceSDKManager.getInstance().appContext.getString(oauth_display_type)
 
+    internal val useWebServerFlow: Boolean
+        get() = with(SalesforceSDKManager.getInstance()) {
+            (useWebServerAuthentication || isBrowserLoginEnabled)
+                    && !(isUsingFrontDoorBridge && frontdoorBridgeCodeVerifier == null)
```
wmathurin, line 152: I believe that stuff is for QR code login without code verifier (and therefore requiring user agent flow).
But we will probably forget those details.
Could we add some comments for future generations ;-)

PR #2736, opened 2025-07-18
@W-17964655: [Android] WSC discovery endpoint to postpone OAuth flow

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  `LoginActivity` is updated to support selecting Salesforce Welcome Discovery URLs plus responding to their callback URL to initiate log in.

  During the discussion in Slack with @wmathurin and @Crebs, we leaned towards having WSC behave and look as much like the existing default login activity as possible.  In particular, app bar and overflow menu are identical.  With that in mind, it was really easy to fold the new logic into our existing activity.

  Here's a video that shows the new login/welcome flow in action.  This updated version of the video today shows the more polished version.

[Screen_recording_20250721_140629.webm](https://github.com/user-attachments/assets/f8bacb22-9a85-4e3b-a2bb-30c7b9107bd2)

### DIFF
--- a/libs/SalesforceSDK/res/xml/servers.xml
+++ b/libs/SalesforceSDK/res/xml/servers.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <servers>
     <server name="Production" url="https://login.salesforce.com" />
-    <server name="Welcome" url="https://welcome.salesforce.com" />
+    <server name="Welcome" url="https://welcome.salesforce.com/discovery" />
     <server name="Sandbox" url="https://test.salesforce.com" />
 </servers>
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -30,6 +30,7 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
 import android.content.res.XmlResourceParser;
+import android.os.Looper;
 
 import androidx.annotation.NonNull;
 import androidx.lifecycle.MutableLiveData;
@@ -167,7 +168,11 @@ public class LoginServerManager {
 		edit.putString(SERVER_URL, server.url);
 		edit.putBoolean(IS_CUSTOM, server.isCustom);
 		edit.apply();
-		selectedServer.postValue(server);
+		if (Looper.myLooper() == Looper.getMainLooper()) {
+			selectedServer.setValue(server);
+		} else {
+			selectedServer.postValue(server);
+		}
 	}
 
 	/**
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -61,7 +61,7 @@ public class LoginServerManager {
 
 	// Default login servers.
 	public static final String PRODUCTION_LOGIN_URL = "https://login.salesforce.com";
-	public static final String WELCOME_LOGIN_URL = "https://welcome.salesforce.com";
+	public static final String WELCOME_LOGIN_URL = "https://welcome.salesforce.com/discovery";
 	public static final String SANDBOX_LOGIN_URL = "https://test.salesforce.com";
 
 	// Keys used in shared preferences.
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -922,9 +922,6 @@ open class LoginActivity : FragmentActivity() {
         intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let { loginHost ->
             val loginUrl = "https://$loginHost"
             loginServerManager.addCustomLoginServer(loginHost, loginUrl)
-            loginServerManager.setSelectedLoginServer(
-                LoginServer(loginHost, loginUrl, true)
-            )
         }
     }
 
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -34,11 +34,15 @@ import android.accounts.AccountManager.KEY_ACCOUNT_AUTHENTICATOR_RESPONSE
 import android.annotation.SuppressLint
 import android.app.Activity
 import android.app.admin.DevicePolicyManager.ACTION_SET_NEW_PASSWORD
+import android.content.Context
 import android.content.Intent
+import android.content.Intent.FLAG_ACTIVITY_CLEAR_TASK
+import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.content.pm.PackageManager.FEATURE_FACE
 import android.content.pm.PackageManager.FEATURE_IRIS
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory.decodeResource
+import android.net.Uri
 import android.net.http.SslError
 import android.net.http.SslError.SSL_EXPIRED
 import android.net.http.SslError.SSL_IDMISMATCH
@@ -130,6 +134,7 @@ import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse
 import com.salesforce.androidsdk.auth.OAuth2.swapJWTForTokens
 import com.salesforce.androidsdk.auth.idp.interfaces.SPManager.Status
 import com.salesforce.androidsdk.auth.idp.interfaces.SPManager.StatusUpdateCallback
+import com.salesforce.androidsdk.config.BootConfig.getBootConfig
 import com.salesforce.androidsdk.config.LoginServerManager.LoginServer
 import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey.ManagedAppCertAlias
 import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey.RequireCertAuth
@@ -151,11 +156,20 @@ import kotlinx.coroutines.launch
 import org.json.JSONObject
 import java.net.URI
 import java.net.URLDecoder
+import java.net.URLEncoder
 import java.security.PrivateKey
 import java.security.cert.X509Certificate
 
 /**
  * Login activity authenticates a user. Authorization happens inside a web view.
+ *
+ * Support for Salesforce Welcome (WSC) Discovery is provided.  If the activity
+ * is started with a valid WSC Discovery URL in the intent data, the web view
+ * will display WSC Discovery.  The activity will return the user to default
+ * log in with the selected login hint and My Domain login server after
+ * completing the WSC Discovery flow.  In addition to WSC Discovery, the
+ * activity may be started using intent extras for login hint and My Domain
+ * login server.  See the extra key constants provided by the activity.
  *
  * Once an authorization code is obtained, it is exchanged for access and
  * refresh tokens to create an account via the account manager which stores
@@ -180,6 +194,7 @@ open class LoginActivity : FragmentActivity() {
             webViewClient = this@LoginActivity.webViewClient
             webChromeClient = this@LoginActivity.webChromeClient
             setBackgroundColor(Color.Transparent.toArgb())
+            settings.domStorageEnabled = true /* Salesforce Welcome Discovery requires this */
             settings.javaScriptEnabled = true
         }
 
@@ -202,8 +217,13 @@ open class LoginActivity : FragmentActivity() {
             SalesforceSDKManager.getInstance().setViewNavigationVisibility(this)
         }
 
-        // Set the Salesforce Welcome Login hint and host for the OAuth authorize URL, if applicable.
-        useLoginHint(intent)
+        // Apply the intent extras' Salesforce Welcome Login hint and host for use in the OAuth authorize URL, if applicable.
+        applySalesforceWelcomeLoginHintAndHost(intent)
+
+        // Apply the URL as the initial login URL if it's a valid Salesforce Welcome Discovery URL.
+        intent.data?.let { uri ->
+            useSalesforceWelcomeDiscoveryMobileUrl(uri)
+        }
 
         /*
          * For Salesforce Identity API UI Bridge support, the overriding
@@ -220,7 +240,7 @@ open class LoginActivity : FragmentActivity() {
             )
         }
 
-        /**
+        /*
          *  The Salesforce Connected App or External Client App consumer key
          *  from the Salesforce Identity API UI Bridge front door URL.  This
          *  is sometimes known as "client id" or "remote access consumer
@@ -308,8 +328,20 @@ open class LoginActivity : FragmentActivity() {
             }
         }
 
-        // Take action on server change.
-        viewModel.selectedServer.observe(this) {
+        // Take action on selected server change.
+        viewModel.selectedServer.observe(this) { selectedServer ->
+
+            // Guard against observing a selected server already provided by the intent data, such as a Salesforce Welcome Discovery mobile URL.
+            val selectedServerUri = selectedServer.toUri()
+            if (intent.data?.host == selectedServerUri.host) {
+                return@observe
+            }
+
+            // Use the URL to switch between default or Salesforce Welcome Discovery log in, if applicable.
+            if (switchDefaultOrSalesforceWelcomeDiscoveryLogin(selectedServerUri)) {
+                return@observe
+            }
+
             if (viewModel.singleServerCustomTabActivity) {
                 // Skip fetching authorization and show custom tab immediately.
                 viewModel.reloadWebView()
@@ -474,27 +506,6 @@ open class LoginActivity : FragmentActivity() {
     }
 
     // endregion
-    // region Salesforce Welcome Login Private Implementation
-
-    /**
-     * Uses the Salesforce Welcome login hint and host in the Intent, if
-     * applicable.
-     */
-    private fun useLoginHint(intent: Intent) {
-
-        viewModel.loginHint = intent.getStringExtra(EXTRA_KEY_LOGIN_HINT)
-        intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let { loginHost ->
-            SalesforceSDKManager.getInstance().loginServerManager.setSelectedLoginServer(
-                LoginServer(
-                    loginHost,
-                    "https://$loginHost",
-                    true
-                )
-            )
-        }
-    }
-
-    // endregion
     // End of Public Functions
 
     protected open fun certAuthOrLogin() {
@@ -567,7 +578,7 @@ open class LoginActivity : FragmentActivity() {
         // Create account and save result before switching to new user
         accountAuthenticatorResult = SalesforceSDKManager.getInstance().userAccountManager.createAccount(userAccount)
 
-        setResult(Activity.RESULT_OK)
+        setResult(RESULT_OK)
         finish()
     }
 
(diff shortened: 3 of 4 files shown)


libs/SalesforceSDK/res/xml/servers.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <servers>
     <server name="Production" url="https://login.salesforce.com" />
-    <server name="Welcome" url="https://welcome.salesforce.com" />
+    <server name="Welcome" url="https://welcome.salesforce.com/discovery" />
```
JohnsonEricAtSalesforce, line 4: @wmathurin - I've set up the MSDK logic to distinguish Salesforce Welcome Discovery URLs by the path element.  According to our convention, that _works_ and doesn't require us to add an UX for the user to specify if they're entering discovery URL in the "Add Connection" bottom sheet.

Alternately, we could add some way for the user to check-off that the scheme and host they've entered is for discovery and then add this path in the MSDK logic.  That seemed more intrusive to the user, though.

JohnsonEricAtSalesforce: I'm open to other ideas.  This seemed to keep it simple, at first blush.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -151,12 +156,21 @@ import kotlinx.coroutines.launch
...
 import java.net.URI
 import java.net.URLDecoder
+import java.net.URLEncoder
 import java.security.PrivateKey
 import java.security.cert.X509Certificate
 
 /**
  * Login activity authenticates a user. Authorization happens inside a web view.
  *
+ * Support for Salesforce Welcome (WSC) Discovery is provided.  If the activity
```
JohnsonEricAtSalesforce, line 164: Here's some light code documentation for login activity's welcome discovery support.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -180,6 +194,7 @@ open class LoginActivity : FragmentActivity() {
             webViewClient = this@LoginActivity.webViewClient
             webChromeClient = this@LoginActivity.webChromeClient
             setBackgroundColor(Color.Transparent.toArgb())
+            settings.domStorageEnabled = true /* Salesforce Welcome Discovery requires this */
```
JohnsonEricAtSalesforce, line 195: We knew this was coming back based on earlier support discussions.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -202,8 +217,13 @@ open class LoginActivity : FragmentActivity() {
             SalesforceSDKManager.getInstance().setViewNavigationVisibility(this)
         }
 
-        // Set the Salesforce Welcome Login hint and host for the OAuth authorize URL, if applicable.
-        useLoginHint(intent)
+        // Apply the intent extras' Salesforce Welcome Login hint and host for use in the OAuth authorize URL, if applicable.
+        applySalesforceWelcomeLoginHintAndHost(intent)
```
JohnsonEricAtSalesforce: I worked hard to think of login activity's use cases and intuitive semantics in naming conventions.  That way I could really extract Salesforce Welcome Discovery into readable method calls that kept that logic separate from the existing logic branches, which are more in-line with each other.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -308,8 +328,20 @@ open class LoginActivity : FragmentActivity() {
...
+        viewModel.selectedServer.observe(this) { selectedServer ->
+
+            // Guard against observing a selected server already provided by the intent data, such as a Salesforce Welcome Discovery mobile URL.
+            val selectedServerUri = selectedServer.toUri()
+            if (intent.data?.host == selectedServerUri.host) {
+                return@observe
+            }
+
+            // Use the URL to switch between default or Salesforce Welcome Discovery log in, if applicable.
+            if (switchDefaultOrSalesforceWelcomeDiscoveryLogin(selectedServerUri)) {
```
JohnsonEricAtSalesforce, line 334: Here's another location where I worked hard to extract the WSC logic from existing logic.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -473,27 +505,6 @@ open class LoginActivity : FragmentActivity() {
...
     }
 
-    // endregion
-    // region Salesforce Welcome Login Private Implementation
-
-    /**
-     * Uses the Salesforce Welcome login hint and host in the Intent, if
-     * applicable.
-     */
-    private fun useLoginHint(intent: Intent) {
```
JohnsonEricAtSalesforce, line 483: This moved into a region with other WSC code that made more sense with the existing region markers in this source.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -881,6 +892,150 @@ open class LoginActivity : FragmentActivity() {
         )
     }
 
+    // region Salesforce Welcome Login Private Implementation
```
JohnsonEricAtSalesforce, line 891: Each of the new WSC specific methods in this region should hopefully be self-explanatory with their source commentary as well.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce, line 334: Usually, this is step #1 where the user selects welcome from the select server or has entered it in add connection.  This'll stop the default login activity and start an instance for welcome just as if the app had started an instance in response to an external link.  I like that the activity never "switches" between default or welcome log in modes except by finishing and starting again.  That keeps the runtime testing really simple.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -202,8 +217,13 @@ open class LoginActivity : FragmentActivity() {
...
         }
 
-        // Set the Salesforce Welcome Login hint and host for the OAuth authorize URL, if applicable.
-        useLoginHint(intent)
+        // Apply the intent extras' Salesforce Welcome Login hint and host for use in the OAuth authorize URL, if applicable.
+        applySalesforceWelcomeLoginHintAndHost(intent)
+
+        // Apply the URL as the initial login URL if it's a valid Salesforce Welcome Discovery URL.
+        intent.data?.let { uri ->
+            useSalesforceWelcomeDiscoveryMobileUrl(uri)
```
JohnsonEricAtSalesforce: This is step #2 where the activity detects a welcome URL and applies it instead of default login.  This is something like what we did for front door URL, but the code is more concise this time.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -890,6 +1045,12 @@ open class LoginActivity : FragmentActivity() {
      */
     open inner class AuthWebViewClient : WebViewClient() {
         override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
+
+            // Use the request's Salesforce Welcome Discovery mobile callback URL, if applicable.
+            if (useSalesforceWelcomeDiscoveryMobileCallbackUrlForDefaultLogin(request.url)) {
```
JohnsonEricAtSalesforce, line 1053: Step #3 is detecting welcome's mobile callback after an environment selection and using the login hint to start default login with the My Domain host.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: Final step #4 is using login hint and host in default log in after the user completes the welcome flow.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -881,6 +892,150 @@ open class LoginActivity : FragmentActivity() {
...
+     * Salesforce Welcome Discovery back to default log in.
+     * @param proposedSelectedServerUrl The proposed selected server URL
+     * @return Boolean true if the provided proposed selected server URL is a
+     * switch from Salesforce Welcome Discovery back to the default log in,
+     * false otherwise.
+     * */
+    private fun isSwitchFromSalesforceWelcomeDiscoveryToDefaultLogin(
+        proposedSelectedServerUrl: Uri
+    ) = viewModel.loginUrl.value?.toUri()?.let {
+        isSalesforceWelcomeDiscoveryMobileCallbackUrl(proposedSelectedServerUrl)
```
wmathurin: Shouldn't that be:
```kotlin
private fun isSwitchFromSalesforceWelcomeDiscoveryToDefaultLogin(
        proposedSelectedServerUrl: Uri
    ) = viewModel.loginUrl.value?.toUri()?.let {
    isSalesforceWelcomeDiscoveryMobileUrl(it)
} ?: false
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -881,6 +892,150 @@ open class LoginActivity : FragmentActivity() {
...
+        val loginServerManager = SalesforceSDKManager.getInstance().loginServerManager
+
+        viewModel.loginHint = intent.getStringExtra(EXTRA_KEY_LOGIN_HINT)
+
+        intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let { loginHost ->
+            val loginUrl = "https://$loginHost"
+            loginServerManager.setSelectedLoginServer(
+                LoginServer(loginHost, loginUrl, true)
+            )
+            loginServerManager.addCustomLoginServer(loginHost, loginUrl)
```
wmathurin, line 923: Should it be added before being selected?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: You are exactly correct and I'd been looking for that while I was testing and seeing something a little off.  It's nice to know someone (if not me 🥴) followed my naming convention and code comments so well. Thanks!

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: Good question.  It works in this order.  That's probably because the user doesn't return to the picker until after both updates are done.  I'll try switching it just to see how it goes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: @wmathurin - Here's the commit with the fix. https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/commits/f36cbd86eda44d7423ea540b4faedcc552a083d8

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: I updated the order in https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/commits/f36cbd86eda44d7423ea540b4faedcc552a083d8

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -167,7 +168,11 @@ public void setSelectedLoginServer(LoginServer server) {
 		edit.putString(SERVER_URL, server.url);
 		edit.putBoolean(IS_CUSTOM, server.isCustom);
 		edit.apply();
-		selectedServer.postValue(server);
+		if (Looper.myLooper() == Looper.getMainLooper()) {
+			selectedServer.setValue(server);
```
JohnsonEricAtSalesforce, line 172: This optimization uses the synchronous `setValue` on the main thread, when applicable, and falls back to the asynchronous `postValue` when off main.  In `LoginActivity`, this makes the UX smoother since we're already on the main thread.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -881,6 +893,150 @@ open class LoginActivity : FragmentActivity() {
...
+        val loginServerManager = SalesforceSDKManager.getInstance().loginServerManager
+
+        viewModel.loginHint = intent.getStringExtra(EXTRA_KEY_LOGIN_HINT)
+
+        intent.getStringExtra(EXTRA_KEY_LOGIN_HOST)?.let { loginHost ->
+            val loginUrl = "https://$loginHost"
+            loginServerManager.addCustomLoginServer(loginHost, loginUrl)
+            loginServerManager.setSelectedLoginServer(
+                LoginServer(loginHost, loginUrl, true)
+            )
```
brandonpage: Doesn't `addCustomLoginServer` already set the LoginServer it creates as the selected option?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -881,6 +889,160 @@ open class LoginActivity : FragmentActivity() {
...
+        }
+    }
+
+    /**
+     * Creates a Salesforce Welcome Discovery mobile URL using the provided
+     * Salesforce Welcome Discovery host and path URL.
+     * @param salesforceWelcomeDiscoveryHostAndPathUrl The Salesforce Welcome
+     * Discovery host and path URL
+     * @return A Salesforce Welcome Discovery mobile URL with all required
+     * parameters
```
brandonpage, line 933: NIT and Unrelated to this PR:  I found out recently that the [official Kotlin style guide](https://kotlinlang.org/docs/coding-conventions.html#documentation-comments) actually _discourages_ using `@param` and `@return` in documentation comments 🤯.  

If we wanted to go that route it would obviously be quite a large change if we wanted to be consistent everywhere, so perhaps we should have a team discussion about it in the near-ish future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -881,6 +889,160 @@ open class LoginActivity : FragmentActivity() {
...
+     * @return Boolean true if a switch between default or Salesforce Welcome
+     * Discovery log is made, false otherwise.
+     */
+    private fun switchDefaultOrSalesforceWelcomeDiscoveryLogin(uri: Uri) =
+
+        // If the selected server has changed to a new Salesforce Welcome Discovery URL and host.
+        if (isSalesforceWelcomeDiscoveryUrlPath(uri)) {
+
+            // Navigate to Salesforce Welcome Discovery.
+            startActivity(
```
brandonpage, line 979: Is there a way to avoid this?  Since we are using single top it shouldn't create multiple LoginActivity's simultaneously and mess up the back stack but it just feels a heavy handed way of changing the login url (tbf I've done similar things in the past).  Maybe there are other reasons I am not seeing?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1210,6 +1379,96 @@ open class LoginActivity : FragmentActivity() {
             )
         }
 
+        // endregion
+        // region Salesforce Welcome Login Private Implementation
+
+        /** The default Salesforce Welcome Discovery mobile callback URL.  This value is fixed until future Salesforce Welcome updates */
+        private const val SALESFORCE_WELCOME_DISCOVERY_MOBILE_CALLBACK_URL = "sfdc://discocallback"
```
brandonpage, line 1382: NIT:  These all feel a bit wordy for me personally.  I don't think they really need Salesforce in the name and I think they would be find with either "WELCOME" or "DISCOVERY", but that's just my 2 cents.

libs/SalesforceSDK/res/xml/servers.xml
brandonpage: The end user wouldn't know if the app developer is using a scheme and host for discovery, right?

JohnsonEricAtSalesforce: Correct.  We also don't have a specification for what this will look like beyond internal use, so it might change.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: It does, indeed.  I'll update that.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: That is really interesting and feels like a loss of helpful formatting based on how Android Studio displays documentation.  They don't seem to offer an explanation, either.  We'll have to ponder that one.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: They key driver was that the app can receive an external link and will need a way to start a "Welcome" `LoginActivity` according to the app's own lifecycle and timing.  It's possible that the "Default" `LoginActivity` is already resumed and will then switch to "Welcome." Or, the app could directly start a new "Welcome" `LoginActivity` at any time, perhaps from an external link and kind of like QR Code Log In.  This gives the app and MSDK one Intent and a consistent approach to start that adjusts to the current state of things on the fly.

I test this two ways.  One is by using the server picker.  The second is by emulating an external link being received in the app's `MainActivity`.  The external link may or may not be for the default, production WSC even.  Before this is used externally, a lot about WSC may change.  So we have some future unknowns to adapt to.

Now that I've seen how Welcome played out like this, I'm looking back at QR Code Log In and wondering if it would be a beneficial way to clean that up as well.  The logic there is not this organized.

It's also kind of like the way the Chrome callback is handled as well, at least from what I saw in `onNewIntent`.

Maybe there's other ways.  Though, now that we have QR Code Log In and Welcome exposing non-default paths to `LoginActivity` via its `Intent` seems to be working really and follows Android basics.

wmathurin:
What happens if a non-supported client app uses disco? I'm guessing the redirect never takes place. Could we/should we handle that case gracefully?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: Re-reading the original question and answering it more directly, this actually _doesn't_ create multiple activities and preserves the back-stack plus the app's main activity really neatly.  Single top will re-use the existing instance with a call to `onNewItent`.  I watched it closely in the debugger to be sure as well 👍🏻

wmathurin:
> > What happens if a non-supported client app uses disco? I'm guessing the redirect never takes place. Could we/should we handle that case gracefully?
> 
> You're correct that `LoginActivity` will simply ignore the WSC URL and return the user to default log in, which is a lot like what we did for QR Code Log In. It checks the `client_id` matches the boot config remote access consumer key.

Where is that logic?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: I updated what little code documentation we have on the other project and found I really like the new style.  The doc definitely became far more readable without all the boiler plate.  Doesn't seem like any information is lost, but things like return value might be a little less glanceable.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1210,6 +1376,96 @@ open class LoginActivity : FragmentActivity() {
...
+        fun isSalesforceWelcomeDiscoveryMobileUrl(
+            context: Context,
+            uri: Uri,
+        ) = isSalesforceWelcomeDiscoveryUrlPath(uri) && uri.queryParameterNames?.contains(
+            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID
+        ) != null && uri.queryParameterNames?.contains(
+            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID
+        ) != null && uri.queryParameterNames?.contains(
+            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CALLBACK_URL
+        ) != null && uri.getQueryParameter(SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID) == getBootConfig(context).remoteAccessConsumerKey
```
JohnsonEricAtSalesforce: @wmathurin - This is the validation logic matching your comment a moment ago.

wmathurin: The server currently will only allow discovery for a handful of consumer keys (see https://git.soma.salesforce.com/globalIdentity/LoginService/blob/domain-disco-2/ui/lwr/src/modules/login/constants/myDomainDiscoveryConstants.ts#L13). I don't really like the idea of hard-coding that in the client side code. Do you know if the server errors out if you attempt discovery with a non-supported client id?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: Okay!  I wasn't thinking about the deep linking but it makes a lot of sense to have a single code path/point of entry to start Welcome.

brandonpage: @JohnsonEricAtSalesforce Speaking of `onNewIntent` I forgot to comment yesterday and ask if you had tested Advanced Auth?  I don't see a reason why it wouldn't work as expected but it is something we should confirm.

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: The server _does not_ error if given an unknown client id.  For instance, you could try this URL in your web browser.

```
https://gidruntime-cell2.sfproxy.gid.dev1-uswest2.aws.sfdc.cl/?client_id=SfdcMobileChatterXXX&client_version=3.6.0&callback_url=sfdc%3A%2F%2Fdiscocallback
```

What it seems to do it ignore the three parameters entirely.  This means it will prompt for the user's password when a selection is made in the Environment Switcher.  That sort of makes sense, but what's odd is that when a valid password is entered the user gets redirected to `stage.welcome.com` instead of authenticated with the instance.  Perhaps that's a bug?

I don't have any logic in MSDK other than ensuring the client id in an externally linked WSC Discovery URL matches what's in the app's boot config plus making sure that WSC Discovery URLs generated by MSDK use the app's boot config.

What could we do here?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: Can you DM instructions on how to test that to me? I agree, it _should_ work but does require testing.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
wmathurin: Maybe in the check above, you could see if the client id is one of [those](https://git.soma.salesforce.com/globalIdentity/LoginService/blob/domain-disco-2/ui/lwr/src/modules/login/constants/myDomainDiscoveryConstants.ts#L13)? Maybe put some big TODO to remove that logic once the server supports all client ids?
If it's not, then you would let them do a login with WSC without the discovery step, which will work as long as they don't expect advanced authentication.

JohnsonEricAtSalesforce: @wmathurin - How does this look? https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/commits/e437641b1eb54ba9400523c30a930a91321169fa

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1427,13 +1426,17 @@ open class LoginActivity : FragmentActivity() {
...
-            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CALLBACK_URL
-        ) != null && uri.getQueryParameter(SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID) == getBootConfig(context).remoteAccessConsumerKey
+        ): Boolean {
+            val clientIdParameter = uri.getQueryParameter(SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID)
+            return isSalesforceWelcomeDiscoveryUrlPath(uri) && uri.queryParameterNames?.contains(
+                SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID
+            ) != null && uri.queryParameterNames?.contains(
+                SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_VERSION
+            ) != null && uri.queryParameterNames?.contains(
+                SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CALLBACK_URL
```
wmathurin, line 1436: Maybe add a comment to indicate that this should be kept in sync with the corresponding client ids list on the server and also have a todo to remove that check once discovery is made available to all client ids?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1427,13 +1426,17 @@ open class LoginActivity : FragmentActivity() {
...
             uri: Uri,
-        ) = isSalesforceWelcomeDiscoveryUrlPath(uri) && uri.queryParameterNames?.contains(
-            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID
-        ) != null && uri.queryParameterNames?.contains(
-            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID
-        ) != null && uri.queryParameterNames?.contains(
-            SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CALLBACK_URL
-        ) != null && uri.getQueryParameter(SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID) == getBootConfig(context).remoteAccessConsumerKey
+        ): Boolean {
+            val clientIdParameter = uri.getQueryParameter(SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID)
```
wmathurin, line 1430: What did you think of @brandonpage suggestion to shorten constant names:
- SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID could just be DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID
- etc

JohnsonEricAtSalesforce: Just to toss it out there, the reason I thought the longer names is helpful is so they key really, really well to the names of the documents and other resources where I (being the newer developer) know where to find help.  Even being a developer for Mobile SDK, our support across so many aspects of the product can be difficult to keep organized.

Let me know what we want to do either way.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: Take a look at https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/commits/ff78c2615eb8f49c016480d31e9f2f30c59d1a2b

wmathurin: LGTM.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: That makes sense... but by that same logic you could literally prefix everything we do with `SALESFORCE`, which wouldn't be helpful.

PR #2738, opened 2025-07-24
@W-19130548: [MSDK Android] Using External Link With Login_Hint And Login_Host Duplicates Login Servers In Server Picker

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This is a set of quick bug fixes based on testing in SApp today.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -192,7 +192,7 @@ public class LoginServerManager {
 	public void addCustomLoginServer(String name, String url) {
 		// Prevent duplicate servers.
 		for (LoginServer existingServer : getLoginServers()) {
-			if (name.equals(existingServer.name) && url.equals(existingServer.url)) {
+			if (url.equals(existingServer.url)) {
 				setSelectedLoginServer(existingServer);
 				return;
 			}
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -1456,6 +1456,7 @@ open class LoginActivity : FragmentActivity() {
             context: Context,
             uri: Uri,
         ): Boolean {
+            if (!uri.isHierarchical) return false
             val clientIdParameter = uri.getQueryParameter(SALESFORCE_WELCOME_DISCOVERY_MOBILE_URL_QUERY_PARAMETER_KEY_CLIENT_ID)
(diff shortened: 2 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -192,7 +192,7 @@ public void useSandbox() {
 	public void addCustomLoginServer(String name, String url) {
 		// Prevent duplicate servers.
 		for (LoginServer existingServer : getLoginServers()) {
-			if (name.equals(existingServer.name) && url.equals(existingServer.url)) {
+			if (url.equals(existingServer.url)) {
```
JohnsonEricAtSalesforce, line 195: The fix here is to remove "name" from the duplicate server check.  When testing a Welcome external link in SApp, this now re-uses the app's provided "Welcome" login server and selects it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1456,6 +1456,7 @@ open class LoginActivity : FragmentActivity() {
             context: Context,
             uri: Uri,
         ): Boolean {
+            if (!uri.isHierarchical) return false
```
JohnsonEricAtSalesforce, line 1459: This is an unrelated fix.  In SApp when navigating from the GUS login server back to any other the others, the login URL is momentarily `about:blank` which throws and exception and crash here without this check.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -151,8 +151,10 @@ fun PickerBottomSheet(pickerStyle: PickerStyle) {
     val onNewLoginServerSelected = { newSelectedServer: Any?, closePicker: Boolean ->
         if (newSelectedServer != null && newSelectedServer is LoginServer) {
             viewModel.showServerPicker.value = !closePicker
-            viewModel.loading.value = true
-            SalesforceSDKManager.getInstance().loginServerManager.selectedLoginServer = newSelectedServer
+            if (newSelectedServer != SalesforceSDKManager.getInstance().loginServerManager.selectedLoginServer) {
```
JohnsonEricAtSalesforce, line 154: Another rider fix here is to prevent unneeded processing when the user taps the already selected login server.  This seems harmless enough in MSDK's template apps.  For SApp, however, this causes the app to hang with login view's progress spinner forever.  While that might be an issue SApp could fix, the case I saw here is that the update simply shouldn't be emitted by MSDK when no data change was made.

wmathurin (approved):

PR #2745, opened 2025-08-15
Fix NPE when attempting push registration upon login.

brandonpage:


wmathurin (approved):
Maybe more readable with
```kotlin
val communityUrl = account.communityUrl
if (!communityUrl.isNullOrBlank()) {
    fields[NETWORK_ID] = communityUrl
}
```

PR #2747, opened 2025-08-19
Upgrade existing users if the app adds a unique account type.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -130,6 +132,9 @@ else if (installedVersion.isGreaterThanOrEqualTo(new SdkVersion(9, 2, 0, false))
             if (installedVersion.isLessThan(new SdkVersion(12, 0, 0, false))) {
                 updateFromBefore12_0_0();
             }
+            if (installedVersion.isLessThan(new SdkVersion(15, 0, 0, false))) {
+                migrateAccountType();
+            }
```
brandonpage, line 132: I know this isn't the normal naming or upgrade time-fame scheme we typically use, but since this requires the app to implement a value it seemed like a better fit.  Alternatively, we could expose it as a function they can choose to run outside of a MSDK upgrade.

wmathurin (commented):
LGTM.
I see it's a draft PR. Do you have more changes planned?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1192,7 +1192,16 @@ open class SalesforceSDKManager protected constructor(
     open val isHybrid = false
 
     /** The authentication account type, which should match authenticator.xml */
-    val accountType = appContext.getString(account_type)
+    val accountType: String by lazy {
+        val type = appContext.getString(account_type)
+        if (type == "com.salesforce.androidsdk") {
+            // TODO: Turn this logline into an assert in 14.0
+            e(TAG, "No app specific account type found.  To ensure users " +
+                    "can login override the \"account_type\" value in your strings.xml.")
```
brandonpage, line 1200: Unrelated, but I **absolutely hate** the these member imports.  `e` was a bad method name to begin with but at least `SalesforceSDKLogger.e` is readable.  There is no way this would have made it through code review if it was anything other than the log calls we are super familiar with.

libs/SalesforceSDK/res/xml/legacy_authenticator.xml
```diff
@@ -0,0 +1,3 @@
+<account-authenticator xmlns:android="http://schemas.android.com/apk/res/android"
+    android:accountType="com.salesforce.androidsdk" android:icon="@drawable/sf__icon"
```
brandonpage, line 2: Hard coded to our default account type so that we can still retrieve the accounts to migrate when the app provides a new one.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/LegacyAuthenticatorService.java
```diff
@@ -0,0 +1,29 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.auth;
+
+class LegacyAuthenticatorService extends AuthenticatorService { }
```
brandonpage, line 29: AndroidManifest throws an error if you try to reuse the same class for more than one service.

wmathurin (approved):

PR #2750, opened 2025-08-27
@W-19307934 Add UserAccount to Logout Broadcast

npars:
### Problem
We have limited knowledge of which account is undergoing logout in SalesforceSDK. The current implementation requires us to make assumptions about which account is undergoing logout which risks introducing race conditions or logic errors. 

### Solution
* Pass UserAccount undergoing logout with the Logout Broadcast

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/util/LogoutCompleteReceiver.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/util/LogoutCompleteReceiver.kt
@@ -58,7 +58,7 @@ abstract class LogoutCompleteReceiver : BroadcastReceiver() {
      * @param reason The reason for the logout. If no reason is available, [LogoutReason.UNKNOWN] is used.
      */
     @Deprecated(
-        message = "Use onLogoutComplete method that includes the userAccount that was logged out.",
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/util/LogoutCompleteReceiver.kt
```diff
@@ -29,20 +29,45 @@ package com.salesforce.androidsdk.util
...
-    protected abstract fun onLogoutComplete(reason: LogoutReason)
+    /**
+     * Called when logout is complete.
+     *
+     * @param reason The reason for the logout. If no reason is available, [LogoutReason.UNKNOWN] is used.
+     */
+    @Deprecated(
+        message = "Use onLogoutComplete method that includes the userAccount that was logged out.",
+        replaceWith = ReplaceWith("onLogoutComplete(reason: LogoutReason, userAccount: UserAccount?)")
+    )
```
brandonpage: NIT: We typically include the MSDK version we intend to remove the API in.  I would assume 14.0 in this case.

npars: Done

wmathurin (approved):

brandonpage (approved):

PR #2752, opened 2025-08-29
Add Clear Caches menu item to login view

wmathurin:
- Add sf__clear_caches string resource
- Implement clearCaches method using webView.clearCache(true)
- Add Clear Caches menu item to login dropdown
- Add unit and UI tests for new functionality

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -28,6 +28,7 @@ package com.salesforce.androidsdk.ui
 
 import android.webkit.CookieManager
 import android.webkit.URLUtil
+import android.webkit.WebView
 import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.derivedStateOf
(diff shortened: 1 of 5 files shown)


brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -241,6 +241,11 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
     open fun clearCookies() =
         CookieManager.getInstance().removeAllCookies(null)
 
+    /** Clear WebView Caches. */
+    open fun clearCaches(webView: android.webkit.WebView) {
+        webView.clearCache(true)
+    }
```
brandonpage, line 248: Do we want to also reload after clearing cache?  I believe we do that when we clear cookies.  

Should be noted that reloading will cycle the web server code verifier but I don't think that will matter.

wmathurin: The caller does the reload, see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2752/files#diff-321c03965ac0663f16817520073fbe2d6def698c5568ab388f1161c35536d069R303

NB: it's the same with clearCookies.
```kotlin
    open fun clearCookies() =
        CookieManager.getInstance().removeAllCookies(null)
```

brandonpage: Whoops, I missed that.  Great!

PR #2753, opened 2025-09-01
Update Maven Central publishing from OSSRH to Central Portal.

brandonpage:
These changes allow us to publish to the new Maven Central Portal via the [OSSRH Staging API](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/), which is documented as:

> The Portal OSSRH Staging API is a partial reimplementation of the OSSRH / Nexus Repository Manager 2 Staging APIs. The intent of this service is to enable publishers who are using existing plugins to have a smooth migration to the Central Publisher Portal. We hope that over time plugins will adopt the [Portal API](https://central.sonatype.org/publish/publish-portal-api/) rather than rely on this service.

It does not sound like this intermediate API is going away anytime soon, but we should eventually upgrade to the new API when the Gradle Plugin we use supports it or an official Gradle Plugin is _finally_ released.

### DIFF
--- a/publish/publish.sh
+++ b/publish/publish.sh
@@ -11,3 +11,4 @@ fi
 ./gradlew libs:SalesforceReact:publishReleasePublicationToSonatypeRepository
 ./gradlew libs:SmartStore:publishReleasePublicationToSonatypeRepository
 ./gradlew libs:MobileSync:publishReleasePublicationToSonatypeRepository
+./gradlew publishToSonatype closeSonatypeStagingRepository
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ buildscript {
 
     dependencies {
         classpath("com.android.tools.build:gradle:8.10.1")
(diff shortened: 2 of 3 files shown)


publish/publish.sh
```diff
@@ -10,4 +10,5 @@ fi
 ./gradlew libs:SalesforceHybrid:publishReleasePublicationToSonatypeRepository
 ./gradlew libs:SalesforceReact:publishReleasePublicationToSonatypeRepository
 ./gradlew libs:SmartStore:publishReleasePublicationToSonatypeRepository
-./gradlew libs:MobileSync:publishReleasePublicationToSonatypeRepository
\ No newline at end of file
+./gradlew libs:MobileSync:publishReleasePublicationToSonatypeRepository
+./gradlew publishToSonatype closeSonatypeStagingRepository
```
brandonpage, line 14: This last line is needed tell the staging API we are done and ready to validate in the new Central Portal.

PR #2762, opened 2025-09-10
@W-19461869: [M1][MSDK13.1] Enable Flexible Server Matching for QR Code Login (Android)

JohnsonEricAtSalesforce:
🧱 _*Under Construction: Not Ready For Final Review*_ 🚧


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -54,7 +54,7 @@
  *
  * @author bhariharan
  */
-public class LoginServerManager {
+public class LoginServerManager implements LoginServerManaging {
```
JohnsonEricAtSalesforce, line 57: @wmathurin - Here's a comparable interface for the protocol we added on iOS.  It does really help the tests and logic objects remain consistent.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+import com.salesforce.androidsdk.config.LoginServerManaging
+import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey.AppServiceHosts
+import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey.OnlyShowAuthorizedHosts
+import com.salesforce.androidsdk.config.RuntimeConfig.getRuntimeConfig
+
+/**
+ * For Salesforce Identity UI Bridge API support, an overriding front door
+ * bridge URL to use in place of the default initial URL.
+ */
+internal class FrontdoorBridgeLoginOverride(
```
JohnsonEricAtSalesforce, line 43: @wmathurin - This is nearly identical to the iOS version we created.

JohnsonEricAtSalesforce: This class encapsulates the otherwise separate front-door bridge properties that were attached to unrelated classes along with all the logic and tests that has accumulated around them.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,111 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.ui
+
+import android.net.Uri
+import com.salesforce.androidsdk.config.LoginServerManaging
+import java.net.URL
+
+internal data class FrontdoorBridgeUrlAppLoginServerMatch(
```
JohnsonEricAtSalesforce: @wmathurin - This is another bit of logic we can use as-is from the iOS code along with all its tests.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -310,7 +310,7 @@ open class LoginActivity : FragmentActivity() {
                 with(SalesforceSDKManager.getInstance()) {
                     // Fetch well known config and load in custom tab if required.
                     fetchAuthenticationConfiguration {
-                        if (isBrowserLoginEnabled) {
+                        if (isBrowserLoginEnabled && viewModel.frontdoorBridgeLoginOverride == null /* Browser-based authentication is applicable when not authenticating with a front-door bridge URL */) {
```
JohnsonEricAtSalesforce: For Android and just like iOS, this is the key fix to enable a front-door bridge URL generated by an advanced authentication app to bypass showing the unneeded native browser.  The user does get logged in successfully when the front door bridge URL is loaded in the default web view.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -100,7 +100,13 @@ internal data class FrontdoorBridgeUrlAppLoginServerMatch(
                 }
             }
         } else {
-            results.add(selectedAppLoginHost)
+            try {
+                val url = URL(selectedAppLoginHost)
+                results.add(url.host)
+            } catch (_: Exception) {
+                // If parsing fails, try to use as-is (might already be just a host)
```
wmathurin: When do you think it would not parse but should still be used?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverrideTest.kt
```diff
@@ -0,0 +1,627 @@
...
+
+        // Assert
+        assertEquals("", override.codeVerifier)
+        assertTrue(override.matchesConsumerKey)
+    }
+
+    @Test
+    fun testEdgeCase_VeryLongCodeVerifier() {
+        // Arrange
+        val longCodeVerifier = "a".repeat(1000) // Very long code verifier
```
wmathurin: Isn't the code verifier a base64-encoded 128-byte randomly generated key?
It should always be the same length. That test probably does not make sense?

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/RuntimeConfig.java
```diff
@@ -69,7 +73,8 @@ public enum ConfigKey {
 
     private static RuntimeConfig INSTANCE = null;
 
-    RuntimeConfig(Context ctx) {
+    @VisibleForTesting(otherwise = PACKAGE_PRIVATE)
```
brandonpage, line 76: I did not know you could could use the Kotlin style named parameter syntax in a Java file.  Must be because it is an annotation?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
brandonpage: I really don't love the very overly documented generated code :/

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+     * For Salesforce Identity UI Bridge API support, indicates if the
+     * overriding front door bridge URL has a host that matches the app's
+     * selected login server.
+     */
+    var matchesLoginHost: Boolean = false
+        private set
+
+    init {
+        val frontdoorBridgeUrlComponents = frontdoorBridgeUrl.toString()
+        val frontdoorBridgeUri = frontdoorBridgeUrlComponents.toUri()
```
brandonpage: We take in `frontdoorBridgeUrl` as a `Uri` to make a `frontdoorBridgeUrlComponents` `String`... and then only use that to make `frontdoorBridgeUri` which is a `Uri`.   

Am I missing something or is this completely pointless?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+     * The selected app login server.  This is intended for test automation only
+     */
+    selectedAppLoginServer: String = SalesforceSDKManager.getInstance().loginServerManager.selectedLoginServer.url,
+
+    /**
+     * The preference for using mobile device management preferences for
+     * allowing the addition and switching of app login servers.  This is
+     * intended for test automation only
+     */
+    addingAndSwitchingLoginServersPerMdm: Boolean = true,
```
brandonpage, line 66: NIT:  Something like `mdmControlledServers` might be better?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+            val frontdoorBridgeUrlClientId = startUri.getQueryParameter("client_id")
+
+            frontdoorBridgeUrlClientId?.let { clientId ->
+                val appConsumerKey = BootConfig.getBootConfig(SalesforceSDKManager.getInstance().appContext).remoteAccessConsumerKey
+                matchesConsumerKey = clientId == appConsumerKey
+            }
+        }
+
+        // Check if the front door URL host matches the app's selected login server
+        val addingAndSwitchingLoginServersAllowedResolved = if (!addingAndSwitchingLoginServersPerMdm) {
```
brandonpage: NIT:  Unnecessary `!`.  Just flip the if and else blocks so it is easier for a human to read.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+     * intended for test automation only
+     */
+    addingAndSwitchingLoginServersPerMdm: Boolean = true,
+
+    /**
+     * The preference for allowing the addition and switching of app login
+     * servers when the MDM preference is ignored.  This is intended for test
+     * automation only
+     */
+    addingAndSwitchingLoginServerOverride: Boolean = false
```
brandonpage, line 73: Why (even for testing) would we ever want to ignore a hard requirement from an MDM?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+    private val addingAndSwitchingLoginServersAllowed: Boolean
+        get() {
+            val runtimeConfig = getRuntimeConfig(SalesforceSDKManager.getInstance().appContext)
+            val onlyShowAuthorizedServers = runtimeConfig.getBoolean(OnlyShowAuthorizedHosts)
+            val mdmLoginServers = try {
+                runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHosts)
+            } catch (_: Exception) {
+                null
+            }
+            return !onlyShowAuthorizedServers && (mdmLoginServers?.isEmpty() != false)
```
brandonpage: Seems like neither of these values tell us if _adding_ is allowed.  Just if switching is allowed/possible.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
```diff
@@ -0,0 +1,151 @@
...
+        appLoginServer?.let { server ->
+            matchesLoginHost = true
+            // Set the login server on the server manager
+            val loginServerManager = SalesforceSDKManager.getInstance().loginServerManager
+            val loginUrl = "https://$server"
+            loginServerManager.addCustomLoginServer(loginUrl, loginUrl)
+        }
+    }
+
+    private val addingAndSwitchingLoginServersAllowed: Boolean
```
brandonpage, line 134: NIT:  Why are these private vals at the bottom of the file?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+            selectedAppLoginServer
+        )
+    }
+
+    private fun appLoginServerForFrontdoorBridgeUrl(
+        frontdoorBridgeUrl: Uri,
+        loginServerManaging: LoginServerManaging,
+        addingAndSwitchingLoginServersAllowed: Boolean,
+        selectedAppLoginServer: String
+    ): String? {
```
brandonpage: Why is this a named function if it is only ever used the once?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+    private fun eligibleAppLoginServersForFrontdoorBridgeUrl(
+        loginHostStore: LoginServerManaging,
+        addingAndSwitchingLoginHostsAllowed: Boolean,
+        selectedAppLoginHost: String
+    ): List<String> {
+        val results = mutableListOf<String>()
+        if (addingAndSwitchingLoginHostsAllowed) {
+            val numberOfHosts = loginHostStore.numberOfLoginServers()
+            for (i in 0 until numberOfHosts) {
+                val server = loginHostStore.loginServerAtIndex(i)
```
brandonpage: Why not just use the existing `loginServers.forEach { server ->  }`???

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                }
+            }
+        } else {
+            try {
+                val url = URL(selectedAppLoginHost)
+                results.add(url.host)
+            } catch (_: Exception) {
+                // If parsing fails, try to use as-is (might already be just a host)
+                results.add(selectedAppLoginHost)
```
brandonpage: How is a string that throws an exception when parsed as a URL ever going to be a valid URL host?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+        return results
+    }
+}
+
+private fun Uri.isMyDomain(): Boolean {
+    return host?.contains(".my.") == true
```
brandonpage, line 107: This should be `.contains(".my.salesforce.com")`, right?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+            loginServerManaging,
+            addingAndSwitchingLoginServersAllowed,
+            selectedAppLoginServer
+        )
+
+        for (eligibleAppLoginServer in eligibleAppLoginServers) {
+            if (frontdoorBridgeUrlHost == eligibleAppLoginServer) {
+                return eligibleAppLoginServer
+            }
+        }
```
brandonpage: Would be more efficient to combine this so you aren't iterating through the list twice.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        if (frontdoorBridgeUrl.isMyDomain()) {
+            val frontdoorBridgeUrlMyDomainSuffix = "my.${frontdoorBridgeUrlHost.split(".my.").last()}"
```
brandonpage: Isn't this always `my.salesforce.com`?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+            if (frontdoorBridgeUrlMyDomainSuffix.isNotEmpty()) {
```
brandonpage: I am surprised the compiler did not complain about this check as it can never be false.  The suffix is always "my." + something.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
```diff
@@ -0,0 +1,117 @@
...
+                for (eligibleAppLoginServer in eligibleAppLoginServers) {
+                    if (eligibleAppLoginServer.endsWith(frontdoorBridgeUrlMyDomainSuffix)) {
```
brandonpage: Doesn't this just match _any_ My Domain?  By this logic "one.my.salesforce.com" matches "two.my.salesforce.com".

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManaging.kt
```diff
@@ -0,0 +1,54 @@
...
+
+package com.salesforce.androidsdk.config
+
+import com.salesforce.androidsdk.config.LoginServerManager.LoginServer
+
+/**
+ * An object that can manage Salesforce Mobile SDK's list of login server. This
+ * is functionally equivalent to MSDK's iOS `SFSDKLoginHostStoring` protocol.
+ */
+internal interface LoginServerManaging {
```
brandonpage, line 36: There is no reason for this to exist.  `loginServerAtIndex` and `numberOfLoginServers` could be replaced with `loginServers.forEach`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
brandonpage: Why is this a class and not a function that returns a single boolean?   It is instantiated just so `appLoginServerMatch` can be checked once.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -430,7 +430,7 @@ open class LoginActivity : FragmentActivity() {
      */
     @Suppress("MemberVisibilityCanBePrivate")
     fun loginWithFrontdoorBridgeUrl(
-        frontdoorBridgeUrl: String,
+        frontdoorBridgeUrl: Uri,
```
brandonpage: This breaks semantic versioning.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: So if the app has QR Code login setup it disables Advanced Authentication? 

Also, please move the comment from inside the `if` to a separate line.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1359,7 +1353,7 @@ open class LoginActivity : FragmentActivity() {
         data class UiBridgeApiParameters(
 
             /** The front door bridge URL */
-            val frontdoorBridgeUrl: String,
+            val frontdoorBridgeUrl: Uri,
```
brandonpage: Again, semver.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -220,20 +215,15 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
...
     }
 
     /** Reloads the WebView with a newly generated authorization URL. */
     open fun reloadWebView() {
-        if (!isUsingFrontDoorBridge) {
-            // The Web Server Flow code challenge makes the authorization url unique each time,
-            // which triggers recomposition.  For User Agent Flow, change it to blank.
-            if (!SalesforceSDKManager.getInstance().useWebServerAuthentication) {
-                loginUrl.value = ABOUT_BLANK
-            }
```
brandonpage: Revert this!  This is a bug.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -253,13 +243,19 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
...
     fun loginWithFrontDoorBridgeUrl(
-        frontdoorBridgeUrl: String,
+        frontdoorBridgeUrl: Uri,
         pkceCodeVerifier: String?,
     ) {
-        isUsingFrontDoorBridge = true
-        frontdoorBridgeServer = with(URI(frontdoorBridgeUrl)) { "${scheme}://${host}" }
-        frontdoorBridgeCodeVerifier = pkceCodeVerifier
-        loginUrl.value = frontdoorBridgeUrl
+        val frontdoorBridgeLoginOverride = FrontdoorBridgeLoginOverride(
```
brandonpage: Shadowed name.  And why are there two?  We set the class level one from the Activity.  Why are we making one local to this function?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
wmathurin: OnlyShowAuthorizedHosts: If true, prevents users from modifying the list of hosts that the Salesforce mobile app can connect to. See [doc](https://developer.salesforce.com/docs/platform/mobile-sdk/guide/oauth-mdm.html#mdm-settings-for-automatic-custom-host-provisioning).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
wmathurin: @brandonpage is right a bit of the proposed logic has been lost:

> [Soft match] Look at part of the hostname in the QR code that comes after .my. and make sure it appears in the currently selected login server also as long as the currently selected login server does not have .my, itself. When the currently login server has a .my. the whole hostname should match. So [mydomain.my.salesforce.com](http://mydomain.my.salesforce.com/) would be allowed if [login.salesforce.com](http://login.salesforce.com/) is currently selected but not if [myotherdomain.my.salesforce.com](http://myotherdomain.my.salesforce.com/) is selected.

So if we have a my domain front door url, and we did not find an exact match, we should look among the eligible login hosts that are not my domain themselves.

The idea is that if the QR says one.my.salesforce.com and you have login.salesforce.com, then you should be let through.
The only weakness is that if you have login.salesforce.com and test.salesforce.com we won't know which one to the my domain "belongs" to. That would however only be a problem if somehow the app disallowed login.salesforce.com and the QR code is for a production my domain (or vice versa with Sandbox).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
wmathurin: Not in the internal test environments.
The my domain could be: konsf.test1.my.pc-rnd.salesforce.com and login pool server would be: login.test1.pc-rnd.salesforce.com.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
brandonpage: Hmmm, `OnlyShowAuthorizedHosts` seems poorly named but fair enough.  Maybe we should add a code comment stating that?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
brandonpage: Ah, that's interesting.  I was going off the official doc.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
brandonpage: Apparently this is not true for [internal test environments](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2762#discussion_r2338487991).  Perhaps we should use regex to check with a wildcard for something like ".my.*salesforce.com"?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
wmathurin: The official doc is correct for sandbox/productions. But internal test / dev envs are different.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
brandonpage: If this is never set outside of tests why does it exists.  Why test both true and false in tests if it can only ever be false in production?

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatchTest.kt
```diff
@@ -0,0 +1,436 @@
...
+
+        // Act
+        val match = FrontdoorBridgeUrlAppLoginServerMatch(
+            frontdoorBridgeUrl = uri,
+            loginServerManaging = mockLoginServerManaging,
+            addingAndSwitchingLoginServersAllowed = true,
+            selectedAppLoginServer = "different.salesforce.com"
+        )
+
+        // Assert - Should match the first server ending with "my.salesforce.com"
```
wmathurin, line 368: Wait. if front door url is: https://company.my.salesforce.com
And login servers are login.salesforce.com, test.my.salesforce.com, custom.my.salesforce.com, then it should soft-match with login.salesforce.com.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: Seems reasonable.  @wmathurin - Thinking of the way `FrontdoorBridgeLoginOverride` grew over time from a single property on an existing class to an entire class of its own in your reviews from iOS, now we have this new `FrontdoorBridgeUrlAppLoginServerMatch` type which we broke out in turn.  It has humble enough beginnings (for now) to make a stand-alone function as Brandon suggested.  Do you see this type getting more complex over time?  If not, we can do as Brandon suggested easily enough.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: I'll get that comment moved.  👍🏻

JohnsonEricAtSalesforce: `frontdoorBridgeLoginOverride` is only set during the lifetime of a front door bridge login event.  At the end of the event, `resetFrontDoorBridgeUrl` is called and all properties related to front door use are cleared (conveniently this is a single object now based on Wolf's iOS reviews).  After attempting the QR Code, the advanced authentication logic is as it was before.

For additional context, users/apps with advanced authentication can authenticate via a front door bridge URL without MSDK needing to show the browser.  This new condition prevents MSDK from displaying the unneeded browser while behind it the user would already be authenticated successfully from the front door bridge URL.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: Of all the interesting things the AI tooling did that worked and didn't, this was the oddest choice - Or so I thought! 🥴. Nothing in my prompt or the existing logic suggested this, from what I saw. I was already planning on removing this before merging.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverrideTest.kt
JohnsonEricAtSalesforce: This test doesn't make sense, agreed.  MSDK only passes this value through and I don't see there's any length restriction that should have prompted the tool to create this test.  I'll remove it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/RuntimeConfig.java
JohnsonEricAtSalesforce: I recall using named parameters with Java annotations all the way back to Java 5 and Hibernate ORM (2005), which used it a lot.  That's an interesting place where named parameters snuck in early and I'd never thought about it 🙂

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
JohnsonEricAtSalesforce: That is indeed completely pointless and I hadn't gotten that far in reviewing the generated output.  Thanks for finding it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeLoginOverride.kt
JohnsonEricAtSalesforce: It is just there for testing, true.  It's use is paired with `addingAndSwitchingLoginServerOverride` which lets the test choose the test value.  An alternate approach could be to pop and interface on `RuntimeConfig` and mock it much as we did for `LoginServerManager`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: @wmathurin had me break this into a separate source as part of the iOS review.  Is was once private to `FrontdoorBridgeLoginOverride`.  We'd though about attaching this function to `LoginServerManager`, but then I'd suggested that class should't have its scope expanded to an awareness of front-door bridge.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: Looks like iOS didn't have a comparable way of accessing this as a list so the tools missed finding this convenience.  Sorry, I hadn't gotten that far yet.  I'll update this.  Thanks!

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: @wmathurin noticed this oddity injected by the generator as well.  I'm still really curious where it got this notion since that wasn't in the prompt or really even hinted at by any code in context. 🤨 I've removed it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: I'll take a look during my next testing pass since the matching logic is tricky.  Stay tuned on this one ⏳

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: Agreed.  We can simplify that.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatch.kt
JohnsonEricAtSalesforce: @wmathurin, that issue with the "belongs to" note was causing the Salesforce app to fail for me though.  The test orgs and users I have would create a QR Code for the specific My Domain host, but the app ships (in test, at least) with preset "login" hosts that the QR Code cannot be used with.

I had these notes in the story and the soft-match as we have it on this branch do work in the app with the test users and orgs I have at the moment.

- If QR is not a my domain, existing login server must match exactly
- If QR is a my domain, existing login server must either match exactly or match everything after the .my.
  1. If adding and switching are disallowed, only let the QR through if its login server "soft-matches" the currently selected login server.
  2. If adding is disallowed but switching is allowed, let the QR through if its login server "soft-matches" any of the login server and switch to it.
  3. If adding is allowed and switching is allowed, try (b) first, but if no match are found add the QR login server and switch to it.

There's definitely complexity to go around on this logic 😅 We might need to break this into a pairing session so we can demonstrate the behavior the app is seeing.

JohnsonEricAtSalesforce: Let's schedule some time to review this interactively in the app 🚀

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManaging.kt
JohnsonEricAtSalesforce: Agreed.  We can replace these two methods which were a carry-over from the iOS implementation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
JohnsonEricAtSalesforce: That one was a hallucination in the tool 🤖 and I'll revert it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
JohnsonEricAtSalesforce: This was also work-in-progress output of the tool that hadn't yet been reviewed 🤖 I'll optimize this ✅

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/FrontdoorBridgeUrlAppLoginServerMatchTest.kt
JohnsonEricAtSalesforce: @wmathurin - This specific rule is where our notion of "matching" the front-door URL's host to another host needs work.  The users I was given by the app cannot log in using the "soft-matched" login server.  Only the exact My Domain host that's encoded in the front-door URL by the Identity API will work.  

This may be the reason only the logic that creates the QR Code and obtains the UI Bridge front-door URL from the API might be able to know what the appropriate login server.  The client code, like we attempted here in MSDK, may not be able to guess the answer.  Even if it could, the server should likely own that decision for additional future-proofing in the code and avoiding redundant client implementation.

We should ask the identity team for more information on this.

JohnsonEricAtSalesforce:
This change has been shelved indefinitely pending systems requirements analysis.

PR #2764, opened 2025-09-12
@W-19628153: [Android][QRCode] When Auth Config "Use the native browser for user authentication on Android" is enabled users logging in via QR code still challenged with browser login

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

This brings the fix for front door bridge login and advanced authentication from https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2762 into a stand-alone pull request.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -311,7 +311,8 @@ open class LoginActivity : FragmentActivity() {
                 with(SalesforceSDKManager.getInstance()) {
                     // Fetch well known config and load in custom tab if required.
                     fetchAuthenticationConfiguration {
-                        if (isBrowserLoginEnabled) {
+                        /* Browser-based authentication is applicable when not authenticating with a front-door bridge URL */
+                        if (isBrowserLoginEnabled && !viewModel.isUsingFrontDoorBridge) {
                             if (useWebServerAuthentication) {
                                 viewModel.loginUrl.value?.let { url -> loadLoginPageInCustomTab(url, customTabLauncher) }
                             } else {


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -311,7 +311,8 @@ open class LoginActivity : FragmentActivity() {
                 with(SalesforceSDKManager.getInstance()) {
                     // Fetch well known config and load in custom tab if required.
                     fetchAuthenticationConfiguration {
-                        if (isBrowserLoginEnabled) {
+                        /* Browser-based authentication is applicable when not authenticating with a front-door bridge URL */
+                        if (isBrowserLoginEnabled && !viewModel.isUsingFrontDoorBridge) {
```
JohnsonEricAtSalesforce, line 315: Since the front door bridge URL will sign the user in automatically, there's no need to display the native browser for username and password authentication if the view model indicates a front door bridge is in use.  Note, this property on the view model is only set for the lifetime of a single front door bridge login event.

brandonpage: Thank you for that note, it preempted the question I was going to ask. 🚀

JohnsonEricAtSalesforce: 🤣 You asked the same question on the previous pull request.  We're creatures of habit!

JohnsonEricAtSalesforce:
Note: This pull request supersedes https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2762 until the requirements for further changes are fully defined.

JohnsonEricAtSalesforce:
The CI test failures are inherited from `dev` and not related to this change.

PR #2766, opened 2025-09-18
Fix use of non-current account in client manager and possible logout of current user.

brandonpage:
This bugfix is not a simple revert of a previous change since I believe the code changed in this PR has been working incorrectly for some time by using the wrong accounts token.  We simply did not notice until the logout issue was added on top.

### DIFF
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/ClientManagerMockTest.kt
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/ClientManagerMockTest.kt
@@ -1,0 +1,428 @@
+package com.salesforce.androidsdk.rest
+
+import android.accounts.Account
+import android.content.Context
+import android.content.Intent
+import androidx.test.filters.SmallTest
+import androidx.test.platform.app.InstrumentationRegistry
+import com.salesforce.androidsdk.accounts.UserAccount
+import com.salesforce.androidsdk.accounts.UserAccountManager
+import com.salesforce.androidsdk.accounts.UserAccountManagerTest
+import com.salesforce.androidsdk.analytics.EventBuilderHelper
+import com.salesforce.androidsdk.app.SalesforceSDKManager
+import com.salesforce.androidsdk.auth.HttpAccess
+import com.salesforce.androidsdk.auth.OAuth2
+import io.mockk.every
+import io.mockk.just
+import io.mockk.mockk
+import io.mockk.mockkObject
+import io.mockk.mockkStatic
+import io.mockk.runs
+import io.mockk.slot
+import io.mockk.spyk
+import io.mockk.unmockkAll
+import io.mockk.verify
+import okhttp3.Call
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.OkHttpClient
+import okhttp3.Response
+import okhttp3.ResponseBody.Companion.toResponseBody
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -417,11 +428,12 @@ public String getNewAuthToken() {
                         if (Looper.myLooper() == null) {
                             Looper.prepare();
                         }
+                        boolean showLoginPage = accounts.length > 1;
                         // Note: As of writing (2024) this call will never succeed because revoke API is an
                         // authenticated endpoint.  However, there is no harm in attempting and the debug logs
                         // produced may help developers better understand the state of their app.
                         SalesforceSDKManager.getInstance()
-                                .logout(null, null, false, OAuth2.LogoutReason.REFRESH_TOKEN_EXPIRED);
+                                .logout(matchingAccount, null, showLoginPage, OAuth2.LogoutReason.REFRESH_TOKEN_EXPIRED);
```
brandonpage, line 436: `logout` will default to the current user if null is passed 😬.  That is intended in most cases, but really bad in this one.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -379,8 +379,20 @@ public AccMgrAuthTokenProvider(ClientManager clientManager, String instanceUrl,
...
+            Account[] accounts = clientManager.getAccounts();
+            Account matchingAccount = null;
+
+            // Find the account for this client.
+            for (Account account : accounts) {
+                UserAccount user = userAccountManager.buildUserAccount(account);
+                if (user != null && lastNewAuthToken.equals(user.getAuthToken())) {
+                    matchingAccount = account;
+                }
+            }
```
brandonpage, line 392: I wish there was a better way to do this.  Perhaps we should associate each ClientManager to an account?

wmathurin: But don't we have an account associated already with ClientManager: `final Account acc = clientManager.getAccount();`

brandonpage: That returns the current account, which is (part of) the problem. If the user we are trying to get a new token for not the current user we will invalidate and refresh for the wrong account.

wmathurin: Tying the ClientManager to account would make sense, but it's quite a change in behavior so probably not a 13.1 change. Maybe we could add a convenience method on UserAccountManager to find a matching account by auth token?

brandonpage:
> @brandonpage do you know how it was working in 12.1.1?

It "worked" in `12.1.1` because the `getNewAuthToken` call would refresh the token for the user you were switching to (technically current user).  This happened because `refreshStaleToken` used the refresh token from the account passed in.  If that operation was successful then we don't call `logout`.  

`12.2.0` onwards `refreshStaleToken` was changed to (correctly?) use the cached refresh token.  This cached token is for the correct user, which means the refresh correctly fails.  When that happens `refreshStaleToken` returns a null user, which leads to `logout` being called with that null user and the current user is removed.

PR #2772, opened 2025-09-24
Use BOM Compose Versions

npars:
Tidy up dependency declaration to use the compose versions from the BOM.

We noticed that the versions were declared separately which increased risk of misalignment for consumers. This is especially important because SalesforceMobileSDK uses Compose APIs marked with `ExperimentalMaterial3Api` which may not be compatible across library releases. A mismatch in library versions may result in a runtime crash.

### DIFF
--- a/libs/SalesforceSDK/build.gradle.kts
+++ b/libs/SalesforceSDK/build.gradle.kts
@@ -11,7 +11,6 @@ plugins {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


wmathurin:
In dev, we have moved to a TOML-based dependency management - see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2891

PR #2774, opened 2025-09-25
Improve getNewAuthToken.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -408,10 +392,31 @@ public String getNewAuthToken() {
...
+            // Only check for matching account inside synchronized thread that
+            // is actually getting the new auth token.
+            UserAccountManager userAccountManager = SalesforceSDKManager.getInstance().getUserAccountManager();
+            Account[] accounts = clientManager.getAccounts();
+            Account matchingAccount = null;
             String newAuthToken = null;
             String newInstanceUrl = null;
-            try {
 
+            if (refreshToken != null) {
```
brandonpage, line 404: If the `refreshToken` is somehow null it seems like that should be a logout scenario, but we would need to cache the account/user in the client manager.  This was mentioned on the previous PR as well, but feels like too large of a change for 13.1.  I will create a 14.0 trust item to better handle this.

PR #2776, opened 2025-09-30
Saving scope in user account

wmathurin:
- token end point response contains scope
- this value is now captured and saved in the user account
- tests were updated


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -667,7 +673,49 @@ public String getTokenFormat() {
...
+	public String getScope() {
+		return scope;
+	}
+
+	/**
+	 * Parses the space-delimited scope string into its individual components.
+	 *
+	 * @return Array of scope strings (empty if scope is null/empty).
+	 */
+	public String[] parseScopes() {
```
brandonpage, line 689: I find Lists to be a lot easier to deal with than Arrays, especially in Kotlin.  Seems like the List `contains` function would be useful for scopes, but you could call `contains()` on the String scope itself as well so really not a big deal.

Edit:  Apparently I should have just kept reading, `hasScope` resolves the above.

brandonpage (approved):

wmathurin:
Only test failure (LoginActivityTest.viewModelIsUsingFrontDoorBridge_DefaultValue_onCreateWithoutQrCodeLoginIntent) is related to changes.

PR #2777, opened 2025-10-02
Allowing no scopes to be provided 

wmathurin:
- Checking the scope in the token endpoint response to make sure we got refresh_token
- Refactored onAuthFlowComplete() to make it testable and added tests for it

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
@@ -33,7 +33,6 @@ import androidx.annotation.VisibleForTesting
 import com.salesforce.androidsdk.R.string.sf__generic_authentication_error
 import com.salesforce.androidsdk.R.string.sf__generic_authentication_error_title
 import com.salesforce.androidsdk.R.string.sf__managed_app_error
-import com.salesforce.androidsdk.R.string.sf__missing_refresh_token_scope_error
 import com.salesforce.androidsdk.accounts.UserAccount
 import com.salesforce.androidsdk.accounts.UserAccountBuilder
 import com.salesforce.androidsdk.accounts.UserAccountManager
@@ -136,14 +135,9 @@ internal suspend fun onAuthFlowComplete(
     // Create a ScopeParser from tokenResponse.scope
     val scopeParser = ScopeParser(tokenResponse.scope)
     
-    // Check that it has the refresh token scope, otherwise call onAuthFlowError
+    // Check that it has the refresh token scope, only warns if it is missing
     if (!scopeParser.hasRefreshTokenScope()) {
         w(TAG, "Missing refresh token scope.")
-        onAuthFlowError(
-            context.getString(sf__generic_authentication_error_title),
-            context.getString(sf__missing_refresh_token_scope_error), null
-        )
-        return
     }
 
     // Check that the tokenResponse.scope contains the identity scope before calling the identity service
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/AuthenticationUtilitiesTest.kt
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/AuthenticationUtilitiesTest.kt
@@ -116,23 +116,6 @@ class AuthenticationUtilitiesTest {
     }
 
     @Test
-    fun testOnAuthFlowComplete_missingRefreshTokenScope_shouldCallError() = runTest {
-        // Given
-        val tokenResponseWithoutRefreshToken = createTokenEndpointResponse(
-            scope = "id" // Missing refresh_token scope
-        )
-
-        // When
-        callOnAuthFlowComplete(tokenResponseWithoutRefreshToken)
-
-        // Then
-        verify { onAuthFlowError.invoke("Error", "Authentication error: missing refresh_token scope on connected app / external client app. Please contact your administrator.", null) }
-        verify(exactly = 0) { onAuthFlowSuccess.invoke(any()) }
-        verify(exactly = 0) { mockUserAccountManager.createAccount(any()) }
-        verify(exactly = 0) { mockUserAccountManager.switchToUser(any()) }
-    }
-
-    @Test
     fun testOnAuthFlowComplete_managedAppRequirement_shouldCallError() = runTest {
         // Given
         val userIdentityWithManagedAppRequirement = createIdServiceResponse(
--- a/libs/SalesforceSDK/res/values/sf__strings.xml
+++ b/libs/SalesforceSDK/res/values/sf__strings.xml
@@ -9,7 +9,6 @@
     <string name="sf__invalid_server_url">Please type in a valid URL</string>
     <string name="sf__access_token_revoked">Your access token has been revoked by the administrator. You will now be logged out.</string>
     <string name="sf__managed_app_error">Authentication only allowed from managed device.</string>
-    <string name="sf__missing_refresh_token_scope_error">Authentication error: missing refresh_token scope on connected app / external client app. Please contact your administrator.</string>
     <string name="sf__jwt_authentication_error">JWT authentication error. Please try again.</string>
 
     <!-- SSL errors -->


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -94,11 +99,55 @@ internal suspend fun onAuthFlowComplete(
...
+        ::setAdministratorPreferences,
+        ::addAccountHelper,
+        ::handleScreenLockPolicy,
+        ::handleBiometricAuthPolicy,
+        ::handleDuplicateUserAccount
+    )
+}
+
+@VisibleForTesting
+internal suspend fun onAuthFlowComplete(
```
wmathurin, line 93: This is like the original onAuthFlowComplete except with pieces refactored out in private helper methods - which are not called directly but are called through lambda parameters.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -144,68 +202,20 @@ internal suspend fun onAuthFlowComplete(
         .clientId(consumerKey)
         .nativeLogin(nativeLogin)
         .build()
-    account.downloadProfilePhoto()
```
wmathurin, line 147: Moved this one into `fun addAccount(account: UserAccount?, context: Context, isTestRun: Boolean, loginServerManager: LoginServerManager)`

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -115,20 +164,29 @@ internal suspend fun onAuthFlowComplete(
...
-            )
-        }
-    }.onFailure { throwable ->
-        w(TAG, "Cannot fetch user identity due to an error.", throwable)
-    }.getOrNull()
+    // Create a ScopeParser from tokenResponse.scope
+    val scopeParser = ScopeParser(tokenResponse.scope)
+    
+    // Check that it has the refresh token scope, otherwise call onAuthFlowError
+    if (!scopeParser.hasRefreshTokenScope()) {
```
wmathurin, line 139: This check is new to this PR.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/AuthenticationUtilitiesTest.kt
```diff
@@ -0,0 +1,327 @@
...
+
+        // Setup mock for UserAccountManager methods
+        every { mockUserAccountManager.createAccount(any()) } returns mockk<android.os.Bundle>()
+        every { mockUserAccountManager.switchToUser(any()) } returns Unit
+        every { mockUserAccountManager.sendUserSwitchIntent(any(), any()) } returns Unit
+
+    }
+
+    @Test
+    fun testOnAuthFlowComplete_blockIntegrationUser_shouldCallError() = runTest {
```
wmathurin, line 104: Tests against the refactored onAuthFlowComplete().

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -387,17 +387,13 @@ public static URI getFrontdoorUrl(URI url,
...
     /**
-     * Computes the scope parameter from an array of scopes. Also adds
-     * the 'refresh_token' scope if it hasn't already been added.
+     * Computes the scope parameter from an array of scopes.
      *
      * @param scopes Array of scopes.
-     * @return Scope parameter.
+     * @return Scope parameter string (possibly empty).
      */
     public static String computeScopeParameter(String[] scopes) {
```
wmathurin, line 395: This is an important change for this PR.
`ScopeParser.computeScopeParameter(scopes)` will leave an empty scopes list unchanged.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/ScopeParser.kt
```diff
@@ -0,0 +1,159 @@
...
+         *   scopes assigned to the connected app / external client app will be requested by default
+         *   (no explicit scope parameter is sent).
+         * - If {@code scopes} is non-empty, ensures {@code refresh_token} is present in the set and
+         *   returns a space-delimited string of unique, sorted scopes.
+         *
+         * @param scopes Array of scopes.
+         * @return Scope parameter string (possibly empty).
+         */
+        @JvmStatic
+        fun computeScopeParameter(scopes: Array<String>?): String {
```
wmathurin, line 65: This is the critical change of this PR.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -681,38 +682,14 @@ public String getScope() {
...
-		if (TextUtils.isEmpty(scopeToCheck)) {
-			return false;
-		}
-		for (final String s : parseScopes()) {
-			if (scopeToCheck.equals(s)) {
-				return true;
-			}
-		}
-		return false;
+		return new ScopeParser(scope).hasScope(scopeToCheck);
```
brandonpage, line 692: ~~The syntax isn't as clean in Java, but I wonder if `ScopeParser` could just be a couple String[] extensions?~~

Edit:  There are several functions in `ScopeParser` so I see its need to be a class.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -94,11 +99,55 @@ internal suspend fun onAuthFlowComplete(
...
+        getRuntimeConfig(SalesforceSDKManager.getInstance().appContext),
+        ::updateLoggingPrefsHelper,
+        ::fetchUserIdentity,
+        ::startMainActivityHelper,
+        ::setAdministratorPreferences,
+        ::addAccountHelper,
+        ::handleScreenLockPolicy,
+        ::handleBiometricAuthPolicy,
+        ::handleDuplicateUserAccount
+    )
```
brandonpage: Since the original `onAuthFlowComplete` is internal you could just update the original and add all of the new values with these as the default parameters.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
brandonpage: Is not having the refresh token scope actually an invalid state/error?  Do you still get the refresh scope if the policy is set to expire immediately?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -357,3 +350,166 @@ private fun logAddAccount(account: UserAccount?) {
...
+}
+
+/**
+ * Helper method to start main activity.
+ */
+private fun startMainActivityHelper() {
+    SalesforceSDKManager.getInstance().appContext.startActivity(Intent(SalesforceSDKManager.getInstance().appContext, SalesforceSDKManager.getInstance().mainActivityClass).apply {
+        setPackage(SalesforceSDKManager.getInstance().appContext.packageName)
+        flags = FLAG_ACTIVITY_NEW_TASK
+    })
```
brandonpage: NIT:  Why remove the `with(SalesforceSDKManager.getInstance())` block?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/ScopeParser.kt
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Utility class for parsing and working with OAuth2 scopes.
+ */
+class ScopeParser {
+
+    companion object {
+        const val REFRESH_TOKEN = "refresh_token"
+        const val ID = "id"
+        private const val SINGLE_SPACE = " "
+        private const val EMPTY_STRING = ""
```
brandonpage: I don't dislike this, but I think we typically just use `""` in code like how 0 isn't really a magic number.  🤷

brandonpage (approved):
Looks good!  Really appreciate all of the tests and the effort put into making `onAuthFlowComplete` more testable!

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: We never allowed it. We would add the refresh_token scope during login if it was not in bootconfig. So login would fail if the connected app did not have it configured. 
I need to try and see what happens if the connected app does not have the scope and I don't check it and also what happens if it is set to expire immediately.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: I can put it back - it was more readable.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/ScopeParser.kt
wmathurin: Agreed, I will get rid of it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: @brandonpage 
I just tried the following:
- an ECA that does not have refresh_token and a client app that does not specify scopes in bootconfig (and without the check above) => login works / token end point response does not include a refresh token 
- an ECA with refresh token set to expire immediately => login works / token end point response includes a refresh token (but it will fail to refresh the access token when used)

I'm okay removing the check. We should include those scenarios in our test plan. Especially to make sure that having that null refresh token does not lead to some unwanted behaviors.

wmathurin:
The only test failure "LoginActivityTest.viewModelIsUsingFrontDoorBridge_DefaultValue_onCreateWithoutQrCodeLoginIntent" is unrelated.

PR #2784, opened 2025-10-08
@W-19843412: [MSDK Android] Login And Developer Information Activities Do Not Pad Display Cutouts

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This relatively easy and safe fix helps two of our activities which recently migrated to Compose also pad for the display cutouts.  I've attached a video walkthrough showing the old behavior and then the updated version, though it doesn't _actually_ show the cutout in the video.  One can see the new padding when watching carefully, though.

  I tested this on a Google Pixel 8 Pro with API 36.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.kt
@@ -31,17 +31,18 @@ import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.navigationBars
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.windowInsetsPadding
+import androidx.compose.foundation.layout.safeDrawing
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.material3.TopAppBar
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.kt
```diff
@@ -76,24 +87,17 @@ class DevInfoActivity : ComponentActivity() {
...
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun DevInfoScreen(devInfoList: List<Pair<String, String>>) {
-    Column(
+fun DevInfoScreen(
+    paddingValues: PaddingValues,
+    devInfoList: List<Pair<String, String>>,
+) {
+    LazyColumn(
+        contentPadding = paddingValues,
```
JohnsonEricAtSalesforce, line 95: For this activity, the crux of the fix is here and follows the Compose guide for Scaffold.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.kt
```diff
@@ -64,7 +66,16 @@ class DevInfoActivity : ComponentActivity() {
 
         setContent {
             MaterialTheme(colorScheme = SalesforceSDKManager.getInstance().colorScheme()) {
-                DevInfoScreen(devInfoList)
+                Scaffold(
```
JohnsonEricAtSalesforce, line 69: @wmathurin, is there any reason this activity didn't use Scaffold before? It will gain a lot of the required padding and layout functionality out-of-the-box if it switches to Scaffold.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -217,16 +221,17 @@ internal fun LoginView(
     )
 
     Scaffold(
-        topBar = topAppBar,
         bottomBar = bottomAppBar,
+        contentWindowInsets = WindowInsets.safeDrawing,
```
JohnsonEricAtSalesforce, line 225: The crux of the fix for login activity is here since it's already using a Scaffold.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -217,16 +221,17 @@ internal fun LoginView(
...
+        topBar = topAppBar,
     ) { innerPadding ->
         if (loading) {
             loadingIndicator()
         }
-
         // Load the WebView as a composable
         AndroidView(
             modifier = Modifier
+                .background(dynamicBackgroundColor.value)
```
JohnsonEricAtSalesforce, line 234: Note, to keep the cutout padding matching the web view (and not showing as black) we need the view model's dynamic web view background color here.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/DevInfoActivity.kt
wmathurin: No reasons.

PR #2785, opened 2025-10-09
@W-19368406: [MSDK] Local dev instant login (Android)

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This adds a new activity which may be used for debug app builds only to authenticate a user via test credentials in automated tests.  The new activity will finish without any action for app release builds.  This is controlled by Salesforce Mobile SDK's existing `isDebugBuild` property.

  This new means of automating authentication for debug testing does not expose production builds to risk and also minimizes access since usernames and passwords are not used.  The test credentials JSON provided by Salesforce Mobile SDK's REST Explorer sample app uses refresh tokens which can be remotely expired and cannot be used for privilege elevation beyond the public APIs.

  From the shell, the new intent may be started using ADB.
```
$ adb shell
$ am start -n <App Package Namepace>/com.salesforce.androidsdk.util.test.TestAuthenticationActivity -e creds '<Credentials JSON>'
```

The "App Package Namespace" is the app's package namespace, which is usually declared in the app module's Gradle script via `android { namespace`.

The "Credentials JSON" can be obtained by running the existing Salesforce Mobile SDK REST Explorer sample app, logging in and then using the "Dev Menu" "Export Credentials To Clipboard" menu item.

Note that the shell invocation must preserve quotations and other potentially interpreted characters from the shell such that the Credentials JSON is received by the activity without alteration.

The new intent could also be launched by UIAutomator2.

```
intent.setComponent(ComponentName("<App Package Namespace>", "com.salesforce.androidsdk.util.test.TestAuthenticationActivity"))
intent.putExtra("creds", "<Credentials JSON>")
context.startActivity(intent)
```

Appium is another means of launching the new intent.

```
capabilities = {
       "platformName": "Android",
       "deviceName": "Android Emulator",
       "appPackage": "<App Package Namespace>",
       "appActivity": "com.salesforce.androidsdk.util.test.TestAuthenticationActivity",
       "appium:processArguments": {
           "args": ["creds", "<Credentials JSON>"]
       }
   }
```


libs/SalesforceSDK/src/debug/AndroidManifest.xml
```diff
@@ -2,4 +2,15 @@
 
     <!-- Debug-Only: Post notifications permission supports the show developer support notification. -->
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+
+    <application>
+
+        <!-- Debug-Only: Test Authentication Activity For Automated Testing -->
```
wmathurin: What's keeping it from being used in production? Should we check that we are in a debug build in the code?

JohnsonEricAtSalesforce: We ought to be able to use the `SalesforceSDKManager` `isDebugBuild` check in the activity and simply finish without action for other build types.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/LoginViewActivityTest.kt
```diff
@@ -118,11 +119,10 @@ class LoginViewActivityTest {
 
     @Test
     fun topAppBar_ChangeServerButton_OpensServerPicker() {
-        var showPicker: MutableState<Boolean>? = null
+        val showPicker = mutableStateOf(false)
```
JohnsonEricAtSalesforce, line 122: This subtle change resolves inspector warnings and should be acceptable for tests.  I'd used the same style where dynamic background color is now used as well.

PR #2788, opened 2025-10-15
@W-19964705: [MSDK Android] Push Notification Registration Failure For User Accounts With Community URL

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This corrects the `network_id` field when registering for SFDC push notifications, which was incorrectly set to the community URL instead of the community id.

### DIFF
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/PushServiceTest.kt
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/PushServiceTest.kt
@@ -35,6 +35,7 @@ import com.salesforce.androidsdk.rest.RestClient.ClientInfo
 import com.salesforce.androidsdk.rest.RestResponse
 import io.mockk.every
 import io.mockk.mockk
+import io.mockk.verify
 import kotlinx.serialization.json.Json.Default.encodeToJsonElement
 import kotlinx.serialization.json.Json.Default.encodeToString
 import kotlinx.serialization.json.JsonArray
@@ -738,6 +739,7 @@ class PushServiceTest {
         val restClient = mockk<RestClient>()
         every { restClient.sendSync(any()) } returns restResponse
 
+        val account = createTestAccount()
         var actualStatus: Int? = null
         val actualId = object : PushService() {
             override fun onPushNotificationRegistrationStatus(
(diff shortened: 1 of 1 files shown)


wmathurin:
Maybe we could add tests for this method by using the approach used in [AuthenticationUtilities](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt#L99) / [AuthenticationUtilitiesTest](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/AuthenticationUtilitiesTest.kt)
 where we have a bunch of parameters  that take lambda functions but default to calling existing helper methods?

JohnsonEricAtSalesforce:
> Maybe we could add tests for this method by using the approach used in [AuthenticationUtilities](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt#L99) / [AuthenticationUtilitiesTest](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/AuthenticationUtilitiesTest.kt) where we have a bunch of parameters that take lambda functions but default to calling existing helper methods?

As it turns out, I wrote a set of tests around `PushService` earlier in the yet.  We could add a test for this value like this commit 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2788/commits/d914477adebc8052f5e4c920bdc15fb6f2f8ca32

mani516:
@wmathurin @JohnsonEricAtSalesforce 

Hi there!

Thanks for identifying and fixing this one. However it seems this fix is not yet available in the latest release i.e., 13.1.1

When can we expect this fix to be released? This is blocking our sdk upgrade from 11.1.0, to support 16kb page sizes, for which the extended deadline is 31 May 2026.

Although, it looks like the above change was related to a code change which was done in 13.1.0 - https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/5e40cff4dcd73c4a851030ed4da4870951cef606

But we were facing issue while migrating to 13.0.2 also (from 11.1.0).

Usecase 1

1. Install app with salesforce sdk 11.1.0
2. Launch app
3. setDelayRegistrationUntilContactKeyIsSet is set to true
4. SFMCSdk initStatus returns Success (1)
5. Login the User
6. Update the app with Salesforce Sdk 13.0.2
7. Launch app
8. SFMCSdk initStatus returns Failure (-1)

Usecase 2

1. Install app with salesforce sdk 11.1.0
2. Launch app
3. setDelayRegistrationUntilContactKeyIsSet is set to true
4. SFMCSdk initStatus returns Success (1)
5. Dont login the user
6. Update the app with Salesforce Sdk 13.0.2
7. Launch app
8. SFMCSdk initStatus still returns Success (1)
9. Login the user
10. Relaunch the app
11. SFMCSdk initStatus returns Failure (-1)

PR #2823, opened 2026-01-20
@W-20978376: User Story - [Android] Remove supportWelcomeDiscovery flag in dev

JohnsonEricAtSalesforce:
🥁 _*Ready For Review*_ 🎸

  This catches 13.2.0 `dev` up with today's commits to 13.1.1 `master`

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -1594,9 +1594,16 @@ open class LoginActivity : FragmentActivity() {
         private val activity: LoginActivity = this@LoginActivity
     ) : Observer<String> {
         override fun onChanged(value: String) {
-            // Guard against observing a pending login server already provided by the intent data, such as a Salesforce Welcome Discovery mobile URL.
             val pendingServerUri = value.toUri()
-            if ((activity.intent.data?.host == pendingServerUri.host && activity.intent.data?.path == pendingServerUri.path) || activity.intent.getStringExtra(EXTRA_KEY_LOGIN_HOST) == pendingServerUri.host) {
+            val intent = activity.intent
+            val data = intent.data
+
+            // Guard against observing a pending login server already provided by the intent data, such as a Salesforce Welcome Discovery mobile URL.
+            val dataHostMatches = data?.host == pendingServerUri.host
+            val dataPathMatches = data?.path == pendingServerUri.path
+            val uriMatches = dataHostMatches && dataPathMatches
+            val extraMatches = intent.getStringExtra(EXTRA_KEY_LOGIN_HOST) == pendingServerUri.host
+            if (uriMatches || extraMatches) {
                 activity.viewModel.previousPendingServer = value
                 return
             }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -111,7 +111,6 @@ import com.salesforce.androidsdk.R.color.sf__background_dark
 import com.salesforce.androidsdk.R.color.sf__primary_color
 import com.salesforce.androidsdk.R.drawable.sf__action_back
 import com.salesforce.androidsdk.R.string.cannot_use_another_apps_login_qr_code
-import com.salesforce.androidsdk.R.string.salesforce_welcome_is_disabled
 import com.salesforce.androidsdk.R.string.sf__biometric_opt_in_title
 import com.salesforce.androidsdk.R.string.sf__generic_authentication_error_title
 import com.salesforce.androidsdk.R.string.sf__jwt_authentication_error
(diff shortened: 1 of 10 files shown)


wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -572,7 +572,9 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
     ) : Observer<String?> {
         override fun onChanged(value: String?) {
             if (!sdkManager.isBrowserLoginEnabled && !viewModel.isUsingFrontDoorBridge && value != null) {
-                val isNewServer = viewModel.loginUrl.value?.startsWith(value) != true
+                val valueUrl = value.toUri()
+                val loginUrl = viewModel.loginUrl.value?.toUri()
+                val isNewServer = loginUrl?.host != valueUrl.host || loginUrl?.path != valueUrl.path
```
brandonpage: This looks like an unrelated behavior change.  Why is it here?

brandonpage (dismissed):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
wmathurin: It's a fix for W-20935841 that we ran into when testing WSC and WSC/disco login. The fix was done in master in the same PR that removed supportWelcomeDiscovery.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
brandonpage: Ah, thank you I wasn't aware of that fix.  Having both welcome.salesforce.com and welcome.salesforce.com/discovery seems like a very niche issue.  And I know for a fact that I explicitly used `startsWith` instead of directly comparing the path to prevent bugs where login would fail from running `getAuthorizationUrl` when we shouldn't have as that causes the PKCE code and verifier to be set with new values.  Perhaps the scenario this was intended to prevent does not exist anymore but I'm not 100% sure.

brandonpage: Circling back:  
- The `startsWith` is _not_ a carry over from the old `OAuthWebviewHelper`.  
- Looking at the current code, `LoginUrlSource` is only ever called when `selectedServer` changes and I do not see a scenario.  If it does not happen mid authentication there should not be any issues with getting new code/verifier.  

Seems like whatever issue this did avoid is no longer possible.

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1596,7 +1597,30 @@ open class LoginActivity : FragmentActivity() {
...
+            if (data?.path?.equals(pendingServerUri.path) == false) {
+                Log.i("CodeCov", "3.1")
+            }
+            if (intent.getStringExtra(EXTRA_KEY_LOGIN_HOST).equals(pendingServerUri.host)) {
+                Log.i("CodeCov", "4.0")
+            }
+            if (!(intent.getStringExtra(EXTRA_KEY_LOGIN_HOST).equals(pendingServerUri.host))) {
+                Log.i("CodeCov", "4.1")
+            }
+            if (((data?.host?.equals(pendingServerUri.host) == true).and(data?.path?.equals(pendingServerUri.path) == true)).or(intent.getStringExtra(EXTRA_KEY_LOGIN_HOST).equals(pendingServerUri.host))) {
```
brandonpage: @JohnsonEricAtSalesforce Was this intended to be checked in?

Edit:  I see the  "(CodeCov Testing - Hold Reviewing Commit)" now.

JohnsonEricAtSalesforce: Thanks for looking for that.  This commit will be replaced with a much simpler commit once I get full coverage.  Even with Windsurf helping, getting through all the logic branches and minimizing their count is still tedious.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
JohnsonEricAtSalesforce: Wow!  Thanks for the detailed analysis ⭐️⭐️⭐️⭐️⭐️

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
JohnsonEricAtSalesforce: @brandonpage - Here's the finalized commit for code coverage, assuming it still produces the coverage I was looking for. https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2823/changes/16daac020380a66be09456e601bd7aa9eab90f24

Most of the updates were suggested by or reviewed by Windsurf.  I wrote the tests by hand, though, since the generated code was overly verbose it seemed.

PR #2827, opened 2026-01-22
Properly handle UI Test Config.

brandonpage:
This PR moves the expected location of the `ui_test_config.json` file to `shared/test/` to match `test_credentials.json` and adds a `ui_test_config.json.sample` file.

### DIFF
--- a/.github/workflows/reusable-ui-workflow.yaml
+++ b/.github/workflows/reusable-ui-workflow.yaml
@@ -10,9 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BUNDLE_GEMFILE: ${{ github.workspace }}/.github/DangerFiles/Gemfile
-      CI_USER_USERNAME: ${{ secrets.CI_USER_USERNAME }}
-      CI_USER_PASSWORD: ${{ secrets.CI_USER_PASSWORD }}
-      TEST_CONFIG: ${{ secrets.UI_TEST_CONFIG }}
     steps:
       - uses: actions/checkout@v4
         if: ${{ inputs.is_pr }}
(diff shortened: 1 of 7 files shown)


shared/test/ui_test_config.json.sample
```diff
@@ -0,0 +1,11 @@
+{
```
wmathurin, line 1: There is also a users section - see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/test_config.json.sample

.github/workflows/reusable-ui-workflow.yaml
```diff
@@ -27,10 +24,11 @@ jobs:
       - name: Install Dependencies
         env:
           TEST_CREDENTIALS: ${{ secrets.TEST_CREDENTIALS }}
+          UI_TEST_CONFIG: ${{ secrets.UI_TEST_CONFIG }}
         run: |
           ./install.sh
           echo $TEST_CREDENTIALS > ./shared/test/test_credentials.json
-          echo $TEST_CONFIG > ./native/NativeSampleApps/AuthFlowTester/src/main/assets/test_config.json
+          echo $UI_TEST_CONFIG > ./shared/test/ui_test_config.json
```
wmathurin, line 31: 👍 ui_test_config is a better name than test_config.
I should rename on iOS also (and change the github actions there also)

wmathurin (approved):

.github/workflows/reusable-ui-workflow.yaml
brandonpage: I debated that name change because I didn't know if this could potentially be used to remove all consumer keys from the repo.  But we can rethink about the naming if we do that in the future.

PR #2832, opened 2026-01-23
@W-21006714: singleServerCustomTabActivity is malfunctioning (13.2)

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This is the 13.2 counterpart for https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2831

  This resolves issues with the function of the `singleServerCustomTabActivity` property which was introduced in MSDK 13.0.0.  The logic worked as intended at that time, but had regressed partially in 13.1.0 and fully in the pending 13.1.1 release.

  The MSDK sample apps behave in exactly the same way with this update as they did in 13.1.0.  Here's a video that demonstrates that in the REST Explorer sample app.  Note that the activity timing in this sample app isn't perfect since this app is not built as the target app would be and defaults to the login view instead of an app activity.
[Screen_recording_20260122_180115.webm](https://github.com/user-attachments/assets/8157c690-4f35-416e-a2a1-13116cb71414)

  Also, this property now has detailed unit tests and high code coverage to prevent future regression.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -1547,6 +1547,7 @@ open class LoginActivity : FragmentActivity() {
                 if (activity.viewModel.singleServerCustomTabActivity) {
                     // Show blank page and spinner until PKCE is done.
                     activity.viewModel.loginUrl.value = ABOUT_BLANK
+                    finish()
                 } else {
                     // Don't show server picker if we are re-authenticating with cookie.
                     activity.clearWebView(showServerPicker = !activity.sharedBrowserSession)
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -215,7 +215,7 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
      * Since browser based authentication requires PKCE (and therefore the Web Server flow) the User Agent flow
      * cannot be used while in this mode.
      */
-    open val singleServerCustomTabActivity = true
+    open val singleServerCustomTabActivity = false
 
     /** Value representing if the back button should be shown on the login view. */
     open val shouldShowBackButton = with(SalesforceSDKManager.getInstance()) {
@@ -622,7 +622,8 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         private val scope: CoroutineScope = viewModelScope,
     ) : Observer<String> {
         override fun onChanged(value: String) {
-            if ((sdkManager.isBrowserLoginEnabled || singleServerCustomTabActivity) && !viewModel.isUsingFrontDoorBridge) {
+            val useBrowserCustomTab = sdkManager.isBrowserLoginEnabled || singleServerCustomTabActivity
+            if (useBrowserCustomTab && !viewModel.isUsingFrontDoorBridge) {
                 scope.launch {
                     viewModel.browserCustomTabUrl.value = viewModel.getAuthorizationUrl(
                         server = value,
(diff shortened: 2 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -215,7 +215,7 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
      * Since browser based authentication requires PKCE (and therefore the Web Server flow) the User Agent flow
      * cannot be used while in this mode.
      */
-    open val singleServerCustomTabActivity = false
+    open val singleServerCustomTabActivity = true
```
brandonpage: This looks like an accidental addition.

JohnsonEricAtSalesforce: Thanks.  I'd already noticed that as well while this is in draft.  I'll have that any much more cleaned up before this is ready for final review and out of draft status.

I'm actually reviewing this in more detail against `dev` since we have much, much better tests there.  I'd recommend holding any further review until this is marked complete.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -571,7 +571,7 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         private val scope: CoroutineScope = viewModelScope,
     ) : Observer<String?> {
         override fun onChanged(value: String?) {
-            if (!sdkManager.isBrowserLoginEnabled && !viewModel.isUsingFrontDoorBridge && value != null) {
+            if (!sdkManager.isBrowserLoginEnabled && !singleServerCustomTabActivity && !viewModel.isUsingFrontDoorBridge && value != null) {
```
JohnsonEricAtSalesforce, line 574: Part 1 is making sure `loginUrl` is not set when using `singleServerCustomTabActivity` since the web view will not be used.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -622,7 +622,8 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         private val scope: CoroutineScope = viewModelScope,
     ) : Observer<String> {
         override fun onChanged(value: String) {
-            if (sdkManager.isBrowserLoginEnabled && !viewModel.isUsingFrontDoorBridge) {
+            val useBrowserCustomTab = sdkManager.isBrowserLoginEnabled || singleServerCustomTabActivity
```
JohnsonEricAtSalesforce, line 625: Part 2 is making sure `customBrowserTabUrl` is set when using `singleServerCustomTabActivity`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1547,6 +1547,7 @@ open class LoginActivity : FragmentActivity() {
                 if (activity.viewModel.singleServerCustomTabActivity) {
                     // Show blank page and spinner until PKCE is done.
                     activity.viewModel.loginUrl.value = ABOUT_BLANK
+                    finish()
```
JohnsonEricAtSalesforce, line 1550: Part 3 is restoring the missing finish call in the activity result.

The repeated cycle of tapping back and returning to a brand new instance of `LoginActivity` which re-displays the browser custom tab is because when `LoginActivity` finishes without authenticating the default behavior of `SalesforceActivity.onResume()` is to detect there is no authenticated REST Client and restart `LoginActivity`.  That's also what caused both 13.0.0 and now this version to briefly show a new `LoginActivity` while authentication is occurring in the custom tab.

The behavior here now matches 13.0.0 so I reckon it resolves the bug.  Would we ever need this to behave better in an app that defaults to login?

brandonpage (approved):
LGTM.  Thanks you for the tests!

PR #2834, opened 2026-02-04
New feature flag for welcome discovery (WD)

wmathurin:


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -901,6 +902,20 @@ open class LoginActivity : FragmentActivity() {
      */
     private fun applySalesforceWelcomeDiscoveryIntent(intent: Intent) {
```
wmathurin, line 905: @JohnsonEricAtSalesforce Does this look like the right place?
From my debugging, this method always gets called: for regular logins, and also for 1st or 2nd part of login with welcome discovery.

wmathurin:
I did not add unit tests.
We could add user agent validation in our tests against AuthFlowTester though.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -901,6 +902,20 @@ open class LoginActivity : FragmentActivity() {
...
+                .registerUsedAppFeature(FEATURE_WELCOME_DISCOVERY_LOGIN);
+        }
+        else {
+            SalesforceSDKManager.getInstance().unregisterUsedAppFeature(
+                FEATURE_WELCOME_DISCOVERY_LOGIN
+            );
+        }
+
+        // Re-apply user agent to WebView
+        webView.settings.userAgentString = SalesforceSDKManager.getInstance().userAgent
```
brandonpage: Not for this PR, just thinking out loud: It would be awesome to use LiveData to make this happen automatically so we don't accidentally forget to update the WebView.  Just like how the login url the WebView uses gets automatically generated when the selected server changes, registering/unregistering a feature flag (and potentially other things) should regenerate and set the value in the WebView.

brandonpage (approved):

brandonpage:
@wmathurin Looks like the `LoginActivityScenarioTest.testWebviewSettings` test needs to be updated to reflect the change to the user agent string.

wmathurin:
> @wmathurin Looks like the `LoginActivityScenarioTest.testWebviewSettings` test needs to be updated to reflect the change to the user agent string.

Fixed by [68cfc80](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2834/commits/68cfc80278ef819930fff5d7d2977fb7f15cde48)

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/LoginActivityScenarioTest.kt
```diff
@@ -375,4 +377,78 @@ class LoginActivityScenarioTest {
             assertTrue(activity.switchDefaultOrSalesforceWelcomeDiscoveryLogin(PRODUCTION_LOGIN_URL.toUri()))
         }
     }
+
```
wmathurin, line 380: New tests checking the feature flag on the user agent for various scenario:
- welcome discovery login server
- my domain login server hint (that happens in part 2 of wsc disco flow)
- my domain login server (should not have the WD feature flag)

wmathurin:
All 574 unit tests are passing - see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2834/checks?check_run_id=62694114940

PR #2835, opened 2026-02-04
@W-19368426: [MSDK] Rewrite Android Screen Lock (a.k.a Passcode) for RTL UI

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This updates the screen lock setup activity to support right-to-left text layouts along with a much heftier effort to migrate from Java to Kotlin, layout using Jetpack Compose and provide comprehensive unit tests.

  All deprecations are resolved as well and the sources should have no inspector warnings.

  To test, provide the strings used by the activity in an RTL language such as Arabic.  Then, set the system language to that language.


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginServerListItem.kt
```diff
@@ -91,6 +91,7 @@ import com.salesforce.androidsdk.R.string.sf__server_picker_click_label
 import com.salesforce.androidsdk.R.string.sf__server_remove_content_description
 import com.salesforce.androidsdk.R.string.sf__server_url_delete
 import com.salesforce.androidsdk.config.LoginServerManager.LoginServer
+import com.salesforce.androidsdk.ui.PADDING_SIZE
```
brandonpage, line 94: The local `PADDING_SIZE` is not removed.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,457 @@
...
+import com.salesforce.androidsdk.security.ScreenLockManager.Companion.SCREEN_LOCK
+import com.salesforce.androidsdk.ui.ScreenLockViewModel.Companion.Factory
+import com.salesforce.androidsdk.ui.components.ScreenLockView
+import com.salesforce.androidsdk.util.SalesforceSDKLogger
+
+/**
+ * An activity that locks the app behind the operating system's provided
+ * authentication.
+ */
+internal class ScreenLockActivity : FragmentActivity() {
```
brandonpage: Technically for Semver this should be `open class ScreenLockActivity` (public).  It wasn't possible for us to use a subclass so it shouldn't matter.  I don't think this was a bad decision, just throwing it out there for the sake of completeness and conversation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockViewModel.kt
```diff
@@ -0,0 +1,105 @@
...
+ * intended for testing purposes only.  Defaults to appropriate values for the
+ * current Android SDK
+ * @param build The Android SDK build. This parameter is intended for testing
+ * purposes only. Defaults to the current Android SDK build
+ */
+internal class ScreenLockViewModel(
+    build: Int = SDK_INT,
+    // TODO: Test coverage needed. ECJ20260205
+    val appName: String = getInstance().appName ?: "App",
+) : ViewModel() {
```
brandonpage: A ViewModel seems like overkill IMO.  Is there a reason it is needed?  

Throwing that out for consideration before you spend a ton of time writing tests.

brandonpage (commented):
Just a couple high level comments while you are in draft.  👍

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
JohnsonEricAtSalesforce: I went ahead and made it public again as I'm getting ready to mark this ready for review soon ✅

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockViewModel.kt
JohnsonEricAtSalesforce: I was wondering as well since at first the view model was relatively light and much of the logic related to screen lock is at the activity/view level.  That said, there are some observables and they will benefit from the life cycle management the view model offers.  It will also be ready should this grow in complexity.  So, I'm leaning towards keeping the view model unless we feel a compelling reason not to.

JohnsonEricAtSalesforce:
Code coverage is 96.60%.  There are a handful of Jetpack Compose function call sites that are beyond the ability of JaCoCo to probe and that's expected from the research I've done.  They are tested by the new tests though and I used tooling to validate that.

The tests I created are comprehensive and should provide reasonable assertions for the behavior migrated from Java.

There are a few places where I introduced new ways of tidying the code so it provides better coverage numbers and generates less noise in the code coverage reports.

JohnsonEricAtSalesforce:
Here're screenshots of the English LTR and Arabic RTL I tested with.
20260206 Screen Lock In Arabic RTL
20260206 Screen Lock In English LTR

libs/SalesforceSDK/res/layout/sf__screen_lock.xml
```diff
@@ -1,68 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
JohnsonEricAtSalesforce, line 3: I did a side-by-side comparison of this XML with the new Compose logic plus compared screenshots of the two versions for parity.

libs/SalesforceSDK/res/values/sf__dimens.xml
```diff
@@ -1,11 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <resources>
-    <!-- Dimensions used in sf__screen_lock.xml -->
```
JohnsonEricAtSalesforce, line 4: I believe these were the only orphaned dimensions after removing the XML layout.

libs/SalesforceSDK/res/values/sf__styles.xml
```diff
@@ -206,16 +206,6 @@
         <item name="android:windowLightStatusBar">false</item>
     </style>
 
-    <style name="SalesforceSDK.ScreenLock.ErrorMessage">
```
JohnsonEricAtSalesforce, line 209: This style was also orphaned when the XML layout for screen lock was removed.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/preview/ScreenLockViewPreview.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.salesforce.androidsdk.R
+import com.salesforce.androidsdk.ui.ScreenLockActivity.Companion.noOp
+import com.salesforce.androidsdk.ui.ScreenLockViewModel
+import com.salesforce.androidsdk.ui.components.ScreenLockView
+import com.salesforce.androidsdk.util.test.ExcludeFromJacocoGeneratedReport
+
+@ExcludeFromJacocoGeneratedReport
```
JohnsonEricAtSalesforce: At a suggestion from our tooling, I moved the preview to a separate source.  There were some proposed benefits, though I later realized those didn't manifest.  That said, it really did look nice having this non-release code extracted so I left it that way.

To exclude this from code coverage, one suggestion was to move previews to the `debug` source root.  However, I then realized we're building and generating coverage from the debug variant anyways.  So, that became a moot point.  It was interesting to note that the `ExcludeFromJacocoGeneratedReport` annotation only applies to the method and not to the default parameter values.  That was when I started looking for a way to exclude the entire source.

For now, this seems fine.  Perhaps we revisit the variants and directory layout in the future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/preview/ScreenLockViewPreview.kt
```diff
@@ -0,0 +1,51 @@
...
+@Preview(showBackground = true)
+@Composable
+fun ScreenLockViewPreview() {
+    ScreenLockView(
+        appName = "App",
+        innerPadding = PaddingValues.Absolute(0.dp),
+        appIcon = painterResource(id = R.drawable.sf__salesforce_logo),
+        viewModel = ScreenLockViewModel(),
+        logoutAction = ::noOp,
```
JohnsonEricAtSalesforce: This was `logout = {}` for a time, but of course JaCoCo flagged the anonymous function as uncovered.  So, I created a utility `noOp` function that is trivially tested elsewhere and eliminates the coverage noise.  It seemed like an easy win.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginServerListItem.kt
JohnsonEricAtSalesforce: When I was basing the new `ScreenLockActivity` off other recent Compose implementations, I noticed this `PADDING_SIZE` and `CORNER_RADIUS` where used in a lot of disparate sources though still defined in `LoginView.kt`.  Doesn't it seem they should be in a common location? I created `UserInterfaceConstants.kt` but we could do something different.  There may be other values as well.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.salesforce.androidsdk.ui.PADDING_SIZE
+import com.salesforce.androidsdk.ui.ScreenLockViewModel
+import com.salesforce.androidsdk.ui.theme.hintTextColor
+
+/**
+ * The set up view when screen lock is required and not enrolled.
+ */
+@Composable
+internal fun ScreenLockView(
+    appIcon: Painter,
```
JohnsonEricAtSalesforce, line 70: The `Painter` reference here is a little interesting, but it was a way to pass in the programmatically derived drawable instance for the app icon which doesn't actually have a resource id.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
```diff
@@ -0,0 +1,134 @@
...
+    appName: String,
+    innerPadding: PaddingValues,
+    logoutAction: () -> Unit,
+    viewModel: ScreenLockViewModel,
```
JohnsonEricAtSalesforce: Yes, trailing commas everywhere to reduce future LoC diff.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
```diff
@@ -0,0 +1,134 @@
...
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(innerPadding),
+    ) {
+
+        // Log out button.
+        Button(
+            colors = ButtonColors(
+                containerColor = Transparent,
```
JohnsonEricAtSalesforce, line 85: I based all the colors, fonts and padding from other recently added Compose layouts.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
```diff
@@ -0,0 +1,134 @@
...
+                contentColor = colorScheme.hintTextColor,
+                disabledContainerColor = Transparent,
+                disabledContentColor = colorScheme.surfaceVariant,
+            ),
+            contentPadding = PaddingValues(PADDING_SIZE.dp),
+            enabled = true,
+            modifier = Modifier.padding(PADDING_SIZE.dp),
+            onClick = logoutAction,
+            shape = RoundedCornerShape(CORNER_RADIUS.dp),
+        ) { Text(color = colorScheme.primary, fontWeight = Normal, fontSize = 17.sp, text = stringResource(sf__logout)) }
```
JohnsonEricAtSalesforce, line 95: For Compose function call sites that JaCoCo doesn't probe well, I did make them a single line to reduce the impact noise on the repo's stats.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
```diff
@@ -0,0 +1,134 @@
...
+            painter = appIcon,
+        )
+
+        Spacer(modifier = Modifier.weight(1f))
+
+        // Setup message text.
+        Text(
+            fontSize = 14.sp,
+            modifier = Modifier.align(CenterHorizontally),
+            text = viewModel.setupMessageText.value ?: stringResource(sf__screen_lock_setup_required, appName),
```
JohnsonEricAtSalesforce, line 114: The setup message and setup button label became the best justification for keeping the view model, in the end.  They're dynamic.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginOptionsActivity.kt
```diff
@@ -1,3 +1,29 @@
+/*
```
JohnsonEricAtSalesforce, line 1: I added the copyright header here since it was missing.  We do want those everywhere, correct?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.java
```diff
@@ -1,295 +0,0 @@
...
-    private Button logoutButton;
-    private Button actionButton;
-
-    @Override
-    public void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-        EdgeToEdge.enable(this);
-
-        // Protect against screenshots.
-        getWindow().setFlags(WindowManager.LayoutParams.FLAG_SECURE,
```
JohnsonEricAtSalesforce, line 89: I carefully migrated all the logic from the previous activity line-by-line and method-by-method to the new activity in a best attempt to avoid missing anything.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+    internal val viewModel: ScreenLockViewModel
+            by viewModels { Factory }
+
+    /** The activity result for the biometric setup */
+    private val biometricSetupActivityResultLauncher = registerForActivityResult(StartActivityForResult(), ::onBiometricSetupActivityResult)
+
+    @Override
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        create()
```
JohnsonEricAtSalesforce, line 112: I really wanted to be able to pull the data levers on the create logic for top-notch `onCreate` unit tests and coverage, but being a lifecycle method there's no opening to it.  So, I moved all the logic to a private/VisibleForTest method with parameters.  That really worked like a charm, I thought.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /**
+     * Implements the creation of the activity.
+     * @param build The Android SDK build. This parameter is intended for
```
JohnsonEricAtSalesforce: The commentary is verbose, yes, but being professional open source I'm really hoping it helps others know the intent and not abuse the contract of the code.  I have often felt it is difficult to interpret the idea behind much of the code in the library, so this could help.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+     * intended for testing purposes only. Defaults to the current
+     * PackageManager instance
+     * @param sdkManager The SalesforceSDKManager to use. This parameter is
+     * intended for testing purposes only. Defaults to the current
+     * SalesforceSDKManager instance
+     */
+    @SuppressLint("NewApi")
+    @VisibleForTesting
+    internal fun create(
+        build: Int = SDK_INT,
```
JohnsonEricAtSalesforce: This activity has a lot of SDK-dependent logic, so I added testable parameters where needed.  The bad thing is that it also requires a "suppress".  Used with discipline, it seems alright and gets that path automatically tested.  I'd be open to hear other ideas on that, though.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+        packageManager: PackageManager = this.packageManager,
+        sdkManager: SalesforceSDKManager = getInstance(),
```
JohnsonEricAtSalesforce, line 131: It's too big a topic for a pull request thread, but I'll go on a limb and wonder if this level of testability might require a dependency injection tool like Hilt.  I used it quite a bit and it was a love 'n hate experience.  It automates this type of test set up a lot, though.  It could a great retro topic after this pull request.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+    @SuppressLint("InlinedApi") // TODO: Remove when min API > 29.
+    @VisibleForTesting
+    internal fun presentBiometricAuthentication(
+        biometricManager: BiometricManager = BiometricManager.from(this),
+        biometricPrompt: BiometricPrompt = getBiometricPrompt(),
+        biometricSetupActivityResultLauncher: ActivityResultLauncher<Intent> = this.biometricSetupActivityResultLauncher,
+        build: Int = SDK_INT,
+    ) {
+        when (biometricManager.canAuthenticate(viewModel.biometricAuthenticators())) {
+            BIOMETRIC_ERROR_NO_HARDWARE, BIOMETRIC_ERROR_SECURITY_UPDATE_REQUIRED, BIOMETRIC_ERROR_UNSUPPORTED, BIOMETRIC_STATUS_UNKNOWN -> {
```
JohnsonEricAtSalesforce, line 236: These codes are not a type safe enumeration.  Should we just make this an `else` so it cannot fall through?  @brandonpage, thoughts?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+
+    companion object {
+        private const val TAG = "ScreenLockActivity"
+
+        /**
+         * An empty function.
+         */
+        internal fun noOp() {}
```
JohnsonEricAtSalesforce: Here's a dandy, code-covered no-op function (which Kotlin doesn't provide) so we can reduce JaCoCo noise.  Perhaps this could live in the new `UserInterfaceConstants.kt` or elsewhere?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockViewModel.kt
```diff
@@ -0,0 +1,112 @@
...
+    // region Computed Values
+
+    /**
+     * The displayed name of the app.
+     * @param sdkManager The SalesforceSDKManager to use. This parameter is
+     * intended for testing purposes only. Defaults to the current
+     * SalesforceSDKManager instance
+     * @return The displayed name of the app
+     */
+    fun appName(
```
JohnsonEricAtSalesforce, line 80: This became more testable as a function than a computed property.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockViewModel.kt
```diff
@@ -0,0 +1,112 @@
...
+        sdkManager: SalesforceSDKManager = SalesforceSDKManager.getInstance(),
+    ) = sdkManager.appName ?: "App"
+
+    /**
+     * Returns the biometric authenticators to use.
+     * @param build The Android SDK build. This parameter is intended for
+     * testing purposes only. Defaults to the current Android SDK build
+     */
+    fun biometricAuthenticators(
```
JohnsonEricAtSalesforce, line 89: This also became more testable as a function than a computed property.

libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -2,15 +2,15 @@
...
+	android:versionCode="91"
 	android:versionName="13.2.0.dev">
 
     <uses-permission android:name="android.permission.GET_ACCOUNTS" />
     <uses-permission android:name="android.permission.USE_CREDENTIALS" />
     <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
     <uses-permission android:name="android.permission.USE_BIOMETRIC" />
 
-    <application>
+    <application android:supportsRtl="true">
```
JohnsonEricAtSalesforce, line 13: Wow, so after the Jetpack Compose migration this was actually the only task left.  Note, without this the app gets into a strange state where it _will_ use the RTL text but used the LTR layout.  It looks _really_ odd.

I reckon this will resource-merge through to the applications and in this case that means all application modules that depend on SalesforceSDK.  Is that acceptable? Would we want to leave this _off_ here and let the app modules take responsibility for adding it?  Is there a case where an app module could be negatively affected? They could override as needed.  I always consider that when updating a library module such as this.

libs/SalesforceSDK/build.gradle.kts
```diff
@@ -22,6 +22,7 @@ dependencies {
     api("androidx.browser:browser:1.8.0") // Update requires API 36 compileSdk
     api("androidx.work:work-runtime-ktx:2.10.3")
 
+    implementation("com.google.accompanist:accompanist-drawablepainter:0.37.3")
```
JohnsonEricAtSalesforce, line 25: Here's some Google code that lets us "remember" the programmatically-derived Drawable for the app icon.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/ScreenLockActivityScenarioTest.kt
```diff
@@ -0,0 +1,936 @@
...
+            activity.sendAccessibilityEvent(
+                accessibilityManager = any(),
+                build = any(),
+                eventText = capture(sendAccessibilityCapturingSlot)
+            )
+        }
+        assertTrue(sendAccessibilityCapturingSlot.captured.contains(activity.getString(sf__screen_lock_auth_failed)))
+    }
+
+    private fun Drawable.renderToBitmap(): Bitmap {
```
JohnsonEricAtSalesforce, line 919: For asserting the content of the app icon, here's a utility method that helps.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/preview/ScreenLockViewPreview.kt
brandonpage: If the Previews aren't in the same file as the view code then why even have them? The whole point is that you can see the changes you are making live as you edit.  You could have `ScreenLockView` and `ScreenLockViewPreview` open side by side but I don't understand the point.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
brandonpage: Passing ViewModels is not recommended.  If it is necessary for testing we should have a `viewModel(factory = ViewModelFactory)` default value so it isn't passed in production.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
brandonpage: Huh?  `ScreenLockViewModel` is internal and `setupMessageText` never set (outside of tests).

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
```diff
@@ -0,0 +1,134 @@
...
+                contentPadding = PaddingValues(PADDING_SIZE.dp),
+                enabled = true,
+                modifier = Modifier
+                    .align(CenterHorizontally)
+                    .padding(PADDING_SIZE.dp),
+                onClick = {
+                    viewModel.setupButtonAction.value()
+                },
+                shape = RoundedCornerShape(CORNER_RADIUS.dp),
+            ) { Text(color = colorScheme.onPrimary, fontSize = 14.sp, fontWeight = Normal, text = viewModel.setupButtonLabel.value ?: stringResource(sf__screen_lock_setup_button)) }
```
brandonpage, line 135: NIT: `fontSize` (3 times in this file) and icon size should probably reference a common const value.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
brandonpage: `ActivityScenario` is used in most of our other test to exercise `onCreate`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+     * @param build The Android SDK build. This parameter is intended for
+     * testing purposes only. Defaults to the current Android SDK build
+     * @param packageManager The PackageManager to use. This parameter is
+     * intended for testing purposes only. Defaults to the current
+     * PackageManager instance
+     * @param sdkManager The SalesforceSDKManager to use. This parameter is
+     * intended for testing purposes only. Defaults to the current
+     * SalesforceSDKManager instance
+     */
+    @SuppressLint("NewApi")
```
brandonpage: Is this necessary?  I would rather have warnings now than accidentally suppress something important in the future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+    @VisibleForTesting
+    internal fun create(
+        build: Int = SDK_INT,
+        packageManager: PackageManager = this.packageManager,
+        sdkManager: SalesforceSDKManager = getInstance(),
+    ) {
+        enableEdgeToEdge()
+
+        // Protect against screenshots.
+        window.setFlags(FLAG_SECURE, FLAG_SECURE)
```
brandonpage, line 136: I would not mind this being in a `if (!SalesforceSDKManager.getInstance().isDebugBuild)`.  The OS prompts are blocked regardless so it might not be needed at all.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
brandonpage: Looking at `BiometricManger.java` the return type has an `@IntDef` annotation:

> Denotes that the annotated element of integer type, represents a logical type and that its value should be one of the explicitly named constants. If the IntDef#flag() attribute is set to true, multiple constants can be combined.

So if a new Android API level added a new value we would get an compile time error when we updated to target it.  So what you have is fine as is.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
brandonpage: Having this in a centralized place sounds good to me.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
```diff
@@ -0,0 +1,490 @@
...
+    /**
+     * A biometric authentication callback.
+     * @param activity The activity to use for authentication.  This parameter
+     * is intended for testing purposes only. Defaults to this inner class
+     * receiver
+     */
+    @VisibleForTesting
+    internal inner class BiometricAuthenticationCallback(
+        private val activity: ScreenLockActivity = this@ScreenLockActivity,
+    ) : AuthenticationCallback() {
```
brandonpage, line 462: Could this just be an inline lambda now?  If that makes testing a lot harder than nbd.

libs/SalesforceSDK/build.gradle.kts
brandonpage: Do we need a 3PP for this?  If it is from Google maybe not?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/preview/ScreenLockViewPreview.kt
JohnsonEricAtSalesforce: I did try the side by side editors and that would be one way to use the suggestion I was prototyping where `@Preview` annotated code could be excluded from code coverage just by moving it from release to debug.  That said, since we're generating coverage from the debug variant the benefit's not there anyways, alas.  I'll recombine the two 👍🏻

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
JohnsonEricAtSalesforce: `setupMessageText` is set from `ScreenLockActivity.setErrorMessage(...)`.  Is that what you were looking for?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
JohnsonEricAtSalesforce: True, though all the tests I'd used on `LoginActivity` were able to manipulate intent extras or data to exercise logic in the activity instance.  In this case, `ScreenLockActivity` doesn't have any external API to accomplish that.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
JohnsonEricAtSalesforce: Perfect-O.  Thanks for looking at that.

libs/SalesforceSDK/build.gradle.kts
JohnsonEricAtSalesforce: We should check.  Can you direct message any instructions you have on how I can review or request that?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/ScreenLockView.kt
JohnsonEricAtSalesforce: We do already have `lifecycle-viewmodel-compose-android` as a dependency and screen lock only has a top-level composable.  If this parameter defaulted to `viewModel()` from that dependency, it would seem to match [Thinking in Compose](https://developer.android.com/develop/ui/compose/mental-model) better.

I'd interpreted the style guide to mean a "screen level" composable was allowed access to the view model while "component level" composables called from other composables were not in order to preserve the "skip as much as possible" guideline.

I'll push a quick commit so we can review what that looks like.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
JohnsonEricAtSalesforce: The warnings would be a bit of noise, but I'd also be willing to track them.  I'll take this off then.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
JohnsonEricAtSalesforce: I'd kept this exactly as it was.  Should we keep the functionality the same for a point release such as 13.2?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
JohnsonEricAtSalesforce: You're correct that it's an extracted class for testing ✅

libs/SalesforceSDK/build.gradle.kts
JohnsonEricAtSalesforce: That 3PP is approved ✅

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ScreenLockActivity.kt
brandonpage: @JohnsonEricAtSalesforce Can we use Mockk?  I think I figured out a (hacky) workaround:

Image

This should let you:

```
mockkObject(ScreenLockBuildConfig)
every { ScreenLockBuildConfig.isAtLeast30() } returns false
```

**Obviously this is not great.** But it seems better than having errors in the IDE:
Image

cc @wmathurin

JohnsonEricAtSalesforce: I'm game to give it a try 🤞🏻 I'll report how it works out soon.

JohnsonEricAtSalesforce: It's an interesting idea!  I took a try at it and made some neat adjustments after I read the docs for `CheckSdkIntAtLeast`.  Notably, I switched each check to a property.   I also extracted the new object to the new constants source.

It seems to achieve the result of letting us mock the SDK level and avoid having inspector errors where we have our own logic to check the level.  I also verified the inspector will still warn us if we don't use our new checks correctly!

@brandonpage, look here 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2835/changes/cc2203a24b8db64809b80cdc18f9959b26498769

brandonpage (approved):
LGTM 🎉

PR #2836, opened 2026-02-06
Removing consumer key from sample apps

wmathurin:
Changes:

- bootconfig.xml for sample apps have been git removed
- when running install.sh, bootconfig.xml files are created for the sample apps using bootconfig.xml.sample as template 
- consumer key and call back url are set on bootconfig.xml / bootconfig.json for the sample apps using environment variables (*) if defined

(*) MSDK_ANDROID_REMOTE_ACCESS_CONSUMER_KEY and MSDK_ANDROID_REMOTE_ACCESS_CALLBACK_URL

**Merge only after**
- https://github.com/forcedotcom/SalesforceMobileSDK-Shared/pull/539 is merged
- shared submodule is updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,7 @@ yarn.lock
 index.android.bundle*
 shared/test/test_credentials.json
 shared/test/ui_test_config.json
+**/bootconfig.xml
+!libs/test/**/bootconfig.xml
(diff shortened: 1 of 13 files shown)


JohnsonEricAtSalesforce (approved):
👍🏻

PR #2837, opened 2026-02-07
Token Migration

brandonpage:
This PR adds a `migrateRefreshToken` function to an new Kotlin `UserAccountManagerExtention` class.  

Example usage:
```
UserAccountManager.getInstance().migrateRefreshToken(
    userAccount = user,  // defaults to current user if not set
    appConfig = config,
    onMigrationSuccess = { user ->
        // success lambda
    },
    onMigrationError = { error, errorDesc, e ->
        // error lambda
    },
)
```

Implementing it in the existing Java class would have required defining a new public interface just for the callback functions since lambda functions can not be directly passed as parameters.  I did not want to introduce the extra public API since I expect we will convert the class to Kotlin soon.  I attempted to convert the class for this PR but the static references converted to companion properties and return type nullability would have introduced minor breaking changes (that would be better left for 14.0).  

`migrateRefreshToken` launches a new `TokenMigrationActivity` that does most of the work.  A new `MigrationCallbackRegistry` object facilitates passing the callback functions from the entry point API to the activity, and a code comment explains why that is necessary.  Most of the code in `TokenMigrationActivity` is the `WebViewClient` implementation that is very similar to the one in `LoginActivity`, but unfortunately they are a little too coupled to share.  There is very little else in the activity because it utilizes the existing `LoginViewModel` to do most of the work.

The Activity UI is very minimal.  When the WebView is loading, it and the Activity background are transparent with a 50% tint.  The same progress indicator from standard login is reused, so if the app has customized it that should be reflected here as well.   

No Interaction Required,                       Migration Failure,                    Interaction Required           
<p float="center">
Screenshot_20260115-154115
Screenshot_20260115_161228
Screenshot_20260115_161154
</p>

TODO:
- [X] More unit tests (`buildAuthWebview`, AuthenticationUtilities, etc)
- [ ] ~~UI Tests~~ - Will send a follow-up PR.
- [ ] More manual testing (beacon, etc)


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/OAuthConfig.kt
```diff
@@ -26,11 +26,15 @@
  */
 package com.salesforce.androidsdk.config
 
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
```
brandonpage, line 32: This is Kotlin magic ✨.  The compiler is able to implement the required functions to satisfy the `Parcelable` interface by itself since this is a `data class`.  

We should do this for `UserAccount` in the future.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/TokenMigrationActivity.kt
```diff
@@ -0,0 +1,324 @@
...
+        val callbackKey = intent.getStringExtra(EXTRA_CALLBACK_ID) ?: run {
+            SalesforceSDKLogger.e(TAG, "Unable to parse MigrationResult callback id.")
+            finish()
+            return
+        }
+        val resultCallback = MigrationCallbackRegistry.consume(callbackKey) ?: run {
+            SalesforceSDKLogger.e(TAG, "Unable to retrieve MigrationResult callback.")
+            finish()
+            return
+        }
```
brandonpage, line 105: This _should_ never happen, but I will admit it is a scary little gap in the flow.  If we somehow hit an error here we can't invoke the error callback.  

If anyone has a clever idea let me know.  I'd rather not call `startActivityForResult` since it has been deprecated for years.  And it's replacement `registerForActivityResult` needs to be called from an Activity.

libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -45,6 +45,12 @@
             android:configChanges="orientation|screenLayout|uiMode|screenSize|smallestScreenSize"
             android:exported="true" />
 
+        <!-- Token Migration Activity -->
+        <activity android:name="com.salesforce.androidsdk.ui.TokenMigrationActivity"
+            android:excludeFromRecents="true"
+            android:theme="@style/AccountSwitcher"
```
brandonpage, line 51: Should I rename this style to something more generic?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/TokenMigrationActivity.kt
JohnsonEricAtSalesforce: I haven't read everything yet, but could you use `this@TokenMigrationActivity` since that's captured in this closure?

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Attempts to migrate the [userAccount] to the provided Connected App or
+ * External Client Application [appConfig].
+ *
+ * This might cause the approve/deny screen to be presented to the user to authorize the
+ * new app. If successful a new set of credentials (refresh token, access token) are obtained
+ * and replace the existing credentials for the user.
+ */
+@Suppress("UnusedReceiverParameter")
+fun UserAccountManager.migrateRefreshToken(
```
JohnsonEricAtSalesforce, line 48: Taking the success and error callbacks as up-front parameters works so no change required.  I wondered if this could benefit from returning a `Result`.  The call would look something like this pseudo-code then:
```
UserAccountManager.getInstance().migrateRefreshToken(
    userAccount = user,  // defaults to current user if not set
    appConfig = config
).onSuccess { user ->
        // success lambda
}.onError { error -> // I saw there's currently three parameters, but maybe they all fold into the error here?
        // error lambda
}
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
```diff
@@ -0,0 +1,108 @@
...
+    val callbackKey = MigrationCallbackRegistry.register(
+        callbacks = MigrationCallbackRegistry.MigrationCallbacks(
+            onMigrationSuccess = loggedOnSuccess,
+            onMigrationError = onMigrationError,
+        )
+    )
+
+    val context = SalesforceSDKManager.getInstance().appContext
+    val intent = Intent(context, TokenMigrationActivity::class.java)
+    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
JohnsonEricAtSalesforce: This works but could be a chance to use `intent.apply {}` since there are so many properties.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/TokenMigrationActivity.kt
```diff
@@ -0,0 +1,324 @@
+/*
+ * Copyright (c) 2026-pffsent, salesforce.com, inc.
```
JohnsonEricAtSalesforce: I don't know how I saw this nor I know when since `pffsent` must be the result of a tear in the space time continuum. 🙃 I've been looking for things like this when my Android keyboard bindings accidentally pop meta-characters in my editor just before a commit!

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
brandonpage: I had the same thought.  Originally I had a function in the AppFlowTester's `RestUtils` class that called `migrateRefreshToken` and returned a `Result` but I ended up just not using it.  

I don't think it makes sense to change the signature to Result since the lambdas passed in flow all the way through `AuthenticationUtilities`.  But I will take a look in case doing so lets us get rid of `MigrationCallbackRegistry`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/TokenMigrationActivity.kt
brandonpage: I have a completely reasonable (not really) explanation why the vast majority of my typos involve extra `f`'s that I'll let you sometime if you want to know lol

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -438,6 +454,12 @@ private fun handleDuplicateUserAccount(
             clearCaches()
             userAccountManager.clearCachedCurrentUser()
 
+            // Remove the existing Account from AccountManager so createAccount can create a fresh one
+            val existingAccount = userAccountManager.buildAccount(duplicateUserAccount)
+            if (existingAccount != null) {
+                SalesforceSDKManager.getInstance().clientManager.removeAccount(existingAccount)
```
wmathurin: What was the impact of not doing the removeAccount until now?

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/TokenMigrationActivity.kt
```diff
@@ -0,0 +1,324 @@
...
+
+    @VisibleForTesting
+    internal fun buildAuthWebview(
+        frontDoorUrl: String,
+        resultCallback: MigrationCallbackRegistry.MigrationCallbacks,
+        instanceServer: String,
+    ): WebView = WebView(this@TokenMigrationActivity).apply {
+        @SuppressLint("SetJavaScriptEnabled") // Required by Salesforce
+        settings.javaScriptEnabled = true
+        settings.userAgentString = SalesforceSDKManager.getInstance().userAgent
```
wmathurin: Shouldn't the original user agent for the web view be part of the custom user agent? See https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt#L207

wmathurin (approved):
LGTM - the UI tests (especially the one combining multi users + migration) will catch anything we missed.

wmathurin:
> LGTM - the UI tests (especially the one combining multi users + migration) will catch anything we missed.

After writing that, I went and checked the UI tests on iOS and realized we don't have a test that does migration with multi users ;-)

JohnsonEricAtSalesforce (approved):
⭐️

brandonpage:
I will send a follow-up PR with UI tests as this one has gotten very large.

PR #2840, opened 2026-02-11
Make LegacyAuthenticatorService public to avoid error log lines.  Make upgrade step public.

brandonpage:


wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKUpgradeManager.java
```diff
@@ -314,11 +314,14 @@ private void updateFromBefore13_0_2() {
...
+    /**
      *  Migrate any accounts with account_type "com.salesforce.androidsdk" to a unique value.
-     *  TODO: Remove this in Mobile SDK 15.0
+     *
+     *  @deprecated Will be removed in Mobile SDK 15.0.0.
      */
-    private void migrateAccountType() {
+    @Deprecated
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    public void migrateAccountType() {
```
brandonpage, line 324: We typically don't make upgrade steps public, but I figured affected customers may want to do this outside of a MSDK upgrade.

PR #2841, opened 2026-02-17
@W-20798759: [MSDK] Add ability to programmatically update servers list

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This adds two new methods to the login server manager: Replace and re-order login server.  For 13.2 point release, this only adds behavior and maintains API compatibility.  To that end, this keeps this source in Java for now and attempts to follow the style of the existing code.

  There are new unit tests to validate and cover the new methods.  Also, [here's a video](https://drive.google.com/file/d/15VMZKhUywKJM1Vnwlg5bPiUfVAv-Oi0D/view?usp=drive_link) where I added a few menu options to test interactively.  Note in the video how the index of the updated login server is corrected when needed so it does not exceed the last index or precede the non-custom login servers which have higher priority and are immutable.  The video doesn't show it, but the reset action also included all the negative tests that are present in the unit tests.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -351,17 +351,16 @@ public class LoginServerManager {
 
 		// Determine the last non-custom login server index.
 		final List<LoginServer> servers = getLoginServers();
-		int lastNonCustomIndex = -1;
+		int firstCustomLoginServerIndex = -1;
 		for (int i = servers.size() - 1; i >= 0; i--) {
-			if (!servers.get(i).isCustom) {
-				lastNonCustomIndex = i;
-				break;
+			if (servers.get(i).isCustom) {
+				firstCustomLoginServerIndex = i;
 			}
 		}
 
 		// Adjust the re-ordered custom login server index to be within bounds.
-		if (updatedIndex <= lastNonCustomIndex) {
-			updatedIndex = lastNonCustomIndex + 1;
+		if (updatedIndex <= firstCustomLoginServerIndex) {
+			updatedIndex = firstCustomLoginServerIndex;
 		} else if (updatedIndex >= servers.size()) {
 			updatedIndex = servers.size() - 1;
 		}
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -26,6 +26,9 @@
  */
 package com.salesforce.androidsdk.config;
 
+import static java.lang.String.format;
+import static java.util.Locale.US;
+
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
@@ -130,7 +133,7 @@ public class LoginServerManager {
 		if (name != null && url != null) {
 			LoginServer server = new LoginServer(name, url, isCustom);
 
-			// Only notify livedata consumers if the value has changed.
+			// Only notify live data consumers if the value has changed.
 			if (!server.equals(selectedServer.getValue())) {
 				selectedServer.postValue(server);
 			}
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -323,6 +326,92 @@ public List<LoginServer> getLoginServersFromPreferences() {
...
+	 * Reorders a custom login server in the list of login servers.
+	 *
+	 * @param originalIndex The original index of the custom login server.  If this is not the index
+	 *                      of a custom login server, this method will do nothing
+	 * @param updatedIndex  The new index of the custom login server.  This must be after the last
+	 *                      non-custom login server and within the updatable bounds of the list.  If
+	 *                      it is not it will be automatically corrected
+	 */
+	@SuppressWarnings("unused")
+	public void reorderCustomLoginServer(
```
JohnsonEricAtSalesforce, line 339: The documentation comment should provide adequate guidance, particularly where the boundary of non-custom and custom login server indexes is guarded.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -323,6 +326,92 @@ public List<LoginServer> getLoginServersFromPreferences() {
...
+			updatedIndex = lastNonCustomIndex + 1;
+		} else if (updatedIndex >= servers.size()) {
+			updatedIndex = servers.size() - 1;
+		}
+
+		// Update the login server list.
+		loginServers.remove(originalIndex);
+		loginServers.add(updatedIndex, originalLoginServer);
+
+		// Edit each login server indexed after the updated index.
```
JohnsonEricAtSalesforce, line 372: This "re-indexing" logic is very similar to what's already in remove login server.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -323,6 +326,95 @@ public List<LoginServer> getLoginServersFromPreferences() {
...
+	 * Replaces one custom login server with another.
+	 *
+	 * @param originalCustomLoginServer The original custom login server.  If this is not a custom
+	 *                                  login server or doesn't match an existing login server this
+	 *                                  method will do nothing.
+	 * @param updatedCustomLoginServer  The updated custom login server.  If this is not a custom
+	 *                                  login server this method will do nothing.
+	 */
+	@SuppressWarnings("unused")
+	public void replaceCustomLoginServer(
```
JohnsonEricAtSalesforce, line 396: The documentation comment here should guide callers around legal parameters.

JohnsonEricAtSalesforce:
Note the checks haven't run tests or code coverage, so there might be a little more work to do.  That said, this ought to be really close to the mark.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -130,7 +133,7 @@ public LoginServer getSelectedLoginServer() {
 		if (name != null && url != null) {
 			LoginServer server = new LoginServer(name, url, isCustom);
 
-			// Only notify livedata consumers if the value has changed.
+			// Only notify live data consumers if the value has changed.
```
brandonpage, line 136: NIT:  I think `LiveData` is probably the most correct.  Google never writes it as two words.

JohnsonEricAtSalesforce:
@wmathurin, I did make a few minor logic updates to preemptively fix bugs as I increased code coverage.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
brandonpage: Is this the interface they wanted?  I was envisioning something like:
```public void reorderLoginServers(List<LoginServer> reorderedList)```

Where they just gave us the list of servers in the order they wanted.

brandonpage: And this only reorders _custom_ login servers.  Is that what they want?

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
brandonpage: Is it not possible to add an `update(name, url)` method to `LoginServer` itself?

PR #2844, opened 2026-02-23
Token Migration UI Tests

brandonpage:
This PR fundamentally changes the direction of our UI Test strategy for this repo:
- The submodule and dependency on `SalesforceSDK-UITests` has been removed.  
- Most of the PageObject code uses `Compose UI Test` APIs which are _extremely_ fast and appear to be very stable.  
- `Espresso Web` is used for most WebView interactions.  
- `UIAutomator2` APIs are only used in the couple places that they are unfortunately necessary (explained in code comments).  
- The use of Android's `Test Orchestrator` also noticeably to helps stability with Firebase Test Lab at the cost of only ~10 seconds per test (now ~30s each). 
- Tests now utilize `ui_test_config.json`.

This PR is quite large, but contains **no behavior changes to production code** and only adds IDs to the existing UI.

Note: The UI Tests CI will fail on this PR because the code and yaml config (that does not change on PR) are out of sync.  You can see my [locally triggered Firebase Runs here](https://console.firebase.google.com/project/mobile-apps-firebase-test/testlab/histories/bh.b2f490897ec66d1a), the first two failures were before `Test Orchestrator`.


install.sh
```diff
@@ -38,10 +38,11 @@ apply_bootconfig_paths() {
         if [ -f "$bootconfig" ]; then
 	    # Substitute env vars if set
 	    if [ -n "${MSDK_ANDROID_REMOTE_ACCESS_CONSUMER_KEY:-}" ]; then
-                gsed -i "s|__CONSUMER_KEY__|${MSDK_ANDROID_REMOTE_ACCESS_CONSUMER_KEY}|g" "$bootconfig"
+	              # sed -i.bak works identically on both BSD sed (macOS) and GNU sed (Linux)
+                sed -i.bak "s|__CONSUMER_KEY__|${MSDK_ANDROID_REMOTE_ACCESS_CONSUMER_KEY}|g" "$bootconfig" && rm -f "$bootconfig.bak"
```
brandonpage, line 42: @wmathurin `gsed` isn't installed on MacOS by default and doesn't exist on linux.  This was the best I could come up with for a command that worked on both platforms but happy to hear suggestions.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/UITestConfig.kt
```diff
@@ -0,0 +1,134 @@
...
+    Json.decodeFromString(
+        string = ResourceReaderHelper.readAssetFile(
+            InstrumentationRegistry.getInstrumentation().targetContext,
+            /* assetFilePath = */ "ui_test_config.json",
+        ) ?: throw Exception("ui_test_config.json file not found.")
+    )
+}
+
+@Serializable
+data class UITestConfig(val loginHosts: List<LoginHost>, val apps: List<AppConfig>) {
```
brandonpage, line 74: Again, Kotlin's `Json` and `Serialization` "plugins" are magic.  `Json.decodeFromString` returns an instance of this data class.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/AuthorizationPageObject.kt
```diff
@@ -0,0 +1,127 @@
...
+import com.salesforce.androidsdk.R as sdkR
+
+private const val TAG = "AuthorizationPageObject"
+private const val MAX_RETRIES = 5
+
+/**
+ * Handles the OAuth authorization "Allow" button that may appear
+ * after login or token migration.
+ */
+class AuthorizationPageObject(composeTestRule: ComposeTestRule) : BasePageObject(composeTestRule) {
```
brandonpage, line 44: This class gave me the most trouble since I have to use `UIAutomator2` to tap the "Allow" button (given it has no resource ID).  I split it into separate login and migration functions so I could use Compose/Espresso to quickly check for signs of if we need to wait for and click the allow button or not.  The slow waiting + quick short circuit in a loop seems to work well but I purposefully left log lines to make debugging easier if it ever fails/flaps.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/TokenMigrationTest.kt
```diff
@@ -0,0 +1,220 @@
...
+
+    val loginPage = LoginPageObject(composeTestRule)
+    val loginOptions = LoginOptionsPageObject(composeTestRule)
+    val app = AuthFlowTesterPageObject(composeTestRule)
+
+    // region Migration within same app (scope upgrade)
+
+    // Migrate within same CA (scope upgrade).
+    @Test
+    fun testMigrate_CA_AddMoreScopes() {
```
brandonpage, line 72: Test comments, names and structure copied directly from iOS with few modifications.

setversion.sh
```diff
@@ -49,45 +49,47 @@ update_top_build_gradle ()
 {
     local file=$1
     local versionName=$2
-    gsed -i "s/version = \"[0-9\.]*\"/version = \"${versionName}\"/g" ${file}
+    # sed -i.bak works identically on both BSD sed (macOS) and GNU sed (Linux)
+    sed -i.bak "s/version = \"[0-9\.]*\"/version = \"${versionName}\"/g" "${file}" && rm -f "${file}.bak"
```
brandonpage, line 53: @wmathurin I could revert this if you are uncomfortable with it.  This script doesn't need to run on CI (at this time).

.github/workflows/reusable-ui-workflow.yaml
```diff
@@ -23,11 +23,11 @@ jobs:
           ref: ${{ github.head_ref }}
       - name: Install Dependencies
         env:
-          TEST_CREDENTIALS: ${{ secrets.TEST_CREDENTIALS }}
+          MSDK_ANDROID_REMOTE_ACCESS_CALLBACK_URL: ${{ secrets.MSDK_ANDROID_REMOTE_ACCESS_CALLBACK_URL }}
```
wmathurin, line 26: On iOS, I don't use/need those because the consumer key are always set through login options in the tests (and they come from ui_test_config.json).

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/AuthFlowTesterPageObject.kt
```diff
@@ -0,0 +1,249 @@
...
+        composeTestRule.onNodeWithContentDescription(ALERT_TITLE_CONTENT_DESC)
+            .assertTextEquals(getString(R.string.request_successful))
+
+        composeTestRule.onNodeWithContentDescription(ALERT_POSITIVE_BUTTON_CONTENT_DESC)
+            .performClick()
+        composeTestRule.waitForIdle()
+    }
+
+    fun getTokens(): Tokens {
+        expandUserCredentialsSection()
```
wmathurin, line 109: On iOS, I ended up having export button that gives me back a JSON with all the values in the section - things were too slow when expanding and reading individual values one by one. But it looks like things are faster on Android?

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/TokenMigrationTest.kt
```diff
@@ -0,0 +1,220 @@
...
+            KnownAppConfig.BEACON_JWT
+        )
+        migrateAndValidate(
+            KnownAppConfig.BEACON_OPAQUE
+        )
+    }
+
+    // endregion
+
+    private fun loginAndValidate(
```
wmathurin, line 184: This method will probably be useful in other login tests (not migration related) - so maybe it should move to a super class or helper class. On iOS, all the test suites derive from BaseAuthflowTester (and the test suites don't interact with the PageObject directly).

setversion.sh
wmathurin: It's fine if it works the same. 
And maybe one day, we will run the release from CI !

wmathurin: (Although we could just install gsed as part of the CI job).

wmathurin: gsed is being used across all the setversion.sh in all the repos - so maybe we keep it unchanged (for now).

wmathurin (approved):

setversion.sh
brandonpage: `gsed` does not exist for Linux, it's just `sed`.  AFAIK GNU make `gsed` because BSD `sed` is a completely different implementation than UNIX.  `sed` is not truly portable but it seems like the command I changed it to has the same result on both.

.github/workflows/reusable-ui-workflow.yaml
brandonpage: Yeah.  It isn't possible (or at least reasonably worth the effort) to truly set the contents of the bootconfig file on Android so I don't have an equivalent "staticAppConfig".  I wanted to have at least one test that uses the bootconfig file alone so I added it to CI.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/AuthFlowTesterPageObject.kt
brandonpage: Compose UI Test skips animations, syncs/synchronizes with Compose so you never have to sleep/wait and automatically scrolls (which was honestly a huge pain with UIAutomator2).  I don't think it would really speed it up much but I will take a look.

brandonpage: Shockingly they are about the same:

Screenshot 2026-02-23 at 8 35 46 PM

Edit:  The above didn't account for some UI navigation overhead.   The more accurate numbers are still close though, with the Json method being ~100ms faster on an API 36 device: 
Screenshot 2026-02-24 at 8 21 58 AM

Unless the difference is larger on API 28 I'd rather just test using the UI since that also ensures its correctness. 

Second Edit:  ^ It's not.  Between 100-200 ms difference.

setversion.sh
wmathurin: We should move to sed then. Maybe have a dedicated story and move all the setversion.sh across all the repos as part of that work?

.github/workflows/reusable-ui-workflow.yaml
wmathurin: Oh yeah, I remember that from the earlier PRs.
It's not ideal to have both platforms diverge on that (but I guess they were already quite different in their handling of bootconfig). It would be better if we could move away from any pre-build setup step. But like you said, it's probably too much work / we don't have enough time.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/AuthFlowTesterPageObject.kt
wmathurin: I'm jealous, on iOS, things are really slow.

wmathurin: That gave me an idea - I'm now disabling animations on iOS (see https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/7677881a496c28063797c0972dda952150e16e57?w=1) - it's a bit faster but not as impressive as Android.

PR #2845, opened 2026-02-25
@W-21148602: Internal Server List does not respect changes made to Servers.xml file

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This updates the login server manager to respond to updates made to the login servers from servers.xml in resources and the login servers provided by Mobile Device Management via the runtime configuration.  Previously, login servers from these two locations were static (for servers.xml) or additions-only (for runtime config) after the app had been installed.

  For a point release, these changes do need to maintain compatibility with the 13.x public API.  To minimize regression risk, they're also maintaining the Java style and deprecated shared preferences persistence already in place.

  Accomplishing the change is non-trivial.  The current logic was not built with mutability in mind for the non-custom servers.  Scaling the existing logic and ensuring the shared preferences data maintains exact integrity did increase the change footprint.

  New unit tests provide functional automated testing of all the changes and 100% patch code coverage.  The overall test coverage for the login server manager is much, much higher than before as well.

  I'll follow-up with inline source commentary as well.  Thanks for reading!

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -48,9 +48,6 @@ import androidx.lifecycle.MutableLiveData;
 import com.salesforce.androidsdk.R;
 import com.salesforce.androidsdk.util.SalesforceSDKLogger;
 
-import org.xmlpull.v1.XmlPullParserException;
-
-import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
@@ -187,10 +184,12 @@ public class LoginServerManager {
 	 * @return The selected login server
 	 */
 	public LoginServer getSelectedLoginServer() {
+		// Fetch the selected login server.
 		final SharedPreferences selectedServerPrefs = ctx.getSharedPreferences(SERVER_SELECTION_FILE, MODE_PRIVATE);
-		final String name = selectedServerPrefs.getString(SERVER_NAME, null);
-		final String url = selectedServerPrefs.getString(SERVER_URL, null);
-		boolean isCustom = selectedServerPrefs.getBoolean(IS_CUSTOM, false);
+		final String selectedServerName = selectedServerPrefs.getString(SERVER_NAME, null);
+		final String selectedServerUrl = selectedServerPrefs.getString(SERVER_URL, null);
+		final boolean selectedServerIsCustom = selectedServerPrefs.getBoolean(IS_CUSTOM, false);
+		final boolean selectedServerHasValidNameAndUrl = selectedServerName != null && selectedServerUrl != null;
 
 		// Refresh the list of mobile device management (MDM) servers from the runtime config.
 		if (isRuntimeConfigAppServiceHostsSet()) {
@@ -200,26 +199,36 @@ public class LoginServerManager {
 		// Get the active list of login servers.
 		final List<LoginServer> loginServers = getLoginServers();
 
-		// Selection has been saved before and is available in the active list of login servers.
-		if (name != null && url != null && loginServers.stream().anyMatch(server -> server.name.equals(name) && server.url.equals(url))) {
-			final LoginServer server = new LoginServer(name, url, isCustom);
+		// Check if the selected login server is available in the active list of login servers.
+		final boolean selectedLoginServerIsAvailable = loginServers.stream().anyMatch(server ->
+				server.name.equals(selectedServerName) && server.url.equals(selectedServerUrl)
+		);
 
-			// Only notify live data consumers if the value has changed.
-			if (!server.equals(selectedServer.getValue())) {
-				selectedServer.postValue(server);
-			}
-		} else {
+		// If the selected login server is valid and is available in the active list of login servers.
+		LoginServer selectedLoginServer = null;
+		if (selectedServerHasValidNameAndUrl) {
+			if (selectedLoginServerIsAvailable) {
+				selectedLoginServer = new LoginServer(selectedServerName, selectedServerUrl, selectedServerIsCustom);
 
-			// First time selection defaults to the first server on the list.
-			if (loginServers != null) {
-				if (!loginServers.isEmpty()) {
-					selectedServer.postValue(loginServers.get(0));
+				// Notify live data consumers if the value has changed.
+				if (!selectedLoginServer.equals(selectedServer.getValue())) {
+					selectedServer.postValue(selectedLoginServer);
 				}
 			}
+		}
 
-			// Stores the selection for the future.
+		// If the selected login server is invalid or not available in the active list of login servers.
+		if (selectedLoginServer == null) {
+
+			// Default to the first login server on the list.
+			if (!loginServers.isEmpty()) {
+				selectedServer.postValue(loginServers.get(0));
+			}
+
+			// Store the selected login server.
 			setSelectedLoginServer(selectedServer.getValue());
 		}
+
 		return selectedServer.getValue();
 	}
 
@@ -321,6 +330,7 @@ public class LoginServerManager {
 
 		int index = servers.indexOf(server);
 
+		// TODO: Unrelated Coverage Needed. ECJ20260303
 		if (allowNonCustomRemoval || server.isCustom && index != -1) {
 			int numServers = servers.size();
 			Deque<LoginServer> stack = new ArrayDeque<>(servers.subList(index + 1, numServers));
@@ -387,10 +397,13 @@ public class LoginServerManager {
     public List<LoginServer> getLoginServersFromRuntimeConfig() {
 		String[] mdmLoginServers = runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHosts);
 		final List<LoginServer> allServers = new ArrayList<>();
+		// TODO: Coverage Needed. ECJ20260303
 		if (mdmLoginServers != null) {
 			String[] mdmLoginServersLabels = runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHostLabels);
 			if (mdmLoginServersLabels == null || mdmLoginServersLabels.length != mdmLoginServers.length) {
+				// TODO: Coverage Needed. ECJ20260303
 				SalesforceSDKLogger.w(TAG, "No login servers labels provided or wrong number of login servers labels provided - using URLs for the labels");
+				// TODO: Coverage Needed. ECJ20260303
 				mdmLoginServersLabels = mdmLoginServers;
 			}
 
@@ -400,6 +413,7 @@ public class LoginServerManager {
 			for (int i = 0; i < mdmLoginServers.length; i++) {
 				final String name = mdmLoginServersLabels[i];
 				final String url = mdmLoginServers[i];
+				// TODO: Coverage Needed. ECJ20260303
 				if (name == null || url == null) { continue; }
 
 				final LoginServer server = new LoginServer(name, url, false);
@@ -412,6 +426,7 @@ public class LoginServerManager {
 				allServers.add(server);
 			}
 		}
+		// TODO: Coverage Needed. ECJ20260303
 		return (!allServers.isEmpty() ? allServers : null);
 	}
 
@@ -574,8 +589,9 @@ public class LoginServerManager {
 			}
 			try {
 				eventType = xml.next();
-			} catch (XmlPullParserException | IOException e) {
+			} catch (Exception e) {
 				SalesforceSDKLogger.w(TAG, "Exception thrown while parsing XML", e);
+				break;
 			}
 		}
 		return loginServers;
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
@@ -37,6 +37,7 @@ import static java.util.Locale.US;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
+import android.content.res.Resources;
 import android.content.res.XmlResourceParser;
 import android.os.Looper;
 
@@ -83,10 +84,17 @@ public class LoginServerManager {
 	@VisibleForTesting
 	public static final String RUNTIME_PREFS_FILE = "runtime_prefs_file";
 
-	private static final String NUMBER_OF_ENTRIES = "number_of_entries";
-	private static final String SERVER_NAME = "server_name_%d";
-	private static final String SERVER_URL = "server_url_%d";
-	private static final String IS_CUSTOM = "is_custom_%d";
+	@VisibleForTesting
+	public static final String NUMBER_OF_ENTRIES = "number_of_entries";
+
+	@VisibleForTesting
+	public static final String SERVER_NAME = "server_name_%d";
+
+	@VisibleForTesting
+	public static final String SERVER_URL = "server_url_%d";
+
+	@VisibleForTesting
+	public static final String IS_CUSTOM = "is_custom_%d";
 
 	/**
 	 * Shared preferences for the selected login server
@@ -204,9 +212,8 @@ public class LoginServerManager {
 
 			// First time selection defaults to the first server on the list.
 			if (loginServers != null) {
-				final LoginServer server = loginServers.get(0);
-				if (server != null) {
-					selectedServer.postValue(server);
+				if (!loginServers.isEmpty()) {
+					selectedServer.postValue(loginServers.get(0));
 				}
 			}
 
@@ -248,12 +255,12 @@ public class LoginServerManager {
 	}
 
 	/**
-	 * Adds a custom login server to the shared pref file.
+	 * Adds a custom login server.
 	 *
-	 * @param name Server name.
-	 * @param url Server URL.
+	 * @param name The login server name
+	 * @param url The login server URL
 	 */
-	public void addCustomLoginServer(String name, String url) {
+	public void addCustomLoginServer(@NonNull String name, @NonNull String url) {
 		// Prevent duplicate servers.
 		for (LoginServer existingServer : getLoginServers()) {
 			if (url.equals(existingServer.url)) {
@@ -369,11 +376,7 @@ public class LoginServerManager {
 	 * server.xml when false
 	 */
 	private boolean isRuntimeConfigAppServiceHostsSet() {
-		try {
-			return runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHosts) != null;
-		} catch (Exception e) {
-			return false;
-		}
+		return runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHosts) != null;
 	}
 
 	/**
@@ -382,20 +385,10 @@ public class LoginServerManager {
 	 */
 	@SuppressWarnings("UnusedReturnValue")
     public List<LoginServer> getLoginServersFromRuntimeConfig() {
-		String[] mdmLoginServers = null;
-		try {
-			mdmLoginServers = runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHosts);
-		} catch (Exception e) {
-			SalesforceSDKLogger.w(TAG, "Exception thrown while attempting to read array, attempting to read string value instead", e);
-		}
+		String[] mdmLoginServers = runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHosts);
 		final List<LoginServer> allServers = new ArrayList<>();
 		if (mdmLoginServers != null) {
-			String[] mdmLoginServersLabels = null;
-			try {
-				mdmLoginServersLabels = runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHostLabels);
-			} catch (Exception e) {
-				SalesforceSDKLogger.w(TAG, "Exception thrown while attempting to read array, attempting to read string value instead", e);
-			}
+			String[] mdmLoginServersLabels = runtimeConfig.getStringArrayStoredAsArrayOrCSV(AppServiceHostLabels);
 			if (mdmLoginServersLabels == null || mdmLoginServersLabels.length != mdmLoginServers.length) {
 				SalesforceSDKLogger.w(TAG, "No login servers labels provided or wrong number of login servers labels provided - using URLs for the labels");
 				mdmLoginServersLabels = mdmLoginServers;
@@ -407,6 +400,8 @@ public class LoginServerManager {
 			for (int i = 0; i < mdmLoginServers.length; i++) {
 				final String name = mdmLoginServersLabels[i];
 				final String url = mdmLoginServers[i];
+				if (name == null || url == null) { continue; }
+
 				final LoginServer server = new LoginServer(name, url, false);
 				persistLoginServer(
 						name,
@@ -516,7 +511,7 @@ public class LoginServerManager {
 	 * @param sharedPreferences The login server shared preferences
 	 * @return The adjusted login server index
 	 */
-	private int getIndexAdjustedToCustomLoginServerBounds(
+	private @NonNull Integer getIndexAdjustedToCustomLoginServerBounds(
 			final Integer index,
 			final SharedPreferences sharedPreferences
 	) {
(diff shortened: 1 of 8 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -92,7 +91,14 @@ public LoginServerManager(Context ctx) {
 				Context.MODE_PRIVATE);
 		runtimePrefs = ctx.getSharedPreferences(RUNTIME_PREFS_FILE,
 				Context.MODE_PRIVATE);
+
+		// Reset non-custom servers from mobile device management (MDM) and servers XML.
+		resetNonCustomLoginServers(runtimePrefs);
```
JohnsonEricAtSalesforce: @wmathurin, @brandonpage and @bbirman, before I update the tests and move this into ready-for-review I wanted to run the concept of the change past you all.

I chatted briefly with @bbirman to review the iOS behavior since the work item specifically hails that logic as working compared to Android's.  Currently, MSDK Android loads the `servers.xml` only once on install.  The work item mentions upgrade, but I didn't see that logic.  More, when MDM provides servers they're additive only.  When MDM drops a server, the app doesn't reflect that.

What I'm trying here and seems to manually test well is:
1. Drop both sets of non-custom servers on startup, including non-custom servers from `servers.xml` and MDM.
2. Load the current set of non-custom servers from either `servers.xml` _or_ MDM according to the existing logic, as much as possible.
3. When MDM is not enabled, preserve the user's custom login servers at the end of the list

I'm doing a little more testing around the behavior of the only show authorized hosts parameter.  I want to be sure the MDM flow still respects the user's custom servers.

The question to answer is: Does this seem like a reasonable update to the existing logic to resolve this work item? I can manipulate `servers.xml` and the MDM data and the app seems to keep the servers list adds, updates or removes non-custom servers just as we'd want.  The logic here is older and the methods are not as self-contained as our newer logic, so I wanted to be sure we all got to review any potential side-effects of this before moving further!

bbirman: For #3 (and pending your authorized hosts testing), on iOS the custom login servers would be preserved if MDM was enabled as long as "onlyShowAuthorizedHosts" is false

brandonpage: This sounds good to me.  

One thought though: Should we only remove MDM servers when network is available?  Security wise, it would be bad it:

1. Put device in airplane mode
2. Open app.  MDM servers are deleted and updated servers cannot be retrieved.  
3. Turn airplane mode off.
4. `LoginServerManager` constructor does not run again.  MDM server list is not respected.

Is that scenario possible?

JohnsonEricAtSalesforce: Thanks and that thought about network is a great topic.  I'll do some research!

JohnsonEricAtSalesforce: I took a manual look and also asked our agents.  RuntimeConfig does cache the policy locally and will give MSDK those values when no network is present ✅

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
 	public static final String WELCOME_LOGIN_URL = "https://welcome.salesforce.com/discovery";
 	public static final String SANDBOX_LOGIN_URL = "https://test.salesforce.com";
 
-	// Keys used in shared preferences.
```
JohnsonEricAtSalesforce, line 71: I've reviewed the diff in Github once already, so my pre-amble is that Github does a terrible job (as it often does) correlating the diff.  If one really wants to navigate the diff more easily, use Android Studio to compare the revisions 🎉

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
-	private static final String SERVER_URL_FILE = "server_url_file";
-	private static final String RUNTIME_PREFS_FILE = "runtime_prefs_file";
-	private static final String NUMBER_OF_ENTRIES = "number_of_entries";
-	private static final String SERVER_NAME = "server_name_%d";
-	private static final String SERVER_URL = "server_url_%d";
-	private static final String IS_CUSTOM = "is_custom_%d";
-	private static final String SERVER_SELECTION_FILE = "server_selection_file";
+	/**
+	 * Shared preferences when non-custom login servers are provided by resources servers.xml
```
JohnsonEricAtSalesforce, line 77: There are some new documentation comments to help many of the less-intuitively named properties and methods be clearer and easily reviewed in Android Studio.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
+	 */
+	@VisibleForTesting
```
JohnsonEricAtSalesforce, line 79: `@VisibleForTesting` appears in many places to enable the new tests.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
+
+	/**
+	 * The Android context used for shared preferences access
+	 */
 	private final Context ctx;
+
+	/**
+	 * The Android runtime configuration used for Mobile Device Management (MDM)
+	 */
+	private final RuntimeConfig runtimeConfig;
```
JohnsonEricAtSalesforce, line 114: Only two properties were added or modified: `runtimeConfig` is one and present for internal testing support.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
+
+	/** The resource id of the servers.xml file */
+	private final int serversXmlResourceId;
```
JohnsonEricAtSalesforce, line 117: Only two properties were added or modified: `serversXmlResourceId` is one and present for internal testing support.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
+		this(ctx, getRuntimeConfig(ctx), servers);
+	}
+
+	/**
+	 * Constructs a new login server manager.
+	 *
+	 * @param ctx The context
+	 */
+	@VisibleForTesting
+	public LoginServerManager(
```
JohnsonEricAtSalesforce, line 144: This new and private parameterized constructor is present for testing dependency injection.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
-				Context.MODE_PRIVATE);
-		runtimePrefs = ctx.getSharedPreferences(RUNTIME_PREFS_FILE,
-				Context.MODE_PRIVATE);
+		this.runtimeConfig = runtimeConfig;
+		this.serversXmlResourceId = serversXmlResourceId;
+		settings = ctx.getSharedPreferences(SERVER_URL_FILE, MODE_PRIVATE);
+		runtimePrefs = ctx.getSharedPreferences(RUNTIME_PREFS_FILE, MODE_PRIVATE);
+
+		// (Re-)initialize non-custom login servers provided by the resources servers.xml.
+		resetNonCustomLoginServers(settings);
```
JohnsonEricAtSalesforce, line 156: A crux change is here in the constructor to (re-)initialize the non-custom login servers from resources `server.xml` when applicable.  Note, this doesn't affect the user's custom login servers.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -68,31 +73,90 @@ public class LoginServerManager {
...
 		initSharedPrefFile();
+
+		// Select a default login server.
 		getSelectedLoginServer();
```
JohnsonEricAtSalesforce, line 160: The updated `getSelectedLoginServer` has new logic to detect when the "selected" login server is no longer in the "available" login servers and select the new default login server.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -118,39 +182,57 @@ public LoginServer getLoginServerFromURL(String url) {
...
-		} else {
+		// Fetch the selected login server.
+		final SharedPreferences selectedServerPrefs = ctx.getSharedPreferences(SERVER_SELECTION_FILE, MODE_PRIVATE);
+		final String selectedServerName = selectedServerPrefs.getString(SERVER_NAME, null);
+		final String selectedServerUrl = selectedServerPrefs.getString(SERVER_URL, null);
+		final boolean selectedServerIsCustom = selectedServerPrefs.getBoolean(IS_CUSTOM, false);
+		final boolean selectedServerHasValidNameAndUrl = selectedServerName != null && selectedServerUrl != null;
+
+		// Refresh the list of mobile device management (MDM) servers from the runtime config.
+		if (isRuntimeConfigAppServiceHostsSet()) {
```
JohnsonEricAtSalesforce, line 199: This refresh of the runtime config login servers wasn't present before but now nicely compliments the following call to `getLoginServers` so the selected login server will reflect when MDM has removed the currently selected login server.  This was an odd looking bug where the "selected" login server might no longer appear in the server picker after an MDM update.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -118,39 +182,57 @@ public LoginServer getLoginServerFromURL(String url) {
...
+			getLoginServersFromRuntimeConfig();
+		}
+
+		// Get the active list of login servers.
+		final List<LoginServer> loginServers = getLoginServers();
+
+		// Check if the selected login server is available in the active list of login servers.
+		final boolean selectedLoginServerIsAvailable = loginServers.stream().anyMatch(server ->
```
JohnsonEricAtSalesforce, line 207: Our agent tooling recommended this `stream` approach to search the list in Java.  This'll be easier in Kotlin, someday.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -187,12 +268,12 @@ public void useSandbox() {
...
-	 * Adds a custom login server to the shared pref file.
+	 * Adds a custom login server.
 	 *
-	 * @param name Server name.
-	 * @param url Server URL.
+	 * @param name The login server name
+	 * @param url The login server URL
 	 */
-	public void addCustomLoginServer(String name, String url) {
+	public void addCustomLoginServer(@NonNull String name, @NonNull String url) {
```
JohnsonEricAtSalesforce, line 276: Though it hadn't been reported, passing null for these parameters was a potential `NullPointerException`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -201,11 +282,12 @@ public void addCustomLoginServer(String name, String url) {
 			}
 		}
 
-		if (getLoginServersFromRuntimeConfig() == null) {
-			persistLoginServer(name, url, true, settings);
-		} else {
-			persistLoginServer(name, url, true, runtimePrefs);
-		}
+		persistLoginServer(
```
JohnsonEricAtSalesforce, line 285: The new `getSharedPreferences()` method lets this be one call to `persistLoginServer` with the correct `SharedPreferences`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -219,109 +301,152 @@ public void reset() {
...
-	 * @param server the server to remove
+	 * @param server                The server to remove
+	 * @param sharedPreferences     The shared preferences to remove the server from
+	 * @param allowNonCustomRemoval Boolean true allows the removal of non-custom login servers and
+	 *                              false does not
 	 */
-	public void removeServer(LoginServer server) {
-		List<LoginServer> servers = getLoginServers();
+	@VisibleForTesting
+	public void removeServer(
```
JohnsonEricAtSalesforce, line 329: This private and parameterized version of `removeServer` supports testing.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -219,109 +301,152 @@ public void reset() {
...
 	 */
 	public List<LoginServer> getLoginServers() {
-		List<LoginServer> allServers = getLoginServersFromRuntimeConfig();
-		if (allServers == null) {
-			allServers = getLoginServersFromPreferences();
-		} else {
-			allServers = getLoginServersFromPreferences(runtimePrefs);
-		}
-		return allServers;
+		return getLoginServersFromPreferences(getSharedPreferences());
```
JohnsonEricAtSalesforce, line 373: `getLoginServers` also gets a lot simpler since it can have one call to `getLoginServersFromPreferences` with the correct `SharedPreferences` parameter.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -219,109 +301,152 @@ public void reset() {
...
 				SalesforceSDKLogger.w(TAG, "No login servers labels provided or wrong number of login servers labels provided - using URLs for the labels");
 				mdmLoginServersLabels = mdmLoginServers;
 			}
-			final List<LoginServer> storedServers = getLoginServersFromPreferences(runtimePrefs);
-			for (int i = 0; i < mdmLoginServers.length; i++) {
-				final String name = mdmLoginServersLabels[i];
-				final String url = mdmLoginServers[i];
+
+			// Reset non-custom servers from Mobile Device Management (MDM).
+			resetNonCustomLoginServers(runtimePrefs);
```
JohnsonEricAtSalesforce, line 416: Another crux change is here in `getLoginServersFromRuntimeConfig` to reset the non-custom login servers from that `SharedPreferences` so that strange things don't happen as the MDM data changes.  My testing from `dev` showed that adding servers worked but deleting servers had no effect.  This also meant that "editing" a server resulted in a duplication.  The list always reflects MDM now.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -219,109 +301,152 @@ public void reset() {
...
+
+			// Null-cleanse MDM login server URLs and names.
+			final List<String> mdmLoginServersList = asList(mdmLoginServers);
+			mdmLoginServersList.removeIf(Objects::isNull);
```
JohnsonEricAtSalesforce, line 420: Our agent tools recommended this approach for null-cleansing the MDM data which aided in simplifying the tests and avoiding code coverage noise.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -349,28 +474,16 @@ public void reorderCustomLoginServer(
...
+		final List<LoginServer> servers = getLoginServers();
+		updatedIndex = getIndexAdjustedToCustomLoginServerBounds(updatedIndex, getSharedPreferences());
 
 		// Update the login server list.
 		loginServers.remove(originalIndex);
 		loginServers.add(updatedIndex, originalLoginServer);
 
 		// Edit each login server indexed after the updated index.
-		final Editor editor = settings.edit();
+		final Editor editor = getSharedPreferences().edit();
```
JohnsonEricAtSalesforce, line 486: This was a latent bug where the wrong `SharedPreferences` was hard-coded even when MDM was active.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -414,6 +527,32 @@ public void replaceCustomLoginServer(
...
 	}
 
+	/**
+	 * Adjusts a login server index to be within the bounds of the custom login servers.
+	 *
+	 * @param index The login server index
+	 * @param sharedPreferences The login server shared preferences
+	 * @return The adjusted login server index
+	 */
+	private @NonNull Integer getIndexAdjustedToCustomLoginServerBounds(
```
JohnsonEricAtSalesforce, line 537: These utility methods to adjust a login server's index to the appropriate bounds of non-custom or custom servers are useful, but for future when `SharedPreferences` is replaced it'd be amazing to see the custom servers stored separately from the non-custom ones entirely.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -435,83 +574,142 @@ private List<LoginServer> getLegacyLoginServers() {
...
-					eventType = xml.next();
-				} catch (XmlPullParserException | IOException e) {
-					SalesforceSDKLogger.w(TAG, "Exception thrown while parsing XML", e);
-				}
+			}
+			try {
+				eventType = xml.next();
+			} catch (Exception e) {
+				SalesforceSDKLogger.w(TAG, "Exception thrown while parsing XML", e);
+				break;
```
JohnsonEricAtSalesforce, line 602: Without this break, this exception turned into infinite recursion and an app hang which I caught in testing while adding coverage to this (previously?) untested line.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -435,83 +574,142 @@ private List<LoginServer> getLegacyLoginServers() {
...
 		}
 		int numServers = servers.size();
 		final Editor edit = settings.edit();
 		for (int i = 0; i < numServers; i++) {
 			final LoginServer curServer = servers.get(i);
-			edit.putString(String.format(Locale.US, SERVER_NAME, i), curServer.name.trim());
-			edit.putString(String.format(Locale.US, SERVER_URL, i), curServer.url.trim());
-			edit.putBoolean(String.format(Locale.US, IS_CUSTOM, i), curServer.isCustom);
-			if (i == 0) {
+			persistLoginServer(
```
JohnsonEricAtSalesforce, line 642: All login server persistence happens now via `persistLoginServer` since it has all the logic needed to correctly order and sanitize the `SharedPreferences`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -435,83 +574,142 @@ private List<LoginServer> getLegacyLoginServers() {
...
+					curServer.isCustom,
+					settings
+			);
+
+			// Set the default login server to the first entry once.
+			if (i == 0 && ctx.getSharedPreferences(SERVER_SELECTION_FILE, MODE_PRIVATE).getAll().isEmpty()) {
 				setSelectedLoginServer(curServer);
 			}
 		}
-		edit.putInt(NUMBER_OF_ENTRIES, numServers);
```
JohnsonEricAtSalesforce, line 492: This `NUMBER_OF_ENTRIES` update happens in `persistLoginServer` now.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -435,83 +574,142 @@ private List<LoginServer> getLegacyLoginServers() {
...
+	 * @param name              The login server name
+	 * @param url               The login server URL
+	 * @param isCustom          boolean true for non-custom (managed) login servers, false for
+	 *                          custom (user-entered) login servers
+	 * @param sharedPreferences The shared preferences
 	 */
-	private void persistLoginServer(String name, String url, boolean isCustom, SharedPreferences sharedPrefs) {
-		if (name == null || url == null) {
-			return;
+	private void persistLoginServer(final @NonNull String name,
```
JohnsonEricAtSalesforce, line 666: Another big crux change is here to `persistLoginServer` which now has the ability to correctly order non-custom and custom login servers and all the other login servers.  This is where the most work actually happened to make the non-custom login server list mutable.

libs/SalesforceSDK/src/com/salesforce/androidsdk/config/LoginServerManager.java
```diff
@@ -520,22 +718,40 @@ private void persistLoginServer(String name, String url, boolean isCustom, Share
 	 * @param prefs SharedPreferences file.
 	 * @return List of all saved servers.
 	 */
-	private List<LoginServer> getLoginServersFromPreferences(SharedPreferences prefs) {
+	@VisibleForTesting
+	public @NonNull List<LoginServer> getLoginServersFromPreferences(final SharedPreferences prefs) {
 		int numServers = prefs.getInt(NUMBER_OF_ENTRIES, 0);
 		if (numServers == 0) {
-			return null;
+			return new ArrayList<>();
```
JohnsonEricAtSalesforce, line 725: I left most other methods returning `null` if the did on `dev` but this one actually made sense to eliminate the null checks at all the private call sites and simplify the logic and tests.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/LoginServerManagerTest.kt
```diff
@@ -0,0 +1,817 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.xmlpull.v1.XmlPullParserException
+
+@RunWith(AndroidJUnit4::class)
+@SmallTest
+class LoginServerManagerTestKt {
```
JohnsonEricAtSalesforce: Where `mockk` was needed, I wrote new tests directly in Kotlin.

JohnsonEricAtSalesforce:
Testing Notes: I tested this via both manual testing on device and via all the new unit tests.  In particular, I also used a temporary method to log the contents of both `SharedPreferences` during testing so I could be 100% sure the data looked rock solid.  That was really tedious, but it did reveal interesting data quality issues that unit tests and even manual testing could miss - there was one bug where data indexed _above_ the number of servers value was orphaned after edits.  The unit tests and even the app didn't detect that, but I saw it by putting my own eyes on it ✅

Testing requires restarting the app after:
1. Adding to `servers.xml`
2. Editing `servers.xml`
3. Removing from `servers.xml`
...And since I didn't have a managed device for the MDM tests I had to update the values obtained from `RuntimeConfig` in the code to...
4. Add servers
5. Edit servers
6. Remove servers

Each of these tests should be tested with a cold start of the app to ensure the (re-)initialization of the `LoginServerManager` correctly restores the servers.

Each of these tests should be performed with and without the presence of custom login servers.

The unit tests provide more destructive testing, such as inserting null values and malformed XML in all kinds of places they shouldn't be.  I spent five days in manual testing and automated testing development since I'd view this as a higher risk change.  This was where I found and resolved a number of `NullPointerException` crashes and one infinite hang of the app when XML is malformed. 😅

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/LoginServerManagerTest.kt
brandonpage: Maybe rename this one to `LoginServerManagerMockTest`?  

At the very least let's make the file name match the class so we don't have `LoginServerManageTest.java` and `LoginServerManageTest.kt`.

brandonpage (approved):

wmathurin (approved):

PR #2847, opened 2026-03-06
Add UI Tests

brandonpage:
Tests added:
- `BootConfigLoginTests` - tests that use the (as currently defined) CA Opaque app in AuthFlowTester's `bootconfig.xml` as is.
- `CAScopeSelectionLoginTests` - Connected App tests with mixed scopes, flow and hybrid token options.
- `ECALoginTests` - ECA Opaque and JWT tests with various login options.
- `BeaconLoginTests` - Beacon Opaque and JWT tests with various login options.
- `MultiUserLoginTests` - Tests with multiple users logged in.  Different app types, scopes, etc are used.  
   - The Token Migration tests in this class also ensure that auth tokens are stored/retained correctly since the account, rest client, etc are recreated on user switch.  This is important since it is not possible to restart or cold boot the app mid test (like iOS does) with Compose UI Test or Espresso.

The new `testMultiUser_tokenMigration_backgroundUser` test uncovered the fact that token migration would always set the specified user as the current user.  That has been fixed in this PR and unit tests have been added to cover the SDK code change.

To make the above test possible the test app UI was updated to allow for user selection when migrating:
share_5015376007694299033

Other Noteworthy Changes:
- AuthFlowTester now has a `README.md` that describes the UI Tests coverage and how to use the app for manual testing.
- CI has been updated to make efficient use of as many test users as are provided in `ui_test_config.json`.  All users will be used at once for single user tests and the `MultiUserLoginTests` are batched to run as many as possible but avoid runs using the same user simultaneously.
- CI will run a handful of individual tests (5 as of now) on a single API level for PRs. 
- The AuthFlowTester debug APK is now archived on every test run.

There are separate WIs for Advanced Auth and Welcome test that will be finished soon.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
@@ -26,9 +26,12 @@
  */
 package com.salesforce.androidsdk.auth
 
+import android.accounts.Account
+import android.accounts.AccountManager
 import android.content.Context
 import android.content.Intent
 import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
+import android.os.Bundle
 import androidx.annotation.VisibleForTesting
 import com.salesforce.androidsdk.R.string.sf__generic_authentication_error
 import com.salesforce.androidsdk.R.string.sf__generic_authentication_error_title
@@ -44,6 +47,7 @@ import com.salesforce.androidsdk.analytics.SalesforceAnalyticsManager
 import com.salesforce.androidsdk.app.Features.FEATURE_BIOMETRIC_AUTH
 import com.salesforce.androidsdk.app.Features.FEATURE_SCREEN_LOCK
 import com.salesforce.androidsdk.app.SalesforceSDKManager
+import com.salesforce.androidsdk.app.SalesforceSDKManager.Companion.encryptionKey
 import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse
 import com.salesforce.androidsdk.auth.OAuth2.addAuthorizationHeader
 import com.salesforce.androidsdk.auth.OAuth2.callIdentityService
@@ -170,27 +174,30 @@ internal suspend fun onAuthFlowComplete(
 
     // Save the user account
     addAccount(account)
-    userAccountManager.createAccount(account)
-    userAccountManager.switchToUser(account)
 
-    // Init user logging
-    updateLoggingPrefs(account)
+    if (tokenMigration) {
+        userAccountManager.persistAccount(account)
+    } else {
+        userAccountManager.createAccount(account)
+        userAccountManager.switchToUser(account)
 
-    // Send User Switch Intent, create user and switch to user.
-    val numAuthenticatedUsers = userAccountManager.authenticatedUsers?.size ?: 0
-    val userSwitchType = when {
-        // We've already authenticated the first user, so there should be one
-        numAuthenticatedUsers == 1 -> USER_SWITCH_TYPE_FIRST_LOGIN
+        // Init user logging
+        updateLoggingPrefs(account)
 
-        // Otherwise we're logging in with an additional user
-        numAuthenticatedUsers > 1 -> USER_SWITCH_TYPE_LOGIN
+        // Send User Switch Intent, create user and switch to user.
+        val numAuthenticatedUsers = userAccountManager.authenticatedUsers?.size ?: 0
+        val userSwitchType = when {
+            // We've already authenticated the first user, so there should be one
+            numAuthenticatedUsers == 1 -> USER_SWITCH_TYPE_FIRST_LOGIN
 
-        // This should never happen but if it does, pass in the "unknown" value
-        else -> USER_SWITCH_TYPE_DEFAULT
-    }
-    userAccountManager.sendUserSwitchIntent(userSwitchType, null)
+            // Otherwise we're logging in with an additional user
+            numAuthenticatedUsers > 1 -> USER_SWITCH_TYPE_LOGIN
 
-    if (!tokenMigration) {
+            // This should never happen but if it does, pass in the "unknown" value
+            else -> USER_SWITCH_TYPE_DEFAULT
+        }
+        userAccountManager.sendUserSwitchIntent(userSwitchType, null)
+
         // Kickoff the end of the flow before storing mobile policy to prevent launching
         // the main activity over/after the screen lock.
         startMainActivity()
@@ -494,4 +501,35 @@ internal fun handleDuplicateUserAccount(
             })
         }
     }
+}
+
+/**
+ * Persists account data to the Android AccountManager without changing
+ * the current user. This is needed for token migration of background
+ * users where [UserAccountManager.createAccount] cannot be used because
+ * it unconditionally calls [UserAccountManager.storeCurrentUserInfo].
+ */
+private fun UserAccountManager.persistAccount(
+    userAccount: UserAccount,
+    accountType: String = SalesforceSDKManager.getInstance().accountType,
+    acctManager: AccountManager = AccountManager.get(SalesforceSDKManager.getInstance().appContext),
+) {
+    val account = Account(userAccount.accountName, accountType)
+    val password = SalesforceSDKManager.encrypt(userAccount.refreshToken, encryptionKey)
+    val created = acctManager.addAccountExplicitly(account, password, /* userdata = */ Bundle())
+
+    // addAccountExplicitly fails if the account already exists, so update the refresh token.
+    if (!created) {
+        acctManager.setPassword(account, password)
+    }
+
+    // Cache auth token to avoid an unnecessary refresh on first access.
+    acctManager.setAuthToken(
+        account,
+        /* authTokenType = */ AccountManager.KEY_AUTHTOKEN,
+        /* authToken = */ SalesforceSDKManager.encrypt(userAccount.authToken, encryptionKey),
+    )
+
+    // Persist all remaining user data via the existing public helper.
+    updateAccount(account, userAccount)
 }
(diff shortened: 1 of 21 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -879,7 +879,7 @@ public TokenEndpointResponse(Map<String, String> callbackUrlParams, List<String>
                 cookieSidClient = callbackUrlParams.get(COOKIE_SID_CLIENT);
                 sidCookieName = callbackUrlParams.get(SID_COOKIE_NAME);
                 parentSid = callbackUrlParams.get(PARENT_SID);
-                tokenFormat = callbackUrlParams.get(TOKEN_FORMAT);
+                tokenFormat = callbackUrlParams.getOrDefault(TOKEN_FORMAT, "");
```
brandonpage, line 882: This was a bug that came out of UI testing.  The `tokenFormat` for opaque tokens would either be empty string or null depending on when/where the user was created so it won't be null anywhere after this change.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/AuthenticationUtilitiesTest.kt
```diff
@@ -370,6 +377,60 @@ class AuthenticationUtilitiesTest {
         verify(exactly = 0) { startMainActivity.invoke() }
     }
 
+    @Test
```
brandonpage, line 380: Unit Tests for SDK change.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -170,27 +174,30 @@ internal suspend fun onAuthFlowComplete(
 
     // Save the user account
     addAccount(account)
-    userAccountManager.createAccount(account)
-    userAccountManager.switchToUser(account)
 
-    // Init user logging
-    updateLoggingPrefs(account)
+    if (tokenMigration) {
+        userAccountManager.persistAccount(account)
```
brandonpage, line 179: This is the main SDK change.  It avoids changing the current user and sending the user switch broadcast that should not be done for token migration.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/ChromeCustomTabPageObject.kt
```diff
@@ -26,62 +26,93 @@
...
-private const val TIMEOUT = 5_000L
+private const val RETRY_COUNT = 3
 
 /**
  * Handles Custom Tab interactions.
  * UiAutomator is required here because the browser (often Chrome) runs in a
  * separate process that Espresso and Compose Test APIs cannot access.
  */
-class CustomTabPageObject {
+class ChromeCustomTabPageObject(composeTestRule: ComposeTestRule): LoginPageObject(composeTestRule) {
```
brandonpage, line 43: The `AdvancedAuthLoginTests` are not included in this PR (still a little org setup to finish and a Firebase complication to work around) but will be in the next.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
```diff
@@ -36,6 +37,18 @@ abstract class BasePageObject(val composeTestRule: ComposeTestRule) {
     fun getString(id: Int) = context.getString(id)
 
     companion object {
-        const val TIMEOUT_MS: Long = 2_000
+        val isFtl: Boolean by lazy {
+            Settings.System.getString(
+                InstrumentationRegistry.getInstrumentation().targetContext.contentResolver,
+                /* name = */ "firebase.test.lab"
+            ) == "true"
+        }
```
brandonpage, line 45: This nifty trick allows us to do things like set different timeout values for local vs CI runs or skip tests in certain scenarios (I am seeing a lot of instances of Chrome crashing on lower API levels in Firebase :/).

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/AuthFlowTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.salesforce.samples.authflowtester.pageObjects.LoginOptionsPageObject
+import com.salesforce.samples.authflowtester.pageObjects.ChromeCustomTabPageObject
+import com.salesforce.samples.authflowtester.pageObjects.LoginPageObject
+import com.salesforce.samples.authflowtester.testUtility.ScopeSelection.EMPTY
+import com.salesforce.samples.authflowtester.testUtility.KnownLoginHostConfig.REGULAR_AUTH
+import com.salesforce.samples.authflowtester.testUtility.KnownLoginHostConfig.ADVANCED_AUTH
+import com.salesforce.samples.authflowtester.testUtility.KnownAppConfig.CA_OPAQUE
+import org.junit.Rule
+
+abstract class AuthFlowTest {
```
brandonpage, line 50: New base test class that handles all of the boiler-plate setup and common verification functions.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/AuthFlowTest.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    val loginOptions = LoginOptionsPageObject(composeTestRule)
+    val app = AuthFlowTesterPageObject(composeTestRule)
+
+    val user: KnownUserConfig by lazy {
+        val minSdk = InstrumentationRegistry.getInstrumentation().targetContext
+            .applicationInfo.minSdkVersion
+        val userNumber = (Build.VERSION.SDK_INT - minSdk) % KnownUserConfig.values().count()
+        KnownUserConfig.values()[userNumber]
+    }
```
brandonpage, line 72: This might look a little strange, but it ensures each test runs uses a different user.  By starting at `minSdkVersion` we can basically count up and never have a collision:

API 28 -> User  1
API 29 -> User  2
...

`reusable-ui-workfow.yaml` runs batches with a max size to match the number of users (for single user tests).

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/AuthFlowTest.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    // For MultiUser tests
+    val otherUser: KnownUserConfig by lazy {
+        val userNumber = (user.ordinal + 1) % KnownUserConfig.values().count()
+        KnownUserConfig.values()[userNumber]
+    }
```
brandonpage, line 78: Multi-User tests are run separately and the bash in `reusable-ui-workfow.yaml` groups API level to avoid collisions.  So the groups look like:

[API 28, API 30], [API 32, API 34], ...  [API 33, API 35], etc.

Since we can only run 2 levels at a time like this I should really add a 6th user, it would decrease test time for multi-user by 50%.

brandonpage:
As always, the tests will not run with the latest yaml configuration for security (but I will update you with the results from my fork).

 @wmathurin @JohnsonEricAtSalesforce As I was experimenting and testing the new (complicated) bash batching that we could avoid the above limitation by moving all of the bash from yaml to scripts.  As long as the name (and parameters) of those scripts don't change the updates would be immediately seen on PRs.  As long as we don't run them before the permission check it should be safe.  It could make local testing a bit easier too.  Thoughts?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: So you will allows the app to migrate the refresh token of a user that is not currently active? Since there might be a screen shown (allow/deny) - wouldn't that be confusing?

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
wmathurin: Nice - in my latest PR, I made the timeout configurable through the environment and I'm giving more time on CI: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4002/changes/62b53f0fca16659e81c7bf94d042734d6d0e1776

wmathurin (approved):

brandonpage:
> As always, the tests will not run with the latest yaml configuration for security (but I will update you with the results from my fork).
> 
> @wmathurin @JohnsonEricAtSalesforce As I was experimenting and testing the new (complicated) bash batching that we could avoid the above limitation by moving all of the bash from yaml to scripts. As long as the name (and parameters) of those scripts don't change the updates would be immediately seen on PRs. As long as we don't run them before the permission check it should be safe. It could make local testing a bit easier too. Thoughts?

The runs triggered locally seem to be going well so I will merge this as soon as I verify the test reporting is working.

Single User:
Screenshot 2026-03-06 at 6 06 06 PM

Multi User:
Screenshot 2026-03-06 at 4 04 49 PM

brandonpage:
I think this work if finished.  I am seeing a 100% pass rate on the last ~10 runs if I allow 1 retry.  On average I see ~1-3 tests out of 40 flap due to environmental issues in CI (usually lower API levels) but always pass on the retry. 

I have updated the bash to merge the original and retry result files to be the format the test reporter expects.  So now tests that pass on the retry are counted as success, but list the tests that flapped:
Screenshot 2026-03-08 at 11 13 56 PM

PR #2851, opened 2026-03-17
@W-18574835: Add unique locators for Add Connection page UI elements

wmathurin:
Added testTag and contentDescription to the Add Connection page components for UI test automation and accessibility:
- Connection name field (testTag: sf__picker_custom_label)
- Connection URL field (testTag: sf__picker_custom_url)
- Apply button (testTag: sf__apply_button)

This enables stable automation while maintaining human-readable screen reader descriptions through separate accessibility string resources.


brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -522,7 +529,9 @@ internal fun AddConnection(
                 modifier = Modifier
                     .fillMaxWidth()
                     .padding(PADDING_SIZE.dp)
-                    .focusRequester(focusRequester),
+                    .focusRequester(focusRequester)
+                    .testTag("sf__picker_custom_label")
```
brandonpage, line 533: I don't know if test tag is actually needed.  Semantics contentDescription always works for me 🤷

PR #2853, opened 2026-03-20
Add advanced auth testing for Beacon apps

wmathurin:
## Summary
- Add `AdvancedAuthBeaconLoginTests` to test Beacon apps via Chrome Custom Tabs
- Make `BeaconLoginTests` extensible with template methods for login host/user config  
- Fix Chrome Custom Tab text entry by clicking fields before setText
- Add pre-flight server reset to ensure WebView context for login options UI

### DIFF
--- a/native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/AdvancedAuthBeaconLoginTests.kt
+++ b/native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/AdvancedAuthBeaconLoginTests.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2026-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of salesforce.com, inc. nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission of salesforce.com, inc.
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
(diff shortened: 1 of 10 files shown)


native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/AuthFlowTest.kt
```diff
@@ -90,6 +90,23 @@ abstract class AuthFlowTest {
         }
     }
 
+    /**
+     * Ensures we're on REGULAR_AUTH server before opening Login Options.
+     * Server selection is "sticky" so previous test might have left it on ADVANCED_AUTH.
+     */
+    private fun ensureRegularAuthServer() {
```
wmathurin, line 97: To bring login options from a chrome custom tab - we need to use UIAutomator. It gets messy since the situation could arise if the last test was using advanced auth even if the current test uses regular auth.
I do something similar on iOS: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift#L76

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/AdvancedAuthBeaconLoginTests.kt
```diff
@@ -0,0 +1,68 @@
...
+ * NB: Requires intent filters in AndroidManifest.xml matching the beacon redirect URIs:
+ * - beaconadvancedopaque://success/done (for Beacon Opaque)
+ * - beaconadvancedjwt://success/done (for Beacon JWT)
+ *
+ * NB: Tests use the second user from ui_test_config.json (advanced_auth host) to avoid
+ * login conflicts when running in parallel with regular auth tests.
+ */
+@RunWith(AndroidJUnit4::class)
+@LargeTest
+class AdvancedAuthBeaconLoginTests : BeaconLoginTests() {
```
wmathurin, line 49: Same approach as iOS - running the same tests but with a different login host (and user).

native/NativeSampleApps/AuthFlowTester/src/main/AndroidManifest.xml
```diff
@@ -32,6 +32,26 @@
                 <category android:name="android.intent.category.BROWSABLE" />
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
+
+            <!-- Intent filter for advanced authentication with Beacon Opaque app -->
+            <intent-filter>
+                <data android:scheme="beaconadvancedopaque"
```
wmathurin, line 38: @brandon-page the existing callback uri work just fine. I did not need to change them.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/AuthFlowTest.kt
brandonpage: Would it be easier/more stable to set the selected server via SalesforceSDKManager in `cleanup()`?

brandonpage (approved):

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/testUtility/AuthFlowTest.kt
wmathurin: You are right ! 
On iOS, the tests run from a separate process, everything has to be done through the UI.

brandonpage:
@wmathurin Btw, I _think_ the syntax error that causes the UI Tests to not run on PR is an extra space after the `\` on the timeout line.  🙄

wmathurin:
> @wmathurin Btw, I _think_ the syntax error that causes the UI Tests to not run on PR is an extra space after the `\` on the timeout line. 🙄

Fixed with https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2853/commits/ef36a2795445b4eb4e2e8a5d84e345679f77b210 - but we will know with the next PR ;-)

PR #2649, opened 2025-01-09
Changing min sdk from 26 to 28

wmathurin:
Changes:
- Changed minSdk from 26 to 28
- Removed checks for 28/Android.P in code
- No longer using BC provider (see https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html)

NB: Also updated GUI driven [triggers](https://app.circleci.com/settings/project/github/forcedotcom/SalesforceMobileSDK-Android/triggers) on Circle CI (removed 26, 27 and added 35)

PR #2857, opened 2026-03-25
Add Claude skill for updating min SDK version

wmathurin:
This skill documents the complete process for updating the minimum Android SDK version across the project, based on lessons learned from PRs #2649 and #2529.

The skill covers:
- Updating all build.gradle.kts files (libs and sample apps)
- Updating gradle.properties for Cordova
- Updating GitHub Actions workflow API ranges
- Cleaning up obsolete API level checks in code
- Updating AndroidManifest.xml permissions
- Managing SDK-related TODO comments
- Verification and PR guidelines

### DIFF
--- a/.claude/skills/update-min-sdk.md
+++ b/.claude/skills/update-min-sdk.md
@@ -95,7 +95,55 @@ When moving from 24 to 26:
 - Simplified system UI visibility settings
 - Removed SalesforceSDK_AccessibleNav style fallback
 
-### 5. Update AndroidManifest.xml (if needed)
+### 5. Review New API Features and Opportunities
+
+**Purpose:** Proactively identify modernization opportunities based on new APIs available at NEW_MIN_SDK.
+
+**Process:**
+1. **Review Android documentation** for the new minimum API level:
+   - Feature summary: `https://developer.android.com/about/versions/{VERSION}/summary`
+   - Migration guide: `https://developer.android.com/about/versions/{VERSION}/migration`
+   - Behavior changes: `https://developer.android.com/about/versions/{VERSION}/behavior-changes-all`
+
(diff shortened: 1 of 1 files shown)


bbirman:
I haven't started for iOS but I was wondering about how to handle version bumps for the templates, are you going to do a separate skill for that?

wmathurin:
> I haven't started for iOS but I was wondering about how to handle version bumps for the templates, are you going to do a separate skill for that?

Yeah we should have skills per repo - then at the end we could have skills in the workspace that refers to the corresponding skills in the various repo.

NB: I'm not writing the skill by hand, I tell claude to draw lessons from relevant past PRs (I provide the list - because their title are not necessarily very consistent) and then I review the output.

brandonpage (approved):
LGTM, but I have a suggestion:  

Can we add a step to take action or make suggestions based on the new min API?  I imagine we often don't have a `TODO` or `if( min == something)` condition simply because we don't know the new API exists.  Perhaps pointing the LLM to something like the [feature and changelist](https://developer.android.com/about/versions/16/summary) or [migration guide](https://developer.android.com/about/versions/16/migration) could be helpful?

wmathurin:
> LGTM, but I have a suggestion:
> 
> Can we add a step to take action or make suggestions based on the new min API? I imagine we often don't have a `TODO` or `if( min == something)` condition simply because we don't know the new API exists. Perhaps pointing the LLM to something like the [feature and changelist](https://developer.android.com/about/versions/16/summary) or [migration guide](https://developer.android.com/about/versions/16/migration) could be helpful?

Added step here: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2857/commits/4eed12a875ee426d4bf8bb47371f4f9e22b48274

PR #2860, opened 2026-03-30
Fix unit tests that fail based on API level.

brandonpage:
- `List.getLast()` is a Java 21 API.  Even though our _project_ supports Java 21, older versions of Android did not.  That means the the use of that API immediately fails when the test runs on older devices.
- API < 30 will always return `BIOMETRIC_WEAK`.


JohnsonEricAtSalesforce (approved):
Thanks! 🤩

PR #2862, opened 2026-04-02
Claude/flutter sdk implementation nc uv2

jaygdesai:


brandonpage:
We appreciate the effort, but there is not enough interest in a Flutter SDK to warrant the maintenance cost at this time.

JayTataGit:
Thanks for the feedback @brandonpage. If most of my applications are written in Flutter and I am migrating to Salesforce as my CRM, what options do I have?

PR #2868, opened 2026-04-15
@W-21933885: [MSDK Android] App Attestation Implementation

JohnsonEricAtSalesforce:
🎸 _*Ready For Final Review - Test Failures Appear Unrelated*_ 🥁

  This adds Mobile SDK Android support for:

1. The new "Challenge" API offered by the Salesforce App Attestation External Client App (ECA) Plug-In
  1.1. The "Challenge" API creates challenges for use as the "hash" when creating Google Play App Integrity API "tokens"
2. Google Play App Integrity API
  2.1. Tokens can be created using the "Challenge" and then passed to authorization and token refresh requests for validation by the App Attestation ECA Plug-In

  The app's entry point is calling a new "updateAppAttestationClient" method on the SalesforceSDKManager with the relevant Google Cloud Project ID.  Once that's provided, the manager and the new AppAttestationClient manage the interaction with the Challenge API and Google Play Integrity API.  The AppAttestationClient provides "Attestation" values to the existing authorization and token refresh logic when applicable.

  Most of the new code follows our existing patterns.  I'll include a source code walk-though in the differences as well.  None of this is set in stone and we can incorporate feedback here or in follow-up work items.

  I've tested this by creating a new test app from the REST Explorer sample.  That app is configured for deployment to a Google Play Developer account that is configured for App Integrity with a matching Google Cloud project.  I've been able to deploy the app to my internal release track.  When installed from there, I'm able to successfully authenticate to the internal MSDK App Attestation Test org I've been given access to.

  Test coverage is nearing 100% which is great given authentication's critical priority and high availability threshold.

  The Google Play App Integrity API has public documentation here https://developer.android.com/google/play/integrity/overview

  There's no user interface component for this change in either mobile or web for screenshots or a demo, alas!

  Here's an example of how the client app could (re-)initialize app attestation each time the selected login server changes.
```
        // Prepare for Salesforce App Attestation.
        getInstance().loginServerManager.selectedServer.observeForever { selectedLoginServer ->
            selectedLoginServer.url.toUri().host.let { selectedLoginServerHost ->
                if (selectedLoginServerHost == "msdkappattestationtestorg.test1.my.pc-rnd.salesforce.com") {
                    Pair<String?, Long?>(selectedLoginServerHost, 762473690072L)
                } else {
                    Pair<String?, Long?>(selectedLoginServerHost, null)
                }
            }.let { appAttestationConfiguration ->
                getInstance().updateAppAttestationClient(appAttestationConfiguration.first ?: "_Ignored_Value_", appAttestationConfiguration.second)
            }
        }
```


JohnsonEricAtSalesforce:
The last mile of code coverage is taking some time to wrap up today.  While I'm waiting on the last couple of CodeCov reports, I'll add some walk-through commentary even though I'd still recommend holding reviews until I mark the pull request Ready For Review.

.github/workflows/reusable-lib-workflow.yaml
```diff
@@ -27,6 +27,10 @@ jobs:
       - name: Install Dependencies
         env:
           TEST_CREDENTIALS: ${{ secrets.TEST_CREDENTIALS }}
+          # On PR runs, only SalesforceReact consumes the bundled index.android.bundle,
+          # so skip the yarn install + react-native bundle step for every other lib to
+          # save ~3-5 min per matrix job. Nightly runs still produce the bundle.
+          SKIP_REACT_NATIVE_BUNDLE: ${{ (inputs.is_pr && inputs.lib != 'SalesforceReact') && '1' || '0' }}
```
JohnsonEricAtSalesforce, line 33: This _is_ unrelated to the feature, but could be useful to consider.  While spending a lot of time watching ~25m CI runs, I asked our tools where a quick win would be in trimming that time.  The analysis found that we're re-running yarn from scratch for all the modules instead of just for `SalesforceReact`.  I believe this brought my run down to ~12m!  I need to verify that over a few more runs, but if that's the case this could be a nice optimization.  @brandonpage?

.github/workflows/reusable-lib-workflow.yaml
```diff
@@ -109,7 +113,6 @@ jobs:
 
           if $IS_PR ; then
             LEVELS_TO_TEST=$PR_API_VERSION
-            RETRIES=1
```
JohnsonEricAtSalesforce, line 112: @brandonpage, a little later you'll see a fix in the `RestClientTest.java` when I had generated by our tools to resolve unreliable behavior I saw in that test.  I've been marking it `@Ignore` a lot since it's so unpredictable.  The fix looks solid - It's so solid that our tools believe we don't need retry anymore.  I'd love to see a "faster fail" on pull request runs since they can take a very long time.  Thoughts?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -226,6 +227,54 @@ open class SalesforceSDKManager protected constructor(
...
+     * The client side implementation of the Salesforce App Attestation External
+     * Client App (ECA) Plugin or null when app attestation is disabled.
+     *
+     * This property is not intended for public use outside of Salesforce Mobile
+     * SDK
+     *
+     * TODO: Make this Kotlin-internal once it is no longer referenced by Java. ECJ20260420
+     */
+    @Volatile
+    var appAttestationClient: AppAttestationClient? = null
```
JohnsonEricAtSalesforce, line 240: A crux change is creating a new object to encapsulate all the things for the new Salesforce "Challenge" API, the Integrity Token Provider, the Token and providing that in the "Attestation" format the auth and token refresh endpoints now expect.  That's here.

Our tools had some great suggestion around making this property thread safe, so I added @Volatile, the private setter and a dedicated lock object based on tool feedback.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -226,6 +227,54 @@ open class SalesforceSDKManager protected constructor(
...
+     * for the app in Google Play Console's Play Integrity API and provided to
+     * the Salesforce App Attestation External Client App Plugin.
+     *
+     * @param apiHostName The Salesforce App Attestation External Client App
+     * (ECA) Plugin Challenge API Host Name.  This usually matches the selected
+     * login server
+     * @param googleCloudProjectId The Google Cloud Project ID or null to
+     * disable Salesforce App Attestation
+     */
+    fun updateAppAttestationClient(
```
JohnsonEricAtSalesforce, line 260: If one reads the description of this pull request, this is the entry point for an app to actually enable App Attestation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AppAttestationClient.kt
```diff
@@ -0,0 +1,234 @@
...
+ * @param integrityManager The Google Play App Integrity API Integrity Manager.
+ * This parameter is intended for testing purposes only. Defaults to a new
+ * instance
+ * @param remoteAccessConsumerKey The Salesforce Connected App (CA) or External
+ * Client App (ECA)remote access consumer key, usually provided by the boot
+ * config
+ * @param restClient The REST client, usually provided by the Salesforce SDK
+ * Manager's unauthenticated REST client
+ */
+class AppAttestationClient(
```
JohnsonEricAtSalesforce, line 66: This object is the heart 'n soul of App Attestation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AppAttestationClient.kt
```diff
@@ -0,0 +1,234 @@
...
+    internal val restClient: RestClient,
+) {
+
+
+    /** The Google Play Integrity API Token Provider */
+    @VisibleForTesting
+    internal var integrityTokenProvider: StandardIntegrityTokenProvider? = null
+
+    init {
+        prepareIntegrityTokenProvider()
```
JohnsonEricAtSalesforce, line 88: This is the "warm up" for Google Play Integrity API, as we often call it in internal discussion and docs.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AppAttestationClient.kt
```diff
@@ -0,0 +1,234 @@
...
+     * requested for the device id.  Then, a Google Play Integrity API Token is
+     * fetched using the "Challenge" as the Request Hash. The resulting token is
+     * encoded into a value usable as the "attestation" parameter in the
+     * Salesforce OAuth authorization request.
+     * @param integrityTokenProvider The Google Play App Integrity API Integrity
+     * Token Provider.  This parameter is intended for testing purposes only
+     * @return The "attestation" value usable in Salesforce OAuth authorization
+     * and token refresh requests or null if the value cannot be created
+     */
+    internal suspend fun createSalesforceOAuthAuthorizationAppAttestation(
```
JohnsonEricAtSalesforce, line 138: The doc comment says it all and having detailed comments has really been adding context for our tools.  This is where our auth and token refresh flows get the new attestation value.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AppAttestationClient.kt
```diff
@@ -0,0 +1,234 @@
...
+        integrityTokenProvider: StandardIntegrityTokenProvider? = this.integrityTokenProvider,
+    ): String? {
+        // Guard to ensure the Google Play Integrity API Integrity Provider was asynchronously resolved or do so synchronously now.
+        val integrityTokenProviderResolved = integrityTokenProvider ?: prepareIntegrityTokenProvider().result
```
JohnsonEricAtSalesforce, line 142: This guard helps resolve the case where LoginActivity generates the authorization URL faster than the integrity token provider can be warmed up.  Our login flow gets to this point fast compared to what I believe Google Integrity API expected according to their documentation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AppAttestationClient.kt
```diff
@@ -0,0 +1,234 @@
...
+            OAuthAuthorizationAttestation(
+                attestationId = deviceId,
+                attestationData = Base64.getEncoder().encodeToString(
+                    integrityTokenResponse.getResult().token().encodeToByteArray()
+                )
+            ).toBase64String()
+        }.getOrElse { e ->
+            // If the Google Play Integrity API failed due to the Integrity Token Provider being expired, re-prepare it once for an inline retry.
+            if ((e as? IntegrityServiceException)?.errorCode == INTEGRITY_TOKEN_PROVIDER_INVALID) {
+                createSalesforceOAuthAuthorizationAppAttestation(
```
JohnsonEricAtSalesforce, line 177: This is a second way we guard by checking for Google Play Integrity API's documented "expired" code and issuing an inline retry.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AppAttestationClient.kt
```diff
@@ -0,0 +1,234 @@
...
+     */
+    @JvmName("createSalesforceOAuthAuthorizationAppAttestationBlocking")
+    fun createSalesforceOAuthAuthorizationAppAttestationBlocking() = runBlocking { createSalesforceOAuthAuthorizationAppAttestation() }
+
+    /**
+     * Fetches a new "Challenge" from the Salesforce App Attestation External
+     * Client App (ECA) Plug-In.
+     * @return The Salesforce App Attestation ECA Plug-In's "Challenge"
+     */
+    internal fun fetchSalesforceMobileAppAttestationChallenge(): String {
```
JohnsonEricAtSalesforce, line 203: This fetches the new "Challenge" value from the org at the very beginning of the flow in our internal docs.  If needed, I can send those internally.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -162,7 +170,9 @@ internal class NativeLoginManager(
             CONTENT_TYPE_HEADER_NAME to CONTENT_TYPE_VALUE_HTTP_POST,
             AUTHORIZATION to "$AUTH_AUTHORIZATION_VALUE_BASIC $encodedCreds",
         )
+        val attestationValue = getInstance().appAttestationClient?.createSalesforceOAuthAuthorizationAppAttestation()
         val authRequestBody = createRequestBody(
+            ATTESTATION to attestationValue,
```
JohnsonEricAtSalesforce, line 175: Here's the attestation parameter in use in Native Login.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -119,14 +123,18 @@ import kotlin.coroutines.suspendCoroutine
...
  */
 internal class NativeLoginManager(
     private val clientId: String,
     private val redirectUri: String,
     private val loginUrl: String,
     private val reCaptchaSiteKeyId: String? = null,
     private val googleCloudProjectId: String? = null,
-    private val isReCaptchaEnterprise: Boolean = false
+    private val isReCaptchaEnterprise: Boolean = false,
+    private val restClient: RestClient = getInstance().clientManager.peekUnauthenticatedRestClient()
```
JohnsonEricAtSalesforce, line 137: I increased test coverage quite a bit here with some injectables, but I didn't go any deeper than the new attestation code.  I left a to-do in the test suite noting that.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -272,8 +281,9 @@ internal class NativeLoginManager(
         }
     }
 
-    private fun createRequestBody(vararg kvPairs: Pair<String, String>): RequestBody {
-        val requestBodyString = kvPairs.joinToString("&") { (key, value) -> "$key=$value" }
+    @VisibleForTesting
+    internal fun createRequestBody(vararg kvPairs: Pair<String, String?>): RequestBody {
+        val requestBodyString = kvPairs.filter { it.second != null }.joinToString("&") { (key, value) -> "$key=${Uri.encode(value)}" }
```
JohnsonEricAtSalesforce, line 286: The new attestation value is the first optional parameter, so I allowed optional values here.  Also, our tools suggested a few other fixes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -306,13 +312,51 @@ public static URI getAuthorizationUrl(
...
+            boolean useHybridAuthentication,
+            URI loginServer,
+            String clientId,
+            String callbackUrl,
+            String[] scopes,
+            String loginHint,
+            String displayType,
+            String codeChallenge,
+            Map<String, String> addlParams,
+            SalesforceSDKManager salesforceSdkManager) {
```
JohnsonEricAtSalesforce, line 346: I also increased coverage a lot here by adding an injectable `SalesforceSDKManager`.  Note, the public API remains the same.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -306,13 +312,51 @@ public static URI getAuthorizationUrl(
...
         final StringBuilder sb = new StringBuilder(loginServer.toString());
+
+        final AppAttestationClient appAttestationClient = salesforceSdkManager.getAppAttestationClient();
+        final String authorizationAppAttestationValue = appAttestationClient != null ? appAttestationClient.createSalesforceOAuthAuthorizationAppAttestationBlocking() : null;
```
JohnsonEricAtSalesforce, line 350: I have been doing a lot of research around the impact of introducing the low-latency token call and the possibly higher latency (though difficult to prefetch) challenge call.  Our tools tell me that internally the two call sites here (the other is for token refresh) are always background thread.  That said, our tools warn that there's nothing stopping a library consumer from hitting this method on main and triggering an ANR or worse.  Thoughts? We're "safe by convention" for now, it seems.  Do we invest more time in this now or defer a task for the future to improve our public API here?

I'm also still thinking over this one, but now's a great time to open the conversation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -536,13 +591,24 @@ public static final Request.Builder addAuthorizationHeader(Request.Builder build
...
     }
 
-    private static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAccessor,
-                                                                  URI loginServer,
-                                                                  FormBody.Builder formBodyBuilder)
+    @VisibleForTesting
+    public static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAccessor,
+                                                                 URI loginServer,
+                                                                 FormBody.Builder formBodyBuilder,
+                                                                 SalesforceSDKManager salesforceSdkManager)
```
JohnsonEricAtSalesforce, line 598: Here's another injectable SalesforceSDKManager so we can increase coverage.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/AppAttestationChallengeApiClient.kt
```diff
@@ -0,0 +1,80 @@
...
+ * Challenge API endpoint.
+ *
+ * See https://docs.google.com/document/d/1MGw0-dO4Q-CJLNuqYBSYKAbEUy484lpLLX20ZIvwU6Y/edit?tab=t.0
+ * TODO: Replace with final documentation when available. ECJ20260311
+ *
+ * @param apiHostName The Salesforce App Attestation Challenge API host
+ * @param restClient The REST client to use
+ */
+@Suppress("unused")
+internal class AppAttestationChallengeApiClient(
```
JohnsonEricAtSalesforce, line 43: This mini-client for the new Challenge API is really similar to the SfapApiClient we built recently.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -445,11 +445,12 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         loading.value = true
 
         with(sdkManager) {
+            val debugOverrideAppConfig = debugOverrideAppConfig
             oAuthConfig = when {
                 // Used by UserAccountManager.migrateRefreshToken/TokenMigrationActivity
                 migrationOAuthConfig != null -> migrationOAuthConfig
                 // Used by LoginOptions
-                isDebugBuild && debugOverrideAppConfig != null -> debugOverrideAppConfig!!
```
JohnsonEricAtSalesforce, line 452: We've had a few variations of this race condition force-unwrap crash.  I recall we fixed it once, then it has evolved a bit and come back.  How's this iteration look?  The key is switching to a single local that is immutable compared to the two references to shared state.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/RestClientTest.java
```diff
@@ -732,36 +729,38 @@ public void testQueryAll() throws Exception {
      * Create new account then look for it using soql.
      * @throws Exception
      */
-    @Test(timeout = 180000) // 3 minutes - test creates 201 accounts which takes time, especially in Firebase Test Lab
+    @Test(timeout = 300000) // 5 minutes - test creates 201 accounts which takes time, especially in Firebase Test Lab
     public void testQueryWithBatchSize() throws Exception {
```
JohnsonEricAtSalesforce, line 733: This test has been unreliable for quite a while (I think?).  So, I asked our tools to suggest an improvement.  It was a really insightful and confidence-inspiring discussion that made sense to me.  @wmathurin or @brandonpage , what do you feel? Is this a meaningful update to make this test reliable and still fulfill its contract?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
JohnsonEricAtSalesforce: I slept on this one and am feeling there's some work needed to polish this.  I also see at least one possible issue recovering from expired/initializing token providers at this point from testing in my app.

Our tools recommend extracting generation of the "attestation" from this method for a lot of reasons I agree with.  It _was_ a clean look for the prototype that didn't require effort by the app to inline the "attestation" generation here, but it doesn't scale and is fragile.

I'm looking to make "attestation" a new parameter in an optional overload of `getAuthorizationUrl`; This is based on direct feedback from our tools.  Also, I'll likely separate `createSalesforceOAuthAuthorizationAppAttestation` into two separate methods so that calling logic can appropriately manage concurrency for the separate tasks of challenge and token fetch.

What timeline can we set for moving `OAuth2.java` into Kotlin and modern patterns?  It'd be a great maintainability lift, without a doubt.

JohnsonEricAtSalesforce: My effort to extract the generation of the attestation parameter value out of this method is in progress.  I have a proof-of-concept deployed to my Play developer account that works.  Now, I'll just get that polished for review.  Stay tuned.

JohnsonEricAtSalesforce: This is done and seems to look a lot better for the caller.  Now, `attestation` is one of the `addlParams` values (Which is what the tools originally suggested; I'd misquoted them earlier).  This means no additional concurrency concerns are being introduced to `getAuthorizationUrl` in this pull request ✅. You'll notice two already off-main callers now take responsibility for populating this parameter: `LoginViewModel.getAuthorizationUrl` and `IDPAuthCodeHelper.getAuthorizationPathForSP` populate this correctly.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -532,17 +562,28 @@ public static final IdServiceResponse callIdentityService(HttpAccess httpAccesso
...
             throws OAuthFailedException, IOException {
+
         final StringBuilder sb = new StringBuilder(loginServer.toString());
         sb.append(OAUTH_TOKEN_PATH);
-        sb.append(QUESTION).append(DEVICE_ID).append(EQUAL).append(SalesforceSDKManager.getInstance().getDeviceId());
+        sb.append(QUESTION).append(DEVICE_ID).append(EQUAL).append(salesforceSdkManager.getDeviceId());
+
+        final AppAttestationClient appAttestationClient = salesforceSdkManager.getAppAttestationClient();
+        final String challenge = appAttestationClient != null ? appAttestationClient.fetchMobileAppAttestationChallenge() : null;
+        final String attestationValue = challenge != null ? appAttestationClient.createAppAttestationBlocking(challenge) : null;
```
JohnsonEricAtSalesforce, line 584: This call site for adding app attestation is already off main, so it was much less of a consideration to inline the challenge and attestation work.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
```diff
@@ -103,10 +104,18 @@ internal class IDPAuthCodeHelper private constructor(
      * Compute relative path of authorization url for SP
      * @return authorization relative path
      */
-    fun getAuthorizationPathForSP(): String? {
+    private suspend fun getAuthorizationPathForSP(): String? {
```
JohnsonEricAtSalesforce: This method is already internal and called by a suspending method, so this formalizes that for the use of the App Attestation Client methods here.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -171,7 +181,7 @@ internal class NativeLoginManager(
         val authRequest = RestRequest(
             POST,
             LOGIN,
-            "$loginUrl$OAUTH_AUTH_PATH", // Full path for unauthenticated request
+            "$loginUrl$OAUTH_AUTH_PATH${attestationValue?.let { "?$ATTESTATION=$it" } ?: ""}", // Full path for unauthenticated request
```
JohnsonEricAtSalesforce: The previous commit to this was adding the attestation value to the form body.  This encodes it directly to the URL string.  Would we want to use a URL builder here or are we comfortable with the string handling?

JohnsonEricAtSalesforce:
A quick manual testing update:
✅ Headful login via the web view
✅ Token refresh
⏳ Headless login via native login

.github/workflows/reusable-lib-workflow.yaml
brandonpage: Nice find!

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/ScreenLockActivityScenarioTest.kt
```diff
@@ -83,12 +87,96 @@ import io.mockk.verify
...
+        IdlingPolicies.setMasterPolicyTimeout(
+            previousTimeout.idleTimeout,
+            previousTimeout.idleTimeoutUnit
+        )
+    }
+}
+
+@Ignore
 @RunWith(AndroidJUnit4::class)
 class ScreenLockActivityScenarioTest {
```
JohnsonEricAtSalesforce, line 90: @brandonpage, this test isn't related to App Attestation, however it has been unreliable and its failures are delaying my workflow.  Our tools have been delivering outstanding results at test updates and the conversation regarding this test was very confidence inspiring.  So, here's a version that will hopefully speed us up with a more predictable test.

I'm trying the same with ScreenLockViewTest as well.  That should be coming soon.  Sorry for the additional changes; It will be great to have a more reliable test portfolio 🤞🏻

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
```diff
@@ -103,10 +107,19 @@ internal class IDPAuthCodeHelper private constructor(
      * Compute relative path of authorization url for SP
      * @return authorization relative path
      */
-    fun getAuthorizationPathForSP(): String? {
+    @VisibleForTesting
+    internal suspend fun getAuthorizationPathForSP(): String? {
```
JohnsonEricAtSalesforce, line 111: These light updates enable testing.

JohnsonEricAtSalesforce:
A few days ago, I did take an earnest try at troubleshooting this and several other intermittent crashes and flake tests that are consistently preventing CI runs and CodeCov generation on this pull request _and_ https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2871 .  The trouble was, the more I tried updates and re-ran that jobs the more and more failures and crashes appeared.  I have an extensive inventory of unreliable tests inventoried on a local branch on my workstation.  There's a lot!

I love the notion of not merging a pull request until all the tests pass and CodeCov reports high.  However, since this seems like a systemic issue should we not block these two pull requests and instead file a follow up to review our CI and Test Suite?  I _am_ able to run all the tests locally and they pass.

Screenshot 2026-04-27 at 11 51 49 screenLockActivity_onAuthError_sendsAccessibilityEvent 1

Screenshot 2026-04-27 at 11 52 02 screenLockActivity_onAuthError_sendsAccessibilityEvent 2

JohnsonEricAtSalesforce:
The is rebased on current `dev` which was a simple update with minimal conflicts.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -446,14 +447,22 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
      * @param server The login server URL
      * @param migrationOAuthConfig The OAuth config to use for migration
      */
-    internal fun generateMigrationAuthorizationPath(
+    internal suspend fun generateMigrationAuthorizationPath (
```
JohnsonEricAtSalesforce, line 450: @brandonpage, I was reviewing the rebase today in detail and noticed this would want to be suspending to keep accommodating the app attestation logic.  It only has one call site that is already suspending, so that seems acceptable.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
```diff
@@ -103,10 +107,19 @@ internal class IDPAuthCodeHelper private constructor(
...
         SalesforceSDKLogger.d(TAG, "Getting authorization url")
         val context = SalesforceSDKManager.getInstance().appContext
         val useHybridAuthentication = SalesforceSDKManager.getInstance().useHybridAuthentication
+
+        // Add Salesforce Mobile App Attestation parameter to authorization URL if applicable.
+        val additionalParams = appAttestationClient?.run {
```
wmathurin, line 117: For client side IDP-SP flow, how should app attestation work? Right now it looks like an attested IDP app could get a refresh token for a non-attested SP app. That could be problematic.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
```diff
@@ -162,16 +169,21 @@ internal class NativeLoginManager(
...
+            val challenge = fetchMobileAppAttestationChallenge()
+            createAppAttestation(challenge) ?: return@run null
+        }
         val authRequestBody = createRequestBody(
             RESPONSE_TYPE to CODE_CREDENTIALS,
             CLIENT_ID to clientId,
             REDIRECT_URI to redirectUri,
             CODE_CHALLENGE to codeChallenge,
         )
+        val queryString = if (attestationValue != null) "?$ATTESTATION=$attestationValue" else ""
```
wmathurin: Why not have it in the request body with the other params?

JohnsonEricAtSalesforce: The server didn't see the parameter in the request body when I tested a while back.  It seems to be specifically coded to receive the parameter in the query parameters.  Should we request a change to look in both locations?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
JohnsonEricAtSalesforce: The part of the "attestation" that is specific to an org is the "challenge" that's generated before requesting the "token" from Google Play Integrity API.  That said, the "attestation" parameter is ignored if it is not needed by the org.  Would that make this a non-issue? What analysis do we have of the cross-matrix between IDP and App Attestation?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
wmathurin, line 260: How is it supposed to work? Is the app supposed to register the googleCloudProjectId before first login? Why is it apiHostName specific?
If it needs to be dynamic and decided at runtime, should we use the approach we did for dynamic consumer key where the app can register a block/lambda to provide a googleCloudProjectId during a login flow once the login server is known?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
wmathurin: It feels weird to have it in the query string. IMHO we should request a change to look for it in the body.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
wmathurin: I believe during a IDP / SP flow, the IDP gets the code and then the SP does the code to token exchange. As long as both use their own attestation values, we should be okay (both app would be attested).

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
brandonpage: Not for this PR, but perhaps it makes sense to move the authorization path generation into LoginViewModel?  (like Refresh Token Migration)

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -488,24 +498,38 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
...
             }
 
-            val jwtFlow = !jwt.isNullOrBlank() && !authCodeForJwtFlow.isNullOrBlank()
+            val currentJwt = jwt
+            val currentAuthCode = authCodeForJwtFlow
+            val jwtFlow = if (currentJwt != null && currentAuthCode != null) {
+                currentJwt.isNotBlank() && currentAuthCode.isNotBlank()
+            } else {
+                false
+            }
```
brandonpage: @JohnsonEricAtSalesforce  I don't understand.  Does this not do the exact same thing in 7 lines of code that was originally 1 line?

JohnsonEricAtSalesforce: You're exactly correct - with the exception that this generates to simpler byte code with fewer logic branches.  This came as a suggestion from our tools and gives us a higher coverage score, at the least.   I'm on the fence much of the time when it comes to cosmetic Kotlin changes that have logic/coverage impacts that improve our coverage.  For this one, I was game to get the coverage win.  If anyone's particularly adverse to it, I'm flexible.

There were two cases like this.  The other, however, remained a one-liner.

JohnsonEricAtSalesforce:
Before making any updates which might intentionally reduce code coverage, here's a screenshot of the "peak" coverage possible on this pull request, some of which was possible largely due to our new tools.  Also, future coverage reports may not generate properly once known unstable tests from upstream are restored to this change set prior to merge.

Screenshot 2026-05-04 at 18 45 18

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
JohnsonEricAtSalesforce: Any call to `/authorize` or `/token` will know include attestation so long as the host app has configured the Google Cloud Project Id and Challenge API Host.  With that in mind, this should be solid.  Do we have a test environment ready to try this out?  If so, @wmathurin, send that to my direct message.

JohnsonEricAtSalesforce: I'll also update the internal test plan documentation with a task to cover this flow.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
JohnsonEricAtSalesforce: Agreed.  I have this updated in https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2868/changes/cc5f7dbc35ea3307cfc3b573ed4e4f1a93d4f625

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Based on our conversation last Friday, I created follow-up [W-22355537](https://gus.lightning.force.com/a07EE00002ZLTyHYAX) to cover this.  That will separate the Google Cloud Project Id into a new and optional Salesforce SDK Manager initialization parameter.  When set, that will allow the manager to immediately begin preparing the Google Play Integrity API Token Provider warm up.  The Salesforce Challenge API Hostname will be resolved by a new callback function which the app registers as a separate Salesforce SDK Manager property.  That callback will be called with the resolved My Domain host after the well-known authentication configuration is fetched.  Only when both are present will the attestation parameter be generated.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
JohnsonEricAtSalesforce: @brandonpage, just checking if you do or don't want to roll this back.  I thought it over a little more and still like the cosmetic code change in exchange for fewer logic branches in the generated byte code and higher coverage.  It's just a preference though, so feel free to make the call.

brandonpage: @JohnsonEricAtSalesforce  I am very much in favor of keeping the code as easy as possible for humans to read and digest.  I think the 1 line removed is easier to quickly understand than the 7 that replaced it.  And when have we ever cared about byte code?

As for code coverage, I am very happy we are pushing to increase it, but this feels artificial.  Adding more code just to have more coverage doesn't accomplish much IMO.

JohnsonEricAtSalesforce: That works.  I'll just roll it back.  It's just a neat experiment in balancing our myriad of expectations 😊

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -488,24 +498,38 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
...
             val additionalParams = when {
-                jwtFlow -> null
+                jwtFlow -> mutableMapOf()
```
brandonpage, line 513: This probably doesn't matter, but I think we need to remove `jwtFlow` in 14.0 since `getFrontdoorUrl` is being removed.  I believe it is the (very) old "Magic Links" flow that was simply carried over when we re-wrote the login flows in 13.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/OAuth2MockTests.kt
```diff
@@ -0,0 +1,218 @@
...
+import okhttp3.ResponseBody.Companion.toResponseBody
+import okio.Buffer
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.net.URI
+
+@RunWith(AndroidJUnit4::class)
+class OAuth2MockTests {
```
brandonpage, line 26: 🎉

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
JohnsonEricAtSalesforce: That commit is pushed 👍🏻

PR #2870, opened 2026-04-23
Add Login for Admin feature

brandonpage:
The Android implementation of this feature is quite a bit more involved and far reaching than [its iOS counterpart](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4022).  I choose to do a mini re-write of how the `LoginViewModel` handles generating urls because:
- `loginUrl`, `browserCustomTabUrl`, the authorization url used for token migration and the authorization url used for front door bridge were all created from a single function that retuned a single url.  With the Web Server flow, running that function a second time generates a new code verifier and will break login using any previously generated url.  I foresee future scenarios where this becomes very problematic if we are starting and cancelling/switching different auth flows.
- With `BrowserCustomTabUrlSource` it was possible to generate an authorization url for `singleServerCustomTabActivity` that did not use the Web Server Flow, which should never be allowed.  
- Adding _another_ boolean flag (like I did on iOS) doubles the matrix of test scenarios and makes the code harder to follow.  This PR aims to do the opposite, removing several instances of `if (!sdkManager.isBrowserLoginEnabled && !singleServerCustomTabActivity && !viewModel.isUsingFrontDoorBridge ... `.

To address the above issues, this PR:
- Changes `getAuthorizationUrl` to `generateAuthorizationUrl` (maybe it should be plural?), which sets both `loginUrl` and `browserCustomTabUrl` and returns nothing. 
- Add a `generateMigrationAuthorizationPath` function to handle Token Migration Authorization Url generation separately.
- Add a separate `frontDoorBridgeUrl` variable so `loginUrl` never needs to be re-generated to avoid using Web Server.  
- Add a new (very simple) `onBrowserCustomTabReady` launching mechanism.

Demo of cancelling 'Login for Admin' vs org triggered Advanced Auth:
<video src="https://github.com/user-attachments/assets/15cf476e-3798-4e07-be3b-cceaf52e7d49" controls="controls" style="max-width: 730px;">
</video>

### DIFF
--- a/libs/SalesforceSDK/res/values/sf__strings.xml
+++ b/libs/SalesforceSDK/res/values/sf__strings.xml
@@ -27,6 +27,7 @@
     <string name="sf__clear_cache">Clear cache</string>
     <string name="sf__reload">Reload</string>
     <string name="sf__launch_idp">Log In with IDP App</string>
+    <string name="sf__login_for_admins">Login for Admins</string>
     <string name="sf__login_with_biometric">Log In with Biometric</string>
     <string name="sf__setup_biometric_unlock">Setup Biometric Unlock</string>
     <string name="sf__back_button_content_description">Back</string>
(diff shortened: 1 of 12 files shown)


JohnsonEricAtSalesforce (approved):
👍🏻

PR #2872, opened 2026-04-23
Cherry Pick Login for Admin feature 

brandonpage:


libs/SalesforceSDK/res/values/sf__strings.xml
```diff
@@ -27,6 +27,7 @@
     <string name="sf__clear_cache">Clear cache</string>
     <string name="sf__reload">Reload</string>
     <string name="sf__launch_idp">Log In with IDP App</string>
+    <string name="sf__login_for_admins">Login for Admins</string>
```
bbirman: this one has "Admins" and [iOS has "Admin"](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/ecd78976c3bc4a07836a152e94400de152c903c5/shared/resources/SalesforceSDKResources.bundle/en.lproj/Localizable.strings#L71), wondering if it would be cleaner to have the same one for both or if they're intentionally different

bbirman (approved):

JohnsonEricAtSalesforce (approved):
👍🏻

libs/SalesforceSDK/res/values/sf__strings.xml
brandonpage: Good catch, that was not intentional.  Do you think it is worth sending another PR to dev?  If I only change it here (it should be removed before 14 is released) that will cause a merge conflict, right?

bbirman: another dev PR sounds good!

brandonpage: Cool, this will be here for Monday: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2873

PR #2876, opened 2026-05-01
Fix auth flow regressions.

brandonpage:
- Ensure `LoginViewModel`'s `oAuthConfig` is set in to the migration config when generating the authorization path so code exchange succeeds
   - New unit test added to catch this.
- Check `isUsingFrontDoorBridge` to ensure CustomTab does not cover QR Code login.
- Remove unnecessary (second) call to `resetFrontDoorBridgeUrl`.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -587,7 +587,6 @@ open class LoginActivity : FragmentActivity() {
         )
 
         viewModel.clearCookies()
-        viewModel.resetFrontDoorBridgeUrl()
         // Displays the error in a toast, clears cookies and reloads the login page
         runOnUiThread {
             makeText(this, "$error : $errorDesc", LENGTH_LONG).show()
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -442,15 +442,19 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
     }
 
     /**
-     * Generates an OAuth authorization URL for token migration.
-     * @param server The login server URL
-     * @param migrationOAuthConfig The OAuth config to use for migration
+     * Generates an OAuth authorization URL for token migration given a login
+     * [server] and [migrationOAuthConfig].
      */
     internal fun generateMigrationAuthorizationPath(
         server: String,
         migrationOAuthConfig: OAuthConfig,
         sdkManager: SalesforceSDKManager = SalesforceSDKManager.getInstance(),
     ): String {
+        // Set oAuthConfig so downstream migration steps (TokenMigrationClientManager's
+        // callback URL match and doCodeExchange's consumerKey/redirectUri) use the
+        // migration config rather than the default boot config.
+        oAuthConfig = migrationOAuthConfig
+
         val codeVerifier = getRandom128ByteKey().also { codeVerifier = it }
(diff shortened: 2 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -569,7 +573,8 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
         loginUrl.value = webViewUrl
 
         // Launch the browser custom tab when applicable.
-        if (sdkManager.isBrowserLoginEnabled || singleServerCustomTabActivity) {
+        if ((sdkManager.isBrowserLoginEnabled || singleServerCustomTabActivity)
+            && !isUsingFrontDoorBridge) {
```
wmathurin, line 577: Could we add some comments for our future self?

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -587,7 +587,6 @@ open class LoginActivity : FragmentActivity() {
         )
 
         viewModel.clearCookies()
-        viewModel.resetFrontDoorBridgeUrl()
```
JohnsonEricAtSalesforce, line 590: @brandonpage , this reset was part of the QR Code Log In implementation.  If I recall, it covered the scenario where a QR Code failed login and the user was returned to the login web view to with instructions to log in view the web view.  If the front door URL is still present though, it acts as an override in that case.

I took a quick look at my fork's recently updated dev and it seems like that logic still needs this reset.  Has anything else changed where we don't need this?

JohnsonEricAtSalesforce (changes requested):
There's one possible change we might need.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: @JohnsonEricAtSalesforce I called this out in the PR description.  It is deleted here because it is also called at line 566.  It doesn't need to be called twice, right?

JohnsonEricAtSalesforce: Gotcha.  Thanks for the double clarification.  That should be a good cleanup.

JohnsonEricAtSalesforce (approved):
👍🏻

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
brandonpage: @wmathurin Hey, sorry I wrote a comment but forgot to push it.  I was going to add:

```
Launch the browser custom tab when applicable.

Whether the User Agent or Web Server Flow is used for Front Door Bridge is determined
by the server, so because Adv Auth requires PKCE we cannot guarantee it is safe/allowed.
```

@JohnsonEricAtSalesforce Is the above statement accurate?  And if so should we just be checking if `frontdoorBridgeCodeVerifier` is null?  

Should we create a 14.0 WI to address this as we move to Adv Auth first?  It could fall under that epic.

JohnsonEricAtSalesforce: Good question, @brandonpage.  The `isUsingFrontDoorBridge` property is a computed check on `frontDoorBridgeUrl.value != null`, which is the definitive check for front door bridge URL status. The URL is set in both user agent flow and web server flow front door bridge URL cases, where the code verifier is only for web server flow.

It looks like you have a winner as I see it in the comment preview above.

PR #2880, opened 2026-05-07
@W-22355537: [MSDK Android] App Attestation Public API

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This update to App Attestation separates the two key data points required to enable or disable the feature from one combined API to a simpler public API for one method and an internal API for the other.

  The first is the Google Cloud Project Id, which is now an optional parameter for the app to set once on initialization.

  The second is the Salesforce Challenge API Host, which is managed by the well-known authentication configuration fetch internally by MSDK.

  All the tests have been updated to reflect this update and I'll deploy my test app to a device for another round of regression testing.


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -176,13 +177,17 @@ import com.salesforce.androidsdk.security.interfaces.ScreenLockManager as Screen
...
+ * Google Play Integrity API and Salesforce App Attestation or null to
+ * disable both features
  */
 open class SalesforceSDKManager protected constructor(
     @JvmField
     protected val context: Context,
     mainActivity: Class<out Activity>,
     private val loginActivity: Class<out Activity>? = null,
     internal val nativeLoginActivity: Class<out Activity>? = null,
+    googleCloudProjectId: Long? = null,
```
JohnsonEricAtSalesforce, line 191: @wmathurin, here's the relocated `googleCloudProjectId` in the `SalesforceSDKManager` constructor that subclasses use.  I noticed that our `RestExplorerApp` sample subclasses, so this gives that app a convenient way to pass this value.

Are there other constructor use cases we'd need to expose?  I've yet to do a complete inventory of all our samples and templates to see what they are up to and wanted to get your feedback early.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -229,50 +234,38 @@ open class SalesforceSDKManager protected constructor(
...
      * TODO: Make this Kotlin-internal once it is no longer referenced by Java. ECJ20260420
      */
-    @Volatile
-    var appAttestationClient: AppAttestationClient? = null
-        @VisibleForTesting
-        internal set
-
-    /** Lock object for synchronized access to the app Attestation Client */
-    private val appAttestationClientLock = Any()
+    val appAttestationClient: AppAttestationClient? by lazy {
```
JohnsonEricAtSalesforce, line 246: @wmathurin - The `AppAttestationClient` initialization is once at construction now.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1958,6 +1956,9 @@ open class SalesforceSDKManager protected constructor(
                     shareBrowserSessionEnabled = false
                 )
 
+                // Disable Salesforce App Attestation for login servers that are not My Domain servers.
+                appAttestationClient?.apiHostName = null
```
JohnsonEricAtSalesforce, line 1989: @wmathurin - I went without the app-provided callback to determine the Challenge API Host.  Here, we seem to know that we're not using a My Domain and that means App Attestation will be disabled.  Is that correct?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1966,6 +1967,9 @@ open class SalesforceSDKManager protected constructor(
                     browserLoginEnabled = authConfig?.isBrowserLoginEnabled ?: false,
                     shareBrowserSessionEnabled = authConfig?.isShareBrowserSessionEnabled ?: false
                 )
+
+                // Consider enabling Salesforce App Attestation for login servers that are My Domain servers.
+                appAttestationClient?.apiHostName = loginServer.toUri().host
```
JohnsonEricAtSalesforce, line 2001: Similar to the comment a few lines before, here we seem to know we are using a My Domain host and enable App Attestation.  How's that look?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/idp/IDPAuthCodeHelper.kt
```diff
@@ -115,7 +115,7 @@ internal class IDPAuthCodeHelper @VisibleForTesting internal constructor(
 
         // Add Salesforce Mobile App Attestation parameter to authorization URL if applicable.
         val additionalParams = appAttestationClient?.run {
-            val challenge = fetchMobileAppAttestationChallenge()
+            val challenge = fetchMobileAppAttestationChallenge() ?: return@run null
```
JohnsonEricAtSalesforce, line 118: It's convenient I'd previously separated the challenge and attestation fetch at the call site, so here it's just a second guard when the Challenge Host is not set.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -229,50 +234,38 @@ open class SalesforceSDKManager protected constructor(
...
-                    restClient = clientManager.peekUnauthenticatedRestClient()
-                )
-            }
-        }
+    ) = googleCloudProjectId?.let { appAttestationGoogleCloudProjectId ->
+        AppAttestationClient(
+            context = appContext,
+            deviceId = deviceId,
+            googleCloudProjectId = appAttestationGoogleCloudProjectId,
+            remoteAccessConsumerKey = getBootConfig(appContext).remoteAccessConsumerKey,
```
wmathurin: Since 13.2, we allow apps to dynamically change the consumer key at runtime. So having a AppAttestationClient only works with the bootconfig's consumer key will not work in all cases. Why does the consumer key need to be a property of the app attestation client?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
wmathurin: (I'm still hoping we can move to attesting on code exchange only - but that will only fly if user agent flow can be blocked for a given ECA which is not possible today --- stay tuned).

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: That's a good point. It was a property for dependency injection, which aids in testability.  I'll find a way to accommodate the dynamic value 👍🏻

JohnsonEricAtSalesforce: @wmathurin - Take a look at https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2880/changes/0cd9064be78fa7efaf5a7bea0cc646bdea9a5e74 where I gave our tools a chance to recommend a way to keep App Attestation Client updated with the latest remote consumer key while not coupling it with a singleton boot config.  The new `RemoteAccessConsumerKeyProvider` looks like a win for that.

I have been judiciously requesting our tools to avoid coupling new code with fixed, untestable singletons.

JohnsonEricAtSalesforce:
Screenshot 2026-05-08 at 15 00 04

I'm attempting to stabilize some of the unreliable tests inherited from upstream dev.  However, that might not be practical without a dedicated follow-up work effort.  For record-keeping when the checks can complete here's a snapshot illustrated the code coverage for the patch.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -229,50 +235,40 @@ open class SalesforceSDKManager protected constructor(
...
-                    restClient = clientManager.peekUnauthenticatedRestClient()
-                )
-            }
-        }
+    ) = googleCloudProjectId?.let { appAttestationGoogleCloudProjectId ->
+        AppAttestationClient(
+            context = appContext,
+            deviceId = deviceId,
+            googleCloudProjectId = appAttestationGoogleCloudProjectId,
+            remoteAccessConsumerKeyProvider = RemoteAccessConsumerKeyProvider {
```
wmathurin: Where/how is the remoteAccessConsumerKeyProvider set up when an [appConfigForLoginHost](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt#L291) was provided ??

JohnsonEricAtSalesforce: How does this look? https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2880/changes/4f9d02c4937e02a59b6a13d5932d902f3d068449

This models the remote consumer key look up after what's currently in login view model.

If that's a winner, should we create a reusable routine for this? Perhaps on the manager?

Note, the tests will fail on this commit.  I figured to give you a chance to review the production code first.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
wmathurin: Agreed on having a reusable routine somewhere

JohnsonEricAtSalesforce:
Excluding the tests from upstream dev that are failing and blocking the CI run, a successful run shows the new tests on this branch at 100% coverage.

Screenshot 2026-05-11 at 22 35 56

JohnsonEricAtSalesforce:
Sorry for the delay in merging since yesterday.  Google Play was not deploying my updated Integrity Test app to my Internal Track.  It just came through and passed my App Attestation regression test. ✅

PR #2884, opened 2026-05-13
Fix Refresh Token Rotation.

brandonpage:


JohnsonEricAtSalesforce (approved):
👍🏻

wmathurin (approved):

JohnsonEricAtSalesforce (approved):
👍🏻

wmathurin (approved):
LGTM.

Maybe new AuthFlowTester tests (in a separate PR if it makes the cherry-picking easier)?

brandonpage:
Unit patch coverage is actually far better than reported here, but miss calculated due to the failing unit test shard.  That test issue is specific to dev so I cannot put it in this PR, which will be cherry picked to master.

PR #2891, opened 2026-05-18
Update to TOML-based dependency management

brandonpage:
In addition to this being the cleaner, more modern management style it should also help keep dependencies in sync to avoid issues like we just had with Kotlin Serialization.


wmathurin (approved):
👍

gradle/libs.versions.toml
```diff
@@ -0,0 +1,127 @@
...
+firebase-messaging = "25.0.2"
+play-integrity = "1.6.0"
+accompanist = "0.37.3"
+
+# SmartStore
+sqlcipher-android = "4.16.0"
+
+# Hybrid / React Native
+cordova-framework = "15.0.0"
+react-android = "0.79.3"
```
wmathurin, line 38: Shouldn't that be 0.81.5 ??

wmathurin: Looking at https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/ae6623905e8cb44de4fbc8e8c96f830e117de0c5#diff-42466aabfa7f38ab451d4e8d9d75bfe29bda36b39cbab102bb9848eb266ce69aL24-L25
I see that when we bumped up the RN version last time, we updated the package.json but not the build.gradle.kts.

PR #2892, opened 2026-05-19
Fix Token Migration silent failure.

brandonpage:
Migration always uses the Web Server regardless of the user's preferred login flow because `LoginViewModel.generateMigrationAuthorizationPath` hard-codes `useWebServerAuthentication = true` by design. The OAuth callback therefore always carries an authorization `code` rather than a token fragment so we should exchange it for tokens unconditionally.


wmathurin (approved):
I think that iOS supports both but I only wrote [tests](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/RefreshTokenMigrationTests.swift#L121) that migrates from using user agent to using web server flow (not the other way around).

Given that user agent is going away and that apps are planning to move to using web server flow before using the new migration APIs, I'm fine with only doing web server flow for refresh token migration.

JohnsonEricAtSalesforce (approved):
👍🏻

PR #2894, opened 2026-05-20
Update React Native to 0.81.5 in gradle dependency catalog

wmathurin:
## Summary
- Updates React Native version from 0.79.3 to 0.81.5 in `gradle/libs.versions.toml`
- Aligns with the version already specified in `libs/SalesforceReact/package.json`

## Context
The `package.json` in `libs/SalesforceReact` already declares React Native 0.81.5, but the gradle dependency catalog was still pointing to 0.79.3. This caused a version mismatch between the npm dependencies and the gradle dependencies.

## Test plan
- [x] Verified build succeeds: `./gradlew :libs:SalesforceReact:build`
- [x] Build completed successfully in 52s with all tasks passing

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -35,7 +35,7 @@ sqlcipher-android = "4.16.0"
 
 # Hybrid / React Native
 cordova-framework = "15.0.0"
-react-android = "0.79.3"
+react-android = "0.81.5"
 
 # Analytics deps
 tape = "1.2.3"


JohnsonEricAtSalesforce (approved):

wmathurin:
RN is broken because of AGP/Gradle version incompatibilities. That will be sorted out in future PRs (by upgrading RN and also possibly lowering our AGP/Gradle version a bit).

PR #2898, opened 2026-05-20
Fix screen lock activity on auth error sends accessibility event

brandonpage:
The test ScreenLockActivityScenarioTest.screenLockActivity_onAuthError_sendsAccessibilityEvent fails consistently on CI (Firebase Test Lab) but passes locally when run in isolation.

**Root cause**

The sibling test screenLockActivity_logoutScreenLockUsers_defaultParameters (line 817 of libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/ScreenLockActivityScenarioTest.kt) calls activity.logoutScreenLockUsers() with default parameters - i.e. against the real UserAccountManager. When prior tests in the same shard leave accounts in the system AccountManager with SCREEN_LOCK=true in shared preferences, this test triggers a real userAccountManager.signoutUser() call. That path interacts with the test app's AuthenticatorService, which on the CI emulator hangs for >200s and produces an ANR (executing service com.salesforce.androidsdk.tests/com.salesforce.androidsdk.auth.AuthenticatorService, waited 200013ms).

After the ANR, the test process is sluggish (skipped frames, GC pauses, surface-sync timeouts). Subsequent ScreenLockActivity-launching tests, including screenLockActivity_onAuthError_sendsAccessibilityEvent, time out or fail. Locally, with no accumulated test accounts, signoutUser is never invoked and the test passes.

**Fix**

Add an @Before setUp() to ScreenLockActivityScenarioTest that calls UserAccountManagerTest.cleanupAccounts(...) to remove any leftover accounts before each test. This guarantees the _defaultParameters test runs against an empty account list, so AuthenticatorService is never invoked.


JohnsonEricAtSalesforce (approved):
⭐️

wmathurin (approved):

brandonpage:
This fix is not working.  I will open a new PR.

PR #2908, opened 2026-05-28
Fix crash: use postValue in resetFrontDoorBridgeUrl

wmathurin:
## Summary

`resetFrontDoorBridgeUrl()` calls `LiveData.setValue()` but can be invoked from a background coroutine dispatcher (via `onAuthFlowError` → `doCodeExchange` error path). This throws `IllegalStateException: Cannot invoke setValue on a background thread`.

Fix: use `postValue(null)` instead of `value = null`.

## Context

The bug was introduced in "Modern Login Architecture (#2653)" on 2025-01-24. It only manifests when the OAuth code exchange fails (e.g., invalid consumer key, network error during exchange) — the happy path never hits this code.

## Test Plan

- [x] Reproduced crash with invalid consumer key on React Native template app
- [x] Verified fix resolves crash — auth error is handled gracefully


JohnsonEricAtSalesforce (approved):
👍🏻

PR #2910, opened 2026-05-29
Update nightly schedule to weekdays at 10:30 PM PT and fix security issues

brandonpage:
## Summary

Implements [W-22712082 — Cleanup CI for all Repos](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002b3viMYAQ/view).

**Schedule change**: Nightly tests now run **every weekday at 10:30 PM PT** (`30 6 * * 2-6` UTC), up from the previous Mon/Fri-only cadence. Catches regressions within 1 business day instead of up to 4.

**Security hardening**: All workflows updated to follow the [GitHub Actions injection-prevention best practices](https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/):

- **All third-party actions SHA-pinned** with the resolved tag in a comment (e.g., `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1`). Closes the supply-chain attack vector where a moving tag could be retargeted.
- **Top-level `permissions:` block** added to every workflow with the principle of least privilege: `contents: read` baseline, `pull-requests: write` only where Danger or `mikepenz/action-junit-report` posts PR comments.
- **`secrets: inherit` replaced with explicit secret pass-through** at every reusable-workflow caller; `on: workflow_call: secrets:` blocks now declare exactly which secrets each reusable expects.
- **All `${{ ... }}` shell interpolation** in `run:` blocks refactored to `env:` variables with quoted shell expansions, so attacker-controlled strings cannot break out of the script context.
- **`pull_request_target`** retained (required for fork PRs to receive secrets) with `# zizmor: ignore[dangerous-triggers]` and an inline comment documenting the Member Check mitigation.
- **`actions/checkout` steps** set `with: persist-credentials: false` to prevent post-step credential exfiltration via `.git/config`.
- **Build-step logs archived on failure** so engineers can diagnose template/sample-app build failures without re-running.
- **Job-level `permissions:` on reusable-workflow callers**: since the reusable workflows declare `permissions: { contents: read, pull-requests: write }` (needed for `mikepenz/action-junit-report` PR comments), the calling jobs in `nightly.yaml` and `pr.yaml` now declare matching permissions explicitly. Without this, GitHub rejects the workflow at startup with `startup_failure` because a caller's effective permissions cannot be less than the called workflow's declared permissions.

After this change, `zizmor --offline` reports 0 High-confidence findings across all four CI workflows.

## Test plan

- [x] Verified locally with `python3 yaml.safe_load`, `actionlint -shellcheck=`, `zizmor --offline`. All clean.
- [x] CI verification: opened test PR on personal fork (`brandonpage/SalesforceMobileSDK-Android#31`) targeting the same `cleanup-ci-w22712082` branch as this PR. The test PR triggers the new workflows against a real source change in `SalesforceAnalytics`. **No regressions observed** — see linked PR for run details.
- [ ] Reviewer to confirm the `permissions:` blocks match the team's expected privilege levels for each workflow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,10 +1,13 @@
 name: Nightly Tests
 
-on: 
+on:
   schedule:
-    - cron: "0 6 * * 2,6"  # cron is UTC, this translates to 10 PM PST Mon and Fri.
+    - cron: "30 6 * * 2-6"  # cron is UTC; 10:30 PM PT every weekday (Mon-Fri).
   # This lets us trigger the workflow from a browser.
   workflow_dispatch:
+
+permissions:
+  contents: read
 
 # Run the native libraries first so that the native and Hybrid MobileSync tests
 # don't run simultaneously (to prevent flappers).
@@ -17,7 +20,10 @@ jobs:
     uses: ./.github/workflows/reusable-lib-workflow.yaml
     with:
       lib: ${{ matrix.lib }}
-    secrets: inherit
+    secrets:
+      TEST_CREDENTIALS: ${{ secrets.TEST_CREDENTIALS }}
+      GCLOUD_SERVICE_KEY: ${{ secrets.GCLOUD_SERVICE_KEY }}
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   android-nightly-Hybrid:
     if: success() || failure()
     needs: [android-nightly]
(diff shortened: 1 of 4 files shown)


wmathurin (approved):

.github/workflows/nightly.yaml
```diff
@@ -1,24 +1,36 @@
 name: Nightly Tests
 
-on: 
+on:
   schedule:
-    - cron: "0 6 * * 2,6"  # cron is UTC, this translates to 10 PM PST Mon and Fri.
+    - cron: "30 6 * * 2-6"  # cron is UTC; 10:30 PM PT every weekday (Mon-Fri).
```
wmathurin, line 5: Should we do Sunday to Thursday instead (so always followed by a working day the next morning) ?

PR #2912, opened 2026-06-01
Handle test_login_domain with or without https:// prefix

wmathurin:
## Summary

`TestCredentials` now accepts `test_login_domain` as either a bare hostname (`login.salesforce.com`) or a full URL (`https://login.salesforce.com`). This allows the same `test_credentials.json` to work on both iOS and Android without platform-specific formatting.

Android adds `https://` if missing. iOS (companion PR) strips it if present.

## Test plan

- [x] Existing tests pass with current `test_credentials.json` format (with `https://`)
- [x] Tests also pass with bare hostname format (without `https://`)


brandonpage (approved):
Awesome.  I was wondering if we would run into something like this with the Android being added to the RN repo.

PR #2913, opened 2026-06-02
@W-22699717: [Android] Use web server flow when app attestation is configured

JohnsonEricAtSalesforce:
## Summary

- Enforce web server flow (PKCE + code exchange) when app attestation is configured (`appAttestationClient != null`), preventing the user-agent flow from being used with attestation
- Remove attestation parameter from the authorize URL in both `generateAuthorizationUrl()` and `generateMigrationAuthorizationPath()` — attestation is now only sent at the token endpoint
- Add new public `initNative(context, mainActivity, googleCloudProjectId)` overload so apps can enable attestation without subclassing `SalesforceSDKManager`

## Motivation

Supporting attestation on the authorize URL is problematic (e.g., if the instance URL was not configured/discovered, or first login for a beacon app). The server is removing attestation validation on the authorize call and replacing it with a check that denies entry if app attestation is required and the user-agent flow is being used. This client-side change enforces web server flow and removes attestation from the authorize URL to align with the new server behavior.

## Changes

| File | Change |
|------|--------|
| `LoginViewModel.kt` | Refactored `useWebServerFlow` from property to function with `sdkManager` param; added `appAttestationClient != null` condition; removed attestation code from both URL generation methods; removed `ATTESTATION` import |
| `LoginActivity.kt` | Updated call site to `useWebServerFlow()` |
| `OAuth2.java` | Updated Javadoc on `getAuthorizationUrl()` to remove attestation references from `addlParams` docs |
| `SalesforceSDKManager.kt` | Added `initNative` overload with `googleCloudProjectId`; added `@JvmOverloads` to secondary constructor for Java compatibility |
| `LoginViewModelTest.kt` | Removed 9 obsolete attestation-on-authorize-URL tests; added 7 `useWebServerFlow` condition-coverage tests; added `useWebServerAuthentication` stub to pre-existing non-relaxed mock test |
| `SalesforceSDKManagerTests.kt` | Added 2 tests for `initNative` with/without `googleCloudProjectId` |
| `TokenMigrationWebViewTest.kt` | Updated mocks for property→function change |

## Test plan

- [x] All 70 `LoginViewModelTest` tests pass
- [x] Both new `initNative` tests in `SalesforceSDKManagerTests` pass
- [x] Full branch coverage of `useWebServerFlow` (all 12 branches covered)
- [x] Full line coverage of new `initNative` overload
- [x] `TokenMigrationWebViewTest` compiles and mocks updated
- [ ] End-to-end login flow testing with attestation-enabled org (requires server-side rollout coordination)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -191,11 +191,19 @@ open class SalesforceSDKManager protected constructor(
     googleCloudProjectId: Long? = null,
 ) : DefaultLifecycleObserver {
 
+    @JvmOverloads
     constructor(
         context: Context,
         mainActivity: Class<out Activity>,
-        loginActivity: Class<out Activity>? = null
-    ) : this(context, mainActivity, loginActivity, nativeLoginActivity = null)
+        loginActivity: Class<out Activity>? = null,
+        googleCloudProjectId: Long? = null,
+    ) : this(
+        context = context,
+        mainActivity = mainActivity,
+        loginActivity = loginActivity,
+        nativeLoginActivity = null,
+        googleCloudProjectId = googleCloudProjectId
+    )
 
     /** The Android context */
     val appContext: Context = context
@@ -1879,6 +1887,28 @@ open class SalesforceSDKManager protected constructor(
             mainActivity,
             loginActivity,
             nativeLoginActivity,
+        )
+
+        /**
+         * Initializes required components with Salesforce App Attestation
+         * enabled. Native apps must call one overload of this method before
+         * using the Salesforce Mobile SDK.
+         *
+         * @param context The Android context
+         * @param mainActivity The app's main activity class
+         * @param googleCloudProjectId The Google Cloud Project ID to use with
+         * Google Play Integrity API and Salesforce App Attestation
+         */
+        @JvmStatic
+        fun initNative(
+            context: Context,
+            mainActivity: Class<out Activity>,
+            googleCloudProjectId: Long,
+        ) = init(
+            context,
+            mainActivity,
+            LoginActivity::class.java,
+            googleCloudProjectId = googleCloudProjectId,
         )
 
         /**
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -57,7 +57,6 @@ import com.salesforce.androidsdk.app.SalesforceSDKManager.Theme.DARK
 import com.salesforce.androidsdk.app.SalesforceSDKManager.Theme.LIGHT
 import com.salesforce.androidsdk.auth.HttpAccess
 import com.salesforce.androidsdk.auth.OAuth2
-import com.salesforce.androidsdk.auth.OAuth2.ATTESTATION
 import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse
 import com.salesforce.androidsdk.auth.OAuth2.exchangeCode
 import com.salesforce.androidsdk.auth.OAuth2.getFrontdoorUrl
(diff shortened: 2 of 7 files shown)


JohnsonEricAtSalesforce:
All checks passed — 826 unit/UI tests executed on Firebase Test Lab (API 35) with 0 failures, 0 errors, 0 flakes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -191,11 +191,19 @@ open class SalesforceSDKManager protected constructor(
     googleCloudProjectId: Long? = null,
 ) : DefaultLifecycleObserver {
 
+    @JvmOverloads
```
JohnsonEricAtSalesforce, line 194: Added `@JvmOverloads` so Java subclasses (e.g. `SalesforceSDKManagerTest.SalesforceSDKTestManager`) can call `super(context, mainActivity, loginActivity)` without explicitly passing the new `googleCloudProjectId` default parameter. Without this, Kotlin default params aren't visible to Java callers.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1881,6 +1889,28 @@ open class SalesforceSDKManager protected constructor(
...
+            context: Context,
+            mainActivity: Class<out Activity>,
+            googleCloudProjectId: Long,
+        ) = init(
+            context,
+            mainActivity,
+            LoginActivity::class.java,
+            googleCloudProjectId = googleCloudProjectId,
+        )
+
```
JohnsonEricAtSalesforce, line 1913: New public API: allows apps to enable Salesforce App Attestation without subclassing `SalesforceSDKManager`. Uses non-nullable `Long` (not `Long?`) to make the opt-in explicit — the null/disabled case is already served by the existing overloads. Delegates to the private `init()` which wires up the singleton and calls `initInternal`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -212,21 +211,25 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
...
+    ): Boolean = with(sdkManager) {
+        // First, an in-use Salesforce Identity API UI Bridge front-door bridge URL takes precedence.
+        if (isUsingFrontDoorBridge) {
+            // A front-door bridge URL accompanied by a PKCE code verifier requires Web Server Flow.  Otherwise, User Agent-Flow must be used.
+            frontdoorBridgeCodeVerifier != null
+        }
+        // Second, when not using a front-door bridge URL, the app's preferences can be used.
+        // App attestation always requires Web Server Flow because attestation is only supported at the token endpoint.
+        else {
+            useWebServerAuthentication || isBrowserLoginEnabled || singleServerCustomTabActivity || appAttestationClient != null
```
JohnsonEricAtSalesforce, line 230: Key behavioral change: `appAttestationClient != null` forces web server flow when attestation is configured. User-agent flow cannot be used because attestation is only supported at the token endpoint, not the authorize URL.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -212,21 +211,25 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
...
-            // First, an in-use Salesforce Identity API UI Bridge front-door bridge URL takes precedence.
-            if (isUsingFrontDoorBridge) {
-                // A front-door bridge URL accompanied by a PKCE code verifier requires Web Server Flow.  Otherwise, User Agent-Flow must be used.
-                frontdoorBridgeCodeVerifier != null
-            }
-            // Second, when not using a front-door bridge URL, the app's preferences can be used.
-            else {
-                useWebServerAuthentication || isBrowserLoginEnabled || singleServerCustomTabActivity
-            }
+    internal fun useWebServerFlow(
```
JohnsonEricAtSalesforce, line 219: Refactored from a computed property to a function with an optional `sdkManager` parameter. This enables direct unit testing with mocks — all 12 branches now have full coverage.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -544,7 +533,7 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
                 else -> webServerAuthorizationUrl.toString()
             }.toString()
 
-            val webViewUrlValue = if (useWebServerFlow) {
+            val webViewUrlValue = if (useWebServerFlow(sdkManager)) {
```
JohnsonEricAtSalesforce, line 536: Passes the local `sdkManager` (already in scope from the function parameter) so the flow decision uses the same instance as the rest of URL generation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -512,14 +509,6 @@ open class LoginViewModel(val bootConfig: BootConfig) : ViewModel() {
             val codeVerifier = getRandom128ByteKey().also { codeVerifier = it }
             val codeChallenge = getSHA256Hash(codeVerifier)
 
-            // Populate the additional parameter map with app attestation, if applicable.
-            sdkManager.appAttestationClient?.run {
-                val challenge = fetchMobileAppAttestationChallenge() ?: return@run
-                val attestation = createAppAttestation(challenge) ?: return@run
-                additionalParams[ATTESTATION] = attestation
-            }
```
JohnsonEricAtSalesforce, line 522: Attestation removed from the authorize URL. The server is removing attestation validation on the authorize call — attestation now only happens at the token endpoint (`OAuth2.makeTokenEndpointRequest`).

wmathurin (approved):
LGTM - were you able to test it end to end against a test server?

JohnsonEricAtSalesforce:
> LGTM - were you able to test it end to end against a test server?

Yes, but though I can see the flow is working as I'd expect it the server still seems to require the attestation parameter on /authorize.  I'll have to follow-up on where and when the end-to-end test can complete.

JohnsonEricAtSalesforce:
> > LGTM - were you able to test it end to end against a test server?
> 
> Yes, but though I can see the flow is working as I'd expect it the server still seems to require the attestation parameter on /authorize. I'll have to follow-up on where and when the end-to-end test can complete.

@wmathurin, I was able to publish my test app to Google Play and confirm it can pass /authorize in the "client controlled" mode without attestation and then the ECA will allow or deny the provided attestation when passed by this new logic to /token.  That's the test I was looking to update according to our formal test plan.  We should be G2G on this one now 🌈

PR #2916, opened 2026-06-04
@W-22701996: [Android] Improve error handling at token refresh

JohnsonEricAtSalesforce:
## Summary

Surfaces specific token endpoint error types (`client_blocked`, `client_blocked_retry`, `invalid_grant`) to apps via broadcast intent extras during token refresh, and handles `client_blocked_retry` without logging out the user.

**Work Item:** [W-22701996](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002b1OJoYAM/view) — [Android] Improve error handling at token refresh

### Problem

Previously, all token endpoint errors (HTTP 400/401/403) during refresh were treated identically — the SDK logged out the user and broadcast `ACCESS_TOKEN_REVOKE_INTENT` with no error details. For App Attestation, the server returns two distinct error types:
- `client_blocked` (terminal — device/app failed integrity checks)
- `client_blocked_retry` (retriable — attestation couldn't be verified but the refresh token is still valid)

Both incorrectly triggered logout. Apps had no way to distinguish error types to show appropriate UI.

### Changes

- **`OAuth2.java`**: Added `CLIENT_BLOCKED_ERROR` and `CLIENT_BLOCKED_RETRY_ERROR` constants. Added `CLIENT_BLOCKED` to `LogoutReason` enum.
- **`ClientManager.java`**: Added `EXTRA_TOKEN_ERROR` and `EXTRA_TOKEN_ERROR_DESCRIPTION` broadcast intent extra constants. Restructured `getNewAuthToken()` to catch `OAuthFailedException` and branch on error type — `client_blocked_retry` skips logout while terminal errors use the appropriate `LogoutReason`. All OAuth failures broadcast error details in intent extras. Modified `refreshStaleToken()` to rethrow `OAuthFailedException` instead of swallowing it.
- **`AuthenticatorService.java`**: Added early check for `client_blocked_retry` to return an error bundle without redirecting to login, before the `isRefreshTokenInvalid()` check that would otherwise treat it as terminal.

### Tests Added

- **`AuthenticatorServiceTest.kt`** (new file, 4 tests):
  - `client_blocked_retry` → returns error bundle (no login redirect)
  - `client_blocked` → redirects to login
  - `invalid_grant` → redirects to login
  - HTTP 500 → returns error bundle (not a refresh token error)

- **`ClientManagerMockTest.kt`** (5 new tests):
  - `client_blocked` → logout with `CLIENT_BLOCKED` reason, extras in broadcast
  - `client_blocked_retry` → NO logout, null token returned, extras in broadcast
  - `invalid_grant` → logout with `REFRESH_TOKEN_EXPIRED`, extras in broadcast
  - Unparseable error response → broadcast without extras
  - `client_blocked_retry` subsequent call → skips `invalidateToken` (null guard)
  - `revokedTokenShouldLogout=false` → no logout, still broadcasts
  - Null instance URL on successful refresh → broadcasts refresh intent

### How Tests Validate Requirements

| Requirement | Test Validation |
|---|---|
| 1. Surface specific error type | All tests verify the error type propagates through the respective path |
| 2. Include error + error_description in broadcast extras | ClientManager tests assert `EXTRA_TOKEN_ERROR` and `EXTRA_TOKEN_ERROR_DESCRIPTION` on captured broadcast intents |
| 3. `client_blocked` → logout with specific reason | `testGetNewAuthToken_ClientBlocked` verifies `LogoutReason.CLIENT_BLOCKED` passed to `logout()` |
| 4. `client_blocked_retry` → no logout, fail waiting requests | `testGetNewAuthToken_ClientBlockedRetry` verifies `logout()` never called, null token returned |

## Test plan

- [x] All existing `ClientManagerMockTest` tests pass (11 pre-existing + new)
- [x] All `AuthenticatorServiceTest` tests pass (new)
- [x] Firebase Test Lab CI passes
- [x] No new lint warnings introduced
- [x] Codecov confirms line coverage for new code paths (100% patch coverage)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -514,13 +554,9 @@ private UserAccount refreshStaleToken(Account account) throws NetworkErrorExcept
...
 
                 return updatedUserAccount;
-            } catch (OAuth2.OAuthFailedException ofe) {
-                if (ofe.isRefreshTokenInvalid()) {
-                    SalesforceSDKLogger.i(TAG, "Invalid Refresh Token: (Error: " +
-                            ofe.getTokenErrorResponse().error + ", Status Code: " +
-                            ofe.getHttpStatusCode() + ")", ofe);
-                }
-                return null;
+            } catch (OAuthFailedException ofe) {
```
JohnsonEricAtSalesforce, line 587: **Requirement 1 (enabler):** Previously this method caught `OAuthFailedException`, logged, and returned `null` — discarding all error context. Now it rethrows so `getNewAuthToken()` can inspect the specific error type and branch accordingly.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -423,41 +438,66 @@ public String getNewAuthToken() {
...
                 final UserAccount userAccount = refreshStaleToken(matchingAccount);
 
-                // NB: userAccount will be null if refresh token is no longer valid
-                newAuthToken = userAccount != null ? userAccount.getAuthToken() : null;
-                newInstanceUrl =  userAccount != null ? userAccount.getInstanceServer() : null;
+                newAuthToken = userAccount.getAuthToken();
+                newInstanceUrl = userAccount.getInstanceServer();
 
                 Intent broadcastIntent;
-                if (newAuthToken == null) {
```
JohnsonEricAtSalesforce, line 435: By the by, Github isn't displaying a usable diff here since the replaced null check eliminated one block of code that was replaced with another a few lines later.  If one really wants to see this difference do a comparison in Android Studio (or anything better than Github!)

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -423,41 +438,67 @@ public String getNewAuthToken() {
...
+                }
+                broadcastIntent.setPackage(SalesforceSDKManager.getInstance().getAppContext().getPackageName());
+                SalesforceSDKManager.getInstance().getAppContext().sendBroadcast(broadcastIntent);
+            } catch (OAuthFailedException ofe) {
+                // tokenError is never null — OAuthFailedException's constructor requires it.
+                final TokenErrorResponse tokenError = ofe.getTokenErrorResponse();
+                final String errorType = tokenError.error;
+                final String errorDesc = tokenError.errorDescription;
+
+                if (!CLIENT_BLOCKED_RETRY_ERROR.equals(errorType)) {
```
JohnsonEricAtSalesforce, line 501: **Requirements 4 and 3:** `client_blocked_retry` is retriable — the server couldn't verify attestation but the refresh token is still valid. We skip logout so the app can re-attest and retry. All other errors (including `client_blocked`) are terminal and trigger logout with the appropriate `LogoutReason`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
```diff
@@ -147,13 +149,14 @@ public Bundle getAuthToken(AccountAuthenticatorResponse response, Account accoun
 
                 return resBundle;
             } catch (OAuthFailedException ofe) {
-                if (ofe.isRefreshTokenInvalid()) {
-                    SalesforceSDKLogger.i(TAG, "Invalid Refresh Token: (Error: " +
-                            ofe.response.error + ", Status Code: " + ofe.httpStatusCode + ")", ofe);
+                SalesforceSDKLogger.i(TAG, "Token endpoint error: (Error: " + ofe.response.error + ", Status Code: " + ofe.httpStatusCode + ")", ofe);
+
+                // Terminal errors (except retriable attestation) redirect to login.
+                if (!CLIENT_BLOCKED_RETRY_ERROR.equals(ofe.response.error) && ofe.isRefreshTokenInvalid()) {
```
JohnsonEricAtSalesforce, line 155: **Requirement 4:** For the AuthenticatorService path, `client_blocked_retry` must NOT redirect to login. Since `isRefreshTokenInvalid()` only checks HTTP status codes (400/401/403) and `client_blocked_retry` returns 400, we check the error type first. The combined condition ensures retriable errors return an error bundle while terminal errors still redirect to login.

JohnsonEricAtSalesforce:
## Rename: `user_blocked` → `client_blocked`

The error type names were updated from `user_blocked`/`user_blocked_retry` to `client_blocked`/`client_blocked_retry` throughout the implementation.

**How it was discovered:** During live testing of the App Integrity Test app (modified RestExplorer) deployed via Google Play Internal Test Track, the token endpoint returned `client_blocked_retry` — not `user_blocked_retry` as originally specified in the work item. The app's toast receiver didn't trigger because it was listening for the wrong error string.

**Why it was needed:** The server-side contract uses `client_blocked` and `client_blocked_retry` as the OAuth error values. The original work item description used `user_blocked` which was a preliminary name that didn't match the final server implementation. All constants, enum values, test strings, comments, and Javadoc have been updated to reflect the actual server contract.

**Files affected:** `OAuth2.java`, `ClientManager.java`, `AuthenticatorService.java`, `ClientManagerMockTest.kt`, `AuthenticatorServiceTest.kt`

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -236,6 +242,7 @@ public enum LogoutReason {
         UNEXPECTED,              // Unexpected error or crash
         UNEXPECTED_RESPONSE,     // Unexpected response from server
         UNKNOWN,                 // Unknown
+        CLIENT_BLOCKED,          // Device/app blocked by server (e.g. failed attestation)
```
JohnsonEricAtSalesforce, line 245: **Requirement 3:** Distinct logout reason for attestation failures. Differentiates from `REFRESH_TOKEN_EXPIRED` so apps and analytics can distinguish "blocked by attestation" from "token expired".

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -60,6 +69,11 @@ public class ClientManager {
 	public static final String ACCESS_TOKEN_REVOKE_INTENT = "access_token_revoked";
     public static final String ACCESS_TOKEN_REFRESH_INTENT = "access_token_refeshed";
     public static final String INSTANCE_URL_UPDATE_INTENT = "instance_url_updated";
+    /** Intent extra: the {@code error} value from the token endpoint response (e.g. "client_blocked", "invalid_grant"). */
+    public static final String EXTRA_TOKEN_ERROR = "token_error";
```
JohnsonEricAtSalesforce, line 75: **Requirements 1 and 2:** These constants surface the token endpoint's `error` and `error_description` fields in the broadcast intent extras, allowing apps to distinguish between failure modes (e.g. show "app blocked" vs "session expired").

JohnsonEricAtSalesforce:
## Coverage Improvement Summary

### Initial Codecov Report (first CI run)
- **Patch coverage:** 84.62% — 6 lines with partial branch coverage
- ClientManager.java: 85.71% (5 partials)
- AuthenticatorService.java: 66.66% (1 partial)

### Changes Made
Commits were added to address coverage gaps:
1. **Tests for uncovered branches** — unparseable error response (null error/description), null token guard (`invalidateToken` after prior `client_blocked_retry`), and HTTP 500 path in AuthenticatorService
2. **Test for `revokedTokenShouldLogout=false` branch** — verifies terminal errors still broadcast without calling logout
3. **Test for null `instanceUrl` on successful refresh** — covers the short-circuit on the compound `&&` condition
4. **Removed dead null check on `tokenError`** — `OAuthFailedException`'s constructor guarantees `tokenError` is never null, eliminating 2 unreachable partial branches

### Final Codecov Report
- **Patch coverage: 100%** — 0 missing, 0 partials
- ClientManager.java: 100% (+5.56% file coverage)
- AuthenticatorService.java: 100% (+38.77% file coverage)
- Project coverage: 66.54% (+0.30% vs base)
- Total partials across project: 1053 (-19 vs base)

### Flaky Test Note
Two CI runs reported `unit-tests-pr` as "failed" due to pre-existing flaky tests on API 35 (passed on rerun both times). These are unrelated to the changes in this PR:
- `reloadWebView_WithWebServerFlow_DoesNotSetAboutBlank`
- `testNullsInSelectedLoginServer`

The CI workflow also reports a permissions error ("Failed to create checks using the provided token") which is expected for fork-based PRs that cannot write check annotations to the upstream repo.

JohnsonEricAtSalesforce:
## Live Verification: `client_blocked_retry` Error Handling

The error handling implementation was verified end-to-end on a physical device using the App Integrity Test app deployed via Google Play Internal Test Track.

**Test procedure:**

1. **Baseline (v47):** The unmodified app with correct attestation was installed and authenticated successfully against the MSDK App Attestation test org — confirming the attestation flow works correctly under normal conditions.

2. **Fail version (v49):** A modified version was installed that deliberately malforms the attestation by omitting the challenge from the integrity token request hash. This produces a valid-looking attestation that will fail server-side verification.

3. **Trigger:** The user's sessions were removed in Salesforce Setup via Session Management, forcing a token refresh on the next org access attempt.

4. **Result:** The token refresh triggered the server to return `client_blocked_retry`. The app's broadcast receiver successfully received the error via `ACCESS_TOKEN_REVOKE_INTENT` with `EXTRA_TOKEN_ERROR = "client_blocked_retry"` and displayed a toast confirming receipt of the retriable attestation error. The user was NOT logged out — the refresh token remained valid, matching the expected behavior for this work item.

**Conclusion:** Requirement 4 ("For `client_blocked_retry`: do NOT logout, surface the error so the app can retry") is verified working end-to-end against the live server.

Here's a screenshot from the test app which shows it receiving the new error.
Screenshot_20260604-145816

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -423,41 +438,67 @@ public String getNewAuthToken() {
...
                         }
-                        boolean showLoginPage = accounts.length == 1;
+                        final boolean showLoginPage = accounts.length == 1;
+                        final LogoutReason reason = CLIENT_BLOCKED_ERROR.equals(errorType)
+                                ? CLIENT_BLOCKED
+                                : REFRESH_TOKEN_EXPIRED;
                         // Note: As of writing (2024) this call will never succeed because revoke API is an
                         // authenticated endpoint.  However, there is no harm in attempting and the debug logs
                         // produced may help developers better understand the state of their app.
                         SalesforceSDKManager.getInstance()
```
wmathurin, line 514: Server side they also have an enum for logout reasons - we need the client side enum to line up. Although in this case, I'm assuming we won't be sending the reason up to the server.
Do you know if the server added CLIENT_BLOCKED to their LogoutReason enum?

JohnsonEricAtSalesforce: Good question. This is being researched internally with the App Attestation development team — I'll follow up here once I have confirmation on whether the server-side LogoutReason enum includes `CLIENT_BLOCKED` and whether alignment is needed. Stay tuned.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -423,41 +438,67 @@ public String getNewAuthToken() {
...
+                if (lastNewAuthToken != null) {
+                    clientManager.invalidateToken(lastNewAuthToken);
+                }
                 final UserAccount userAccount = refreshStaleToken(matchingAccount);
 
-                // NB: userAccount will be null if refresh token is no longer valid
-                newAuthToken = userAccount != null ? userAccount.getAuthToken() : null;
-                newInstanceUrl =  userAccount != null ? userAccount.getInstanceServer() : null;
+                newAuthToken = userAccount.getAuthToken();
+                newInstanceUrl = userAccount.getInstanceServer();
```
brandonpage, line 463: Why was this null check removed?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
brandonpage: Why was `if (newAuthToken == null)` removed?

JohnsonEricAtSalesforce: There's a comment about this elsewhere. The null was the driver of the logic when the exception was swallowed by the old code, as I recall from this morning. It's no longer the case now that the exception is detailed and propagated for the app or others to receive. I'm away from my desk for a moment, so let me know if that doesn't make sense and I'll provide the full write up in about an hour.

wmathurin (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
brandonpage: AFAIK `newAuthToken` (and `userAccount` above) can still be null.

JohnsonEricAtSalesforce: You're right. I initially dismissed this because `refreshStaleToken` can't return null (it throws instead), but I overlooked that `userAccount.getAuthToken()` could still return null if the token endpoint returned success without an `access_token` field. The old `if (newAuthToken == null)` branch was providing a safety net for that case, and removing it changed the behavior — a null token would incorrectly broadcast `ACCESS_TOKEN_REFRESH_INTENT` instead of revoking.

I've addressed this by adding a `MalformedTokenException` thrown from `refreshStaleToken` when `tr.authToken` is null after a successful refresh. This feeds into a combined `catch (OAuthFailedException | MalformedTokenException e)` block that handles both cases with a single logout + revoke broadcast path.

A new exception type was needed because `OAuthFailedException`'s constructor is package-private to `com.salesforce.androidsdk.auth`, so `ClientManager` (in `.rest`) can't construct one. Rather than widening the constructor's visibility (which would expose an internal API), `MalformedTokenException` is a package-private exception in `ClientManager` that represents this specific malformed-response edge case. If you'd prefer, I can widen `OAuthFailedException`'s constructor instead and eliminate the new type — let me know.

This is an improvement over the old `if (newAuthToken == null)` approach because:
- The error is surfaced at the point of detection (`refreshStaleToken`) rather than handled implicitly in the caller as one of several if/else-if conditions
- The null-token case is named and explicit — it's clearly a malformed response, not a normal control flow branch sitting alongside "instance URL changed" and "token refreshed"
- The combined catch eliminates redundant logout code between the two failure modes while keeping the `client_blocked_retry` skip-logout logic in one place

See commit: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2916/commits/d6f7689da

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
JohnsonEricAtSalesforce: The null check on `userAccount` was removed because `refreshStaleToken` now throws on all failure paths instead of returning null. That part is safe — the only way to reach this line is if the refresh succeeded and returned a valid `UserAccount`.

However, the null check on the *result* of `userAccount.getAuthToken()` (which the old ternary provided) was incorrectly removed. I'd conflated "userAccount can't be null" with "authToken can't be null" — two different things. That case is now handled explicitly by a `MalformedTokenException` thrown inside `refreshStaleToken` before the account is persisted, which enters the same catch/logout/broadcast path as `OAuthFailedException`.

Moving this check into `refreshStaleToken` is better than the original approach because it keeps the success path in `getNewAuthToken` clean — the try block only deals with the happy path (token refreshed, broadcast success), while all failure modes (OAuth errors, malformed responses, network errors) are handled by their respective catch blocks. The old code mixed a failure condition (`newAuthToken == null`) into the same if/else chain as success conditions (`instance URL changed`, `token refreshed`), making it harder to reason about which paths represent errors vs normal outcomes.

See commit: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2916/commits/d6f7689da

brandonpage: > The null check on userAccount was removed because refreshStaleToken now throws on all failure paths instead of returning null.

`refreshStaleToken` a Java method _without_ an annotation guaranteeing the the returned value is not null.  You are removing the guardrails that would prevent a crash from happening if the underlying code is changed.  

It's great that it can't be null _ right now_, but we have had issues with NPEs in this exact spot that were not caught until production users hit them because of specific timings we could not reproduce with tests.  As we are removing Java from the project, the classes that remain need to be defensive.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -423,41 +438,84 @@ public String getNewAuthToken() {
...
+                    errorType = tokenError.error;
+                    errorDesc = tokenError.errorDescription;
+                } else {
+                    errorType = null;
+                    errorDesc = null;
+                }
+
+                if (!CLIENT_BLOCKED_RETRY_ERROR.equals(errorType)) {
+                    // Terminal error (client_blocked, invalid_grant, malformed token, etc.) — logout.
+                    if (clientManager.revokedTokenShouldLogout) {
```
brandonpage, line 503: Unrelated, but I hate `clientManager.revokedTokenShouldLogout` and fundamentally believe it should not exist.  It violates our security posture and there is no iOS equivalent. 

cc @wmathurin

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
wmathurin: +1 on being defensive
We have been bitten before when changing this code (months after the change of course).

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
wmathurin: It's private though. Was it introduced to allow unit testing?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
JohnsonEricAtSalesforce: Fair point — I've added both a `@NonNull` annotation on `refreshStaleToken`'s return type and a defensive null check at the call site:

```java
@NonNull
private UserAccount refreshStaleToken(Account account) throws ...
```

```java
final UserAccount userAccount = refreshStaleToken(matchingAccount);

if (userAccount == null) {
    throw new MalformedTokenException("refreshStaleToken returned null");
}
```

The annotation makes the contract explicit for tooling and future maintainers. The null check provides the runtime safety net — if a future change or timing issue causes a null return, it routes cleanly into the `MalformedTokenException` catch (logout + revoke broadcast) instead of an NPE.

Also added a test (`testGetNewAuthToken_NullUserAccount_LogsOut`) that uses MockK to force `refreshStaleToken` to return null, verifying the defensive path triggers logout + revoke broadcast correctly.

See commit: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2916/commits/ddf7563fb

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
JohnsonEricAtSalesforce: Agreed this is out of scope for this PR, but it's worth tracking. @brandonpage @wmathurin would either of you want to create a follow-up user story to evaluate removing `revokedTokenShouldLogout`?

JohnsonEricAtSalesforce:
## Ready to Squash

This PR has been approved by @wmathurin and all reviewer feedback has been incorporated:

- **Defensive null checks** restored per @brandonpage and @wmathurin's feedback — `refreshStaleToken` return value and `tr.authToken` are both guarded with `MalformedTokenException`, annotated `@NonNull`, and tested via MockK
- **Error type renamed** from `user_blocked` to `client_blocked` after live verification against the server
- **100% patch coverage** maintained across all iterations
- **Live end-to-end verification** completed on a physical device via Google Play Internal Test Track

The PR has been approved since June 4. The `revokedTokenShouldLogout` discussion (raised as unrelated by @brandonpage) can be tracked as a follow-up work item before release.

Ready to squash and merge to dev.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
brandonpage: @wmathurin the boolean we read is private, [but it is set via public API](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/699f7119ab8a0ae837246952daeac6524c0c1ff6/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java#L92).

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
brandonpage: Did we follow up?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
wmathurin: That flag is actually super old: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/195

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
JohnsonEricAtSalesforce: There's an instant message thread internally that I believe is tracking this.  I'll post internally to verify ✅

PR #2921, opened 2026-06-09
Add diagnostic warning for Lightning URL code exchange failure

JohnsonEricAtSalesforce:
## Summary

Currently, using the Lightning domain instead of the My Domain as the login server leads to an unintuitive error response from the server. It isn't clear to the developer or user what is wrong or what action should be taken to address the issue. This was reported by an app in development and led to confusion and delays. Also, it can be confusing when migrating from user agent flow to web server flow since user agent flow would not surface this issue.

## Context

This change does not fix the underlying misconfiguration — it surfaces it clearly to developers and in logs so it can be resolved quickly.

## Test plan

- [x] Build passes (`./gradlew :libs:SalesforceSDK:build`)
- [x] Manual verification: reproduced against Lightning URL — the `unsupported_grant_type` error triggers the new toast and log
- [x] Manual verification: non-Lightning URL auth errors still show the generic toast
- [x] Note: `onAuthFlowError` has no existing instrumented test coverage; the lightning URL detection was verified through manual reproduction against a test org

## Test coverage note

Automated test coverage for this diagnostic path was evaluated and the gap accepted. The `onAuthFlowError` method is a protected method on `LoginActivity` that requires a full activity lifecycle to invoke, depends on `LiveData` state from the view model (`selectedServer.value`), and produces output (toasts and logcat) that is not easily assertable in instrumented tests. Extracting the Lightning URL detection into a standalone testable utility was considered but deemed over-engineering for a diagnostic-only change with no branching logic beyond show/don't-show a message.

### DIFF
--- a/libs/SalesforceSDK/res/values/sf__strings.xml
+++ b/libs/SalesforceSDK/res/values/sf__strings.xml
@@ -10,7 +10,7 @@
     <string name="sf__access_token_revoked">Your access token has been revoked by the administrator. You will now be logged out.</string>
     <string name="sf__managed_app_error">Authentication only allowed from managed device.</string>
     <string name="sf__jwt_authentication_error">JWT authentication error. Please try again.</string>
-    <string name="sf__lightning_url_code_exchange_error">Can\'t use Lightning URL for code exchange. Use My Domain login server URL.</string>
+    <string name="sf__lightning_url_code_exchange_error">Lightning URLs are not supported for OAuth code exchange. Use your My Domain URL instead.</string>
 
     <!-- SSL errors -->
     <string name="sf__ssl_error">SSL error: %s.</string>
--- a/libs/SalesforceSDK/res/values/sf__strings.xml
+++ b/libs/SalesforceSDK/res/values/sf__strings.xml
@@ -10,6 +10,7 @@
     <string name="sf__access_token_revoked">Your access token has been revoked by the administrator. You will now be logged out.</string>
     <string name="sf__managed_app_error">Authentication only allowed from managed device.</string>
     <string name="sf__jwt_authentication_error">JWT authentication error. Please try again.</string>
+    <string name="sf__lightning_url_code_exchange_error">Can\'t use Lightning URL for code exchange. Use My Domain login server URL.</string>
 
     <!-- SSL errors -->
     <string name="sf__ssl_error">SSL error: %s.</string>
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -587,9 +587,19 @@ open class LoginActivity : FragmentActivity() {
         )
 
         viewModel.clearCookies()
+        val isLightningTokenEndpointFailure = e is OAuthFailedException
+            && e.tokenErrorResponse.error == "unsupported_grant_type"
+            && viewModel.selectedServer.value?.contains(".lightning.") == true
+        if (isLightningTokenEndpointFailure) {
+            w(TAG, "Code exchange failed with unsupported_grant_type against Lightning URL: ${viewModel.selectedServer.value}. Lightning URLs do not support authorization_code grant type. Use a My Domain login server URL instead.")
+        }
         // Displays the error in a toast, clears cookies and reloads the login page
         runOnUiThread {
-            makeText(this, "$error : $errorDesc", LENGTH_LONG).show()
+            if (isLightningTokenEndpointFailure) {
+                makeText(this, "Can't use Lightning URL for code exchange. Use My Domain login server URL.", LENGTH_LONG).show()
+            } else {
+                makeText(this, "$error : $errorDesc", LENGTH_LONG).show()
+            }
             viewModel.reloadWebView()
         }
     }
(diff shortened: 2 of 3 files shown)


JohnsonEricAtSalesforce:
Here's a video of the new toast in action. I left the original toast on here so both could be compared. The committed version only display the new toast.
[Screen_recording_20260608_224718.webm](https://github.com/user-attachments/assets/39176bc9-3d8f-4a6a-abd1-4b5da4250ecf)

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -583,9 +587,19 @@ open class LoginActivity : FragmentActivity() {
...
+            && e.tokenErrorResponse.error == "unsupported_grant_type"
+            && viewModel.selectedServer.value?.contains(".lightning.") == true
+        if (isLightningTokenEndpointFailure) {
+            w(TAG, "Code exchange failed with unsupported_grant_type against Lightning URL: ${viewModel.selectedServer.value}. Lightning URLs do not support authorization_code grant type. Use a My Domain login server URL instead.")
+        }
         // Displays the error in a toast, clears cookies and reloads the login page
         runOnUiThread {
-            makeText(this, "$error : $errorDesc", LENGTH_LONG).show()
+            if (isLightningTokenEndpointFailure) {
+                makeText(this, "Can't use Lightning URL for code exchange. Use My Domain login server URL.", LENGTH_LONG).show()
```
brandonpage: This needs to be a localizable string.

JohnsonEricAtSalesforce: Addressed in 8908783 — moved the toast message to `sf__lightning_url_code_exchange_error` in `sf__strings.xml`.

The agent missed this because it didn't self-check against the Code Review Checklist before committing, even though the localization requirement is documented there. To prevent this in the future, b27aa97 adds an explicit instruction to the Agent Behavior Guidelines requiring a checklist self-review before every commit.

libs/SalesforceSDK/res/values/sf__strings.xml
```diff
@@ -10,6 +10,7 @@
     <string name="sf__access_token_revoked">Your access token has been revoked by the administrator. You will now be logged out.</string>
     <string name="sf__managed_app_error">Authentication only allowed from managed device.</string>
     <string name="sf__jwt_authentication_error">JWT authentication error. Please try again.</string>
+    <string name="sf__lightning_url_code_exchange_error">Can\'t use Lightning URL for code exchange. Use My Domain login server URL.</string>
```
brandonpage: I wouldn't be a bad idea to run this string by Steven before merge.

brandonpage (approved):

libs/SalesforceSDK/res/values/sf__strings.xml
JohnsonEricAtSalesforce: Steven's review of the string has been requested asynchronously via internal messaging. Will hold on merge until he weighs in.

wmathurin: We have stories about using the instance url for the refresh flow - shouldn't we be using the instance url also for the code exchange too?

brandonpage: @wmathurin Don't we get instance url _after_ login?

JohnsonEricAtSalesforce: Brandon is correct — the instance URL isn't available at code exchange time. The authorization code exchange happens *during* login, before the token response provides the instance URL.

During this session's investigation, we confirmed the flow:
1. User initiates login against the configured login server URL
2. Authorization endpoint returns a code
3. `exchangeCode()` in `OAuth2.java` sends `grant_type=authorization_code` to the login server's `/services/oauth2/token` endpoint
4. The token response returns the instance URL (along with access token, refresh token, etc.)

So at step 3, the instance URL doesn't exist yet — the SDK only has the configured login server URL. The refresh flow *can* use the instance URL because it runs after login has already completed and the instance URL is stored in the account credentials.

This diagnostic change intentionally doesn't try to fix the URL — it just surfaces the misconfiguration clearly. A future enhancement could normalize the login server URL before code exchange (Option A discussed during investigation), but that's a separate effort.

wmathurin: I guess you are right - that means I was incorrect in thinking we could get rid of discovery flow. For instance for app attestation, we need to get a challenge from the instance server which implies either using a configured my domain login server or discovering where we are going through a discovery flow.

wmathurin: It also means that username based log in through login.salesforce.com (or test.salesforce.com) will not work with app attestation - for the same reason.

wmathurin: Thinking some more, I believe we get redirected to the callback url from the instance URL, so we should be able to capture that information during login even though we don't do it now.

JohnsonEricAtSalesforce: Thanks for the thorough analysis. Future work items have been created as a result of internal discussion threads to track instance URL availability during login and how that affects both recently implemented authentication features (app attestation) and upcoming ones. This PR stays scoped to the diagnostic warning — the broader instance URL capture work will be addressed separately.

JohnsonEricAtSalesforce:
CI failure is `LoginViewModelTest.loginUrl_UpdatesOn_selectedServerChange` — a known flaky test (LiveData timing race). It's unrelated to this change and has an active PR to resolve it separately.

wmathurin (approved):

libs/SalesforceSDK/res/values/sf__strings.xml
JohnsonEricAtSalesforce: Steven reviewed the string and proposed: "Lightning URLs are not supported for OAuth code exchange. Use your My Domain URL instead." Updated in b0ca509 (corrected a typo from his original — "Lighting" → "Lightning").

PR #2923, opened 2026-06-10
W-22917333: Rename beacon_child_consumer_{key,secret} to auto_installed_app_org_consumer_{key,secret}

wmathurin:
## Summary

Updates the token endpoint wire-format keys and adds a legacy fallback for servers that haven't yet rolled out version 264.

When parsing the token response, the SDK now:
1. Checks for `auto_installed_app_org_consumer_{key,secret}` (new field name)
2. Falls back to `beacon_child_consumer_{key,secret}` if the new field is absent

A `TODO` comment marks both fallback constants and the fallback logic for removal once server version 264 has rolled out everywhere.

All code symbol names (`BEACON_CHILD_CONSUMER_KEY/SECRET` constants, `beaconChildConsumerKey` field/getter, builder methods, etc.) are **unchanged**.

## Files changed
- `OAuth2.java` — new constant values + legacy fallback constants + parsing updated with new-then-old fallback
- `UserAccount.java` — `BEACON_CHILD_CONSUMER_KEY/SECRET` constant values updated
- `AuthenticatorService.java` — `KEY_BEACON_CHILD_CONSUMER_KEY/SECRET` constant values updated
- `LogRedactor.kt` — redaction regex pattern string updated
- `UserAccountTest.java` — test map key strings updated
- `SalesforceLoggerTest.java` — test JSON key strings updated

## Test plan
- [x] Android SDK and SalesforceAnalytics APKs build cleanly
- [ ] `connectedAndroidTest` passes (requires connected device)
- [ ] AuthFlowTester instrumentation tests pass (running externally)

## Related
- iOS counterpart: forcedotcom/SalesforceMobileSDK-iOS#4057
- GUS: W-22917333

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
@@ -97,8 +97,8 @@ public class UserAccount {
 	public static final String CLIENT_ID = "clientId";
 	public static final String PARENT_SID = "parentSid";
 	public static final String TOKEN_FORMAT = "tokenFormat";
-	public static final String BEACON_CHILD_CONSUMER_KEY = "beacon_child_consumer_key";
-	public static final String BEACON_CHILD_CONSUMER_SECRET = "beacon_child_consumer_secret";
+	public static final String BEACON_CHILD_CONSUMER_KEY = "auto_installed_app_org_consumer_key";
+	public static final String BEACON_CHILD_CONSUMER_SECRET = "auto_installed_app_org_consumer_secret";
 	public static final String SCOPE = "scope";
 
 	private static final String TAG = "UserAccount";
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -215,6 +215,9 @@ public class OAuth2 {
     private static final String TOKEN_FORMAT = "token_format";
     private static final String BEACON_CHILD_CONSUMER_SECRET = "auto_installed_app_org_consumer_secret";
     private static final String BEACON_CHILD_CONSUMER_KEY = "auto_installed_app_org_consumer_key";
+    // TODO: Remove legacy fallback constants once server version 264 has rolled out everywhere.
+    private static final String LEGACY_BEACON_CHILD_CONSUMER_SECRET = "beacon_child_consumer_secret";
+    private static final String LEGACY_BEACON_CHILD_CONSUMER_KEY = "beacon_child_consumer_key";
 
     public static final DateFormat TIMESTAMP_FORMAT;
     static {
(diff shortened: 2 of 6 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -286,8 +286,8 @@ public class UserAccount {
 			clientId = object.optString(CLIENT_ID, null);
 			parentSid = object.optString(PARENT_SID, null);
 			tokenFormat = object.optString(TOKEN_FORMAT, null);
-			beaconChildConsumerKey = object.optString(BEACON_CHILD_CONSUMER_KEY, null);
-			beaconChildConsumerSecret = object.optString(BEACON_CHILD_CONSUMER_SECRET, null);
+			autoInstalledAppOrgConsumerKey = object.optString(AUTO_INSTALLED_APP_ORG_CONSUMER_KEY, null);
```
JohnsonEricAtSalesforce: **Upgrade path for existing stored accounts**

You've probably already considered this — for users who authenticated via an auto-installed app on 13.1+, their account data is persisted under the old `"beacon_child_consumer_key"` key. Between app upgrade and first successful token refresh, deserialization here won't find the value under the new key name, so `getClientIdForRefresh()` would fall back to the configured `clientId`.

Is the expectation that the refresh still succeeds with the configured `clientId` in that scenario (server accepts either), or would a fallback read be worth adding?
```java
autoInstalledAppOrgConsumerKey = object.optString(AUTO_INSTALLED_APP_ORG_CONSUMER_KEY,
    object.optString("beacon_child_consumer_key", null));
```

(Same would apply to the secret, the Bundle constructor, and `UserAccountManager.decryptUserData`.)

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -97,8 +97,8 @@ public class UserAccount {
 	public static final String CLIENT_ID = "clientId";
 	public static final String PARENT_SID = "parentSid";
 	public static final String TOKEN_FORMAT = "tokenFormat";
-	public static final String BEACON_CHILD_CONSUMER_KEY = "beacon_child_consumer_key";
-	public static final String BEACON_CHILD_CONSUMER_SECRET = "beacon_child_consumer_secret";
+	public static final String AUTO_INSTALLED_APP_ORG_CONSUMER_KEY = "auto_installed_app_org_consumer_key";
```
JohnsonEricAtSalesforce: **Scope vs. GUS story — confirming intent**

I'm sure you decided this deliberately — the GUS story mentions keeping Java identifiers unchanged with no public API changes, but the PR renames constants, getters, and builder methods (public API break for 14.0). Just confirming this was an intentional expansion beyond what the story's agent wrote, since it seems like the right call for a major version.

native/NativeSampleApps/AuthFlowTester/src/main/java/com/salesforce/samples/authflowtester/components/UserCredentialsView.kt
```diff
@@ -160,8 +160,8 @@ fun UserCredentialsView(currentUser: UserAccount?) {
         }
 
         InfoSection(title = BEACON) {
-            InfoRowView(label = BEACON_CHILD_CONSUMER_KEY, value = currentUser?.beaconChildConsumerKey)
-            InfoRowView(label = BEACON_CHILD_CONSUMER_SECRET, value = currentUser?.beaconChildConsumerSecret, isSensitive = true)
+            InfoRowView(label = AUTO_INSTALLED_APP_ORG_CONSUMER_KEY, value = currentUser?.autoInstalledAppOrgConsumerKey)
```
JohnsonEricAtSalesforce: **Minor:** The UI section header still says "Beacon" (line 60). Probably intentional to keep it short, but mentioning in case it was just overlooked in the sweep.

wmathurin:
I decided to only change the value of the fields we look for in the response and keep beacon in our code and tests. The changes were getting large and would make cherry picking harder. Further more AutoInstalledApp is a mouthful compared to Beacon.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -1000,11 +1003,16 @@ public TokenEndpointResponse(Response response, List<String> additionalOauthKeys
                 tokenFormat = parsedResponse.optString(TOKEN_FORMAT);
 
                 // Beacon child fields expected when using a beacon app and web server flow
+                // TODO: Remove LEGACY_BEACON_CHILD_CONSUMER_* fallback once server version 264 has rolled out everywhere.
```
wmathurin, line 1006: That way we can go against older instances and it should still work.

JohnsonEricAtSalesforce (approved):
👍🏻

PR #2931, opened 2026-06-17
Serialize OAuth token refresh per-account to prevent RTR logout.

brandonpage:


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -397,63 +434,134 @@ public AccMgrAuthTokenProvider(ClientManager clientManager, String instanceUrl,
...
-                if (gettingAuthToken) {
-                    try {
-                        lock.wait();
-                    } catch (InterruptedException e) {
-                        SalesforceSDKLogger.w(TAG, "Exception thrown while getting new auth token", e);
+            // The matching loop and the no-match early-out MUST run before any shared-state
+            // election so that a no-match path (e.g. account removed during refresh) never
+            // marks a RefreshState as refreshing — preserving the no deadlock fix and
+            // logout-during-refresh semantics.
+            final UserAccountManager userAccountManager = SalesforceSDKManager.getInstance().getUserAccountManager();
```
wmathurin, line 459: Should we add a helper function on UserAccountManager to return the matching user given a refresh token?

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
```diff
@@ -366,8 +367,44 @@ public interface RestClientCallback {
...
+            // object (rather than synchronizing on the RefreshState reference itself) makes the
+            // intent explicit and avoids the "synchronization on local variable" inspection.
+            final Object lock = new Object();
+            boolean refreshing;
+            String newAuthToken;        // winner's fresh access token (null on failure)
+            String newInstanceUrl;      // winner's instance URL (losers need it; see RestClient.refreshAccessToken)
+            String rotatedRefreshToken; // refresh token after rotation, for losers to adopt
+            long lastRefreshTime = -1;
+            boolean failed;
+        }
```
wmathurin, line 397: `state.failed` is written in two places in the `finally` block but never read anywhere — losers determine failure by checking `state.newAuthToken == null`. Either remove it, or use it (e.g. log a warning in `adoptWinnerResult` when `state.failed` is true) to avoid confusion.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/ClientManagerMockTest.kt
```diff
@@ -923,5 +938,414 @@ class ClientManagerMockTest {
...
+            every { orgId } returns "orgId"
+        }
+        val mockClientManager = mockk<ClientManager>(relaxed = true) {
+            every { accounts } returns arrayOf(mockAccount)
+        }
+        every { mockUserAccountManager.currentUser } returns mockUser
+        every { mockUserAccountManager.buildUserAccount(mockAccount) } returns mockUser
+
+        // Provider still holds the OLD access token and OLD refresh token (it was idle during
+        // the burst). It must match the account on the OLD refresh token, which is what it has.
```
wmathurin: Comment says "OLD access token and OLD refresh token" but the provider below is constructed with `newRefreshToken`, not the old one. The actual scenario is "old access token, already-in-storage (new) refresh token". The comment should be corrected to avoid misleading the reader about what the test actually exercises.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/ClientManagerMockTest.kt
```diff
@@ -923,5 +938,414 @@ class ClientManagerMockTest {
...
+    /** Deterministically blocks until [count] threads are parked in WAITING/TIMED_WAITING. */
+    private fun awaitThreadsParked(threads: List<Thread>, count: Int) {
+        val deadline = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(5)
+        while (System.currentTimeMillis() < deadline) {
+            val parked = threads.count {
+                it.state == Thread.State.WAITING || it.state == Thread.State.TIMED_WAITING
+            }
+            if (parked >= count) return
+            Thread.yield()
+        }
```
wmathurin, line 1429: `Thread.yield()` in a tight deadline loop burns CPU and can be flaky on loaded CI machines where yield starvation delays state transitions. `Thread.sleep(1)` per iteration is cheaper and less sensitive to scheduler pressure.

wmathurin (commented):
The core concurrency model is solid — the winner/loser election, spurious-wakeup loop, single `finally` publish path, and live-token read in `refreshStaleToken` are all correct. A few issues below worth addressing before merging.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/ClientManager.java
brandonpage: I considered that but it won't be needed when we make the [API change to ClientManager's constructor to associate it with a UserAccount from the start](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002MQk78YAD/view).

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/ClientManagerMockTest.kt
brandonpage: Good catch, thank you.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/rest/ClientManagerMockTest.kt
brandonpage: I updated to `Thread.sleep(50)` as I don't think 1ms is even feasible w/ schedule overhead.

libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -303,6 +303,26 @@ public OAuthRefreshInterceptor getOAuthRefreshInterceptor() {
...
+     * <p>Hybrid/WebView callers that previously performed their own token-endpoint request should
+     * call this instead, then read the updated values from {@code UserAccount} to set cookies —
+     * avoiding a second, parallel refresh that would rotate the refresh token out from under the
+     * SDK and trigger a spurious logout.
+     *
+     * @throws IOException on network failure or if the refresh token was revoked.
+     */
+    public synchronized void refreshAccessToken() throws IOException {
+        oAuthRefreshInterceptor.refreshAccessToken();
+    }
```
brandonpage, line 324: @wmathurin @JohnsonEricAtSalesforce The is simple but important new API.

wmathurin (commented):
Great fix for the RTR logout bug overall. 
I kept digging with Claude to see if having the two separate `synchronized(state.lock)` could lead to unwanted edge cases.
It came up with one. It's very unlikely to happen but the fix makes sense, so we should probably adopt it. Below are the details.

**Consecutive-cycle race: a loser can miss a result that was already published**

**The gap**

Between Block 1 (election) and Block 2 (publish), the lock is intentionally released so the network call doesn't hold it. `refreshing=true` is the "slot taken" flag during that gap. This is correct.

The issue arises at the transition between two consecutive refresh cycles. After Winner A publishes in Block 2:
```
A: refreshing=false, newAuthToken=published, notifyAll(), releases lock
   ← RACE: Loser X (just woken) and Winner B (arriving fresh) both compete for state.lock →
B: wins lock in Block 1, sees refreshing=false, sets refreshing=true, newAuthToken=null
X: acquires lock, sees refreshing=true (Winner B!), parks again with whatever deadline time remains
```
Winner B clears `state.newAuthToken=null` as part of becoming winner — correctly, to avoid re-publishing a stale result next cycle. But this means Loser X, even if it was nanoseconds behind, finds nothing to adopt and is now on the hook for Winner B's full network round-trip, counted against a deadline that started ticking during Winner A's cycle.

If Winner A consumed most of the 30s budget and Winner B's call takes longer than what remains, Loser X times out and returns null — even though Winner A's refresh succeeded.

**Proposed fix**

Add a freshness check inside Block 1, before committing to a new winner election:
```java
synchronized (state.lock) {
    if (state.refreshing) {
        // existing loser path — unchanged
    }
    // If a refresh completed very recently, adopt its result instead of starting a new cycle.
    // This closes the consecutive-cycle race where a loser woken by notifyAll() loses the
    // lock race to a new winner, which clears the just-published result before the loser
    // can read it.
    if (state.newAuthToken != null &&
            System.currentTimeMillis() - state.lastRefreshTime < RECENT_REFRESH_THRESHOLD_MILLIS) {
        adoptWinnerResult(state);
        return state.newAuthToken;
    }
    // Become winner — unchanged
    state.refreshing = true;
    state.newAuthToken = null;
    ...
}
```
`RECENT_REFRESH_THRESHOLD_MILLIS` can be small — a few seconds is enough, since the race window is the time between `notifyAll()` and a loser re-acquiring the lock, which is microseconds in practice. The threshold just needs to be larger than that window.

This makes `state.newAuthToken` serve dual purpose: election result for waiting losers, and short-lived cache for threads that arrive just after a cycle completes. Winner B becomes a no-op adopter rather than starting a redundant network call, and Loser X's deadline is no longer at risk from a cycle it had no connection to.

brandonpage:
> Great fix for the RTR logout bug overall. I kept digging with Claude to see if having the two separate `synchronized(state.lock)` could lead to unwanted edge cases. It came up with one. It's very unlikely to happen but the fix makes sense, so we should probably adopt it. Below are the details.

@wmathurin Great find.  I originally had a different approach to fix this (adding a generational counter) that I think better guarantees correctness in the worst case, but was definitely less elegant in some situations.  I was able to combine the best of both into [this fix](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2931/changes/34f0be02e3aac62c99467d8384183089fc2c2658).

wmathurin (approved):

PR #2933, opened 2026-06-19
Fix Login for Admin and Welcome Discovery incompatibility

brandonpage:
## Summary

Fixes the interaction between **Login for Admin (LFA)** and the **Welcome Discovery** 2-phase login flow. Previously the LFA menu item was offered even on the Welcome Discovery Phase 1 page (welcome.salesforce.com/discovery), where the registered OAuth callback (`sfdc://discocallback`) is **not app-unique** — any app on the device that registers the `sfdc://` scheme could intercept the response. This change hides LFA in Phase 1, keeps it available in Phase 2 (the discovered My Domain), guards `LoginActivity.launchLoginForAdminsAction` so a programmatic invocation is also a no-op for Welcome Discovery, and fixes a related reload bug.

### 1. LFA hidden on Welcome Discovery Phase 1
- `LoginView` (Compose) now passes `onLoginForAdmins = null` to `DefaultTopAppBar` when the VM-level `selectedServer` LiveData matches `LoginActivity.isSalesforceWelcomeDiscoveryUrlPath`.
- `DefaultTopAppBar` already conditionally renders the menu item when `onLoginForAdmins != null` — the test wrapper signature was widened to nullable to mirror that contract.

### 2. LFA action is also a no-op for Welcome Discovery (defense-in-depth)
- `LoginActivity.launchLoginForAdminsAction()` now short-circuits with a single `w(TAG, ...)` log when `LoginServerManager.selectedLoginServer.url` matches the discovery path. Even if a host app calls the (deprecated) public method directly, it cannot launch a Custom Tab against welcome.salesforce.com.
- The `@Deprecated` annotation, KDoc, and visibility are unchanged. KDoc gains one line noting the no-op.

### 3. Reload / Clear Cookies / Clear Cache / re-select Welcome → returns the WebView to Phase 1
- `LoginViewModel.reloadWebView()` checks `LoginServerManager.selectedLoginServer` (the source of truth — distinct from `viewModel.selectedServer`, which during Phase 2 holds the discovered My Domain). When it is Welcome Discovery, the function:
  - resets `dynamicBackgroundColor` to white synchronously (so the top app bar background flips immediately, not after `onPageFinished`),
  - rebuilds the Phase-1 mobile URL via the relocated `generateSalesforceWelcomeDiscoveryMobileUrl`,
  - realigns `viewModel.selectedServer.value` back to the Welcome URL so the title and the menu gating recompose,
  - returns early so the Phase-2 auth-URL regen path is skipped.
- `PickerBottomSheet` adds an equality-branch handler: re-selecting Welcome Discovery while it is already the selected server now calls `viewModel.reloadWebView()` (otherwise the picker would dismiss with no observable change).

### 4. Top app bar title is now reactive
- `LoginView` previously read `viewModel.defaultTitleText` (a plain Kotlin getter that calls `LiveData.getValue()`), which is invisible to Compose's snapshot system, so the title never recomposed when `selectedServer` changed. The composable now `observeAsState()`s `selectedServer` and `loginUrl` and reads them directly inside the title computation, so the bar updates in lock-step with reload/re-select.

### 5. Structural-only relocation
- `generateSalesforceWelcomeDiscoveryMobileUrl` moved from a `private fun` on `LoginActivity` to an `internal fun` on `LoginViewModel` so both the activity (intent dispatch) and the VM (reload) can share one implementation. Bytes-identical Phase-1 URL — verified by S19 in the manual test plan.
- `SALESFORCE_WELCOME_DISCOVERY_MOBILE_CALLBACK_URL` widened from `private` to `internal` to support the relocation. No public API surface change.

## Test plan — manual verification checklist

A full manual test plan with 20 numbered scenarios was authored alongside this work (kept locally; not committed). Reviewers should walk through the following scenarios on **at least Android API 31 and API 36**, running the AuthFlowTester sample app unless noted otherwise.

### Menu gating
- [x] **S1** — Welcome Discovery selected, Phase 1 loaded → ⋮ menu does NOT contain \"Login for Admin\".
- [x] **S2** — Production / Sandbox / custom My Domain selected → ⋮ menu DOES contain \"Login for Admin\".
- [x] **S3** — Phase-2 (post-discovery) My Domain → ⋮ menu DOES contain \"Login for Admin\".

### Reload bug fix (the WebView returns to Phase 1 on each of these)
- [x] **S4** — Phase 2 → tap Reload → WebView navigates to `https://welcome.salesforce.com/discovery?client_id=...&client_version=...&callback_url=sfdc%3A%2F%2Fdiscocallback`.
- [x] **S5** — Phase 2 → server picker → re-select \"Welcome\" → same end state as S4.
- [x] **S6** — Phase 2 → Clear Cookies → Phase 1 reloads, cookies cleared.
- [x] **S7** — Phase 2 → Clear cache → Phase 1 reloads, cache cleared.

### Phase-1 → Phase-2 transition (regression guard)
- [x] **S8** — Welcome → enter email → land in Phase 2 → ⋮ menu now CONTAINS \"Login for Admin\" (no menu reopen needed).

### LFA still works in Phase 2
- [x] **S10** — Verify across all variations: a Custom Tab is NEVER launched against welcome.salesforce.com. Logcat audit shows zero hits on `welcome.salesforce.com` in the `loadLoginPageInCustomTab` path.

### Multi-user
- [x] **S11** — User A logged in via Welcome Discovery → add User B via Production → ⋮ menu reflects User B's server (LFA visible).
- [x] **S12** — User A logged in via Welcome Discovery → add User B also via Welcome Discovery → Phase 1 is fresh, no auto-advance to a stale Phase 2.

### Configuration changes
- [x] **S13** — Phase 1 → rotate device → menu still hides LFA in both orientations; WebView preserves Phase 1.
- [x] **S14** — Phase 2 → rotate → menu still shows LFA → tap Reload → returns to Phase 1; rotate back → menu now hides LFA.
- [x] **S15** — Toggle dark/light theme in either phase → URL unchanged, menu list unchanged, top-bar colors update.

### Backward compatibility (apps without Welcome Discovery)
- [x] **S18** — `RestExplorer` (no Welcome in `servers.xml`) → menu list unchanged from prior release; Reload, Clear Cookies, Clear cache, Login for Admin all behave identically to the pre-change build.

### URL-builder relocation regression guard
- [x] **S19** — Capture the Phase-1 URL via WebView client log → bytes-identical to a `dev`-baseline capture: same scheme/host/path, same query parameters in the same order (`client_id`, `client_version`, `callback_url`), `client_version` UTF-8 encoded with `%2F` for slashes, `callback_url=sfdc%3A%2F%2Fdiscocallback`.

### Cross-platform parity
- [x] **S25 Ultra** — Re-run S1, S4, S9 on the iOS twin once it lands; mark as deferred with a linked iOS PR if it has not yet merged.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -772,7 +772,14 @@ open class LoginActivity : FragmentActivity() {
      */
     @Deprecated(message = "This function will be replaced by a permanent solution in 14.0.")
     fun launchLoginForAdminsAction() {
-        val selectedServer = SalesforceSDKManager.getInstance().loginServerManager.selectedLoginServer?.url?.toUri()
+        // Guard against launching a Custom Tab on the Welcome Discovery Phase 1 page, whose
+        // callback (sfdc://discocallback) is not app-unique.  This MUST use the same signal as the
+        // menu-item visibility gating in LoginView (viewModel.selectedServer), NOT
+        // LoginServerManager.selectedLoginServer: the latter stays welcome.salesforce.com through
+        // BOTH phases, while selectedServer is the discovered My Domain in Phase 2 — where LFA is
+        // valid and the menu item is shown.  Keying off LoginServerManager here would no-op the
+        // action in Phase 2 even though the menu offers it.
+        val selectedServer = viewModel.selectedServer.value?.toUri()
         if (selectedServer?.let { isSalesforceWelcomeDiscoveryUrlPath(it) } == true) {
             w(TAG, "launchLoginForAdminsAction is a no-op for Welcome Discovery; " +
                     "LFA requires an app-unique OAuth callback")
@@ -1234,8 +1241,15 @@ open class LoginActivity : FragmentActivity() {
                     return@evaluateJavascript
                 }
 
-                viewModel.dynamicBackgroundColor.value = validateAndExtractBackgroundColor(result)
-                    ?: return@evaluateJavascript
+                viewModel.dynamicBackgroundColor.value =
+                    if (url?.toUri()?.let { isSalesforceWelcomeDiscoveryUrlPath(it) } == true) {
+                        // The Welcome Discovery Phase 1 page (welcome.salesforce.com/discovery)
+                        // reports a dark computed background in some states (e.g. after logout),
+                        // which would tint the top and bottom app bars black.
+                        Color.White
+                    } else {
+                        validateAndExtractBackgroundColor(result) ?: return@evaluateJavascript
+                    }
 
                 // Ensure Status Bar Icons are readable no matter which OS theme is used.
                 val titleTextColorLight = viewModel.titleTextColor?.luminance()?.let { it < 0.5 }
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -150,7 +150,6 @@ import org.json.JSONObject
 import java.lang.String.format
 import java.net.URI
 import java.net.URLDecoder
-import java.net.URLEncoder
 import java.security.PrivateKey
 import java.security.cert.X509Certificate
 
@@ -768,9 +767,17 @@ open class LoginActivity : FragmentActivity() {
      *  Launches the current login server in a Custom Tab. This allows Admins
      *  to authenticate with a Passkey or by other methods that require
      *  Advanced/Browser Authentication.
+     *
+     *  This function is a no-op for Welcome Discovery.
      */
     @Deprecated(message = "This function will be replaced by a permanent solution in 14.0.")
     fun launchLoginForAdminsAction() {
+        val selectedServer = SalesforceSDKManager.getInstance().loginServerManager.selectedLoginServer?.url?.toUri()
+        if (selectedServer?.let { isSalesforceWelcomeDiscoveryUrlPath(it) } == true) {
+            w(TAG, "launchLoginForAdminsAction is a no-op for Welcome Discovery; " +
+                    "LFA requires an app-unique OAuth callback")
+            return
+        }
         val loginUrl = viewModel.browserCustomTabUrl.value ?: return
         loadLoginPageInCustomTab(loginUrl, adminLoginCustomTabLauncher)
     }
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -299,8 +300,69 @@ open class LoginViewModel(
...
+                        generateSalesforceWelcomeDiscoveryMobileUrl(selectedLoginServerUri).toString()
+                    // Reset VM-level selectedServer back to the Welcome URL so the top app bar
+                    // title (via defaultTitleText), menu gating in Compose (LoginView reads
+                    // viewModel.selectedServer to hide "Login for Admin" on Phase 1), and any
+                    // other observers of selectedServer realign with Phase 1.  In Phase 2 this
+                    // value had been overwritten with the discovered My Domain by
+                    // applySalesforceWelcomeLoginHintAndHost → applyPendingServer.
+                    val welcomeUrl = selectedLoginServerUri.toString()
+                    if (selectedServer.value != welcomeUrl) {
+                        selectedServer.value = welcomeUrl
```
wmathurin, line 353: The ordering of `loginUrl.value =` (line 350) before `selectedServer.value =` (line 360) is load-bearing: `LoginUrlSource`'s same-host short-circuit relies on seeing the new `loginUrl` host before `selectedServer` changes, so it suppresses a spurious auth-URL regen. This invariant is currently documented only in the large block comment above — consider adding a short inline comment at each of the two assignment lines (e.g. `// Must precede selectedServer reset — see LoginUrlSource same-host short-circuit`) so a future refactor doesn't accidentally reorder them or extract one without the other.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -768,9 +767,17 @@ open class LoginActivity : FragmentActivity() {
      *  Launches the current login server in a Custom Tab. This allows Admins
      *  to authenticate with a Passkey or by other methods that require
      *  Advanced/Browser Authentication.
+     *
+     *  This function is a no-op for Welcome Discovery.
      */
     @Deprecated(message = "This function will be replaced by a permanent solution in 14.0.")
     fun launchLoginForAdminsAction() {
+        val selectedServer = SalesforceSDKManager.getInstance().loginServerManager.selectedLoginServer?.url?.toUri()
+        if (selectedServer?.let { isSalesforceWelcomeDiscoveryUrlPath(it) } == true) {
```
wmathurin, line 783: The Compose-layer visibility gate (`isWelcomeDiscoveryServer` in `LoginView.kt`) is keyed off `viewModel.selectedServer`, while this programmatic guard is keyed off `LoginServerManager.selectedLoginServer`. These are intentionally different state sources (documented well in `reloadWebView`), but they create a subtle window: in phase 2 the Compose layer correctly shows the LFA item (VM's `selectedServer` is the My Domain), yet if this guard fires anyway (e.g. in a race or from a direct external call), the action silently no-ops while the user sees the item as active. A brief inline comment here noting the dual-source-of-truth intent would help future readers understand why the two guards differ rather than assuming one is stale.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -160,6 +167,8 @@ fun LoginView() {
         }
     }
 
+    val isWelcomeDiscoveryServer = selectedServer.value
+        ?.let { LoginActivity.isSalesforceWelcomeDiscoveryUrlPath(it.toUri()) } == true
```
wmathurin, line 173: `isWelcomeDiscoveryServer` is computed from `viewModel.selectedServer`, which in phase 2 holds the resolved My Domain — so this correctly evaluates to `false` and shows the LFA item. However, the LFA guard in `LoginActivity.launchLoginForAdminsAction()` uses `LoginServerManager.selectedLoginServer` instead. These are consistent for all described scenarios, but a short comment here (e.g. `// During phase 2, selectedServer is the My Domain, so this is false and LFA is shown — see LoginActivity guard for the programmatic defense-in-depth check`) would make the dual-source-of-truth explicit to a future reader rather than leaving it implicit.

brandonpage: ~~Yes, sorry I have this change locally but had not pushed yet while I was wrapping up some other testing.  Pushed now.~~

I meant to put that on the `LoginActivity` comment.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: This should also be fixed by my last commit.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
wmathurin: Thanks for the fix — keying both layers off  removes the inconsistency. However, the `LoginView.kt` comment doesn't appear in the diff for commit `8da7d06` (the only changed files are `LoginActivity.kt`, `LoginViewModel.kt`, and `LoginActivityTest.kt`). Could you check whether it got included?

wmathurin (approved):

PR #2936, opened 2026-06-24
W-17440369: Use instanceUrl host when refreshing OAuth tokens

wmathurin:
## Summary

Token refresh previously always POSTed to `loginServer` (e.g. `login.salesforce.com`). This breaks DPoP `htu` binding (RFC 9449 §4.3) when the login-pool redirects to the user's My Domain, and adds an unnecessary round-trip on every refresh for Bearer flows.

Mirrors the iOS fix from [PR #4074](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4074).

## Changes

**`OAuth2.java`**
- Adds `overrideLoginServerIfNeeded(loginServer, instanceServer, communityId, communityUrl)` static helper with precedence: `communityUrl > instanceServer > loginServer`. `instanceServer` is null until after the first token response, so the code-exchange path falls through to `loginServer` naturally with no special-casing.
- Updates `getOpenIDToken()` to accept and apply the same precedence (new parameters: `instanceServer`, `communityId`, `communityUrl`).

**`ClientManager.java`** — `AccMgrAuthTokenProvider.refreshStaleToken()`
- Uses `overrideLoginServerIfNeeded` before calling `refreshAuthToken`
- Adds target-host log line (host only, no token values)

**`AuthenticatorService.java`** — `Authenticator.getAuthToken()`
- Same update as ClientManager

**`OAuth2OverrideLoginServerTests.kt`** — 6 new unit tests:
1. Instance server populated → returns instance server
2. Instance server null → returns login server
3. Community ID + URL set → returns community URL (even when instance server is also set)
4. Community ID set but URL null → falls back to instance server
5. All null (code exchange path) → returns login server
6. Malformed community URL → falls back to instance server

**`OAuth2Test.java`** — Updated `testGetOpenIDToken` call to match new `getOpenIDToken` signature.

## Test plan

- [x] 6 new `OAuth2OverrideLoginServerTests` pass on emulator (API 36)
- [x] Existing `OAuth2MockTests` (6 tests) pass
- [x] `LoginServerManagerTest` and `ScopeParserTest` pass
- [x] Compiles clean (only pre-existing deprecation warnings, none from this PR)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
@@ -134,8 +134,14 @@ public class AuthenticatorService extends Service {
 
             try {
                 final Map<String,String> addlParamsMap = originalUserAccount.getAdditionalOauthValues();
+                final URI tokenServer = OAuth2.overrideLoginServerIfNeeded(
+                        originalUserAccount.getLoginServer(),
+                        originalUserAccount.getInstanceServer(),
+                        originalUserAccount.getCommunityId(),
+                        originalUserAccount.getCommunityUrl());
+                SalesforceSDKLogger.i(TAG, "Initiating token refresh to host: " + tokenServer.getHost());
                 final OAuth2.TokenEndpointResponse tr = OAuth2.refreshAuthToken(HttpAccess.DEFAULT,
-                        new URI(originalUserAccount.getLoginServer()), originalUserAccount.getClientIdForRefresh(), originalUserAccount.getRefreshToken(), addlParamsMap);
+                        tokenServer, originalUserAccount.getClientIdForRefresh(), originalUserAccount.getRefreshToken(), addlParamsMap);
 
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
```diff
@@ -134,8 +134,14 @@ public Bundle getAuthToken(AccountAuthenticatorResponse response, Account accoun
 
             try {
                 final Map<String,String> addlParamsMap = originalUserAccount.getAdditionalOauthValues();
+                final URI tokenServer = OAuth2.overrideLoginServerIfNeeded(
+                        originalUserAccount.getLoginServer(),
+                        originalUserAccount.getInstanceServer(),
+                        originalUserAccount.getCommunityId(),
+                        originalUserAccount.getCommunityUrl());
```
brandonpage: NIT: Seems like it might be cleaner to just pass the user account?

brandonpage (approved):

PR #2940, opened 2026-06-26
W-23159884: Add stable test tags to login screen and server/account picker

brandonpage:
### What & why
Login-screen and server/account-picker UI tests previously anchored on **localized** `contentDescription`/text strings (the `sf__*` string resources). Those anchors break under non-English locales and right-to-left layouts, which is the accessibility-anchor concern tracked by the parent story.

This change introduces stable, **locale-invariant** `Modifier.testTag` anchors and routes all test/page-object code through them.

### Changes
- **New public API** — `LoginViewTestTags` object (`com.salesforce.androidsdk.ui.components`): the single source of truth for the locale-invariant test-tag resource-ids. Intentionally public so SDK instrumented tests, sample-app page objects (AuthFlowTester), and external UTAM page objects all reference the same constants. Replaces the former `internal const val TAG_*` block in `LoginView.kt`.
- **Production UI** — applied/migrated tags in `LoginView.kt` (more-options button, back button, the 6 overflow menu items, loading indicator, login button), `PickerBottomSheet.kt` (server/account picker containers, back/close buttons, add-connection button, custom label/URL fields, apply button), and `LoginServerListItem.kt` (server delete + confirm-delete targets).
- **Tests** — migrated `PickerBottomSheetTest.kt` (SalesforceSDKTest) and `LoginPageObject.kt` (AuthFlowTester) to anchor on the tags. Row-identity matchers that assert rendered content (`hasText(name) and hasText(url)`) are intentionally retained; Espresso WebView fields and the native AlertDialog path are untouched (not Compose / not taggable).

### Reviewer notes
- **Public API addition** (`LoginViewTestTags`) — flagged per SDK escalation rules. No metalava/API-validator dump exists in the repo, so nothing to regenerate.
- **No `sf__strings.xml` change** — these are resource-id test anchors, not user-facing localized strings.
- The composition roots that host these nodes already opt in with `testTagsAsResourceId = true`, so the tags surface to UIAutomator2/UTAM as Android `resource-id`s.

### Testing
- Compile-verified: production, SalesforceSDKTest androidTest, and AuthFlowTester androidTest source sets all build clean.
- Instrumented suites run in CI on Firebase Test Lab.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
@@ -90,6 +90,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.core.net.toUri
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.shadow
@@ -98,9 +99,11 @@ import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalInspectionMode
+import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
@@ -234,6 +237,7 @@ fun LoginView() {
     )
 }
 
+@OptIn(ExperimentalComposeUiApi::class)
 @Composable
 internal fun LoginView(
     dynamicBackgroundColor: MutableState<Color>,
@@ -254,6 +258,9 @@ internal fun LoginView(
     )
 
     Scaffold(
+        // Expose Compose testTags as Android resource-ids so UI automation (UIAutomator2/UTAM)
+        // can anchor on the stable, locale-invariant tags rather than localized contentDescriptions.
+        modifier = Modifier.semantics { testTagsAsResourceId = true },
         bottomBar = bottomAppBar,
         contentWindowInsets = WindowInsets.safeDrawing,
         topBar = topAppBar,
@@ -285,7 +292,7 @@ internal fun LoginView(
 }
 
 
-@OptIn(ExperimentalMaterial3Api::class)
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalComposeUiApi::class)
 @Composable
 internal fun DefaultTopAppBar(
     backgroundColor: Color,
@@ -324,6 +331,7 @@ internal fun DefaultTopAppBar(
                         disabledContainerColor = Color.Transparent,
                         disabledContentColor = Color.Transparent,
                     ),
+                    modifier = Modifier.testTag(LoginViewTestTags.MORE_OPTIONS_BUTTON),
                 ) {
                     Icon(Icons.Default.MoreVert, contentDescription = moreOptionsDescription)
                 }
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginViewTestTags.kt
```diff
@@ -0,0 +1,121 @@
...
+ * page-object code can anchor on them in any locale — including right-to-left layouts.
+ *
+ * This object is the single source of truth for those values. It is intentionally public so that
+ * SDK instrumented tests, sample-app page objects (e.g. AuthFlowTester), and external UTAM page
+ * objects can all reference the same constants instead of duplicating the literal strings.
+ *
+ * IMPORTANT: These are test anchors only. Do NOT localize them and do NOT reuse them as
+ * user-facing strings.
+ */
+object LoginViewTestTags {
```
brandonpage, line 47: This might appear to be an over investment given our future Login UX plans, but I actually see it being a useful bridge.  If/when we introduce a new SDK provided default UI in `15.0` we can ensure these tags are used.  And since they are public consuming apps can use them in their UI implementation if they chose to create their own.  This make UI tests easier to migrate and hopefully ease the transition.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginViewTestTags.kt
```diff
@@ -0,0 +1,121 @@
...
+    /** Save / apply button in the picker's "Add Connection" form. */
+    const val APPLY_BUTTON = "sf__apply_button"
+    // endregion
+
+    // region Server list item
+    /** Trash icon on a custom server row that arms the slide-to-delete affordance. */
+    const val SERVER_DELETE_BUTTON = "sf__server_delete_button"
+
+    /** Revealed confirm-delete target on a custom server row. */
+    const val SERVER_CONFIRM_DELETE_BUTTON = "sf__server_confirm_delete_button"
```
wmathurin, line 133: Both `SERVER_DELETE_BUTTON` and `SERVER_CONFIRM_DELETE_BUTTON` are shared by every row in the list — all delete buttons get the same tag by design. That's expected Compose behaviour, but it will surprise the first person writing a test that targets a specific server row. Worth adding a one-liner KDoc note here (and on `SERVER_DELETE_BUTTON` above) pointing to the row-identity pattern:

```kotlin
// To target a specific row, combine with a content matcher:
// onNode(hasTestTag(SERVER_DELETE_BUTTON) and hasAncestor(hasText("MyOrg")))
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -422,7 +436,9 @@ internal fun MenuItem(
             )
         },
         onClick = onClick,
-        modifier = Modifier.semantics { contentDescription = text }
+        modifier = Modifier
+            .then(if (testTag != null) Modifier.testTag(testTag) else Modifier)
```
wmathurin, line 440: Minor style nit: the `.then(Modifier)` wrapping is a no-op and slightly obscures intent. A more idiomatic form avoids it:

```kotlin
modifier = (if (testTag != null) Modifier.testTag(testTag) else Modifier)
    .semantics { contentDescription = text }
```

Functionally identical — just a readability preference.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
```diff
@@ -234,6 +237,7 @@ fun LoginView() {
     )
 }
 
+@OptIn(ExperimentalComposeUiApi::class)
```
wmathurin, line 240: `testTagsAsResourceId` is still experimental API, so this `@OptIn` is needed for now. Worth leaving a TODO so it's easy to find and clean up once Compose promotes it to stable:

```kotlin
// TODO: remove @OptIn once testTagsAsResourceId is promoted to stable Compose API
@OptIn(ExperimentalComposeUiApi::class)
```

Same applies to the `@OptIn` in `DefaultTopAppBar` and in `PickerBottomSheet`.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
brandonpage: I would argue that is worse style.  Everything in compose is chain-able but your replacement is not.  

If I had two conditional properties I would have to use a `.then` for the second on no matter what. For example:

```
modifier = (if (testTag != null) Modifier.testTag(testTag) else Modifier)
    .then(<second conditional>)
    .semantics { contentDescription = text }
```

is worse than:

```
modifier = Modifier
    .then(if (testTag != null) Modifier.testTag(testTag) else Modifier)
    .then(<second conditional>)
    .semantics { contentDescription = text }
```

wmathurin: I agree - that was not great feedback - I should have kept this one when reviewing the review.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
brandonpage: I think a TODO is unnecessary.  The `OptIn` becomes a warning when unused.

wmathurin (approved):

brandonpage:
Tests will fail on the latest run because the test org goes down for maintenance every day ~4 PST.   All test passed on the first commit and my second commit only added comments.

PR #2944, opened 2026-06-29
feat(W-22695293): DPoP proof JWT at token exchange (Phase 2, Android)

wmathurin:
## Summary

- **New package `com.salesforce.androidsdk.auth.dpop`** with three classes:
  - `DPoPURLHelper` — RFC 9449 §4.2 URL canonicalization (strips query + fragment)
  - `DPoPKeyManager` — Android Keystore EC P-256 keypair lifecycle, scoped per `credentialsIdentifier`
  - `DPoPProofBuilder` — compact JWS construction (JWK export of affineX/Y, jti via SecureRandom, DER→raw R||S signature)
- **`OAuth2.java`** — new `makeTokenEndpointRequest` / `exchangeCode` / `refreshAuthToken` overloads that accept a `credentialsIdentifier`; attaches `DPoP` header when `SalesforceSDKManager.isUseDPoP() == true`; parses `token_type` from token response; `addAuthorizationHeader` now selects `DPoP` or `Bearer` scheme based on token type
- **`SalesforceSDKManager.kt`** — new `isUseDPoP()` / `setUseDPoP(Boolean)` opt-in flag
- **`UserAccount.java` + `UserAccountBuilder.kt`** — new `dpopScope` (keypair alias UUID) and `tokenType` fields, persisted through JSON/Bundle serialization
- **`LoginViewModel.kt` + `AuthenticationUtilities.kt`** — UUID generated at login start, threaded through `exchangeCode`, stored on `UserAccount` via `UserAccountBuilder`
- **`AuthenticatorService.java`** — passes `dpopScope` to `refreshAuthToken` so the same keypair is reused on refresh
- **`RestClient.java` / `ClientManager.java`** — `OAuthRefreshInterceptor` now holds and applies `tokenType`; `AuthTokenProvider.getTokenType()` default method propagates it through refresh

## Test plan

- [ ] `DPoPURLHelperTest` — 4 canonicalization cases
- [ ] `DPoPKeyManagerTest` — 5 tests: generate/reload idempotency, deletion, per-alias isolation, alias prefix
- [ ] `DPoPProofBuilderTest` — 7 tests: JWS structure, claims, signature verification, JWK coordinate size, jti uniqueness
- [ ] `OAuth2MockTests` — 8 new DPoP tests: header present/absent, null credentialsIdentifier, token type persisted, `DPoP`/`Bearer`/null auth scheme selection, request completes gracefully on keystore error

## Out of scope for this story (W-22695293 Phase 2)

- Nonce challenge/retry — Phase 4 (W-22697878)
- API call proofs with `ath` claim — Phase 3
- Automated UI tests — Phase 5

## iOS equivalent

W-22695307 (closed) on `upstream/dpop` branch — this PR brings Android to parity.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2026-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of salesforce.com, inc. nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission of salesforce.com, inc.
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.androidsdk.auth.dpop
+
+import android.security.keystore.KeyGenParameterSpec
+import android.security.keystore.KeyProperties
+import java.security.KeyPair
+import java.security.KeyPairGenerator
+import java.security.KeyStore
+import java.security.PrivateKey
+import java.security.interfaces.ECPublicKey
+import java.security.spec.ECGenParameterSpec
+
+object DPoPKeyManager {
+
+    private const val ANDROID_KEYSTORE = "AndroidKeyStore"
+
+    fun generateOrLoadKeyPair(alias: String): KeyPair {
+        val keyStore = KeyStore.getInstance(ANDROID_KEYSTORE).apply { load(null) }
+        val existingKey = keyStore.getKey(alias, null)
+        if (existingKey != null) {
+            val privateKey = existingKey as PrivateKey
+            val publicKey = keyStore.getCertificate(alias).publicKey as ECPublicKey
+            return KeyPair(publicKey, privateKey)
+        }
+        val keyPairGenerator =
+            KeyPairGenerator.getInstance(KeyProperties.KEY_ALGORITHM_EC, ANDROID_KEYSTORE)
+        val spec = KeyGenParameterSpec.Builder(
+            alias,
+            KeyProperties.PURPOSE_SIGN or KeyProperties.PURPOSE_VERIFY
+        )
+            .setAlgorithmParameterSpec(ECGenParameterSpec("secp256r1"))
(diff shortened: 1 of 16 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
```diff
@@ -0,0 +1,72 @@
...
+        )
+            .setAlgorithmParameterSpec(ECGenParameterSpec("secp256r1"))
+            .setDigests(KeyProperties.DIGEST_SHA256)
+            .setUserAuthenticationRequired(false)
+            .build()
+        keyPairGenerator.initialize(spec)
+        return keyPairGenerator.generateKeyPair()
+    }
+
+    fun deleteKeyPair(alias: String) {
```
JohnsonEricAtSalesforce, line 64: This method exists but is never called outside of test teardown. When users log out or accounts are removed via `SalesforceSDKManager.removeAccount()`, the DPoP keypair remains in Android Keystore indefinitely, causing keystore bloat over time (especially for multi-user scenarios or frequent login/logout cycles).

**Cross-platform finding:** The iOS implementation (commit `55987b138` on `upstream/dpop`) **does clean up keypairs on logout** in `SFUserAccountManager.m`:

```swift
if (user.credentials.identifier.length > 0) {
    [SFSDKDPoPKeyStore.shared deleteForCredentials:user.credentials];
    [SFSDKDPoPNonceCache.shared clearForScope:user.credentials.identifier];
}
```

**Questions:**
1. Was the Android cleanup intentionally deferred, or is this an oversight?
2. Should we match iOS behavior and add cleanup before SDK 14.0 ships?
3. If cleanup is planned for Phase 3/4, can you note that in the PR description?

**Suggested approach:** Add `deleteKeyPair()` call in `SalesforceSDKManager.removeAccount()` around line 1232, using `userAccount?.dpopScope` as the credentials identifier.

JohnsonEricAtSalesforce (commented):
Review comment posted by AI agent. See inline comment for details.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -667,16 +669,20 @@ open class LoginViewModel(
             }
             val verifier = if (isUsingFrontDoorBridge) frontdoorBridgeCodeVerifier else codeVerifier
 
+            val credentialsIdentifier = java.util.UUID.randomUUID().toString()
```
JohnsonEricAtSalesforce, line 672: ## Naming: `dpopScope` is Misleading

**Files:** `UserAccount.java:125-132`, `UserAccountBuilder.kt:113`

The field `dpopScope` is semantically incorrect — it's a **keystore alias identifier** (UUID), not an OAuth scope.

**Cross-platform:** iOS uses `credentialsIdentifier` for this purpose.

**Recommendation:** Rename to `credentialsIdentifier` or `dpopCredentialsIdentifier` to match iOS and clarify intent.

---

## Question: Stable vs. Ephemeral Identifier

iOS reuses the existing `credentials.identifier` (stable across logins). Android generates `UUID.randomUUID()` per login (this line), creating a new keypair each time.

**Is the "new UUID per login" behavior intentional?** If so, what's the rationale vs. iOS? If not, should we align by reusing a stable identifier?

**Impact:** With cleanup now implemented (4efd4a6a3), keypairs are properly deleted on logout ✅. However, the "new UUID per login" approach still means more keystore churn than iOS:
- User logs out → logs back in = new keypair + keystore write/delete cycle
- iOS approach: same keypair reused across sessions = less keystore operations

Not a bug, but worth discussing for platform consistency and performance.

JohnsonEricAtSalesforce (approved):
👍

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
wmathurin: Thanks — addressed the rename in 23d451f6c:

- `UserAccount.dpopScope` → `UserAccount.credentialsIdentifier` (field, getter/setter, JSON/Bundle key)
- `UserAccountBuilder.dpopScope(...)` → `UserAccountBuilder.credentialsIdentifier(...)`
- `UserAccount.DPOP_SCOPE` constant → `UserAccount.CREDENTIALS_IDENTIFIER`
- All callers updated (`AuthenticationUtilities`, `AuthenticatorService`, `ClientManager`, `SalesforceSDKManager`)

The rest of the Android codebase already uses `credentialsIdentifier` (see `LoginViewModel.doCodeExchange`, `OAuth2.makeTokenEndpointRequest`, `DPoPKeyManager.aliasForCredentialsIdentifier`), so this brings the persisted account field in line with everything else and matches iOS terminology.

---

On the UUID-per-login question: it is intentional, and the keystore-churn difference vs. iOS is smaller than it looks.

- A fresh UUID is only generated on a **new login** (`doCodeExchange`, after a successful code exchange). Token **refreshes** carry the existing `credentialsIdentifier` forward via `populateFromUserAccount(...)` in `AuthenticatorService`/`ClientManager`, so the keypair is reused for the full lifetime of the account.
- The keypair is now cleaned up on logout (4efd4a6a3 + this PR), so the steady-state cost of "log out → log back in" is one keystore delete + one keystore create per login cycle, not unbounded growth.
- iOS reuses `credentials.identifier` (a stable per-user id) so a logout/login pair on iOS re-creates the keypair under the same alias — that's still a delete + create, just under a deterministic name rather than a fresh UUID.

So the iOS approach saves nothing on logout/login churn; the only real difference is that iOS could in principle skip the keypair regeneration if the alias is reused without intervening cleanup, which isn't a flow we want to rely on for DPoP (the spec encourages fresh keys per session).

Happy to align on `credentials.identifier`-style naming if there's a stable per-user id available on Android at this point in the flow, but the UUID was chosen because at code-exchange time we don't yet have `userId`/`orgId` — those come back later from the identity service. Switching to a userId-derived alias would require deferring the key binding until after identity, which complicates the proof JWT path.

PR #2949, opened 2026-06-30
feat(W-22697744): DPoP proof JWT on API calls (Phase 3)

wmathurin:
## Summary

- **`DPoPProofBuilder`**: `buildPayload` now includes the `ath` claim (`base64url(SHA-256(accessToken))`) when `accessToken` is non-null/non-empty, and the `nonce` claim when non-null/non-empty (was silently ignored before)
- **`RestClient`**: new `attachDPoPProofHeader` helper attaches a DPoP proof on every outgoing API request when `tokenType == "DPoP"`; `OAuthRefreshInterceptor.buildAuthenticatedRequest` also calls `attachDPoPProofIfNeeded` so the 401-retry path gets a fresh proof too; `credentialsIdentifier` threaded through all constructor chains
- **`ClientManager.peekRestClient`**: now passes `userAccount.getTokenType()` and `userAccount.getCredentialsIdentifier()` to `RestClient`
- **`OAuth2.callIdentityService`**: new overload attaches DPoP proof (with `ath`) for the identity endpoint when token is DPoP-bound; `TokenEndpointResponse` carries `credentialsIdentifier` for downstream callers; `makeTokenEndpointRequest` sets it on the response
- **`AuthenticationUtilities`**: `fetchUserIdentity` passes `tokenType` and `credentialsIdentifier` from the token response to the identity service call
- **`DPoPProofBuilderTest`**: 4 new `ath` claim tests (present, null, empty, deterministic)

## Test plan

- [ ] `DPoPProofBuilderTest` — all 11 tests pass (7 existing + 4 new `ath` tests)
- [ ] `DPoPKeyManagerTest` — all 5 tests pass (unchanged)
- [ ] `DPoPURLHelperTest` — all 4 tests pass (unchanged)
- [ ] Manual: log in with a DPoP-enabled org and confirm API calls carry `Authorization: DPoP <token>` and `DPoP: <proof>` headers with `ath` claim
- [ ] Manual: confirm Bearer orgs are unaffected (no `DPoP` header, `Authorization: Bearer <token>`)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -380,9 +407,32 @@ public Request buildRequest(RestRequest restRequest) {
                 builder.addHeader(entry.getKey(), entry.getValue());
             }
         }
+
+        // Attach DPoP proof header for DPoP-bound access tokens
+        final String authToken = oAuthRefreshInterceptor.getAuthToken();
+        final String tokenType = oAuthRefreshInterceptor.tokenType;
+        final String credentialsIdentifier = oAuthRefreshInterceptor.credentialsIdentifier;
+        attachDPoPProofHeader(builder, restRequest.getMethod().toString(), url.toString(), authToken, tokenType, credentialsIdentifier);
```
JohnsonEricAtSalesforce: The DPoP proof looks like it gets attached twice on every authenticated request. `buildRequest()` stamps it here, but every request also flows through `OAuthRefreshInterceptor.buildAuthenticatedRequest()`, which calls `attachDPoPProofIfNeeded()` and does `builder.header(DPoP, proof)` — and since `header()` replaces rather than appends, the proof built here is discarded and re-signed. That's two EC P-256 keystore signings per request, the first one thrown away.

I traced the send paths to confirm nothing relies on this copy: `sendAsync` and both `sendSync` overloads dispatch through the cached `okHttpClient` (which has the interceptor registered via `addInterceptor`), and even the WebSocket path (`newWebSocket()`, also fed by `buildRequest()`) runs application interceptors on the upgrade call. The only non-intercepted client is the `"unauthenticated"` one, which is never DPoP-bound.

The interceptor copy is the one to keep — the 401-retry path re-runs `buildAuthenticatedRequest()` after refresh, so the proof must be rebuilt there for a fresh `jti`/`iat`. Suggest dropping the `attachDPoPProofHeader` call (and the private helper) from `buildRequest()` and relying solely on the interceptor. If both are intentional, maybe collapse them into one shared helper so they can't drift.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -598,8 +598,42 @@ public static IdServiceResponse callIdentityService(HttpAccess httpAccessor,
...
+        addAuthorizationHeader(builder, authToken, tokenType);
+        if (DPOP.equals(tokenType) && credentialsIdentifier != null && SalesforceSDKManager.getInstance().isUseDPoP()) {
+            try {
+                final String htu = DPoPURLHelper.INSTANCE.canonicalize(identityServiceIdUrl);
+                final String alias = DPoPKeyManager.INSTANCE.aliasForCredentialsIdentifier(credentialsIdentifier);
+                final KeyPair keyPair = DPoPKeyManager.INSTANCE.generateOrLoadKeyPair(alias);
+                final String proof = DPoPProofBuilder.INSTANCE.buildProof("GET", htu, keyPair, null, authToken);
+                builder.header(DPOP, proof);
+            } catch (Exception e) {
+                SalesforceSDKLogger.e(TAG, "Failed to attach DPoP header for identity service, proceeding without it", e);
```
JohnsonEricAtSalesforce: Small cross-platform consistency question on the identity-service path. If proof attachment throws here we log and proceed — but the request still carries `Authorization: DPoP <token>` with no `DPoP` proof header, which the server will reject, so the fallback can't really succeed.

iOS made the identity coordinator (`SFIdentityCoordinator`) fail-closed for exactly this in commit `97a62410a` — it bails out and surfaces the error rather than sending an unusable request. Worth noting the resource-API path is fine as-is: both platforms are fail-open there (iOS `SFRestRequest` logs and proceeds too), so this is just about the identity call site matching iOS.

Not asserting a change — this touches auth/credential handling, so flagging for a conscious call: either match iOS and surface the failure on the identity path, or keep a uniform fail-open policy across all DPoP paths and drop a one-line comment noting it's intentional. What's your preference?

JohnsonEricAtSalesforce (changes requested):
**Phase 3 DPoP review.** The core crypto is solid — `ath` is computed as `base64url(SHA-256(access_token))` byte-for-byte identically to iOS, base64url flags match, and the null/empty guards are right. I verified iOS Phase 3 parity (`DPoPRequestDecorator` / PR 4060) and ran the DPoP instrumented suite locally (DPoPProofBuilderTest 11/11, DPoPKeyManagerTest 5/5, DPoPURLHelperTest 4/4). This also closes the Phase 2 "unused `ath`/`nonce` params" thread. 🎉

Requesting changes on one item — a redundant double proof-attach on the primary REST path (each request signs twice and discards the first). Plus a minor cross-platform consistency question on the identity-service fail-open behavior; not a blocker, just flagging for a conscious decision.

JohnsonEricAtSalesforce (commented):
Thanks for the quick turnaround, @wmathurin. Confirmed both items are addressed: the double proof-attach is gone (`buildRequest()` now defers entirely to the interceptor — 👍), and `callIdentityService` is fail-closed with the iOS `97a62410a` reference. Nice catch on the AccountManager `credentialsIdentifier`/`tokenType` persistence gap too, and the round-trip regression tests for it pass locally.

One blocker before this is green: I ran the SalesforceSDK instrumented suite at this head and `UserAccountTest.testConvertAccountToBundle` and `testConvertAccountToJSON` both fail — the shared test-account builder now populates the two new fields, but the expected Bundle/JSON in those two tests weren't updated to match. Details inline. Everything else (DPoP suites + the new `UserAccountManagerTest` round-trips) passes 45/47.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -522,6 +524,8 @@ public static UserAccount createTestAccount() {
                 .beaconChildConsumerKey(TEST_BEACON_CHILD_CONSUMER_KEY)
                 .beaconChildConsumerSecret(TEST_BEACON_CHILD_CONSUMER_SECRET)
                 .scope(TEST_SCOPE)
+                .credentialsIdentifier(TEST_CREDENTIALS_IDENTIFIER)
```
JohnsonEricAtSalesforce, line 531: Adding `credentialsIdentifier` / `tokenType` to this shared `createTestAccount()` builder means every account built here now serializes those two keys — which breaks two existing tests that compare against hardcoded expected values:

- `testConvertAccountToBundle` (line 131) → `createTestAccountBundle()` (line 442)
- `testConvertAccountToJSON` (line 159) → `createTestAccountJSON()` (line 396)

Both fail locally at this head with e.g. *"Key credentialsIdentifier is missing in the expected bundle"* — the assertion sees `credentialsIdentifier`/`tokenType` on the `actual` side only. To get the suite green, add the two keys to both expected builders, mirroring how `scope` is set:

```java
// in createTestAccountJSON() (after UserAccount.SCOPE, ~line 430):
object.put(UserAccount.CREDENTIALS_IDENTIFIER, TEST_CREDENTIALS_IDENTIFIER);
object.put(UserAccount.TOKEN_TYPE, TEST_TOKEN_TYPE);

// in createTestAccountBundle() (after UserAccount.SCOPE, ~line 480):
object.putString(UserAccount.CREDENTIALS_IDENTIFIER, TEST_CREDENTIALS_IDENTIFIER);
object.putString(UserAccount.TOKEN_TYPE, TEST_TOKEN_TYPE);
```

(`UserAccount.CREDENTIALS_IDENTIFIER` / `TOKEN_TYPE` are the existing key constants at `UserAccount.java:108-109`.) Verified locally that these are the only two hardcoded-expected serialization tests affected — the new `UserAccountManagerTest` round-trips build both sides from the same source, so they already pass.

JohnsonEricAtSalesforce:
@wmathurin — a heads-up on CI coverage for the DPoP rollout, and a small ask.

The phase PRs targeting the `dpop` integration branch don't appear to be running the **Pull Request** test workflow. On this PR (head `d26cb65b`) the only checks are `SAST` and `salesforce-cla` — no Firebase Test Lab / instrumented-test job. Same story on the prior phase PR 2944. So each phase is merging into `dpop` without the per-PR unit/instrumented coverage and codecov gate that PRs into `dev` get.

The config *looks* right: the `pr.yaml` on the `dpop` branch already lists `dpop` in the `pull_request_target` allowlist (the temporary entry you added in `1b1ef00a3`). But because the repo's default branch is `dev`, and `dev`'s `pr.yaml` allowlist is still just `[dev, master]`, I think GitHub is evaluating the `pull_request_target` trigger against the **default-branch** copy of the workflow — which doesn't include `dpop` — so the event never matches and no run is created. That would explain why there's not even a skipped/awaiting-approval run to point at.

If that's the cause, the fix is to add `dpop` to the allowlist on the **`dev`** copy of `.github/workflows/pr.yaml` (same temporary-entry treatment, removed when DPoP merges back to `dev`), rather than only on the `dpop` branch. Happy to open that one-line PR against `dev` if you'd like.

Worth doing now rather than at merge-back because this isn't the last phase — there are still three Android PRs queued to target `dpop` after this one: **W-22697878** (nonce support), **W-22698013** (automated UI tests), and **W-23192897** (dev-info-screen DPoP state), all still open. So the coverage gap compounds across every remaining phase, and Phase 4's nonce challenge/response is exactly the kind of protocol-state logic where per-PR CI earns its keep.

As a concrete example of the debt: locally I found Phase 3 shipped with two red serialization tests (`UserAccountTest.testConvertAccountToBundle` / `testConvertAccountToJSON`) that CI would have caught immediately — per-PR CI on `dpop` would surface this class of thing at the point it's introduced instead of all at once when `dpop` lands on `dev`.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
wmathurin: Fixed in commit `f980913b5` — added `credentialsIdentifier` and `tokenType` to both `createTestAccountJSON()` and `createTestAccountBundle()`. `testConvertAccountToBundle` and `testConvertAccountToJSON` now pass.

wmathurin:
Thanks for flagging this — you're right about the root cause. Opened #2950 which adds `dpop` to the `branches` allowlist on the `dev` copy of `pr.yaml`. Once that merges, all remaining phase PRs targeting `dpop` (W-22697878, W-22698013, W-23192897) will get full CI.

JohnsonEricAtSalesforce:
@wmathurin — once the CI-enable change (2950) merges to `dev`, this PR won't pick up the Pull Request workflow on its own: `pull_request_target` is evaluated against the `dev` copy of `pr.yaml` at event time, and there's no prior run here to re-trigger. A quick **Close → Re-Open** on this PR after that merges should fire a fresh `reopened` event and kick off the full test suite (Firebase Test Lab + codecov + Danger) against the current head. A new commit/push would do the same if you have one handy — either works.

Would appreciate the close/re-open once the CI PR is in, so we get a green (or informative) CI run here before this lands. Thanks!

wmathurin:
> One blocker before this is green: I ran the SalesforceSDK instrumented suite at this head and `UserAccountTest.testConvertAccountToBundle` and `testConvertAccountToJSON` both fail — the shared test-account builder now populates the two new fields, but the expected Bundle/JSON in those two tests weren't updated to match. Details inline. Everything else (DPoP suites + the new `UserAccountManagerTest` round-trips) passes 45/47.

Should be addressed by https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2949/commits/223fdc3087c3cc4ac354466392c14f44e3a50717

JohnsonEricAtSalesforce (approved):
Approving Phase 3. All three items from the earlier review are resolved and verified:

- **P2.1** (double proof-attach) — `buildRequest()` now defers entirely to the interceptor. 👍
- **P2.2** (identity-service fail-open) — `callIdentityService` is fail-closed with the iOS `97a62410a` reference.
- **P2.3** (serialization tests) — the `credentialsIdentifier`/`tokenType` fixtures are updated, and the broader DPoP test-debt in `LoginViewModelMockTest` / `AuthenticatorServiceTest` / `ClientManagerMockTest` is fixed too.

The full **SalesforceSDK unit suite is green** at this head (`0e1d6b8dd`).

On the red `ui-tests-pr` job: I pulled the AuthFlowTester results and it's an environmental/harness flake, not a Phase-3 issue. 49 of the failures are the identical `AssertionError: Timed out after 15000ms waiting for node: "user_creds_section"` across every login-flow class (Beacon/ECA/CA/BootConfig/Admin), which is the harness's credentials section never rendering — the `ath`/proof changes don't touch that path. Confirmed it's PR-independent: this same `user_creds_section` timeout shows up on unrelated PRs in the same window (e.g. the W-23195021 RTR PR), and the `ui-tests-pr` job flips green/red across PRs regardless of content (it passed on other PRs around the same time). Recommend a re-run of the UI job to clear it — happy to help dig if it reproduces after a re-run.

PR #2951, opened 2026-07-01
@W-19280949: [ASA] FY26Q2 ASA Mobile Shield (aka Mobile Bypass) [1.2] Sensitive Data Stored Unencrypted in Local Storage (Android)

JohnsonEricAtSalesforce:
### Problem

`PushService.enqueuePushNotificationsRegistrationWork` serialized the full `UserAccount` (via `userAccount.toJson()`) into the push-registration worker's input `Data`. WorkManager persists worker input `Data` unencrypted in its Room database (`androidx.work.workdb`), so the user's **auth token, refresh token, sid, and session cookies were written to disk in plaintext**.

### Fix

**Data minimization.** The enqueuer now persists only the non-sensitive `ORG_ID` and `USER_ID`. The worker re-resolves the full `UserAccount` from secure `AccountManager` storage via `UserAccountManager.getUserFromOrgAndUserId(orgId, userId)`. No sensitive token or cookie material is ever written to WorkManager's unencrypted store.

- Absent identifiers still specify "all authenticated users."
- Present identifiers that no longer resolve (e.g. the user logged out before the work ran) fail the work rather than silently widening scope; the SDK re-enqueues registration work with `REPLACE` on the next foreground/login, so a discarded worker is automatically replaced.

**Enqueue-after-persist ordering.** Because the worker now re-resolves from `AccountManager`, push registration must be enqueued only after the account has been persisted. On login, registration was previously enqueued from within the `addAccount` helper — before `onAuthFlowComplete` persisted the account via `createAccount`/`persistAccount` — so the worker ran before the account was written and re-resolution failed (registration recovered only on a later foreground re-enqueue). The registration call is moved out of the `addAccount` helper and invoked in `onAuthFlowComplete` after persistence on both the normal (`createAccount`) and token-migration (`persistAccount`) branches. It is injected as `registerForPush` so the ordering is unit-testable.

### Testing

- **Unit/instrumented:** `PushNotificationsRegistrationChangeWorkerTest` covers unresolvable → failure, absent → success (all users), resolvable → success (re-resolution), missing action → failure. `AuthenticationUtilitiesTest` adds a regression test asserting the account is persisted before push registration is enqueued (`verifyOrder`). Full `AuthenticationUtilitiesTest` suite passes (32/32).
- **On-device (manual):** Verified on an emulator across three full login → push-receipt → logout cycles: push registration resolves the account on the first pass, push messages are received, and the logout deregistration re-resolves the account before it is removed.

### Notes

The first commit (`abafde4a4`) is a preserved encrypt-in-transit fallback; the subsequent commits supersede it with the data-minimization approach, which avoids persisting sensitive material entirely.


JohnsonEricAtSalesforce:
**Summary of changes since initial post**

The fix evolved from the original approach into a smaller, safer change:

- **Data minimization (core fix).** Instead of encrypting the full credentials in WorkManager data, we now enqueue only the non-sensitive `orgId` + `userId`. The worker re-resolves the full `UserAccount` from `AccountManager`, so no auth token, refresh token, sid, or cookies are ever written to local storage.
- **Login ordering fix.** Push registration is now enqueued *after* the account is persisted. Previously it could be enqueued before persistence, so the worker's re-resolution raced account creation and registration failed at login. Verified on-device across multiple logout/login/push cycles.
- **Simplification.** Removed an injectable test-only seam and inlined the registration call, keeping the production path straightforward.

**Note on the last CI run:** the failure on the previous commit was a MockK signature mismatch in `LoginViewModelMockTest` caused by an interim parameter that has since been removed. Verified locally: `LoginViewModelMockTest` (22 tests) and `AuthenticationUtilitiesTest` (31 tests) both pass.

Holding reviews for now — **we'll mark this ready for review once CI is green.**

wmathurin (approved):
LGTM

libs/SalesforceSDK/src/com/salesforce/androidsdk/push/PushNotificationsRegistrationChangeWorker.kt
```diff
@@ -62,9 +60,27 @@ internal class PushNotificationsRegistrationChangeWorker(
...
+         * This is safe because the SDK re-enqueues registration work with
+         * REPLACE on the next app foreground or login, so a discarded worker is
+         * automatically replaced.
+         */
+        val orgId = inputData.getString("ORG_ID")
+        val userId = inputData.getString("USER_ID")
+        val userAccountManager = SalesforceSDKManager.getInstance().userAccountManager
+        val userAccount = if (orgId == null && userId == null) {
+            null /* User account is optional where null specifies all accounts */
+        } else {
```
brandonpage, line 81: @JohnsonEricAtSalesforce What about the upgrade scenario?  

If the user logged in before 14.0 and the app has `ReRegisterPeriodically` set then the worker will get here with `USER_ACCOUNT` data (not org and id) so all users will be re-registered?

JohnsonEricAtSalesforce: Thanks for looking! We'll file a follow-up to investigate.

PR #2955, opened 2026-07-02
DPoP Phase 4: nonce support (W-22697878)

wmathurin:
## Summary

Phase 4 of the Android DPoP implementation: **nonce support** (W-22697878), mirroring the iOS SDK.

**Key design decision (confirmed by DPoP implementation team):** nonces are issued exclusively by the `/token` endpoint — not by resource servers. This means:
- Only `OAuth2.java` (token exchange, token refresh, identity endpoint) harvests `DPoP-Nonce` headers and retries on `use_dpop_nonce` challenges.
- `OAuthRefreshInterceptor` only *reads* the cached nonce proactively when building DPoP proofs — it never writes to the cache and never retries on nonce challenges.

### New class: `DPoPNonceCache`

Thread-safe singleton (`ConcurrentHashMap`) keyed by `(credentialsIdentifier, host)`. Populated exclusively by `OAuth2.java`; read by `attachDPoPProofIfNeeded()` in the interceptor on every API call.

### `OAuth2.java` — token endpoint nonce flow

1. Include cached nonce in initial DPoP proof (proactive, avoids round-trip on warm path).
2. Harvest `DPoP-Nonce` from every response.
3. On `use_dpop_nonce` (400/401): nonce already cached; rebuild proof; retry once; fail-closed.

Same pattern applied to `callIdentityService()`.

### `RestClient.OAuthRefreshInterceptor`

`attachDPoPProofIfNeeded()` reads the cached nonce when building the proof. No harvest, no retry — resource-server 401s fall through to the normal token-refresh path (which already handles nonce via `OAuth2.java`).

### Tests

- `DPoPNonceCacheTest` (114 lines): store/get/clear, per-host isolation, thread-safety
- `OAuthRefreshInterceptorNonceTest` (4 tests): proactive nonce use, no-harvest contract, Bearer passthrough
- `ClientManagerMockTest`: one-line fix — `registerUsedAppFeature(String, UserAccount?)` returns `Unit`; stub changed from `returns true` to `just runs`

All 29 DPoP tests pass.

## Test plan

- [ ] `./gradlew :libs:SalesforceSDK:connectedDebugAndroidTest` compiles cleanly
- [ ] DPoPKeyManagerTest, DPoPProofBuilderTest, DPoPURLHelperTest (20 tests) pass
- [ ] DPoPNonceCacheTest passes
- [ ] OAuthRefreshInterceptorNonceTest passes

> **Note:** `testGetNewAuthToken_InterruptedLoser_ReturnsCleanlyWithoutWinnerBody` crashes with a native SIGABRT — confirmed pre-existing on upstream `dev` and `dpop` branches, not introduced here.

### DIFF
--- a/docs/auth/token-lifecycle.md
+++ b/docs/auth/token-lifecycle.md
@@ -1,0 +1,323 @@
+# Token Lifecycle: Code Exchange, Refresh, DPoP, and RTR
+
+This document describes how the Android SDK acquires and renews OAuth tokens, how DPoP
+proof JWTs are attached, how nonces are managed, and how Refresh Token Rotation (RTR) is
+handled safely under concurrency.
+
+The primary classes are:
+- **`OAuth2.java`** — all calls to the Salesforce token and identity endpoints
+- **`ClientManager.java`** (`AccMgrAuthTokenProvider`) — RTR-safe token refresh coordination
+- **`RestClient.java`** (`OAuthRefreshInterceptor`) — per-request auth header attachment and
+  automatic refresh on 401
+- **`DPoPProofBuilder.kt`**, **`DPoPKeyManager.kt`**, **`DPoPURLHelper.kt`**,
+  **`DPoPNonceCache.kt`** — DPoP proof generation and nonce caching
+
+---
+
+## 1. Code Exchange (initial login)
+
+**Entry point:** `OAuth2.exchangeCode()`
+
+After the user completes the OAuth2 authorization code flow in the login WebView,
+`LoginActivity` calls `exchangeCode()` with the authorization `code` and `code_verifier`
+(PKCE). This method:
+
+1. Builds a `POST /services/oauth2/token` form body with `grant_type=authorization_code`
+   (or `hybrid_auth_code` when hybrid authentication is enabled).
+2. If `isUseDPoP()` is `true` and a `credentialsIdentifier` was generated at login start:
+   - Looks up (or generates) the EC keypair via `DPoPKeyManager.generateOrLoadKeyPair()`.
+   - Proactively reads any cached nonce via `DPoPNonceCache.get(credentialsIdentifier, tokenHost)` — null on first ever login.
+   - Builds a DPoP proof JWT via `DPoPProofBuilder.buildProof()` and attaches it as the `DPoP` header.
+3. Sends the request via `makeTokenEndpointRequest()` (shared with refresh).
+4. On success: the response carries an `access_token`, `refresh_token`, and `token_type`
+   (`"DPoP"` or `"Bearer"`). A `DPoP-Nonce` header may also be present — it is harvested
+   into the cache immediately.
+5. On `use_dpop_nonce` (400/401 with that error body): the nonce was just harvested (step 4);
+   rebuild the proof with it and retry once. Fail-closed on a second nonce failure.
+6. `TokenEndpointResponse` is returned to `LoginActivity`, which persists the tokens into
+   `AccountManager` and calls `callIdentityService()` to fetch the user's identity record.
+
+**DPoP note:** the `credentialsIdentifier` is a UUID generated at login start and stored on
+`UserAccount` as the `dpopScope` AccountManager extra. It is the stable key for the DPoP
+keypair and nonce cache for this user's session.
+
+---
+
+## 2. Token Refresh
+
+**Entry point:** `OAuth2.refreshAuthToken()`
+
+When an API call returns 401, `OAuthRefreshInterceptor` calls `refreshAccessToken()` which
+delegates to `AccMgrAuthTokenProvider.getNewAuthToken()`. That method:
+
+1. **Matches the account** by scanning `AccountManager` for the one whose stored refresh token
+   equals this provider's `refreshToken`. Fails early (returns null) if the account has been
+   removed — this ensures a removed-account path never accidentally sets `refreshing = true`
+   and deadlocks waiting threads.
+
+2. **Acquires the RTR lock** — see §4 below.
+
+3. **Recheck-under-lock guardrail:** before making a network call, re-reads the current tokens
+   from `AccountManager`. If either the access token or the refresh token in storage has
+   already advanced past what this provider last used, a concurrent winner already refreshed —
+   adopt their result without a redundant network POST. This is a correctness guardrail under
+   RTR, not an optimisation: every needless POST rotates the refresh token and widens the
+   stale-token logout window.
+
+4. **`refreshStaleToken()`** → **`OAuth2.refreshAuthToken()`** builds a
+   `POST /services/oauth2/token` form body with `grant_type=refresh_token`. DPoP proof
+   and nonce handling here are identical to code exchange: proactive nonce inclusion,
+   harvest from response, retry-once on `use_dpop_nonce`.
+
+5. On success: broadcasts `ACCESS_TOKEN_REFRESH_INTENT` (or `INSTANCE_URL_UPDATE_INTENT` if
+   the instance URL changed), publishes the new tokens to the per-account `RefreshState`,
+   and wakes waiting losers.
+
+6. On terminal failure (`invalid_grant`, `client_blocked`): broadcasts
+   `ACCESS_TOKEN_REVOKE_INTENT` and calls `SalesforceSDKManager.logout()`.
+
+---
+
+## 3. DPoP Proof Attachment
+
+Every outgoing API call goes through `OAuthRefreshInterceptor.buildAuthenticatedRequest()`,
+which calls `attachDPoPProofIfNeeded()`:
+
+```
+attachDPoPProofIfNeeded(builder, method, url):
+  if tokenType != "DPoP"  →  return (Bearer path, unmodified)
+  if !isUseDPoP()         →  return
+  if credentialsIdentifier == null  →  return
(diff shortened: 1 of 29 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestClient.java
```diff
@@ -913,12 +914,14 @@ private void attachDPoPProofIfNeeded(Request.Builder builder, String method, Str
             if (credentialsIdentifier == null || credentialsIdentifier.isEmpty()) return;
             try {
                 final String htu = DPoPURLHelper.INSTANCE.canonicalize(url);
+                final String host = HttpUrl.get(url).host();
                 final String alias = DPoPKeyManager.INSTANCE.aliasForCredentialsIdentifier(credentialsIdentifier);
                 final java.security.KeyPair keyPair = DPoPKeyManager.INSTANCE.generateOrLoadKeyPair(alias);
-                final String proof = DPoPProofBuilder.INSTANCE.buildProof(method, htu, keyPair, null, authToken);
+                final String nonce = DPoPNonceCache.INSTANCE.get(credentialsIdentifier, host);
```
JohnsonEricAtSalesforce, line 920: The nonce cache is keyed by `(credentialsIdentifier, host)` with a strict host match, whereas iOS keys by `(htu, scope)` **and** provides a `latest(forScope:)` fallback so a resource-server call reuses the token-endpoint nonce even when it was harvested under a different host. iOS documents this as deliberately avoiding a `use_dpop_nonce` round-trip on the first call.

On Android, the initial code-exchange harvests its nonce under the **login** host, so the first API call after login (before any refresh) reads under the **instance** host and misses — the proof goes out with no `nonce` claim. This is benign under the current Salesforce contract (resource servers never challenge, so there's nothing to satisfy), and the steady-state refresh path caches under the instance host correctly. But it's a real divergence from the iOS design.

Question: is the host-based keying (vs. iOS's htu + `latest(scope)` fallback) intentional? And is the login-vs-instance host split for community users (communityUrl host) accounted for? Happy to be told this is by design — just want to confirm it was a conscious choice rather than an accidental asymmetry.

*This comment was generated by an AI agent on behalf of @JohnsonEricAtSalesforce.*

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -706,15 +736,42 @@ public static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAcce
...
                 SalesforceSDKLogger.e(TAG, "Failed to attach DPoP header, proceeding without it", e);
             }
         }
 
         final Request request = requestBuilder.build();
-        final Response response = httpAccessor.getOkHttpClient().newCall(request).execute();
+        Response response = httpAccessor.getOkHttpClient().newCall(request).execute();
+
+        // Harvest nonce from every response (proactive caching for next call).
+        if (credentialsIdentifier != null) {
```
JohnsonEricAtSalesforce: The nonce **harvest** block gates only on `credentialsIdentifier != null`, while the attach and retry blocks additionally gate on `isUseDPoP()` and `DPOP.equals(tokenType)`. It's harmless (the `store` is a no-op when no `DPoP-Nonce` header is present), but the asymmetry is a little surprising to read. Same applies to the identity-service harvest at line 644. Consider aligning the guards, or a one-line comment noting the harvest is intentionally unconditional.

*This comment was generated by an AI agent on behalf of @JohnsonEricAtSalesforce.*

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -623,18 +625,45 @@ public static IdServiceResponse callIdentityService(HttpAccess httpAccessor,
...
+
+        // Harvest nonce from every response.
+        if (credentialsIdentifier != null) {
+            final String responseNonce = response.header("DPoP-Nonce");
+            if (responseNonce != null && !responseNonce.isEmpty()) {
+                DPoPNonceCache.INSTANCE.store(credentialsIdentifier, identityHost, responseNonce);
+            }
+        }
+
+        // Nonce challenge: server requires a nonce. Retry once (fail-closed on second failure).
```
JohnsonEricAtSalesforce, line 655: The nonce-retry block in `callIdentityService` isn't wrapped in try/catch, unlike the corresponding retry in `makeTokenEndpointRequest` (which catches `Exception`). If `buildProof` / keystore access throws a `RuntimeException` here, it would propagate out of a method that only declares `IOException`. Wrapping this block to match the token-endpoint path would keep the two nonce flows symmetric.

*This comment was generated by an AI agent on behalf of @JohnsonEricAtSalesforce.*

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -706,15 +736,42 @@ public static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAcce
...
         }
 
         final Request request = requestBuilder.build();
-        final Response response = httpAccessor.getOkHttpClient().newCall(request).execute();
+        Response response = httpAccessor.getOkHttpClient().newCall(request).execute();
+
+        // Harvest nonce from every response (proactive caching for next call).
+        if (credentialsIdentifier != null) {
+            final String responseNonce = response.header("DPoP-Nonce");
+            if (responseNonce != null && !responseNonce.isEmpty()) {
```
JohnsonEricAtSalesforce: `if (responseNonce != null && !responseNonce.isEmpty())` can use the idiom already established in this file. `OAuth2.java` imports `android.text.TextUtils` (line 30) and uses `TextUtils.isEmpty(...)` at lines 355, 1161, and 1231 — and it returns `true` for both `null` and `""`, so this is behaviorally identical:

```java
final String responseNonce = response.header("DPoP-Nonce");
if (!TextUtils.isEmpty(responseNonce)) {
    DPoPNonceCache.INSTANCE.store(credentialsIdentifier, tokenHost, responseNonce);
}
```

Applies to both harvest blocks (this one and the identity-service block around line 646). Purely a readability/consistency change, no behavior difference.

*This comment was generated by an AI agent on behalf of @JohnsonEricAtSalesforce.*

JohnsonEricAtSalesforce (approved):
**APPROVE** — Phase 4 nonce support looks solid and is consistent with the shipped iOS reference implementation (W-22697880).

Verified during review:
- **Design confirmed against iOS:** the "nonces are issued only by the `/token` endpoint, never by resource servers" contract is stated verbatim in iOS `DPoPNonceCache.swift`, so the interceptor's read-only / no-harvest / no-retry contract is correct and cross-platform-consistent.
- **Runtime trace:** in steady state, token refresh targets the instance host (via `overrideLoginServerIfNeeded`), which is the same host resource calls use — so a harvested nonce is cached where the interceptor reads it. The warm path works.
- **Tests:** all 29 DPoP instrumented tests pass locally at HEAD `9358ed2` on both API 35 and API 36 emulators. CI `unit-tests-pr (SalesforceSDK)` passes; the `ui-tests-pr` failure is the known AuthFlowTester environmental flake (same `Request Successful` assertion failing across many unrelated login suites that require live ECAs), not a regression from this change.

Four non-blocking comments below (one design question + three minor consistency points). None hold up the approval.

wmathurin:
## End-to-end test status

Created `DPoPLoginTests.kt` in AuthFlowTester to exercise the full flow without going through Login Options (bootconfig already points to `sdb6-2.test1.my.pc-rnd.salesforce.com`). The test does: login → REST API call → revoke → REST API call → revoke → REST API call.

**Current status: blocked on server-side deployment.**

The token endpoint works end-to-end — the client sends a proof, receives `use_dpop_nonce` 400, harvests the nonce, retries, and gets back `token_type: DPoP`. But every subsequent resource-server call returns:

```
HTTP 401 {"message":"INVALID_JWT_FORMAT","errorCode":"INVALID_AUTH_HEADER"}
WWW-Authenticate: Token
```

The `WWW-Authenticate: Token` (not `WWW-Authenticate: DPoP error=...`) is the key signal — this is a generic token scheme rejection, not a DPoP-specific proof validation error. The resource server on this org does not yet recognize `Authorization: DPoP <token>`.

**Root cause hypothesis: W-22953617** ("Backport DPoP and Central Auth") is the server-side story that backports `CentralAuthValidationService` with DPoP support to build 262. It is currently **In Progress**. The org `sdb6-2` appears to be on a build where the token endpoint already issues DPoP-bound tokens but the resource server hasn't yet been updated to validate them.

The client-side proof is cryptographically correct (verified offline with Node.js crypto): valid ES256 signature, correct `ath`, clean `htu` (fixed the `\/` JSON escaping in this commit), correct `nonce`. Nothing more to fix client-side until W-22953617 lands on this org.

Will re-run the UI test once W-22953617 is deployed to `sdb6-2`.

docs/auth/token-lifecycle.md
```diff
@@ -0,0 +1,323 @@
+# Token Lifecycle: Code Exchange, Refresh, DPoP, and RTR
```
brandonpage, line 1: This is awesome!

brandonpage (approved):

PR #2956, opened 2026-07-02
Add flag to force advanced authentication

brandonpage:
## Summary

Adds a new `forceAdvancedAuthentication` flag on `SalesforceSDKManager` that forces advanced (browser-based / Chrome Custom Tab) authentication for interactive login, regardless of the target server's My Domain auth-configuration — including standard login servers such as `login.salesforce.com`.

- **Flag ON (new default):** Advanced Auth is always used for interactive login. Welcome Discovery (phase 1) and non-My-Domain/standard endpoints behave as described below.
- **Flag OFF:** Legacy behavior — Advanced Auth is used only when the server's My Domain auth-configuration opts into it.

Apps that don't set the flag get the new "force on" default; apps needing the old behavior must explicitly set it OFF. The property is marked `@Deprecated` — it exists to bridge the transition and will be removed in 15.0 when WebView login is removed entirely.

## Changes

- **`SalesforceSDKManager`**: new `@Deprecated` `forceAdvancedAuthentication` property (defaults `true`). Reworked the auth-modality decision in `fetchAuthenticationConfiguration` so the flag overrides the My-Domain *enable* gate while still honoring the server's `shareBrowserSession` value where one exists. Standard login servers skip the auth-config fetch and gate browser login solely on the flag; the Welcome Discovery host and malformed servers are never forced. Added the flag to dev-support info.
- **`LoginOptionsActivity`**: new "Force Advanced Authentication" dev-menu toggle wired to the flag; made the composable's `SalesforceSDKManager` injectable so the Compose preview renders without the singleton.
- **`sf__strings.xml`**: one new string, `sf__login_options_force_advanced_auth_toggle_content_description`.
- **Tests**: new `SalesforceSDKManagerTests` cases covering the flag on/off × My Domain / standard / Welcome Discovery / non-HTTPS matrix; a `LoginViewModelTest` case asserting the custom tab always uses the Web Server flow (PKCE, never `response_type=token`); `LoginOptionsActivityTest` coverage for the new toggle. Also made the dev-menu `getDevActions` tests deterministic (stub the current user instead of depending on device account state).
- **AuthFlowTester**: instrumented-test harness updated so login paths drive the Chrome Custom Tab under the new default, with matching redirect intent filters in the manifest.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -1596,7 +1596,7 @@ open class SalesforceSDKManager protected constructor(
 //                "Use Web Server Authentication" to "$useWebServerAuthentication",
 //                "Use Hybrid Authentication Token" to "$useHybridAuthentication",
 //                "Support Welcome Discovery" to "$supportsWelcomeDiscovery",
-//                "Force Advanced Authentication", "$forceAdvancedAuthentication",
+//                "Force Advanced Authentication" to "$forceAdvancedAuthentication",
 //                "Browser Login Enabled" to "$isBrowserLoginEnabled",
 //                "IDP Enabled" to "$isIDPLoginFlowEnabled",
 //                "Identity Provider" to "$isIdentityProvider",
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -481,6 +481,11 @@ open class SalesforceSDKManager protected constructor(
     @set:Synchronized
     var useHybridAuthentication = true
 
+    // Backing field for [forceAdvancedAuthentication].  The SDK reads this directly so its own
+    // internal use of the flag doesn't trigger the deprecation warning on the public property.
+    @Volatile
+    private var _forceAdvancedAuthentication = true
+
     /**
      * Forces advanced (browser based) authentication to always be used for login, regardless of
      * the target server's auth configuration, including standard login servers with no My Domain.
@@ -489,7 +494,11 @@ open class SalesforceSDKManager protected constructor(
     @Deprecated("Will be removed in 15.0 when WebView login is removed entirely.")
     @get:JvmName("shouldForceAdvancedAuthentication")
     @set:Synchronized
-    var forceAdvancedAuthentication = true
+    var forceAdvancedAuthentication: Boolean
+        get() = _forceAdvancedAuthentication
+        set(value) {
+            _forceAdvancedAuthentication = value
+        }
 
     // Used to ensure the webview is reloaded when Dev Menu Login Options are changed.
     internal var loginDevMenuReload = false
@@ -1561,7 +1570,7 @@ open class SalesforceSDKManager protected constructor(
             "User Agent", userAgent,
             "Use Web Server Authentication", "$useWebServerAuthentication",
             "Use Hybrid Authentication Token", "$useHybridAuthentication",
-            "Force Advanced Authentication", "$forceAdvancedAuthentication",
+            "Force Advanced Authentication", "$_forceAdvancedAuthentication",
             "Browser Login Enabled", "$isBrowserLoginEnabled",
             "IDP Enabled", "$isIDPLoginFlowEnabled",
             "Identity Provider", "$isIdentityProvider",
@@ -1596,7 +1605,7 @@ open class SalesforceSDKManager protected constructor(
 //                "Use Web Server Authentication" to "$useWebServerAuthentication",
 //                "Use Hybrid Authentication Token" to "$useHybridAuthentication",
 //                "Support Welcome Discovery" to "$supportsWelcomeDiscovery",
-//                "Force Advanced Authentication" to "$forceAdvancedAuthentication",
+//                "Force Advanced Authentication" to "$_forceAdvancedAuthentication",
 //                "Browser Login Enabled" to "$isBrowserLoginEnabled",
 //                "IDP Enabled" to "$isIDPLoginFlowEnabled",
 //                "Identity Provider" to "$isIdentityProvider",
@@ -2167,7 +2176,7 @@ open class SalesforceSDKManager protected constructor(
                     // Standard login servers have no auth-config to source a shared-session value from, so
                     // browser login is gated solely on the force flag and shared session stays false.
                     setBrowserLoginEnabled(
-                        browserLoginEnabled = forceAdvancedAuthentication,
+                        browserLoginEnabled = _forceAdvancedAuthentication,
                         shareBrowserSessionEnabled = false
                     )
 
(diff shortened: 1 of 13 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -2134,28 +2070,42 @@ open class SalesforceSDKManager protected constructor(
...
+                    setBrowserLoginEnabled(
+                        browserLoginEnabled = forceAdvancedAuthentication || (authConfig?.isBrowserLoginEnabled ?: false),
+                        shareBrowserSessionEnabled = authConfig?.isShareBrowserSessionEnabled ?: false
+                    )
 
-                // Consider enabling Salesforce App Attestation for login servers that are My Domain servers.
-                appAttestationClient?.apiHostName = loginServer.toUri().host
+                    // Consider enabling Salesforce App Attestation for login servers that are My Domain servers.
+                    appAttestationClient?.apiHostName = loginServer.toUri().host
+                }
```
brandonpage, line 2194: This ~20 lines is just about the only production code change in the entire PR.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -480,6 +481,16 @@ open class SalesforceSDKManager protected constructor(
     @set:Synchronized
     var useHybridAuthentication = true
 
+    /**
```
wmathurin, line 489: Unrelated cleanup: `withTimeoutOrNull(5000L)` → `withTimeoutOrNull(5000L.milliseconds)` is a correct API upgrade (the `Duration`-based overload is preferred), but it adds an otherwise-unnecessary `kotlin.time.Duration.Companion.milliseconds` import and is not required for the feature. Fine to keep if intentional — just worth calling out for reviewers.

wmathurin (commented):
Three small items noted during review.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1584,6 +1596,7 @@ open class SalesforceSDKManager protected constructor(
 //                "Use Web Server Authentication" to "$useWebServerAuthentication",
 //                "Use Hybrid Authentication Token" to "$useHybridAuthentication",
 //                "Support Welcome Discovery" to "$supportsWelcomeDiscovery",
+//                "Force Advanced Authentication", "$forceAdvancedAuthentication",
```
wmathurin: Syntax inconsistency in the commented-out block: this entry uses `,` where every other entry uses `to`:

```kotlin
// All other entries:
//    "Use Web Server Authentication" to "$useWebServerAuthentication",
// This new entry:
//    "Force Advanced Authentication", "$forceAdvancedAuthentication",
```

Should be:
```kotlin
//                "Force Advanced Authentication" to "$forceAdvancedAuthentication",
```

Dead code, but looks wrong.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerTests.kt
```diff
@@ -318,6 +304,203 @@ class SalesforceSDKManagerTests {
...
+
+    @Test
+    fun fetchAuthenticationConfiguration_ForceFlagOffAndServerOptsIn_EnablesBrowserLogin() {
+
+        // With the force flag off, browser login follows the server's auth-config - the legacy
+        // opt-in path must continue to work.
+        SalesforceSDKManager.getInstance().forceAdvancedAuthentication = false
+
+        SalesforceSDKManager.getInstance().isBrowserLoginEnabled = false
+        SalesforceSDKManager.getInstance().isShareBrowserSessionEnabled = false
```
wmathurin, line 446: This test uses `setSelectedLoginServer` + no `loginServerUrl` override, while all other refactored tests in this file pass `loginServerUrl` directly. Both approaches work and teardown’s `loginServerManager.reset()` handles cleanup, but the inconsistency is a little surprising.

If the intent is specifically to exercise the `loginServerManager.selectedLoginServer` code path (i.e. the no-override branch of `fetchAuthenticationConfiguration`), a short comment explaining that would make it clear this was deliberate rather than accidental.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
brandonpage: Good catch, I just pasted that in from above so it wasn't forgotten when we did Dev Support Info upgrade work.  I'll update it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -480,6 +481,16 @@ open class SalesforceSDKManager protected constructor(
...
 
+    /**
+     * Forces advanced (browser based) authentication to always be used for login, regardless of
+     * the target server's auth configuration, including standard login servers with no My Domain.
+     * Defaults to true.
+     */
+    @Deprecated("Will be removed in 15.0 when WebView login is removed entirely.")
+    @get:JvmName("shouldForceAdvancedAuthentication")
+    @set:Synchronized
+    var forceAdvancedAuthentication = true
```
wmathurin: `@Deprecated` on a property the SDK itself reads internally will generate deprecation warnings in every build and test run. The standard pattern is to back it with a private field and use that internally:

```kotlin
@Volatile
private var _forceAdvancedAuthentication = true

@Deprecated("Will be removed in 15.0 when WebView login is removed entirely.")
@get:JvmName("shouldForceAdvancedAuthentication")
@set:Synchronized
var forceAdvancedAuthentication: Boolean
    get() = _forceAdvancedAuthentication
    set(value) { _forceAdvancedAuthentication = value }
```

`fetchAuthenticationConfiguration` (and any other internal call site) reads `_forceAdvancedAuthentication` with no warning. App callers who touch `forceAdvancedAuthentication` get the deprecation nudge as intended. Tests that need to flip the flag can either access the internal field or add `@Suppress("DEPRECATION")`.

wmathurin (approved):

JohnsonEricAtSalesforce (approved):
## ✅ Approve

Reviewed as a peer at HEAD `f38d1ea`. Approving on the code merits.

**How I arrived at approval:**
- **Traced the runtime path, not just the diff.** Confirmed the reworked auth-modality decision in `fetchAuthenticationConfiguration` matches all five acceptance criteria: Welcome-Discovery phase-1 bypass, phase-2 post-discovery My Domain honoring the flag, standard-server force, and `shareBrowserSession` always sourced from the server (the flag can't clobber it).
- **Verified the key security invariant.** The Chrome Custom Tab always launches from `browserCustomTabUrl`, which is built from the Web Server Flow authorization URL (`response_type=code` + PKCE). Forcing advanced auth can never downgrade the browser flow to the User-Agent/implicit flow — and the new `LoginViewModelTest` case locks that in.
- **Test coverage is thorough.** Flag on/off × My Domain (opt-in / opt-out / shared-session) / standard / Welcome-Discovery / non-HTTPS, default-true, the dev-menu toggle, and dev-support info are all covered. The `getDevActions` determinism fix (stubbing the current user) is a real improvement over depending on device account state.
- **Deprecation & localization are handled correctly.** The flag ships `@Deprecated` with a named 15.0 removal target (intentional transitional bridge), and the one new string carries the `sf__` prefix.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
brandonpage: This removes a warning by upgrading to the newer `Duration` API.  I don't think an extra import is of any concern...

brandonpage:
The UI Tests are failing due to a server error, I will continue to investigate and fix that separately.

PR #2963, opened 2026-07-10
W-23192897: surface DPoP runtime state in developer info screen

wmathurin:
## Summary
- Adds `jwkThumbprint(ECPublicKey)` to `DPoPProofBuilder` — RFC 7638 JWK SHA-256 thumbprint of the EC P-256 public key
- Extends `DevSupportInfo.parseUserInfoSection()` with three new rows in the Current User section:
  - **OAuth Token Type** — always shown (`"Bearer"` or `"DPoP"`)
  - **DPoP Nonce** — DPoP sessions only; shows cached nonce or `"None"`
  - **DPoP Key Thumbprint** — DPoP sessions only; RFC 7638 thumbprint matching the server-bound `jkt` claim, or `"Unavailable"` if key cannot be loaded
- Updates `createFromLegacyDevInfos()` to include the three new keys for backward-compat path
- Adds `docs/dev-info-screen.md` — reference doc for the developer info screen architecture

## Test plan
- [ ] Non-DPoP session: `OAuth Token Type = "Bearer"`, no DPoP Nonce or Thumbprint rows
- [ ] DPoP session: all three rows present; nonce shows `"None"` when cache empty; thumbprint shows `"Unavailable"` if key load fails
- [ ] `jwkThumbprint()` produces a valid 43-char base64url string
- [ ] All 21 `DevSupportInfoTest` tests pass (verified locally on emulator)
- [ ] Lint clean

### DIFF
--- a/docs/dev-info-screen.md
+++ b/docs/dev-info-screen.md
@@ -1,0 +1,191 @@
+# Developer Info Screen
+
+The Salesforce Mobile SDK ships a built-in developer / diagnostic screen accessible from the login UI overflow menu. It surfaces SDK configuration, the current user's OAuth state, and runtime configuration so developers can verify SDK behaviour without needing a debugger.
+
+---
+
+## Accessing the Screen
+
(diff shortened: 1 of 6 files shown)


sfdctaka:
We should remove `W-XXXX` from the title.

sfdctaka (approved):
LGTM!

PR #2965, opened 2026-07-13
feat(dpop): send dpop_jkt in /authorize for my-domain servers (W-23406836)

wmathurin:
### Core feature: \`dpop_jkt\` in \`/authorize\` (W-23406836)

- Eagerly generates the DPoP EC key pair in \`LoginViewModel.generateAuthorizationUrl()\` when \`useDPoP=true\`, so the RFC 7638 JWK thumbprint is available before the \`/authorize\` URL is built
- Adds \`dpop_jkt=<thumbprint>\` to the \`/authorize\` request query params (my-domain servers only — pool servers \`login.salesforce.com\`, \`test.salesforce.com\`, \`welcome.salesforce.com\` do not support DPoP code binding)
- Stores the generated \`credentialsIdentifier\` as \`pendingCredentialsIdentifier\` on the ViewModel so \`doCodeExchange()\` reuses the same key pair alias, preserving the binding from authorize → token exchange
- Applies the same fix to \`generateMigrationAuthorizationPath()\`

Without this fix the server rejects with:
\`\`\`
error=invalid_request&error_description=missing required dpop_jkt for code binding
\`\`\`

### Supporting refactors

- Extracted \`LoginServerManager.isPoolServer()\` helper to centralise the pool-server check used by the \`dpop_jkt\` guard
- Extracted \`addDpopJktIfNeeded()\` helper to remove duplication between the authorize and migration URL paths (previously \`maybeAddDpopJkt\`)
- Cleared stale \`dpop_jkt\` and deleted the orphaned AndroidKeyStore key when re-entering a server, preventing a key-alias mismatch on subsequent logins
- Fixed an accidental \`_forceAdvancedAuthentication=false\` regression introduced during refactoring

### AuthFlowTester improvements

- Added missing AndroidManifest intent filters for \`ecajwtrtr\`, \`ecaopaquertr\`, \`ecajwtdpop\`, and \`ecajwtdpoprtr\` redirect URI schemes — the four RTR/DPoP apps in \`ui_test_config.json\` had no filters, so OAuth callbacks would never be received
- Added full \`DPoPLoginTests\` suite covering: basic DPoP login (hybrid and non-hybrid), RTR, multi-user unique tokens, migration, restart, and Login-for-Admins with DPoP
- Fixed BW/DP/MU/RT flag assertions throughout \`AuthFlowTest\`: \`forceAdvancedAuthentication\` was not forwarded as \`expectAdvancedAuth\` in \`loginAndValidate\`, \`adminLoginAndValidate\`, \`assertRevokeAndRefreshWorks\`, \`migrateAndValidate\`, \`restartAndValidateUser\`, and \`switchToUserAndValidateUser\`
- Removed stale nonce-change assertion: server issues DPoP nonces with a 24 h TTL, not one per \`/token\` response
- Added \`forceAdvancedAuthentication\` parameter to \`loginAndValidate\` for tests that need the in-app WebView path
- Ignored \`testECAJwtDPoPRtr_Hybrid\`: blocked by the same W-22512846 server limitation (hybrid grant rejected when RTR + JWT enabled); \`testMigrate_ECAJwtDPoP_To_ECAJwtDPoPRtr\` uses \`useHybridAuthToken=false\` as a workaround for the same reason

## Test plan
- [x] All non-skipped tests in \`DPoPLoginTests\` pass locally on Android Studio
- [x] \`generateAuthorizationUrl_WhenUseDPoP_AddsDpopJkt_ToUrl\` — URL contains \`dpop_jkt=\` when \`useDPoP=true\`
- [x] \`generateAuthorizationUrl_WhenNotUseDPoP_DoesNotAddDpopJkt_ToUrl\` — no \`dpop_jkt\` when \`useDPoP=false\`
- [x] \`doCodeExchange_WhenPendingCredentialsIdentifierSet_ReusesIt\` — same identifier used in both authorize and code exchange
- [x] \`doCodeExchange_WhenNoPendingCredentialsIdentifier_GeneratesNew\` — fallback path unaffected
- [x] \`./gradlew :libs:SalesforceSDK:compileDebugKotlin\` — clean
- [x] \`./gradlew :libs:test:SalesforceSDKTest:compileDebugAndroidTestKotlin\` — clean

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -549,7 +549,7 @@ open class LoginViewModel(
 
         val additionalParameters = mutableMapOf<String, String>()
 
-        maybeAddDpopJkt(server, sdkManager, additionalParameters)
+        addDpopJktIfNeeded(server, sdkManager, additionalParameters)
 
         val authorizationUrl = OAuth2.getAuthorizationUrl(
             /* useWebServerAuthentication = */ true,
@@ -596,7 +596,7 @@ open class LoginViewModel(
             val codeVerifier = getRandom128ByteKey().also { codeVerifier = it }
             val codeChallenge = getSHA256Hash(codeVerifier)
 
-            maybeAddDpopJkt(server, sdkManager, additionalParams)
+            addDpopJktIfNeeded(server, sdkManager, additionalParams)
 
             val webServerAuthorizationUrl = OAuth2.getAuthorizationUrl(
                 /* useWebServerAuthentication = */ true,
@@ -728,7 +728,7 @@ open class LoginViewModel(
      * Pool servers (login.salesforce.com, test.salesforce.com, welcome.salesforce.com) do not
      * support DPoP code binding and reject the parameter.
      */
-    private fun maybeAddDpopJkt(
+    private fun addDpopJktIfNeeded(
         server: String,
         sdkManager: SalesforceSDKManager,
         params: MutableMap<String, String>,
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -736,8 +736,20 @@ open class LoginViewModel(
         val isMyDomainServer = server != PRODUCTION_LOGIN_URL
             && server != SANDBOX_LOGIN_URL
             && server != WELCOME_LOGIN_URL
-        if (!sdkManager.useDPoP || !isMyDomainServer) return
+        if (!sdkManager.useDPoP || !isMyDomainServer) {
+            // Clear any stale dpop_jkt and its key from a previous server-picker entry.
+            params.remove("dpop_jkt")
+            pendingCredentialsIdentifier?.let {
+                DPoPKeyManager.deleteKeyPair(DPoPKeyManager.aliasForCredentialsIdentifier(it))
+            }
+            pendingCredentialsIdentifier = null
+            return
+        }
         runCatching {
+            // Delete any orphaned key from a prior server-picker navigation before generating a new one.
+            pendingCredentialsIdentifier?.let {
+                DPoPKeyManager.deleteKeyPair(DPoPKeyManager.aliasForCredentialsIdentifier(it))
+            }
             val credId = java.util.UUID.randomUUID().toString()
             val alias = DPoPKeyManager.aliasForCredentialsIdentifier(credId)
             val keyPair = DPoPKeyManager.generateOrLoadKeyPair(alias)
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
@@ -61,9 +61,14 @@ import com.salesforce.androidsdk.auth.OAuth2.TokenEndpointResponse
 import com.salesforce.androidsdk.auth.OAuth2.exchangeCode
 import com.salesforce.androidsdk.auth.OAuth2.getFrontdoorUrl
 import com.salesforce.androidsdk.auth.defaultBuildAccountName
+import com.salesforce.androidsdk.auth.dpop.DPoPKeyManager
+import com.salesforce.androidsdk.auth.dpop.DPoPProofBuilder
 import com.salesforce.androidsdk.auth.onAuthFlowComplete
 import com.salesforce.androidsdk.config.BootConfig
 import com.salesforce.androidsdk.config.LoginServerManager.LoginServer
+import com.salesforce.androidsdk.config.LoginServerManager.PRODUCTION_LOGIN_URL
+import com.salesforce.androidsdk.config.LoginServerManager.SANDBOX_LOGIN_URL
+import com.salesforce.androidsdk.config.LoginServerManager.WELCOME_LOGIN_URL
 import com.salesforce.androidsdk.config.OAuthConfig
 import com.salesforce.androidsdk.config.RuntimeConfig.ConfigKey.OnlyShowAuthorizedHosts
 import com.salesforce.androidsdk.config.RuntimeConfig.getRuntimeConfig
@@ -80,6 +85,7 @@ import kotlinx.coroutines.withContext
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import java.net.URI
 import java.net.URLEncoder
+import java.security.interfaces.ECPublicKey
 import kotlin.coroutines.CoroutineContext
 
 /**
@@ -201,6 +207,10 @@ open class LoginViewModel(
 
     /** Additional Auth Values used for login. */
     open var additionalParameters = hashMapOf<String, String>()
+
+    /** Credentials identifier pre-generated for a pending DPoP login; reused in doCodeExchange(). */
+    @VisibleForTesting
+    internal var pendingCredentialsIdentifier: String? = null
 
     /** JWT string used for JWT Auth Flow. */
     var jwt: String? = null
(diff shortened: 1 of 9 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -708,6 +723,32 @@ open class LoginViewModel(
...
+        params: MutableMap<String, String>,
+    ) {
+        val isMyDomainServer = server != PRODUCTION_LOGIN_URL
+            && server != SANDBOX_LOGIN_URL
+            && server != WELCOME_LOGIN_URL
+        if (!sdkManager.useDPoP || !isMyDomainServer) return
+        runCatching {
+            val credId = java.util.UUID.randomUUID().toString()
+            val alias = DPoPKeyManager.aliasForCredentialsIdentifier(credId)
+            val keyPair = DPoPKeyManager.generateOrLoadKeyPair(alias)
```
sfdctaka, line 751: **Edge case — AndroidKeyStore alias leak on re-entry.** Every call to this helper mints a fresh UUID and persists a new EC keypair; `generateOrLoadKeyPair`'s existing-alias branch can never hit because the alias is brand-new. Only the last-computed `pendingCredentialsIdentifier` is consumed by `doCodeExchange` — earlier ones stay orphaned in AndroidKeyStore until app uninstall.

Realistic trigger: user opens login, picks a different my-domain host from the server picker, picks another, then logs in. Three keystore aliases persisted, two orphaned. Not catastrophic (P-256 entries are small), but unbounded over the app's lifetime and can hit StrongBox slot limits on some devices.

Cheapest fix would be: before overwriting `pendingCredentialsIdentifier`, call `DPoPKeyManager.deleteKeyPair(previousAlias)`. A cleaner but more invasive approach: compute the thumbprint from an ephemeral in-memory keypair here, and only persist to AndroidKeyStore in `doCodeExchange` once the auth code is in hand.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -708,6 +723,32 @@ open class LoginViewModel(
...
+     */
+    private fun addDpopJktIfNeeded(
+        server: String,
+        sdkManager: SalesforceSDKManager,
+        params: MutableMap<String, String>,
+    ) {
+        val isMyDomainServer = server != PRODUCTION_LOGIN_URL
+            && server != SANDBOX_LOGIN_URL
+            && server != WELCOME_LOGIN_URL
+        if (!sdkManager.useDPoP || !isMyDomainServer) return
```
sfdctaka: **Edge case — stale `dpop_jkt` can leak across re-entries.** The happy path a few lines below overwrites any previous value, so the common my-domain → my-domain case is fine. But when a call takes this early-return branch after a prior happy-path call left `dpop_jkt` in `additionalParameters` (the open `var` on line 209, shared across invocations), the stale value survives and is sent on the next `/authorize`.

Trigger: DPoP-enabled my-domain login → user backs out → picks `login.salesforce.com` (or session flips `useDPoP=false`). The pool-server `/authorize` then receives a `dpop_jkt` it doesn't support.

One-liner defense: `params.remove("dpop_jkt")` at the top of the function, before the guard. Consider also clearing `pendingCredentialsIdentifier` on the early-return path so the two stay in sync.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -669,7 +683,8 @@ open class LoginViewModel(
             }
             val verifier = if (isUsingFrontDoorBridge) frontdoorBridgeCodeVerifier else codeVerifier
 
-            val credentialsIdentifier = java.util.UUID.randomUUID().toString()
+            val credentialsIdentifier = pendingCredentialsIdentifier?.also { pendingCredentialsIdentifier = null }
```
sfdctaka, line 684: **Edge case — pending id is cleared before `exchangeCode` returns.** If `exchangeCode` throws (transient network / server hiccup) and the flow re-enters `doCodeExchange` with the same auth code, the second attempt falls through to the `?: UUID.randomUUID()` branch. That UUID doesn't correspond to the AndroidKeyStore alias bound to the `dpop_jkt` sent on the `/authorize` request, so the server-side code binding check will fail.

Whether this is reachable depends on the retry surface — if `doCodeExchange` is truly single-shot per auth-code, this is a non-issue. If any retry path can re-invoke it, moving the `pendingCredentialsIdentifier = null` assignment into the success path (or into `onAuthFlowComplete`) preserves the binding across retries.

sfdctaka (commented):
Non-blocking observations on four edge cases in `addDpopJktIfNeeded` and `doCodeExchange`. All happy-path flows look correct; these only bite under specific re-entry / retry sequences. Flagging for awareness, not requesting changes.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
wmathurin: Fixed in `a62f5b3c4`. Before generating a new keypair, we now call `DPoPKeyManager.deleteKeyPair(aliasForCredentialsIdentifier(it))` on the previous `pendingCredentialsIdentifier` — so repeated server-picker navigation only ever leaves one live entry. The same deletion also happens on the early-return path (pool server or DPoP disabled) to keep the keystore and `pendingCredentialsIdentifier` in sync.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
wmathurin: Also fixed in `a62f5b3c4`. The early-return branch now calls `params.remove("dpop_jkt")` and clears `pendingCredentialsIdentifier` (with keystore cleanup) before returning, so switching from a my-domain server to a pool server can't leave a stale `dpop_jkt` behind.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
wmathurin: `doCodeExchange` is single-shot in practice: it's invoked exactly once by `onWebServerFlowComplete`, which is called from the redirect handler that receives the auth code. The auth code is single-use on the server side, so retrying with the same code would fail at the token endpoint regardless. Leaving the clear on the `?.also { ... }` line for now — if a retry path is ever added, moving the null-out into `onAuthFlowComplete` (as you suggest) would be the right place.

sfdctaka (approved):
LGTM!

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -708,6 +723,44 @@ open class LoginViewModel(
...
+     * support DPoP code binding and reject the parameter.
+     */
+    private fun addDpopJktIfNeeded(
+        server: String,
+        sdkManager: SalesforceSDKManager,
+        params: MutableMap<String, String>,
+    ) {
+        val isMyDomainServer = server != PRODUCTION_LOGIN_URL
+            && server != SANDBOX_LOGIN_URL
+            && server != WELCOME_LOGIN_URL
```
brandonpage: NIT: I think we do this check in a couple places.  Could be worth creating a static utility somewhere (if we don't already have one?).

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
wmathurin: Done here: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2965/commits/f9b9d5b3157fc2a8070ea2dce2d86f4215ee6cac

wmathurin:
@brandonpage what do you think of https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2965/changes/7d3b035f4b5bb6ffe396916a099af07c7bc76685 ? (It probably could have gone to a separate PR though).

brandonpage:
> @brandonpage what do you think of [7d3b035](https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/7d3b035f4b5bb6ffe396916a099af07c7bc76685) ? (It probably could have gone to a separate PR though).

Seems fine.  I didn't want to over invest in opt-out WebView scenarios, but there is no harm in it.

PR #2971, opened 2026-07-17
fix(ManageSpaceActivity): call clearApplicationUserData() after logout

### LINKED ISSUE #425: Clear Data Functionality Not Working in ManageSpaceActivity
**Steps to Reproduce:**
1. Go to Android Settings > Apps > SFApp > Storage 
2. Tap "Clear Data" 
3. Observe the custom dialog 
4. Tap "Yes" to confirm clearing data 

**Expected Behaviour:** 
- The app data should be cleared 
- The app should either close or return to the login screen 
- The OS should handle the data clearing natively 

**Actual Behavior:** 
- The "Yes" button doesn't respond 
- No data is cleared 

Image

**Additional Context:** 
- SDK Version: 13.0.2 
- React Native Version: 0.74.7
- Android API Level: 35

wmathurin:
## Summary
- The \"Yes\" button in \`ManageSpaceActivity\` called \`logout(account=null)\` which only logs out the **current** user. With multiple accounts logged in, other accounts never received SDK-level teardown (SmartStore cleanup, DPoP key deletion, token revocation) before \`clearApplicationUserData()\` wiped everything at the OS level.
- Replaced the single-user logout with a loop over all authenticated users via \`UserAccountManager.getAuthenticatedUsers()\`, calling \`signoutUser()\` per account so each account goes through proper SDK cleanup.
- Falls back to \`logout(account=null)\` when \`getAuthenticatedUsers()\` returns null/empty (fresh install / already logged out).
- Extracted the confirm logic into \`logoutAllUsersAndClearData()\` with lambda parameters for testability, and \`clearApplicationUserData()\` as an \`open\` method so tests can override it without the OS killing the test process.

Fixes https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/425

## Test plan
- [x] Unit tests: \`ManageSpaceActivityTest\` (4 tests — single user, multiple users, null list, empty list) — all passing
- [ ] Build a React Native app with the fixed Android SDK
- [ ] Go to Android Settings > Apps > [app name] > Storage > Clear Data
- [ ] Tap \"Yes\" with a **single** account logged in — verify app data is cleared and app returns to login screen
- [ ] Tap \"Yes\" with **multiple** accounts logged in — verify all accounts are gone and app starts clean
- [ ] Tap \"No\" — verify app continues running normally

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ManageSpaceActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ManageSpaceActivity.kt
@@ -103,6 +103,7 @@ open class ManageSpaceActivity : ComponentActivity() {
                     showLoginPage = false,
                     reason = USER_LOGOUT
                 )
+                clearApplicationUserData()
             },
             titleText = stringResource(sf__manage_space_title),
             textText = stringResource(sf__manage_space_confirmation),
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ManageSpaceActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ManageSpaceActivity.kt
@@ -27,8 +27,6 @@
 package com.salesforce.androidsdk.ui
 
 import android.annotation.SuppressLint
-import android.app.ActivityManager
-import android.content.Context
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
@@ -55,6 +53,7 @@ import com.salesforce.androidsdk.R.string.sf__manage_space_confirmation
 import com.salesforce.androidsdk.R.string.sf__manage_space_logout_no
 import com.salesforce.androidsdk.R.string.sf__manage_space_logout_yes
 import com.salesforce.androidsdk.R.string.sf__manage_space_title
+import com.salesforce.androidsdk.accounts.UserAccount
 import com.salesforce.androidsdk.app.SalesforceSDKManager.Companion.getInstance
 import com.salesforce.androidsdk.auth.OAuth2.LogoutReason.USER_LOGOUT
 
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/ManageSpaceActivity.kt
```diff
@@ -103,6 +105,7 @@ open class ManageSpaceActivity : ComponentActivity() {
                     showLoginPage = false,
                     reason = USER_LOGOUT
                 )
+                (getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager).clearApplicationUserData()
```
brandonpage: What happens if there are multiple users logged in?  `account = null` will only result in the current user being logged out.  

And what do we not clear already on logout?

wmathurin: Good catch — you're right on both counts.

**Multiple users**: `logout(account=null)` resolves to `clientMgr.account` → the current user only. Other accounts never got SDK teardown before `clearApplicationUserData()` wiped their data at the OS level.

**Fix**: Now calls `UserAccountManager.getAuthenticatedUsers()` and loops, calling `signoutUser(user, activity, false, USER_LOGOUT)` per account so each one goes through proper SDK cleanup (SmartStore, DPoP key deletion, token revocation). Falls back to `logout(account=null)` only when the user list is null/empty (fresh install / already logged out).

**What logout() clears that clearApplicationUserData() would already wipe**: The OS call is a superset of SDK logout — it wipes all SharedPrefs, databases, files, caches, then restarts the process. Calling `signoutUser` first still matters for server-side cleanup (push un-registration, refresh token revocation) and to give the SDK a clean teardown path before the process is killed.

4 unit tests added: single user, 3 users, null list, empty list.

brandonpage (approved):

PR #2974, opened 2026-07-31
@W-23201591: [Android] Surface RTR state in developer info screen

JohnsonEricAtSalesforce:
## Summary

**Refresh Token Rotation (RTR)** is a server-driven security feature: when the token endpoint
issues a new refresh token in place of the one that was presented, the old refresh token is
retired. This pull request surfaces that RTR state in the Salesforce Mobile SDK **developer info
screen** (the on-device diagnostics screen reachable from the developer support menu) so that
developers can confirm, on-device, whether Refresh Token Rotation is active for the currently
logged-in user and when the refresh token last rotated.

It adds a new **"RTR"** section to the developer info screen with two rows:

- **RTR Active** — `true` or `false`, reflecting whether the Refresh Token Rotation per-user
  feature flag (an entry the SDK records per user under the code `RT`) is registered for the
  current user; shows `N/A` when no user is logged in.
- **Last Rotation** — the date and time of the most recent confirmed rotation, formatted as an
  ISO 8601 timestamp (the internationally standardized date-time format, for example
  `2026-07-30T12:34:56Z`); shows `Never` until the first rotation, or `N/A` when no user is
  logged in.

To make the rotation timestamp durable across app restarts, a new `lastTokenRotationTime` field
is persisted on the `UserAccount` object. It travels through the full account persistence path,
including the encrypt-on-write and decrypt-on-read handling used to store account data in
Android's `AccountManager`. The timestamp is stamped at the existing point in `ClientManager`
where a rotation is confirmed.

This work builds on the Refresh Token Rotation per-user feature flag (feature code `RT`) that
landed in work item W-23195021. It is the Android counterpart to the iOS work item W-23201588,
and uses the same field labels and display semantics.

Work item: **W-23201591**

## Acceptance criteria

- [x] **AC1** — Per-user fields show `N/A` when no user is logged in. *(Verified by unit test and on-device.)*
- [x] **AC2** — Before any token rotation, the section shows `RTR Active: false` and `Last Rotation: Never`. *(Verified by unit test and by the on-device demo below.)*
- [x] **AC3** — After a confirmed rotation, `RTR Active` shows `true` and `Last Rotation` shows a valid timestamp. *(Verified by unit test and by the on-device demo below.)*
- [x] **AC4** — The rotation timestamp survives an app restart, because it is persisted on the `UserAccount` object rather than held only in memory. *(Verified by an encrypt/decrypt round-trip unit test.)*

> **Demo:** A live end-to-end run against a Refresh Token Rotation–enabled org — showing the
> `false`/`Never` before-state, the Force Token Refresh action, and the resulting `true` +
> timestamp after-state — is attached in a follow-up comment on this pull request, along with
> before/after stills.

## Changes

**Production (7 files)**
- `UserAccount.java` — new `lastTokenRotationTime` field carried across the full persistence path (the string constant, the field, the JSON constructor, the `Bundle` constructor, the getter and setter, `toJson`, and `toBundle`), mirroring how the existing `tokenType` field is handled.
- `UserAccountBuilder.kt` — a `lastTokenRotationTime(...)` builder method, plus the populate-from-existing-account and allow-unset wiring.
- `UserAccountManager.java` — encrypt on write and decrypt on read when storing the field in Android's `AccountManager` user data.
- `AuthenticatorService.java` — a new `KEY_LAST_TOKEN_ROTATION_TIME` user-data key.
- `ClientManager.java` — stamps `setLastTokenRotationTime` at the point where a rotation is confirmed (before the primary `updateAccount` call), then registers the Refresh Token Rotation feature.
- `SalesforceSDKManager.kt` — an `isUserFeatureRegistered(...)` helper; the RTR section wired into the `devSupportInfo` getter; a debug-only **"Force Token Refresh"** developer action; and an extracted `forceTokenRefresh(user)` function to make that action testable. `forceTokenRefresh` takes an optional `restClient` parameter that defaults to the current user's client, so tests can inject a mock without any production-visibility change (matching the existing `invokeServerNotificationAction` pattern).
- `DevSupportInfo.kt` — a `parseRtrSection(...)` function that builds the "RTR" section.

**Tests (6 files)** — 134 scoped tests, all passing:
- `DevSupportInfoTest.kt`, `UserAccountTest.java`, `UserAccountManagerTest.java`, `SalesforceSDKManagerTests.kt`, `ClientManagerMockTest.kt`, `DevInfoActivityTest.kt`.

## Testing

- **134 scoped tests pass** on an Android API 36 emulator: 124 unit and mock-based tests (`SalesforceSDKManagerTests`, `DevSupportInfoTest`, `UserAccountTest`, `UserAccountManagerTest`, `ClientManagerMockTest`) plus 10 instrumented UI tests (`DevInfoActivityTest`).
- **Patch coverage is 84.9%**, which clears the 80% threshold enforced by Codecov (the code-coverage service that reports on each pull request) for the SalesforceSDK library. Here, "patch coverage" means the percentage of the lines changed by this pull request that are exercised by tests. The only uncovered lines are the 8-line block that dispatches work onto a Kotlin coroutine (a background task) and shows an Android `Toast` (a brief on-screen message); that block is not practically unit-testable, and all of its real logic lives in the tested `forceTokenRefresh` function.
- `./gradlew :libs:SalesforceSDK:lintDebug` (the Android lint static-analysis check) runs clean, with no new warnings.

## Reviewer escalation flags (per the repository's CLAUDE.md "Stop and Flag" guidance)

- **New serialized field on `UserAccount`**, plus the encrypt/decrypt handling in `UserAccountManager` — this touches the account persistence path.
- **Change to the token-refresh path** — the `ClientManager` rotation block was reordered to stamp the timestamp before the primary account write.
- **New public API (additive and backward-compatible):** `UserAccount.getLastTokenRotationTime()` / `setLastTokenRotationTime(...)`, and `UserAccountBuilder.lastTokenRotationTime(...)`.
- **Debug-only "Force Token Refresh" developer action** — reviewer opt-in.
- **No `sf__strings.xml` changes** — the developer-info labels are hardcoded, consistent with the existing sections on that screen.

---

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
@@ -107,6 +107,7 @@ public class UserAccount {
 	public static final String FEATURE_FLAGS = "feature_flags";
 	public static final String CREDENTIALS_IDENTIFIER = "credentialsIdentifier";
 	public static final String TOKEN_TYPE = "tokenType";
+	public static final String LAST_TOKEN_ROTATION_TIME = "lastTokenRotationTime";
 
 	private static final String TAG = "UserAccount";
 	private static final String FORWARD_SLASH = "/";
@@ -156,6 +157,7 @@ public class UserAccount {
 	private String scope;
 	private String credentialsIdentifier;
 	private String tokenType;
+	private String lastTokenRotationTime;
 	private Set<String> featureFlags = new java.util.HashSet<>();
 
 	/**
@@ -301,6 +303,7 @@ public class UserAccount {
 			scope = object.optString(SCOPE, null);
 			credentialsIdentifier = object.optString(CREDENTIALS_IDENTIFIER, null);
 			tokenType = object.optString(TOKEN_TYPE, null);
+			lastTokenRotationTime = object.optString(LAST_TOKEN_ROTATION_TIME, null);
 			additionalOauthValues = MapUtil.addJSONObjectToMap(object, additionalOauthKeys, additionalOauthValues);
 		}
 	}
@@ -360,6 +363,7 @@ public class UserAccount {
 			scope = bundle.getString(SCOPE);
 			credentialsIdentifier = bundle.getString(CREDENTIALS_IDENTIFIER);
 			tokenType = bundle.getString(TOKEN_TYPE);
+			lastTokenRotationTime = bundle.getString(LAST_TOKEN_ROTATION_TIME);
 			additionalOauthValues = MapUtil.addBundleToMap(bundle, additionalOauthKeys, additionalOauthValues);
 		}
 	}
@@ -785,6 +789,29 @@ public class UserAccount {
 	public void setTokenType(String tokenType) {
 		this.tokenType = tokenType;
 	}
+
+	/**
+	 * Returns the ISO-8601 timestamp of the last confirmed Refresh Token
+	 * Rotation (RTR) for this user, or null if the refresh token has never
+	 * been rotated.
+	 *
+	 * @return Last token rotation timestamp, or null if not yet rotated.
+	 */
+	public String getLastTokenRotationTime() {
+		return lastTokenRotationTime;
+	}
+
+	/**
+	 * Sets the ISO-8601 timestamp of the last confirmed Refresh Token
+	 * Rotation (RTR).
+	 *
+	 * @param lastTokenRotationTime ISO-8601 timestamp of the last confirmed
+	 *                              rotation.
+	 */
+	public void setLastTokenRotationTime(String lastTokenRotationTime) {
+		this.lastTokenRotationTime = lastTokenRotationTime;
+	}
+
 	/**
 	 * Returns the beacon child consumer key.
 	 *
@@ -1096,6 +1123,7 @@ public class UserAccount {
 			object.put(SCOPE, scope);
 			if (credentialsIdentifier != null) object.put(CREDENTIALS_IDENTIFIER, credentialsIdentifier);
 			if (tokenType != null) object.put(TOKEN_TYPE, tokenType);
+			if (lastTokenRotationTime != null) object.put(LAST_TOKEN_ROTATION_TIME, lastTokenRotationTime);
 			if (!featureFlags.isEmpty()) {
 				org.json.JSONArray flagsArray = new org.json.JSONArray();
 				for (String f : featureFlags) flagsArray.put(f);
@@ -1164,6 +1192,7 @@ public class UserAccount {
 		object.putString(SCOPE, scope);
 		if (credentialsIdentifier != null) object.putString(CREDENTIALS_IDENTIFIER, credentialsIdentifier);
 		if (tokenType != null) object.putString(TOKEN_TYPE, tokenType);
+		if (lastTokenRotationTime != null) object.putString(LAST_TOKEN_ROTATION_TIME, lastTokenRotationTime);
 		object = MapUtil.addMapToBundle(additionalOauthValues, additionalOauthKeys, object);
 		return object;
 	}
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -53,7 +53,9 @@ import android.text.TextUtils.join
 import android.view.WindowInsetsController.APPEARANCE_LIGHT_STATUS_BARS
 import android.webkit.CookieManager
 import android.webkit.URLUtil.isHttpsUrl
+import android.widget.Toast
 import androidx.annotation.VisibleForTesting
+import androidx.annotation.VisibleForTesting.Companion.PRIVATE
 import androidx.annotation.VisibleForTesting.Companion.PROTECTED
 import androidx.compose.material3.ColorScheme
 import androidx.compose.runtime.Composable
@@ -147,6 +149,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.Default
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import java.lang.String.CASE_INSENSITIVE_ORDER
@@ -1386,6 +1389,21 @@ open class SalesforceSDKManager protected constructor(
     fun isGlobalFeatureRegistered(appFeatureCode: String) = features.contains(appFeatureCode)
 
     /**
+     * Returns true if the feature code is registered for the given user
+     * (falling back to the current user when [user] is null). Reads the
+     * per-user feature set that backs the user agent's ftr_ token, so this
+     * reflects features such as RTR that are registered per account.
+     *
+     * @param appFeatureCode The app feature code
+     * @param user The user account, or null to use the current user
+     */
+    internal fun isUserFeatureRegistered(appFeatureCode: String, user: UserAccount? = null): Boolean {
+        val resolvedUser = user ?: userAccountManager.currentUser ?: return false
+        val key = "${resolvedUser.orgId}/${resolvedUser.userId}"
+        return perUserFeatures[key]?.contains(appFeatureCode) == true
+    }
+
+    /**
      * Adds a per-user app feature code for reporting in the user agent header.
      * Falls back to the global set when user is null.
      * @param appFeatureCode The app feature code
@@ -1568,9 +1586,55 @@ open class SalesforceSDKManager protected constructor(
                     })
                 }
             }
+
+            /*
+             * Debug-only helper: proactively drive the SDK's standard
+             * token-refresh path so developers can observe Refresh Token
+             * Rotation (RTR) state update in the dev info screen without
+             * waiting for the access token to expire naturally. This whole
+             * menu is only shown when isDevSupportEnabled() is true (debug
+             * builds by default).
+             */
+            actions["Force Token Refresh"] = object : DevActionHandler {
+                override fun onSelected() {
+                    val user = userAccountManager.currentUser ?: return
+                    CoroutineScope(Default).launch {
+                        val message = forceTokenRefresh(user)
+                        withContext(Main) {
+                            Toast.makeText(appContext, message, Toast.LENGTH_LONG).show()
+                        }
+                    }
+                }
+            }
         }
 
         return actions
+    }
+
+    /**
+     * Drives the SDK's standard token-refresh path for [user] so developers
+     * can observe Refresh Token Rotation (RTR) state update in the dev info
+     * screen without waiting for the access token to expire naturally. Backs
+     * the debug-only "Force Token Refresh" dev action.
+     *
+     * @param user The user whose access token should be refreshed.
+     * @param restClient The REST client to refresh. Defaults to the user's
+     * client; overridable so tests can supply a mock without a network call.
+     * @return A human-readable result message suitable for a Toast. Never
+     * throws — any refresh failure is caught, logged, and returned as a
+     * message (with a null-message fallback to the exception's simple class
+     * name).
+     */
+    @VisibleForTesting(otherwise = PRIVATE)
+    internal fun forceTokenRefresh(
+        user: UserAccount,
+        restClient: RestClient = clientManager.peekRestClient(user)
+    ): String = try {
+        restClient.refreshAccessToken()
+        "Token refresh complete — check RTR section in dev info"
+    } catch (ex: Exception) {
+        e(TAG, "Force Token Refresh failed", ex)
+        "Token refresh failed: ${ex.message ?: ex.javaClass.simpleName}"
     }
 
     /** Information to display in the developer support dialog */
@@ -1637,7 +1701,20 @@ open class SalesforceSDKManager protected constructor(
 //
 //  TODO: Replace devSupportInfo with the above implementation when devSupportInfos is removed in 14.0.
     open val devSupportInfo: DevSupportInfo
-        get() = DevSupportInfo.createFromLegacyDevInfos(devSupportInfos)
+        get() = DevSupportInfo.createFromLegacyDevInfos(devSupportInfos).apply {
+            /*
+             * Surface Refresh Token Rotation (RTR) state so developers can
+             * verify whether RTR is active for the current user's session and
+             * when the token last rotated.
+             */
+            val currentUser = userAccountManager.cachedCurrentUser
+            additionalSections.add(
+                DevSupportInfo.parseRtrSection(
+                    currentUser = currentUser,
+                    rtrActive = currentUser != null && isUserFeatureRegistered(Features.FEATURE_RTR, currentUser),
+                )
+            )
+        }
 
     /** Sends the logout completed intent */
     private fun sendLogoutCompleteIntent(logoutReason: LogoutReason, userAccount: UserAccount?) =
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -1618,19 +1618,23 @@ open class SalesforceSDKManager protected constructor(
      * the debug-only "Force Token Refresh" dev action.
      *
      * @param user The user whose access token should be refreshed.
-     * @param restClient The REST client to refresh. Defaults to the user's
-     * client; overridable so tests can supply a mock without a network call.
+     * @param restClient The REST client to refresh, or null to resolve the
+     * user's client via [ClientManager.peekRestClient]. Tests can supply a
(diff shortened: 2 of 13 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1385,6 +1388,21 @@ open class SalesforceSDKManager protected constructor(
...
+    /**
+     * Returns true if the feature code is registered for the given user
+     * (falling back to the current user when [user] is null). Reads the
+     * per-user feature set that backs the user agent's ftr_ token, so this
+     * reflects features such as RTR that are registered per account.
+     *
+     * @param appFeatureCode The app feature code
+     * @param user The user account, or null to use the current user
+     */
+    internal fun isUserFeatureRegistered(appFeatureCode: String, user: UserAccount? = null): Boolean {
```
JohnsonEricAtSalesforce, line 1400: **New read accessor for per-user feature flags.** RTR-active state is stored as the per-user feature flag `RT` that `ClientManager` registers on a confirmed rotation. The backing `perUserFeatures` map is `private` and the existing public surface only lets you *write* (`registerUsedAppFeature`) or read the whole aggregated user-agent string — there was no narrow "is this one code set for this user?" read.

Rather than widen `perUserFeatures` visibility, this adds a focused `internal` accessor. It resolves the target user (explicit arg → current user → `false` if neither), builds the same `"orgId/userId"` key the writers use (`registerUsedAppFeature`/`unregisterUsedAppFeature`/`getUserAgent`), and does a null-safe, thread-safe (`ConcurrentHashMap`/`ConcurrentSkipListSet`), case-insensitive membership test. The optional `user` param defaults to the current user for ergonomics; the sole caller today passes the user explicitly, and the fallback branch is covered by its own test.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1568,11 +1586,57 @@ open class SalesforceSDKManager protected constructor(
...
+     * client; overridable so tests can supply a mock without a network call.
+     * @return A human-readable result message suitable for a Toast. Never
+     * throws — any refresh failure is caught, logged, and returned as a
+     * message (with a null-message fallback to the exception's simple class
+     * name).
+     */
+    @VisibleForTesting(otherwise = PRIVATE)
+    internal fun forceTokenRefresh(
+        user: UserAccount,
+        restClient: RestClient = clientManager.peekRestClient(user)
```
JohnsonEricAtSalesforce: **Default-argument seam instead of widening production surface.** `forceTokenRefresh` backs the debug-only "Force Token Refresh" dev action; it drives the SDK's standard refresh path so a developer can watch RTR state update without waiting for natural token expiry. It's the testable core — the dev-action lambda only dispatches it on a coroutine and shows the result in a `Toast`.

To make it testable without a network call, `restClient` defaults to `clientManager.peekRestClient(user)` (the production path, unchanged at the `forceTokenRefresh(user)` call site) and tests pass a mock `RestClient`. This mirrors the existing `invokeServerNotificationAction(..., restClient: RestClient = clientManager.peekRestClient(...))` idiom in this same file. An earlier revision made `clientManager` `open` for a test subclass to override — this seam replaces that, so there is **no production-visibility change**. It's `@VisibleForTesting(otherwise = PRIVATE) internal`, and it never throws: any failure is caught, logged, and returned as a message.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1637,7 +1701,20 @@ open class SalesforceSDKManager protected constructor(
...
     open val devSupportInfo: DevSupportInfo
-        get() = DevSupportInfo.createFromLegacyDevInfos(devSupportInfos)
+        get() = DevSupportInfo.createFromLegacyDevInfos(devSupportInfos).apply {
+            /*
+             * Surface Refresh Token Rotation (RTR) state so developers can
+             * verify whether RTR is active for the current user's session and
+             * when the token last rotated.
+             */
+            val currentUser = userAccountManager.cachedCurrentUser
+            additionalSections.add(
```
JohnsonEricAtSalesforce, line 1724: **Why the RTR section is appended via `additionalSections` rather than a new field.** `createFromLegacyDevInfos(devSupportInfos)` returns a `DevSupportInfo` that is already fully populated — its five standard sections (basic info, auth config, boot config, current user, runtime config) are immutable `val`s set at construction. `additionalSections` is the only mutable member, and it's the designed extension point for sections the fixed schema didn't anticipate. So `.apply { additionalSections.add(...) }` is the one legal, minimal-touch way to inject RTR post-construction.

Deliberately **not** done: (a) editing `createFromLegacyDevInfos`, which is marked `// TODO: Remove in 14.0`; (b) adding a dedicated `rtrSection` field, which would change the data-class signature and require wiring into both the legacy path and the future implementation.

⚠️ **14.0 migration note for reviewers:** the commented-out future `devSupportInfo` getter just above builds `DevSupportInfo` via its structured constructor and does **not** append this RTR section. Whoever swaps that in when `devSupportInfos` is removed must carry the RTR `additionalSections.add(...)` over, or RTR silently drops from the screen.

JohnsonEricAtSalesforce:
## Demo — RTR state in the developer info screen

Short screen recording of the feature running end to end in RestExplorer against an
RTR-enabled org, plus before/after stills.

> [!TIP]
> The clip is brief and moves quickly — playing it at **half speed** (0.5×) makes the
> before/after states easy to read.

**What it shows:**
1. **Before (AC2):** the RTR section reads `RTR Active: false` / `Last Rotation: Never`
   (User Agent `ftr_AI.SP.UA`).
2. **Action:** tapping the debug **Force Token Refresh** dev action triggers a token
   refresh; a toast confirms *"Token refresh complete — check RTR section in dev info."*
3. **After (AC3):** the RTR section updates to `RTR Active: true` / `Last Rotation:`
   a valid ISO-8601 timestamp, and the User Agent flips to `ftr_AI.RT.SP.UA` (RT feature
   flag now registered for the user).

https://github.com/user-attachments/assets/ba27e386-63a9-41a1-87f8-da59bdc007be

> [!NOTE]
> The `Last Rotation` timestamp in the before/after **stills** differs from the one in the
> **video** — the stills and the clip were captured on separate token-rotation runs, and
> every rotation stamps the current time. The exact value isn't meaningful; what matters is
> that it transitions from `Never` to a valid timestamp on a confirmed rotation.

rtr_after_true_timestamp_v2
rtr_before_false_never_v2

AC1 (per-user fields show `N/A` with no user) and AC4 (timestamp persists across app
restart) are covered by the automated tests in this PR.

JohnsonEricAtSalesforce:
### Note on CodeCov reporting

This pull request has **no CodeCov status or coverage comment**. I dug into the CI logs, and this is
a **pre-existing, repo-wide CI regression** — not a gap in this change's coverage, and (correcting an
earlier version of this note) **not** a fork-permissions issue.

**Root cause:** CodeCov isn't a standalone check; it's an upload step at the end of each
`test-android` job (`.github/workflows/reusable-lib-workflow.yaml`). The upload has nothing to send
because the `convertedCodeCoverage` Gradle task runs `SKIPPED` on every job: its Jacoco
`executionData` input (`$rootDir/firebase/**/coverage.ec`, per `libs/SalesforceSDK/build.gradle.kts`)
is empty. When the tests were moved to Firebase Test Lab client-side **sharding**
(`--test-targets-for-shard`, added in the Dec 2025 "Shard tests" change), coverage collection was
never adapted — under a sharded matrix FTL doesn't write `coverage.ec` back to the pulled `/sdcard`
path (the only artifact that comes back per shard is `googletest/internal_use/test_args.dat`). So
Jacoco has no execution data → task skipped → no `convertedCodeCoverage.xml` → the CodeCov CLI logs
`No coverage reports found` and `Failed to run upload-coverage`. `fail_ci_if_error: false` makes it
silent rather than a red check.

**Evidence it's not this PR:** the identical `SKIPPED → No coverage reports found → Failed to run
upload-coverage` chain appears on recent internal-branch PRs (#2969 `dpop`, #2973 `codex/*`), across
**all** libraries (SalesforceSDK, MobileSync, SalesforceHybrid), and on the oldest CI run still
available (early July). It has been broken for every PR since sharding landed.

**Coverage for this change** was measured locally: **84.9% patch coverage** on the changed lines,
clearing the 80% threshold. 134 scoped tests pass on an API 36 emulator. The only uncovered lines are
the ~8-line coroutine + `Toast` dispatch block, whose real logic lives in the separately-tested
`forceTokenRefresh` function. Happy to share the local Jacoco report.

The CI fix (collect the per-shard `.ec` files, or merge sharded execution data before the Jacoco
convert step) is out of scope for this PR — flagging so the absent CodeCov comment isn't misread and
so the team can track the coverage-pipeline break separately.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
```diff
@@ -785,6 +789,29 @@ public String getTokenType() {
...
+	/**
+	 * Sets the ISO-8601 timestamp of the last confirmed Refresh Token
+	 * Rotation (RTR).
+	 *
+	 * @param lastTokenRotationTime ISO-8601 timestamp of the last confirmed
+	 *                              rotation.
+	 */
+	public void setLastTokenRotationTime(String lastTokenRotationTime) {
+		this.lastTokenRotationTime = lastTokenRotationTime;
+	}
```
brandonpage, line 813: It will be nice when everything is Kotlin and we can make things like this internal.  Not necessary for this PR, but curious if you have an opinion on [RestrictTo](https://developer.android.com/reference/androidx/annotation/RestrictTo)?  It does not actually prevent someone from using the API, but adds a stern lint warning/error indicating that we do not want them to.

JohnsonEricAtSalesforce:
### 🔎 Root-cause analysis: CodeCov coverage has been silently broken repo-wide since test sharding

Posting this separately from the coverage note above because it's a **CI-pipeline issue that affects
every PR and every library**, not something specific to this change — hoping it saves the next person
the dig.

**Symptom:** No CodeCov status or coverage comment appears on PRs. There's no red check, because the
failure is swallowed.

**Root cause — the sharded Firebase Test Lab run never produces the `coverage.ec` the Jacoco convert
step consumes:**

1. CodeCov upload is a *step* at the end of each `test-android` job
   (`.github/workflows/reusable-lib-workflow.yaml`), not a standalone check.
2. It uploads `libs/<LIB>/build/reports/jacoco/convertedCodeCoverage/convertedCodeCoverage.xml`,
   produced by the Gradle `convertedCodeCoverage` Jacoco task
   (`libs/SalesforceSDK/build.gradle.kts`), whose `executionData` input is
   `$rootDir/firebase/**/coverage.ec`.
3. Since tests moved to Firebase Test Lab client-side **sharding**
   (`--test-targets-for-shard`; commit `4c9d84433` "Shard tests", 2025-12-01), the sharded matrix no
   longer writes `coverage.ec` back to the pulled `/sdcard` path. The only per-shard artifact that
   returns under `artifacts/sdcard/` is `googletest/internal_use/test_args.dat` — **no `.ec` file at
   all**. (The `--environment-variables coverage=true,coverageFile=/sdcard/coverage.ec` +
   `--directories-to-pull=/sdcard` flags predate sharding and no longer collect anything.)
4. With empty `executionData`, Gradle marks the task **`SKIPPED`**, so no XML is written.
5. The CodeCov CLI then logs
   `not_found_files: [...convertedCodeCoverage.xml]` → `Error: No coverage reports found` →
   `Failed to run upload-coverage`.
6. `fail_ci_if_error: false` on the `codecov/codecov-action` step makes all of the above **silent** —
   the job still passes and nothing surfaces on the PR.

**Evidence it's environment-wide (not one branch/lib/fork):** the identical
`SKIPPED → No coverage reports found → Failed to run upload-coverage` chain appears on

- recent internal-branch PRs — #2969 (`dpop`), #2973 (`codex/*`) — i.e. **not** fork-permissions;
- **all** libraries — SalesforceSDK, MobileSync, SalesforceHybrid;
- the oldest CI run still retained (early July).

So it has been broken for every PR since sharding landed.

**Suggested fix (out of scope for this PR):** adapt coverage collection to the sharded layout —
pull each shard's coverage file (`.../shard_*/artifacts/.../coverage.ec`, or whatever FTL emits per
shard now) into `./firebase/` and point the Jacoco task's `executionData` at all of them, so the
merge covers every shard. Worth also flipping `fail_ci_if_error` (or adding a guard) so a future
coverage break is loud instead of silent.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
sfdctaka: Could `peekRestClient` throw?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1568,11 +1586,57 @@ open class SalesforceSDKManager protected constructor(
...
+
+            /*
+             * Debug-only helper: proactively drive the SDK's standard
+             * token-refresh path so developers can observe Refresh Token
+             * Rotation (RTR) state update in the dev info screen without
+             * waiting for the access token to expire naturally. This whole
+             * menu is only shown when isDevSupportEnabled() is true (debug
+             * builds by default).
+             */
+            actions["Force Token Refresh"] = object : DevActionHandler {
```
brandonpage, line 1598: Great idea.

brandonpage: @sfdctaka Should we add this to iOS as well?

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
brandonpage: @JohnsonEricAtSalesforce If you cannot add RTR to the new impl comment above, please add a comment noting that RTR needs to be added so it does not get lost.

brandonpage (approved):
LGTM!  

But please add a comment to the new dev info impl so RTR status does not get lost.  I also would not mind seeing this added to the AuthFlowTester app's UI, but we can do that another day 😉

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
sfdctaka: @brandonpage Ditto. Let me file a ticket for it and we can take care of it next week.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Good catch — yes. `peekRestClient(user)` throws `AccountInfoNotFoundException` (a `RuntimeException`) when there's no account, the user is mid-logout, or auth-token/URL/id data is missing. Because it was a default argument it was evaluated *before* the `try`, so it could escape uncaught — contradicting the "never throws" contract. Fixed by resolving the client inside the `try` (`(restClient ?: clientManager.peekRestClient(user))`) and correcting the KDoc, plus a regression test covering the default path. Commit: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2974/commits/dd055af5a7e8bcd75a7406f781dfc8e711c55642

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Done — added a note in the commented-out 14.0 `devSupportInfo` implementation (and expanded the `TODO`) that the RTR `additionalSections.add(...)` from the live getter must be carried over, so it can't be missed when the block is swapped in. Commit: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2974/commits/dd055af5a7e8bcd75a7406f781dfc8e711c55642

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccount.java
JohnsonEricAtSalesforce: Agree that's the right direction for library-internal API that has to stay `public` from Java. I'd rather not add it to just these two accessors here, since the sibling `UserAccount` accessors (e.g. `tokenType`) aren't annotated and a partial application would be inconsistent — better as a uniform sweep across the account accessors. Filed **W-23667824** to track that. Out of scope for this PR.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Sounds good — thanks for filing the iOS ticket, @sfdctaka. Happy to help with the parity work when it's up. Likewise the AuthFlowTester RTR UI is a nice follow-up for another day.

sfdctaka (approved):
LGTM!

PR #2978, opened 2026-08-06
feat(W-23201584): surface App Attestation state in developer info screen

wmathurin:
## Summary

- **`Features.java`**: adds `FEATURE_APP_ATTESTATION = "AA"` constant
- **`DevSupportInfo.kt`**: adds `parseAppAttestationSection()` companion function with 5 rows:
  - Attestation Enabled (`appAttestationClient != null && apiHostName != null`)
  - API Host (`apiHostName` or "N/A")
  - Google Cloud Project ID (or "N/A")
  - Integrity Provider Ready (`integrityTokenProvider != null` → true/false/N/A)
  - Feature Flag (AA) (true/false/N/A based on current user)
- **`SalesforceSDKManager.kt`**: wires `additionalSections.add(parseAppAttestationSection(...))` in `devSupportInfo` getter alongside the RTR section

## Tests

Added 7 unit tests in `DevSupportInfoTest.kt`:
- No client → all fields N/A or false
- Client without `apiHostName` → Attestation Enabled: false
- Client with host, provider not ready → Enabled: true, Provider Ready: false
- Fully configured → all fields populated
- Feature flag: N/A when no user, false when user has no AA flag, true when flag registered

## Related
- W-23201584
- Companion iOS PR: forcedotcom/SalesforceMobileSDK-iOS#4123 (W-23201582)

PR #2979, opened 2026-08-06
feat(auth): add FEATURE_APP_ATTESTATION per-user flag (W-23195018)

wmathurin:
## Summary

- Adds `FEATURE_APP_ATTESTATION = "AA"` to `Features.java`
- Registers the transient global flag at the two attestation entry points:
  - `OAuth2.java`: browser/web-server flow inside `if (attestationValue != null)`
  - `NativeLoginManager.kt`: native login flow inside `if (attestationValue != null)`
- Promotes the transient global to per-user in `LoginActivity.onAuthFlowSuccess()`, following the same two-phase pattern as `QR` and `WD`

## Test plan

- [ ] `SalesforceSDKManagerTests` — all AA tests pass (verified locally with `./gradlew :libs:SalesforceSDK:build`)
- [ ] `test_givenAppAttestationGlobalFlagSet_whenOnAuthFlowSuccessPromotes_thenAAFlagPerUserAndGlobalCleared` — new test
- [ ] `test_givenNoAppAttestation_whenOnAuthFlowSuccessRuns_thenAAFlagAbsentFromUserAgent` — new test
- [ ] Verify `AA` appears in the user agent after login with app attestation enabled
- [ ] Verify `AA` persists in the user agent across app restarts
- [ ] Verify `AA` is absent for users who logged in without attestation

GUS: W-23195018


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/Features.java
```diff
@@ -47,6 +47,7 @@ public class Features {
     public static final String FEATURE_WELCOME_DISCOVERY_LOGIN = "WD";
     public static final String FEATURE_RTR = "RT";
     public static final String FEATURE_DPOP = "DP";
+    public static final String FEATURE_APP_ATTESTATION = "AA";
```
brandonpage, line 50: This change is also here: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2978/changes

brandonpage (approved):

PR #2981, opened 2026-08-07
fix(dpop): gate proof attachment on per-credential state

sfdctaka:
## Summary

DPoP proof attachment was gated by the process-wide `SalesforceSDKManager.isUseDPoP()` flag. When the flag was flipped off while a DPoP-bound credential was still in use, subsequent refresh calls and REST/identity requests silently dropped the proof, producing a 401 loop.

Gate on `tokenType == "DPoP"` **or** `DPoPKeyManager.hasKeyPair(alias)` instead. The global flag continues to govern *new* logins only; once a credential is bound, every request for it carries a proof regardless of flag state (belt-and-suspenders).

## Changes

**SDK**
- `DPoPKeyManager` — new `hasKeyPair(alias)` side-effect-free presence check; new `shouldAttachDPoP(alias, tokenType)` belt-and-suspenders predicate.
- `OAuth2` — `refreshAuthToken` widened with a `tokenType` parameter (backwards-compat overload preserved); flag guards at the refresh, `use_dpop_nonce` retry, and identity-service paths replaced with the new predicate.
- `RestClient` / `ClientManager` / `AuthenticatorService` — thread `tokenType` through so the outbound-API guard honors bound credentials.

**Tests**
- `OAuth2DPoPTest` — new unit tests: post-flip refresh still attaches proof; Bearer never attaches; `use_dpop_nonce` retry re-attaches proof and consumes the harvested nonce; identity-service call attaches DPoP even with flag off; Bearer identity call never carries DPoP.
- `RestClientDPoPGateTests` — new unit tests: DPoP-bound credential attaches proof on outbound API request; Bearer credential does not.
- `DPoPKeyManagerTest` — coverage for the `hasKeyPair` presence check.
- `DPoPLoginTests` (`AuthFlowTester` androidTest) — UI test locking in flag-off-then-refresh end-to-end.

## Backwards compatibility

- Public API additions only. Existing `refreshAuthToken` overload without `tokenType` is preserved and delegates to the widened signature.
- The global `SalesforceSDKManager.isUseDPoP()` flag continues to work as before for new logins.

## Test plan

- [x] `:libs:SalesforceSDK:connectedAndroidTest` — 401/411 pass; the 11 baseline failures are pre-existing (10 need `test_credentials.json`, 1 MockK stub setup) and unrelated.
- [x] New DPoP-gate unit tests pass under `connectedAndroidTest`.
- [ ] UI test (`AuthFlowTester DPoPLoginTests`) — pending clean-device run (blocked locally on a Chrome Custom Tabs rendering flake unrelated to this change).
- [ ] Full-suite CI regression on merge.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -498,7 +498,8 @@ public class OAuth2 {
 
     /**
      * Gets a new auth token using the refresh token. Overload that accepts a
-     * credentials identifier so DPoP proof can be attached when enabled.
+     * credentials identifier so DPoP proof can be attached when enabled. Delegates
+     * to the fully-parameterized overload with a null token type.
      *
      * @param httpAccessor HttpAccess instance.
      * @param loginServer Login server.
@@ -516,6 +517,34 @@ public class OAuth2 {
                                                          Map<String,String> addlParams,
                                                          @Nullable String credentialsIdentifier)
             throws OAuthFailedException, IOException {
+        return refreshAuthToken(httpAccessor, loginServer, clientId, refreshToken, addlParams,
+                credentialsIdentifier, null);
+    }
+
+    /**
+     * Gets a new auth token using the refresh token. Fully-parameterized overload
+     * that accepts the credential's persisted token type. When the credential is
+     * DPoP-bound the refresh request carries a DPoP proof — independent of the
+     * global {@code isUseDPoP} switch, which only gates new logins.
+     *
+     * @param httpAccessor HttpAccess instance.
+     * @param loginServer Login server.
+     * @param clientId Client ID.
+     * @param refreshToken Refresh token.
+     * @param addlParams Additional parameters.
+     * @param credentialsIdentifier Identifier used to look up the DPoP keypair, or null.
+     * @param tokenType Token type persisted on the credential (e.g. "DPoP" or null / "Bearer").
+     * @return Token response.
+     *
+     * @throws OAuthFailedException See {@link OAuthFailedException}.
+     * @throws IOException See {@link IOException}.
+     */
+    public static TokenEndpointResponse refreshAuthToken(HttpAccess httpAccessor, URI loginServer,
+                                                         String clientId, String refreshToken,
+                                                         Map<String,String> addlParams,
+                                                         @Nullable String credentialsIdentifier,
+                                                         @Nullable String tokenType)
+            throws OAuthFailedException, IOException {
         final FormBody.Builder builder = new FormBody.Builder();
         final boolean useHybridAuthentication = SalesforceSDKManager.getInstance().shouldUseHybridAuthentication();
         final String grantType = useHybridAuthentication ? HYBRID_REFRESH : REFRESH_TOKEN;
@@ -531,7 +560,7 @@ public class OAuth2 {
                 }
             }
         }
-        return makeTokenEndpointRequest(httpAccessor, loginServer, builder, SalesforceSDKManager.getInstance(), credentialsIdentifier);
+        return makeTokenEndpointRequest(httpAccessor, loginServer, builder, SalesforceSDKManager.getInstance(), credentialsIdentifier, tokenType);
     }
 
     /**
@@ -620,7 +649,7 @@ public class OAuth2 {
             throws IOException {
         final Request.Builder builder = new Request.Builder().url(identityServiceIdUrl).get();
         addAuthorizationHeader(builder, authToken, tokenType);
-        if (DPOP.equals(tokenType) && credentialsIdentifier != null && SalesforceSDKManager.getInstance().isUseDPoP()) {
+        if (DPoPKeyManager.INSTANCE.shouldAttachDPoP(credentialsIdentifier, tokenType)) {
             try {
                 final String htu = DPoPURLHelper.INSTANCE.canonicalize(identityServiceIdUrl);
                 final String alias = DPoPKeyManager.INSTANCE.aliasForCredentialsIdentifier(credentialsIdentifier);
@@ -672,7 +701,7 @@ public class OAuth2 {
                                                                  FormBody.Builder formBodyBuilder,
                                                                  SalesforceSDKManager salesforceSdkManager)
             throws OAuthFailedException, IOException {
-        return makeTokenEndpointRequest(httpAccessor, loginServer, formBodyBuilder, salesforceSdkManager, null);
+        return makeTokenEndpointRequest(httpAccessor, loginServer, formBodyBuilder, salesforceSdkManager, null, null);
     }
 
     @VisibleForTesting
@@ -682,6 +711,18 @@ public class OAuth2 {
                                                                  FormBody.Builder formBodyBuilder,
                                                                  SalesforceSDKManager salesforceSdkManager,
                                                                  @Nullable String credentialsIdentifier)
+            throws OAuthFailedException, IOException {
+        return makeTokenEndpointRequest(httpAccessor, loginServer, formBodyBuilder, salesforceSdkManager, credentialsIdentifier, null);
+    }
+
+    @VisibleForTesting
+    @WorkerThread
+    public static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAccessor,
+                                                                 URI loginServer,
+                                                                 FormBody.Builder formBodyBuilder,
+                                                                 SalesforceSDKManager salesforceSdkManager,
+                                                                 @Nullable String credentialsIdentifier,
+                                                                 @Nullable String tokenType)
             throws OAuthFailedException, IOException {
 
         final StringBuilder sb = new StringBuilder(loginServer.toString());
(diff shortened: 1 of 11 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
```diff
@@ -69,4 +70,38 @@ object DPoPKeyManager {
...
+     * by `credentialsIdentifier`. Returns true when either the credential's tokenType is "DPoP"
+     * or a DPoP key pair has already been minted for the credential. Either signal alone is
+     * sufficient — they cover the transient window between `/authorize` (which mints the key
+     * pair before tokenType is written) and `/token` (which writes tokenType after the key pair
+     * has been used to sign the proof).
+     */
+    fun shouldAttachDPoP(credentialsIdentifier: String?, tokenType: String?): Boolean {
+        if (credentialsIdentifier.isNullOrEmpty()) return false
+        if (DPOP_TOKEN_TYPE == tokenType) return true
+        return hasKeyPairForCredentialsIdentifier(credentialsIdentifier)
```
wmathurin, line 126: **[MAJOR] False-positive DPoP on Bearer re-auth with stale key pair**

`shouldAttachDPoP` returns `true` whenever a key pair exists in the Android KeyStore for the given `credentialsIdentifier`, regardless of `tokenType`. If a user re-authenticates as Bearer with the same `credentialsIdentifier` (same org + user ID) and the old DPoP key pair isn't cleaned up, this predicate returns `true` for a Bearer credential — attaching a DPoP proof the server doesn't expect.

The spec confirms this path exists: the `DP` per-user flag is "unregistered if user migrates from DPoP to non-DPoP session."

Suggested fix — short-circuit on explicit non-DPoP `tokenType` before the KeyStore check. This also eliminates the KeyStore I/O on every REST call for Bearer credentials:

```kotlin
fun shouldAttachDPoP(credentialsIdentifier: String?, tokenType: String?): Boolean {
    if (credentialsIdentifier.isNullOrEmpty()) return false
    if (DPOP_TOKEN_TYPE == tokenType) return true
    if (tokenType != null) return false  // explicit non-DPoP type — fast exit, no KeyStore I/O
    // tokenType is null — transition window between /authorize and /token
    return hasKeyPairForCredentialsIdentifier(credentialsIdentifier)
}
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -703,8 +744,9 @@ public static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAcce
         final String tokenHost = HttpUrl.get(refreshPath).host();
         final RequestBody body = formBodyBuilder.build();
         final Request.Builder requestBuilder = new Request.Builder().url(refreshPath).post(body);
+        final boolean attachDPoP = DPoPKeyManager.INSTANCE.shouldAttachDPoP(credentialsIdentifier, tokenType);
 
-        if (credentialsIdentifier != null && salesforceSdkManager.isUseDPoP()) {
+        if (attachDPoP) {
```
wmathurin, line 756: **[MAJOR] `isUseDPoP()` flag semantics changed without explicit callout**

Previously, `isUseDPoP()` acted as a global circuit breaker: flipping it to `false` immediately stopped all DPoP proof attachment for any request. After this change, flipping it off has no effect on existing DPoP-bound credentials — they continue to use DPoP for the lifetime of the session (or until key pairs are cleaned up).

If an operator or test harness flips the flag expecting to disable DPoP globally (e.g., in a degraded-mode or emergency scenario), existing sessions will silently continue sending DPoP proofs.

Please add a Javadoc note to `isUseDPoP()` (and the setter) explicitly documenting this:

```java
/**
 * Controls whether new logins initiate DPoP key generation and DPoP-bound token requests.
 *
 * NOTE: This flag does NOT affect credentials that are already DPoP-bound. An existing
 * DPoP credential will continue to send DPoP proofs regardless of this flag's value,
 * because the server holds a DPoP-bound token and requires a proof on every request.
 * To fully disable DPoP for a user, the credential must be re-authenticated as Bearer.
 */
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
```diff
@@ -38,6 +38,7 @@ import java.security.spec.ECGenParameterSpec
 object DPoPKeyManager {
 
     private const val ANDROID_KEYSTORE = "AndroidKeyStore"
+    private const val DPOP_TOKEN_TYPE = "DPoP"
```
wmathurin: **[MINOR] `DPOP_TOKEN_TYPE` duplicates a constant already defined in `OAuth2.java`**

RFC 9449's token type string `"DPoP"` is case-sensitive. Having two independent definitions (`DPOP_TOKEN_TYPE` here and `DPOP` / equivalent in `OAuth2.java`) risks the two drifting apart if one is edited without the other.

Please consolidate — either expose this as a `@JvmField` constant here and reference it from `OAuth2.java`, or move it to a shared constants file. One source of truth for a protocol-defined string.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -682,6 +711,18 @@ public static TokenEndpointResponse makeTokenEndpointRequest(HttpAccess httpAcce
                                                                  SalesforceSDKManager salesforceSdkManager,
                                                                  @Nullable String credentialsIdentifier)
             throws OAuthFailedException, IOException {
+        return makeTokenEndpointRequest(httpAccessor, loginServer, formBodyBuilder, salesforceSdkManager, credentialsIdentifier, null);
+    }
+
+    @VisibleForTesting
```
wmathurin: **[MINOR] `@VisibleForTesting` on the canonical production implementation is misleading**

This 6-arg overload is the real implementation, called by production code in `refreshAuthToken`. `@VisibleForTesting` conventionally means "this would be private/package-private but we need to access it from tests" — but here it is the primary entry point, not a test-only backdoor.

The annotation should be removed from this overload. If there is a desire to prevent direct external callers from using the 6-arg form, that is better communicated via Javadoc than via a test-visibility annotation that contradicts its actual usage.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
```diff
@@ -69,4 +70,38 @@ object DPoPKeyManager {
...
+     * Returns true iff a key pair is already present in the AndroidKeyStore for the given alias.
+     * Side-effect-free — does NOT mint a key pair on miss.
+     */
+    fun hasKeyPair(alias: String): Boolean {
+        if (alias.isEmpty()) return false
+        return try {
+            val keyStore = KeyStore.getInstance(ANDROID_KEYSTORE).apply { load(null) }
+            keyStore.containsAlias(alias)
+        } catch (e: Exception) {
+            false
```
wmathurin, line 96: **[NIT] Silent exception swallow masks real KeyStore errors**

`catch (e: Exception) { false }` makes a hardware KeyStore error (device lock, entitlement failure, `KeyStoreException`) indistinguishable from "key not present." The return value `false` means "no DPoP — proceed without proof," which is the same 401-loop failure this PR is fixing, but now with no diagnostic signal.

Please add at least a `warn`-level log:

```kotlin
} catch (e: Exception) {
    SalesforceSDKLogger.w(TAG, "DPoPKeyManager: KeyStore lookup failed for $credentialsIdentifier", e)
    false
}
```

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManagerTest.kt
```diff
@@ -99,4 +101,65 @@ class DPoPKeyManagerTest {
...
+
+        // Valid id, no tokenType, no key material → false.
+        assertFalse(DPoPKeyManager.shouldAttachDPoP(id, null))
+        assertFalse(DPoPKeyManager.shouldAttachDPoP(id, "Bearer"))
+
+        // Once a key pair exists, hasKeyPair leg kicks in for null / Bearer tokenType.
+        DPoPKeyManager.generateOrLoadKeyPair(alias)
+        assertTrue(DPoPKeyManager.shouldAttachDPoP(id, null))
+        assertTrue(DPoPKeyManager.shouldAttachDPoP(id, "Bearer"))
+        assertTrue(DPoPKeyManager.shouldAttachDPoP(id, "DPoP"))
```
wmathurin, line 158: **Same issue as the iOS counterpart: this assertion contradicts the `shouldAttachDPoP` implementation.**

`shouldAttachDPoP` fast-exits for any non-null, non-DPoP `tokenType` before doing any KeyStore I/O:
```kotlin
if (tokenType != null) return false // explicit non-DPoP type — fast exit, no KeyStore I/O
```

So `shouldAttachDPoP(id, "Bearer")` returns `false` regardless of whether key material is present — that is intentional and correct. This assertion will fail.

Change `assertTrue` → `assertFalse` here (and fix the comment on the preceding line accordingly):

```kotlin
// Key material present, but an explicit "Bearer" tokenType takes priority — no proof.
assertFalse(DPoPKeyManager.shouldAttachDPoP(id, "Bearer"))
```

The implementation is correct; only the test expectation needs to change.

wmathurin (approved):

PR #2982, opened 2026-08-07
feat(security): remove RSA PKCS1 cipher mode fallback — partial (W-17711863)

wmathurin:
## Summary
Partial removal of the PKCS1 cipher mode fallback. The following are removed in this PR:
- `RSA_PKCS1` as the active push notification cipher — `decryptWithRSAMultiCipherNodes` deleted; `PushNotificationDecryptor` now calls `decryptWithRSABytes` with `RSA_OAEP_SHA256` directly
- `legacyCreateKeysIfNecessary` test helper in `KeyStoreWrapper`

**Not removed in this PR:** the PKCS1 fallback in `SalesforceKeyGenerator.generateUniqueIdIfNoneStored` — deferred to 15.0 (W-23774852). Apps upgrading directly from SDK ≤12.x to 14.0 have AES keys in SharedPreferences encrypted with PKCS1; removing that fallback would force a logout on that path. Since DPoP rolls out in 14.0 and customers are being asked to upgrade, stacking an unexpected logout on top of that transition is a poor experience. The 12.x → 13.x → 14.x path is fully unaffected.

## Test plan
- [ ] `./gradlew :libs:SalesforceAnalytics:assembleDebug :libs:SalesforceSDK:assembleDebug` — clean build
- [ ] `./gradlew :libs:SalesforceAnalytics:test :libs:SalesforceSDK:test` — unit tests pass
- [ ] Instrumented tests (`KeyStoreWrapperTest`, `SalesforceKeyGeneratorTest`) pass on device/emulator


libs/SalesforceSDK/src/com/salesforce/androidsdk/security/SalesforceKeyGenerator.java
```diff
@@ -167,11 +167,7 @@ private synchronized static String generateUniqueIdIfNoneStored(String name, int
             if (uniqueId == null) {
                 final PrivateKey privateKey = KeyStoreWrapper.getInstance().getRSAPrivateKey(LEGACY_KEYPAIR_ALIAS);
                 uniqueId = Encryptor.decryptWithRSA(privateKey, encryptedUniqueId, Encryptor.CipherMode.RSA_OAEP_SHA256);
-                // Decryption failed - must have been encrypted with legacy key with old cipher mode
```
wmathurin, line 170: **⚠️ Upgrade risk: 12.x → 14.0 direct upgrade will cause logout**

Removing this PKCS1 fallback is safe for apps that ran 13.0 at least once (the 13.0 migration re-encrypted SharedPreferences data from LEGACY+PKCS1 to OAEP on first run). However, any app upgrading directly from ≤12.x to 14.0 will hit this code with PKCS1-encrypted data, fail both decryption steps, generate a new unique ID, and lose its session — effectively a forced logout.

This matters more than usual because we are actively pushing customers to upgrade to 14.0 for DPoP rollout. A double-hit of DPoP + unexpected logout on the same upgrade is a bad experience.

**Recommendation:** defer this specific removal to 15.0, once the 12.x install base has had a full release cycle to migrate through 13.x. The other changes in this PR (`decryptWithRSAMultiCipherNodes` removal and push notification PKCS1 path) are safe to ship now.

wmathurin (commented):
**⚠️ Upgrade risk: 12.x → 14.0 direct upgrade will cause logout**

Removing this PKCS1 fallback is safe for apps that ran 13.0 at least once (the 13.0 migration re-encrypted SharedPreferences data from LEGACY+PKCS1 to OAEP on first run). However, any app upgrading directly from ≤12.x to 14.0 will hit this code with PKCS1-encrypted data, fail both decryption steps, generate a new unique ID, and lose its session — effectively a forced logout.

This matters more than usual because we are actively pushing customers to upgrade to 14.0 for DPoP rollout. A double-hit of DPoP + unexpected logout on the same upgrade is a bad experience.

**Recommendation:** defer this specific removal to 15.0, once the 12.x install base has had a full release cycle to migrate through 13.x. The other changes in this PR (removing `decryptWithRSAMultiCipherNodes` and the push notification PKCS1 path) are safe to ship now.

brandonpage (approved):

PR #2983, opened 2026-08-08
Make login server picker non-dismissable (W-23731759)

brandonpage:
## Summary

Advanced Authentication became the default in 14.0, which surfaced two Android login-server-picker defects that could strand a user on a blank login screen. This change makes the login server picker **non-dismissable** and surfaces the navigation/debug affordances the user needs inside the picker, plus fixes a re-select no-op and a couple of related WebView issues. All behavior is flag-independent (works whether `forceAdvancedAuthentication` is on or off).

## Changes

- **Lock the login picker** — disable the X close button and gate swipe/hide dismiss (`confirmValueChange`) for `PickerStyle.LoginServerPicker` only; `UserAccountPicker` behavior unchanged.
- **Conditional back button inside the picker**, reusing the existing `LoginViewModel.shouldShowBackButton` + `LoginActivity.handleBackBehavior()` logic so a user who can leave (existing account / not biometric-locked) still can.
- **Keep the debug Login Options entry point reachable** while the picker is shown (debug builds only).
- **Fix re-select** — choosing the already-selected server now regenerates the authorization URL via `reloadWebView()` instead of being swallowed by the same-value / same-host guards.
- **Flicker fix** — when a Custom Tab is the login surface, keep the WebView blank so the login URL never briefly renders behind the picker.
- **Fix Welcome Discovery reload** — the discovery URL is deterministic, so re-generating it yielded an identical string and Compose skipped the WebView reload. Blank `loginUrl` before reassigning it so the reload button reliably reloads.

## Testing

Run on a physical Pixel 8 Pro via the test-shard mechanism:

- `LoginViewModelTest` (90 tests) — pass
- `PickerBottomSheetTest` + `LoginViewActivityTest` + `LoginActivityTest` (75 tests) — pass

New regression tests cover the flicker fix (WebView stays blank when browser login is enabled, loads the URL when disabled) and the Welcome Discovery reload (loginUrl emits blank then the discovery URL).

GUS: W-23731759


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -830,6 +833,7 @@ open class LoginActivity : FragmentActivity() {
         loadLoginPageInCustomTab(loginUrl, adminLoginCustomTabLauncher)
     }
 
+    @OptIn(ExperimentalInitialNavigationCanLeaveBrowser::class)
```
wmathurin, line 915: **Experimental API — add a note on the risk.** `ExperimentalInitialNavigationCanLeaveBrowser` could be removed or API-broken in a future `androidx.browser` release without a deprecation cycle. A short inline comment saying this is best-effort hardening (the Custom Tab still works without it) would tell future maintainers why the `@OptIn` is safe to keep.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1361,6 +1372,21 @@ open class LoginActivity : FragmentActivity() {
         // region LoginWebviewClient Constants
 
         internal const val ABOUT_BLANK = "about:blank"
+
+        /**
+         * The background color of login.salesforce.com, used as the Custom Tab toolbar color
+         * (unless the app overrides it via [LoginViewModel.topBarColor]).
+         */
+        internal val LOGIN_BACKGROUND_COLOR = Color(red = 244, green = 246, blue = 249)
```
wmathurin, line 1429: **Hardcoded color — acknowledge the staleness risk.** `Color(244, 246, 249)` is a point-in-time snapshot of `login.salesforce.com`'s background. If Salesforce redesigns that page this will silently mismatch. A one-liner saying the staleness is acceptable (e.g. "only visible for the ~200 ms Custom Tab launch") tells the next reader why it's intentional.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -222,6 +225,19 @@ internal fun TestablePickerBottomSheet(
         viewModel.loading.value = true
     }
 
+    // The picker's own header is the only reachable UI while the sheet is shown (the top app
+    // bar sits behind the modal scrim), so the login-exit back button and the debug Login
+    // Options entry are surfaced here, reusing the existing WebView-login logic unmodified.
+    val showLoginBackButton = pickerStyle == PickerStyle.LoginServerPicker && viewModel.shouldShowBackButton
+    val onLoginBackButtonClick: () -> Unit = { (activity as? LoginActivity)?.handleBackBehavior() }
+    val showDevSupport = with(SalesforceSDKManager.getInstance()) {
+        return@with if (pickerStyle == PickerStyle.LoginServerPicker && isDebugBuild && isDevSupportEnabled()) {
```
wmathurin: **Nit: `return@with` is a no-op here.** The last expression of a `with` block is already its return value, so the explicit label adds noise without changing behaviour. Simplify to a plain `if/else` expression and drop the `with` wrapper entirely:\n```kotlin\nval showDevSupport = if (pickerStyle == PickerStyle.LoginServerPicker && isDebugBuild && isDevSupportEnabled()) {\n    { showDevSupportDialog(activity) }\n} else {\n    null\n}\n```

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/PickerBottomSheet.kt
```diff
@@ -291,7 +313,7 @@ internal fun PickerBottomSheet(
 
     LaunchedEffect(Unit) {
         coroutineScope.launch {
```
wmathurin, line 315: **Nit: unnecessary `.toLong()` chain.** If `SLOW_ANIMATION_MS` is an `Int`, `kotlin.time.Duration.Companion.milliseconds` accepts `Int` directly — `SLOW_ANIMATION_MS.milliseconds` is sufficient.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/PickerBottomSheetTest.kt
```diff
@@ -427,6 +431,196 @@ class PickerBottomSheetTest {
...
+            PickerBottomSheetTestWrapper(
+                pickerStyle = PickerStyle.LoginServerPicker,
+                showLoginBackButton = false,
+            )
+        }
+
+        composeTestRule.onNodeWithTag(LoginViewTestTags.PICKER_LOGIN_BACK_BUTTON).assertDoesNotExist()
+    }
+
+    @OptIn(ExperimentalMaterial3Api::class)
```
wmathurin, line 539: **Missing assertion — this test can pass silently even on regression.** The swipe is performed but there's no assertion that the sheet reached `Hidden` afterwards. If `confirmValueChange` were accidentally broken to always block `Hidden`, this test would still pass. Follow the pattern in `accountPicker_closeButton_present_andDismisses`: capture `sheetState` and assert `sheetState.currentValue == SheetValue.Hidden` after `waitForIdle()`.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/LoginActivityTest.kt
```diff
@@ -512,4 +545,144 @@ class LoginActivityTest {
...
+        every { sdkManager.nativeLoginActivity } returns null
+        val bioAuthManager = mockk<BiometricAuthenticationManager>(relaxed = true)
+        every { bioAuthManager.locked } returns false
+        every { sdkManager.biometricAuthenticationManager } returns bioAuthManager
+        val userAccountManager = mockk<UserAccountManager>(relaxed = true)
+        every { sdkManager.userAccountManager } returns userAccountManager
+        every { userAccountManager.authenticatedUsers } returns null
+
+        val viewModel = mockk<LoginViewModel>(relaxed = true)
+        every { viewModel.shouldShowBackButton } returns false
```
wmathurin: **Misleading mock — `viewModel.shouldShowBackButton` is never read by `handleBackBehavior()`.** The back behaviour is controlled entirely by `SalesforceSDKManager` state (biometric lock + authenticated users). This stub doesn't influence the code path under test — remove it or replace with a comment explaining it's inert, so the next reader doesn't assume it matters.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/LoginActivityTest.kt
```diff
@@ -512,4 +545,144 @@ class LoginActivityTest {
...
+        every { sdkManager.nativeLoginActivity } returns null
+        val bioAuthManager = mockk<BiometricAuthenticationManager>(relaxed = true)
+        every { bioAuthManager.locked } returns false
+        every { sdkManager.biometricAuthenticationManager } returns bioAuthManager
+        val userAccountManager = mockk<UserAccountManager>(relaxed = true)
+        every { sdkManager.userAccountManager } returns userAccountManager
+        every { userAccountManager.authenticatedUsers } returns listOf(mockk(relaxed = true))
+
+        val viewModel = mockk<LoginViewModel>(relaxed = true)
+        every { viewModel.shouldShowBackButton } returns true
```
wmathurin: **Same misleading mock as in `test_handleBackBehavior_fromPicker_noAccounts_movesTaskToBack`.** `viewModel.shouldShowBackButton` is stubbed but `handleBackBehavior()` reads from `userAccountManager.authenticatedUsers` (line above), not from the ViewModel. Remove or document.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: The comment already acknowledges the static color and **_not_** only visible for a short time.  Any other color (white/black/etc) would be static since we can no longer guarantee the WebView is being loaded to run the javascript to check the background color.  Even if we always loaded the WebView (we are trying remove) it is not worth waiting for the page to load before launching the custom tab.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: ~~I guess I will remove it as it does not appear to work today anyway.~~

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/ui/PickerBottomSheetTest.kt
brandonpage: `assertDoesNotExist` is used.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: No, we need this.

wmathurin (approved):

PR #2984, opened 2026-08-11
Remove deprecated devSupportInfos, migrate devSupportInfo to structured builder

brandonpage:
## Summary
- Removes the previously-deprecated `SalesforceSDKManager.devSupportInfos` flat `List<String>` API (deprecated for removal in 14.0) and its supporting `DevSupportInfo.createFromLegacyDevInfos` conversion helper.
- `devSupportInfo` now builds `DevSupportInfo` directly via the structured constructor instead of round-tripping through the legacy list, and still appends the RTR and App Attestation `additionalSections`.
- `SmartStoreSDKManager` now overrides `getDevSupportInfo()` (structured) instead of the removed `getDevSupportInfos()`.
- Fixed a KDoc misplacement where `parseRtrSection` had no doc comment and `parseAppAttestationSection` had two.
- Updated `LoginOptionsActivityTest` and `DevSupportInfoTest` to drop coverage of the removed legacy-list API.

## Breaking change
This removes the deprecated public `devSupportInfos` getter and `DevSupportInfo.createFromLegacyDevInfos`. Any external subclass of `SalesforceSDKManager` that overrode or called `devSupportInfos` will need to migrate to `devSupportInfo` (the structured `DevSupportInfo` data class).  However, it is very unlikely many consumers exist. 

## Test plan
- [x] `:libs:SalesforceSDK:compileDebugKotlin` and `:libs:SmartStore:compileDebugJavaWithJavac` build clean
- [x] Full `:libs:SalesforceSDK:build` and `:libs:SmartStore:build` (including lint) pass
- [x] Instrumented tests pass on device: `LoginOptionsActivityTest` (18/18), `DevSupportInfoTest` (30/30), `SalesforceSDKManagerTests` (58/58, RTR/AA wiring unaffected)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1738,98 +1738,53 @@ open class SalesforceSDKManager protected constructor(
...
+            ).apply {
+                /*
+                 * Surface Refresh Token Rotation (RTR) state so developers can
+                 * verify whether RTR is active for the current user's session and
+                 * when the token last rotated.
+                 */
+                additionalSections.add(
+                    DevSupportInfo.parseRtrSection(
+                        currentUser = currentUser,
+                        rtrActive = currentUser != null && isUserFeatureRegistered(Features.FEATURE_RTR, currentUser),
```
brandonpage, line 1777: I think once there is no way to disable RTR we can drop the `RTR Active` key and move the `Last Rotation` key to the main `Authentication Configuration` section.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1738,98 +1738,53 @@ open class SalesforceSDKManager protected constructor(
...
-                DevSupportInfo.parseAppAttestationSection(
-                    appAttestationClient = appAttestationClient,
-                    currentUser = currentUser,
-                    aaFeatureActive = currentUser != null && isUserFeatureRegistered(Features.FEATURE_APP_ATTESTATION, currentUser),
+                additionalSections.add(
+                    DevSupportInfo.parseAppAttestationSection(
+                        appAttestationClient = appAttestationClient,
+                        currentUser = currentUser,
+                        aaFeatureActive = currentUser != null && isUserFeatureRegistered(Features.FEATURE_APP_ATTESTATION, currentUser),
+                    )
```
brandonpage, line 1785: App Attestation could be moved to a dedicated section in `DevSupportInfo` if we want.

wmathurin (approved):
LGTM

PR #2986, opened 2026-08-12
Add auth_trigger and sdkInfo params to Custom Tab authorize URL

brandonpage:
## Summary
- Appends `sdkInfo` (SDK user-agent string) and `auth_trigger` (why native browser login was chosen) query params to the OAuth authorize URL, Chrome Custom Tab path only
- WebView path is unchanged — it already sends the real User-Agent header
- `auth_trigger` priority: `login_for_admin` > `force_advanced_auth` > `org_config` (MDM forces cert auth via a different code path on Android, so `mdm` is unreachable here)

## Test plan
- [ ] `LoginActivityTest` / `BrowserLoginTelemetryTest` pass
- [ ] Manual login via Custom Tab confirms both params present in the authorize request

GUS: W-23789315

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -1032,8 +1032,7 @@ open class LoginActivity : FragmentActivity() {
             customTabsIntent.intent.setPackage(customTabBrowser)
         }
 
-        // Add prompt=login to prevent the browser cookie from bypassing login if it exists.
-        val urlString = if (sharedBrowserSession) loginUrl else loginUrl + PROMPT_LOGIN
+        val urlString = buildCustomTabAuthorizeUrl(loginUrl, completedViaAdminCustomTab)
 
         runCatching {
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -1047,6 +958,30 @@ open class LoginActivity : FragmentActivity() {
...
+        isAdminLogin: Boolean,
+        sdkManager: SalesforceSDKManager = SalesforceSDKManager.getInstance(),
+    ): String {
+        @Suppress("DEPRECATION")
+        val authTrigger = selectAuthTrigger(
+            isAdminLogin,
+            isMdmForcedBrowserLogin(),
+            sdkManager.forceAdvancedAuthentication,
+        )
+        val sdkInfo = Uri.encode(sdkManager.getUserAgent(""))
```
wmathurin, line 1068: I believe it's only the MSDK user agent unlike iOS (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4128#discussion_r3763112074).

wmathurin (approved):

PR #2988, opened 2026-08-13
fix(W-23836447): allow dpop_jkt on login pool servers

wmathurin:
## Summary

- Removes the `LoginServerManager.isPoolServer(server)` guard from `addDpopJktIfNeeded()` in `LoginViewModel.kt` — pool servers now support DPoP authorization code binding
- `welcome.salesforce.com/discovery` is unaffected (discovery resolves a my-domain server before `/authorize` is called)
- Also fixes a key-stability bug: `addDpopJktIfNeeded` was regenerating a new key pair on every call; during pool-server login it is called twice (once per `/authorize`), so the auth code's `dpop_jkt` used Key A but the token-exchange proof used Key C → `invalid_dpop_proof`. Fix: reuse `pendingCredentialsIdentifier` if already set within the same login flow.

## Known limitation — server-side bug W-23864247

During testing against `login.test1.pc-rnd.salesforce.com`, the token exchange returns HTTP 400 `invalid_dpop_proof` even though the client's DPoP proof is cryptographically correct and its JWK thumbprint exactly matches the `dpop_jkt` sent in `/authorize`. This is a server-side issue filed as **W-23864247** (Auth Protocols team). `testECAJwtDPoP_ViaLoginPoolServer` is `@Ignore`d until the server fix is confirmed.

## Changes

**Production code**
- `LoginViewModel.kt` — remove `isMyDomainServer` check; only `!sdkManager.useDPoP` exits early now; fix `pendingCredentialsIdentifier` reuse; updated KDoc

**Unit tests**
- `LoginViewModelTest.kt` — add `generateAuthorizationUrl_WhenUseDPoP_AndPoolServer_AddsDpopJktToUrl` asserting `dpop_jkt` present for `login.salesforce.com` when DPoP enabled

**UI tests (AuthFlowTester)**
- `ui_test_config.json.sample` — add top-level `loginPoolHost` field
- `UITestConfig.kt` — add `loginPoolHost: String?` to `UITestConfig`; add `requireLoginPoolHost()` accessor
- `AuthFlowTest.kt` — add `useLoginPoolHost: Boolean = false` to `loginAndValidate()`; when true, calls `changeServerByUrl(testConfig.requireLoginPoolHost())` while credentials come from `knownLoginHostConfig`
- `DPoPLoginTests.kt` — add `testECAJwtDPoP_ViaLoginPoolServer` (`@Ignore` pending W-23864247)
- `RefreshTokenMigrationTests.kt` — update override signature to include `useLoginPoolHost`

## Test plan

- [x] `./gradlew :libs:SalesforceSDK:compileDebugKotlin` — BUILD SUCCESSFUL
- [x] `./gradlew :libs:SalesforceSDK:compileDebugAndroidTestKotlin` — BUILD SUCCESSFUL
- [x] `./gradlew :native:NativeSampleApps:AuthFlowTester:compileDebugAndroidTestKotlin` — BUILD SUCCESSFUL
- [x] `LoginViewModelTest` — new pool-server test passes on emulator
- [ ] `DPoPLoginTests.testECAJwtDPoP_ViaLoginPoolServer` — blocked on server bug W-23864247

## GUS

W-23836447 — [Android] Allow dpop_jkt on login pool servers (/authorize code binding)


native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/ECALoginTests.kt
```diff
@@ -83,4 +83,14 @@ fun testECAJwt_SubsetScopes_NotHybrid() {
 fun testECAJwt_AllScopes() {
         loginAndValidate(knownAppConfig = ECA_JWT, scopeSelection = ALL)
     }
+
+    // region ECA Pool Server Tests
+
+    // Login via the pool server without DPoP and verify the session is valid.
+    @Test
+    fun testECAJwt_ViaLoginPoolServer() {
+        loginAndValidate(knownAppConfig = ECA_JWT, useLoginPoolHost = true)
```
JohnsonEricAtSalesforce, line 97: This test is active (not `@Ignore`d) and fails in the `ui-tests-pr` CI job for HEAD with `java.lang.Exception: loginPoolHost not found in ui_test_config.json` — `useLoginPoolHost = true` routes into `requireLoginPoolHost()` (UITestConfig.kt:86), which throws when the key is absent. The spec (Workspace 76) required adding `loginPoolHost` to the live `ui_test_config.json` and to `shared/test/ui_test_config.json.sample`; the sample here still has only `loginHosts`/`apps`. Please either provision `loginPoolHost` in the CI config and add it to the `.sample`, or guard this test so it skips gracefully when the key is missing (as the DPoP pool test does with `@Ignore`), so a green `ui-tests` run reflects real coverage.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginViewModel.kt
```diff
@@ -713,17 +713,21 @@ open class LoginViewModel(
...
+     * Only generates a new key pair when starting a fresh login flow.
      */
     private fun addDpopJktIfNeeded(
         server: String,
         sdkManager: SalesforceSDKManager,
         params: MutableMap<String, String>,
     ) {
-        val isMyDomainServer = !LoginServerManager.isPoolServer(server)
-        if (!sdkManager.useDPoP || !isMyDomainServer) {
+        if (!sdkManager.useDPoP) {
```
JohnsonEricAtSalesforce: Optional / for parity with the iOS twin: iOS keeps an explicit early-return for `welcome.salesforce.com` in its equivalent helper, whereas this relaxation relies solely on the call graph (discovery is handled in `reloadWebView()` before `generateAuthorizationUrl`, so this method is never reached with the discovery host). That reasoning is correct today and the KDoc documents it well — this is just a suggestion to add a one-line defensive `welcome` guard for belt-and-suspenders symmetry with iOS, in case a future caller wires `generateAuthorizationUrl` differently. Not blocking.

JohnsonEricAtSalesforce (changes requested):
Reviewed as the cross-platform peer to the approved iOS twin (PR 4130, W-23836436), against the design spec (Workspace 76) and a runtime trace of the DPoP `/authorize` → `/token` key binding.

**The production change is correct — this request is narrowly about a self-added test that is red in CI, not about the auth logic.** Removing the pool-server carve-out so `dpop_jkt` is sent to `login`/`test.salesforce.com` when DPoP is enabled matches (a) the server team's confirmation in the spec that pool servers now support DPoP authorization-code binding, and (b) the iOS change, which makes the identical relaxation. The key-reuse addition (`pendingCredentialsIdentifier`) keeps the `/authorize` `dpop_jkt` and the token-exchange proof bound to the same key pair across server-picker re-entries — I traced this and could not find a path that publishes a thumbprint for one key and signs with another; on a crypto soft-fail no `dpop_jkt` is emitted and the token exchange also attaches no proof, so both hops stay self-consistent. No credential-leak or key-mismatch issue found, and the changed line's unit coverage is green.

**Requested change (blocking): the PR adds an active test that fails deterministically in CI.** `ECALoginTests.testECAJwt_ViaLoginPoolServer` is new here, not `@Ignore`d, and fails every run of the `ui-tests-pr` job with `loginPoolHost not found in ui_test_config.json` — `useLoginPoolHost = true` routes into `requireLoginPoolHost()`, which throws when the key is absent. The spec (Workspace 76) called for adding `loginPoolHost` to both the live `ui_test_config.json` and `shared/test/ui_test_config.json.sample`; neither landed (the sample still has only `loginHosts`/`apps`). Please make the job green by either (a) provisioning `loginPoolHost` in the CI config and adding it to the `.sample`, or (b) guarding the test to skip gracefully when the key is absent (as the DPoP pool test does with `@Ignore`). A PR shouldn't merge adding its own permanently-red test.

Two non-blocking notes:

1. (Consistency, optional) Android drops the explicit `welcome.salesforce.com` guard that the iOS twin keeps. I confirmed via the call graph that `reloadWebView()` handles Welcome Discovery before `generateAuthorizationUrl`, so `addDpopJktIfNeeded` is never reached with the discovery host today — the current code is correct and the KDoc explains it well. Consider keeping a one-line defensive `welcome` check anyway for parity with iOS and resilience if a future caller wires `generateAuthorizationUrl` differently. Purely optional.

2. (Affirm) The `@Ignore` on `testECAJwtDPoP_ViaLoginPoolServer` for the server-side `invalid_dpop_proof` bug (W-23864247) is appropriate and matches the iOS `XCTSkip`. It does not mask a client regression: the client-side contract (sending `dpop_jkt` on the pool `/authorize`) is still covered by the green `LoginViewModelTest` unit test.

For context, the other `ui-tests` failures in this run are unrelated to this PR — the known Custom Tab / touch-injection infra flakes, plus `testECAJwtDPoP_And_NonDPoP_MultiUser_FlagOff_IndependentProofs`, a multi-user token-format test that already exists on `dev` (merged in 6edf6f7cc) and is not in a line this PR changes. Only `testECAJwt_ViaLoginPoolServer` is attributable to this PR.

No public API, localization, or SQLCipher surface is touched. This PR sits on OAuth/credential paths (a CLAUDE.md escalation class), so it also warrants a human sign-off on the auth change.

JohnsonEricAtSalesforce (approved):
Re-reviewed at HEAD c6abcff74 — both requested changes are resolved, and the DPoP change remains sound. Clearing my earlier REQUEST_CHANGES.

**① The blocking issue is fixed.** `ECALoginTests.testECAJwt_ViaLoginPoolServer` is now `@Ignore`d with a comment documenting the re-enable path — the graceful-skip option from my comment. The proof is in CI: `ui-tests-pr / test-android` is green on this HEAD, where it was deterministically red before. All checks (unit-tests, ui-tests, test-orchestrator, SAST, credential scanning) pass.

**② The optional consistency note was adopted.** `addDpopJktIfNeeded` now guards with `LoginServerManager.WELCOME_LOGIN_URL == server`, bringing Android to parity with the iOS twin's explicit Welcome early-return. I confirmed the guard is live — `server` is the raw selected-server URL passed into both `generateAuthorizationUrl` overloads, and it compares by exact equality against the `"https://welcome.salesforce.com/discovery"` constant.

Nice addition beyond what I asked for: `test_givenDPoPEnabled_whenGenerateAuthorizationUrlCalledTwice_thenDpopJktIsStable` asserts the `dpop_jkt` and `pendingCredentialsIdentifier` stay stable across a pool → my-domain redirect. That directly locks in the key-reuse contract at the heart of W-23836447 (the `/authorize` binding and the token-exchange proof must use the same key), with real assertions — a good regression guard.

One note, not blocking:

(Follow-up) The skip is the right call for now, but `shared/test/ui_test_config.json.sample` still doesn't carry `loginPoolHost`. Consider filing a quick follow-up to provision the key in CI and add it to the `.sample`, then remove the `@Ignore`, so the pool-server path regains real e2e coverage and the skip doesn't get forgotten.

wmathurin:
NB: All UI tests passing locally.

PR #2989, opened 2026-08-13
Auto-present biometric opt-in/unlock (automaticPresentation) — port iOS #4041

brandonpage:
Ports the iOS biometric **automatic presentation** feature ([iOS PR #4041](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4041)) to the Android Mobile SDK.

Related work item: W-21981313 (iOS counterpart).

## Summary

Adds an `automaticPresentation` property to the public `BiometricAuthenticationManager` interface. When enabled (the default, matching iOS), the SDK:

- **Automatically presents the biometric opt-in dialog after a fresh login**, if the user hasn't yet responded to it.
- **Automatically triggers the biometric unlock prompt when the app locks**, if the user has already opted in.

`automaticPresentation` defaults to `true` to match iOS behavior. It is an in-memory property (not persisted) and resets to `true` on process restart.

### Opt-in decision tracking ("declined" vs. "never asked")

The opt-in gating distinguishes a user who **declined** biometric unlock (chose "Use Password") from one who has **never been asked**, so a user who declined is not re-prompted on every subsequent login. This is done via a new internal `hasBiometricOptInDecision()` that checks `SharedPreferences.contains(USER_BIO_OPT_IN)` rather than reading the boolean with a default. The stored opt-in decision is cleared on logout via the existing per-account cleanup path (covered by a regression test).

## Changes

- **`security/interfaces/BiometricAuthenticationManager.kt`** — new public `var automaticPresentation: Boolean` on the interface (KDoc'd).
- **`security/BiometricAuthenticationManager.kt`** — implements `automaticPresentation` (default `true`); adds internal `hasBiometricOptInDecision()`.
- **`ui/LoginActivity.kt`** — auto-presents the opt-in dialog on fresh login via `onAuthFlowFinished()` (gated on `enabled && automaticPresentation && !hasBiometricOptInDecision()`), deferring completion until the dialog is dismissed; extracts `shouldAutoPresentBiometricOnCreate()` to gate auto-triggering the unlock prompt on lock (adds the `automaticPresentation` check to the existing conditions).
- **`ui/LoginViewModel.kt` / `ui/components/LoginView.kt` / `ui/BiometricAuthOptInPrompt.kt`** — wire the opt-in dialog through the Compose login UI and the `onAuthFlowFinished` completion callback.
- **`auth/AuthenticationUtilities.kt`** — thread the `onAuthFlowFinished` callback through the auth-flow-complete path.
- Converted dialog from XML to Compose:

Screenshot 2026-08-12 at 2 32 48 PM

## Acceptance criteria coverage

**Scenario A — auto opt-in dialog after login**
- A1: `automaticPresentation=true`, not opted in, fresh login → dialog presented ✅
- A2: `automaticPresentation=true`, already opted in, fresh login → no dialog ✅
- A3: `automaticPresentation=false`, not opted in, fresh login → no dialog ✅
- A4: `automaticPresentation=true`, not opted in, token refresh → no dialog (fresh login only) ✅

**Scenario B — auto biometric prompt on lock**
- B1: `automaticPresentation=true`, opted in, lock → prompt shown automatically ✅
- B2: `automaticPresentation=true`, not opted in, lock → no auto-present ✅
- B3: `automaticPresentation=false`, opted in, lock → no auto-present ✅
- B4: `automaticPresentation=false`, not opted in, lock → no auto-present ✅

**Scenario D — existing behavior (regression)**
- D1: `automaticPresentation=false` → all existing behavior unchanged ✅
- D2: lock/unlock without `automaticPresentation` → user manually taps biometric button ✅
- D3: opt-in without `automaticPresentation` → app must explicitly call the opt-in dialog ✅

## Public API note

This adds a new property to the public `BiometricAuthenticationManager` interface (`var automaticPresentation`). No existing signatures change; the default preserves prior behavior only where callers had already opted into biometric auth. Flagging for API review per SDK policy.

## Test plan
- New/updated unit tests in `BiometricAuthenticationManagerTest.kt`, `LoginActivityTest.kt`, `LoginActivityScenarioTest.kt`, `LoginViewModelMockTest.kt`, and `AuthenticationUtilitiesTest.kt` cover the Scenario A/B/D combinations above and the opt-in-decision / cleanup-on-logout behavior.
- `./gradlew :libs:SalesforceSDK:compileDebugAndroidTestKotlin` succeeds.
- Full relevant instrumented suites pass on device (physical Pixel 8 Pro) and emulator (Pixel 9 Pro API 35).

## Localization
No new user-facing strings added in this PR (the opt-in dialog strings already exist in `sf__strings.xml`).


brandonpage:
Calling it out here, [0e6a458](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2989/commits/0e6a45818e406390afd9867922bd883dfba43505) removes the `presentOptInDialog` API.  This used the old XML layout and there is really no good way to offer the API with Compose.  Despite having iOS parity, this was provided strictly for convenience since the APIs needed for the app to create their own AlertDialog (or better yet settings toggle) are available, which would be a far better experience.

wmathurin (commented):
**1. Two parallel opt-in dialogs (XML + Compose) — clarify or consolidate**

The PR description says the dialog was "converted from XML to Compose," but the legacy `BiometricAuthOptInPrompt` XML `DialogFragment` is still live: `presentOptInDialog(fragmentManager)` in the impl still constructs it. The new auto-present path uses the Compose `BiometricOptInDialog`; apps calling `presentOptInDialog` directly get the old XML one. These two implementations will drift. Either:
- Route `presentOptInDialog` to the Compose component and remove the XML dialog + layout, **or**
- Update the description — "converted" overstates what happened.

---

**2. Dead code in `BiometricAuthOptInPrompt`**

The `onDismissed: () -> Unit = {}` param and `onDismiss` override added to the legacy `DialogFragment` are never wired (only the default empty lambda is ever passed). If the XML dialog is staying for now, this addition should be removed; if it's being deleted, that resolves itself.

---

**3. Advanced-Auth (Custom Tab) path skips auto opt-in — diverges from iOS**

`completeAdvAuthFlow` passes `onAuthFlowFinished = { proceed -> proceed(); finishLoginFlow() }` (no dialog), while WebView/token paths pass `::onAuthFlowFinished`. On iOS, the opt-in dialog is presented inside the `dismissAuthViewControllerIfPresentForScene:completion:` callback, which fires after **all** login types — webview, native login, and Advanced Auth alike; iOS doesn't special-case Custom Tab. Is skipping the dialog on Android's Custom Tab path a deliberate Android-specific decision, or an unintended parity gap worth closing?

---

**4. Add `withContext(Dispatchers.Main)` around the Compose state write — matches iOS pattern**

`onAuthFlowFinished` fires inside `CoroutineScope(Default).launch` with no main-thread dispatch. The iOS equivalent (`dismissAuthViewControllerIfPresentForScene:completion:`) opens with an explicit `![NSThread isMainThread]` guard and `dispatch_sync`s to main before the completion block runs — the UI call is guaranteed on the main thread. Even though current Compose snapshot-state writes tolerate off-thread assignment, wrapping `showBiometricOptInDialog.value = true` and the `else`-branch `startMainActivity()`/`finish()` in `withContext(Dispatchers.Main)` would make the threading contract explicit and mirror the iOS pattern.

wmathurin (commented):
The new commit (`0e6a458`) addresses issues 1 and 2 — `BiometricAuthOptInPrompt`, its layout, and `presentOptInDialog` are all removed. Clean resolution.

Issues 3 (Advanced Auth path skips opt-in dialog — parity with iOS) and 4 (`withContext(Dispatchers.Main)` for the Compose state write) are still open.

brandonpage:
> The new commit (`0e6a458`) addresses issues 1 and 2 — `BiometricAuthOptInPrompt`, its layout, and `presentOptInDialog` are all removed. Clean resolution.
> 
> Issues 3 (Advanced Auth path skips opt-in dialog — parity with iOS) and 4 (`withContext(Dispatchers.Main)` for the Compose state write) are still open.

The intersection of Bio Auth and Adv Auth has never worked on either platform.  I have a separate WI (spec and test plan PR will be sent today) to cover that work.

Android/Kotlin and iOS/Swift have very different threading paradigms.   We should not be seeking parity as doing so will cause crashes in one or the other for the vast majority of scenarios.  Calling `finish()` on the main thread isn't a bad suggestion so I'll add it, but the justification provided is completely wrong.

wmathurin (approved):

PR #2990, opened 2026-08-13
fix(test): handle non-dismissable login picker in Chrome Custom Tab page object

wmathurin:
## Summary

**Root cause**: PR #2983 (merged Aug 11) made the login-server picker non-dismissable — it removed the close button from \`LoginServerPicker\` and blocked swipe-to-dismiss. After a Chrome Custom Tab closes (via \`backOutToLoginActivity\`), the SDK raises the picker. Tests that then called \`openLoginOptions()\` or \`changeServerByUrl()\` were failing because both methods tried to tap \`MORE_OPTIONS_BUTTON\` in the top app bar, which is behind the picker's modal scrim.

**Fixes:**
- **\`ChromeCustomTabPageObject.backOutToLoginActivity()\`**: simplified — closes the tab and waits for either the picker or the top bar to be reachable, without trying to dismiss the picker (which is intentionally non-dismissable).
- **\`LoginPageObject.openLoginOptions()\`**: detects whether the picker is showing before tapping the top bar. If it is, routes through \`PICKER_DEV_SUPPORT_BUTTON\` (the debug-only "Developer Support" button in the picker header, added alongside the non-dismissable picker change) instead of the blocked \`MORE_OPTIONS_BUTTON\` path.
- **\`LoginPageObject.changeServerByUrl()\`**: detects whether the picker is already showing (e.g. left up by \`backOutToLoginActivity\`) and skips the open-picker steps, selecting the server directly.

## Test plan

- [x] \`testECAJwt_SubsetScopes_NotHybrid\` — previously failing, now passes (22s on Pixel 9 Pro XL AVD API 35)
- [ ] Broader test suite (other tests that exercise the \`backOutToLoginActivity → openLoginOptions/changeServerByUrl\` path) should be validated in CI

### DIFF
--- a/native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
+++ b/native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
@@ -53,8 +53,8 @@ abstract class BasePageObject(val composeTestRule: ComposeTestRule) {
 
         /**
          * Extended timeout for Espresso WebView actions ([retryWebAction]) that wait for
-         * server-rendered login page content. The Salesforce sandbox login page can take up to
-         * ~30 s to render interactive form elements after [onPageFinished] fires; this budget
+         * server-rendered login page content. The Salesforce sandbox login page can take
+         * 20–30 s to render interactive form elements after [onPageFinished] fires; this budget
          * covers that latency with headroom for both local emulators and Firebase Test Lab.
          */
         val WEBVIEW_ACTION_TIMEOUT_MS: Long by lazy {
--- a/native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
+++ b/native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
@@ -44,11 +44,21 @@ abstract class BasePageObject(val composeTestRule: ComposeTestRule) {
             ) == "true"
         }
         val TIMEOUT_MS: Long by lazy {
-            if (isFtl) 15_000 else 10_000
+            if (isFtl) 20_000 else 15_000
         }
 
         val SLEEP_TIME_MS: Long by lazy {
             if (isFtl) 5_000 else 2_500
+        }
+
+        /**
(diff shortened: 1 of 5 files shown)


native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/BasePageObject.kt
```diff
@@ -44,11 +44,21 @@ abstract class BasePageObject(val composeTestRule: ComposeTestRule) {
...
         }
+
+        /**
+         * Extended timeout for Espresso WebView actions ([retryWebAction]) that wait for
+         * server-rendered login page content. The Salesforce sandbox login page can take up to
+         * ~30 s to render interactive form elements after [onPageFinished] fires; this budget
+         * covers that latency with headroom for both local emulators and Firebase Test Lab.
+         */
+        val WEBVIEW_ACTION_TIMEOUT_MS: Long by lazy {
+            if (isFtl) 60_000 else 45_000
```
brandonpage, line 61: Have you actually seen it take 30 seconds?  I don't think I have.  It really does not matter much since this is the _max_ time it will wait.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/LoginPageObject.kt
```diff
@@ -102,6 +107,76 @@ open class LoginPageObject(composeTestRule: ComposeTestRule): BasePageObject(com
...
+        try {
+            composeTestRule.waitUntil(timeoutMillis = TIMEOUT_MS) {
+                composeTestRule
+                    .onAllNodesWithTag(LoginViewTestTags.MORE_OPTIONS_BUTTON)
+                    .fetchSemanticsNodes()
+                    .isNotEmpty()
+            }
+        } catch (_: androidx.compose.ui.test.ComposeTimeoutException) {
+            // Best-effort: if the top bar is not reachable within the timeout the caller's
+            // subsequent actions will fail with descriptive messages.
```
brandonpage, line 126: NIT: couldn't hurt to log here.

brandonpage (approved):

brandonpage:
We may still need to update the test run timeouts [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/9cb632744b3f8cd10d721e5ae36aea329fd4575e/.github/workflows/reusable-ui-workflow.yaml#L123) and the other 3 places in the same file.

PR #2991, opened 2026-08-13
[Android Crash] Fix EventsObservable.notifyEvent — UPCR impact 0.076%, 86 users affected

JohnsonEricAtSalesforce:
## What

`EventsObservable.notifyEvent(Event)` could throw `ConcurrentModificationException` when an observer's `onEvent` callback registered or unregistered an observer (or fired a nested `notifyEvent`) on the dispatching thread.

## Why it crashed

`EventsObservable extends android.database.Observable<EventsObserver>`, whose backing `mObservers` is an `ArrayList`. `notifyEvent` iterated that list inside `synchronized (mObservers)`. The monitor only serializes *cross-thread* access — it does nothing against *single-thread re-entrancy*: `synchronized` is reentrant for the owning thread, so an observer callback running mid-iteration could re-acquire the lock and structurally modify the list. The next `Iterator.next()` then saw `modCount != expectedModCount` and threw. With two or more observers registered, this reliably crashed.

Observed in the field on Android (Field Service): among the top crashes for the app (~2.3% of crashes, ranked fifth), affecting 86 users.

## Fix

Take a snapshot of the observer list under the lock, then dispatch outside the lock:

```java
final List<EventsObserver> observers;
synchronized (mObservers) {
    observers = new ArrayList<>(mObservers);
}
for (EventsObserver o : observers) {
    o.onEvent(evt);
}
```

Iterating a copy means a re-entrant register/unregister no longer mutates the list being iterated, and invoking callbacks outside the monitor avoids holding the lock across arbitrary observer code. Delivery semantics for a single dispatch are now well-defined: the set of observers notified is fixed at the moment the event is dispatched.

A side effect worth noting: the previous code held the monitor across the entire iteration, which serialized concurrent cross-thread dispatches — a second thread's `notifyEvent` would block until the first thread's full iteration completed. Narrowing the lock to just the snapshot copy is a consequence of dispatching outside the lock, and it means two threads can now dispatch concurrently and may invoke `onEvent` on the same observer instances at the same time. In practice dispatches originate on the main thread, so this is benign. Dispatching outside the lock is kept intentionally for a concrete benefit: it avoids holding the `mObservers` monitor across arbitrary observer callback code, so a blocking or lock-acquiring `onEvent` cannot stall other threads' register/unregister/dispatch, and it removes a lock-ordering deadlock hazard. Structural changes to `mObservers` remain fully serialized.

`notifyEvent(Event)` is `private`, so this is an internal-only change — **no public API surface changes**.

## Tests

New instrumented test `EventsObservableTest` (5 tests), all green locally on the `remaining` shard emulator.

Re-entrancy regression guards — each throws `ConcurrentModificationException` against the old code and passes against the fix:

- [x] `testObserverUnregisteringItselfDuringDispatchDoesNotThrow`
- [x] `testObserverUnregisteringAnotherDuringDispatchDoesNotThrow` — also asserts the unregistered observer, captured in the snapshot, still receives the in-flight event
- [x] `testObserverRegisteringAnotherDuringDispatchDoesNotThrow` — also asserts the observer added mid-dispatch is not in the snapshot and does not receive the event
- [x] `testObserverFiringNestedNotifyDuringDispatchDoesNotThrow`

Delivery-semantics guard — passes on both old and new code, pinning that the snapshot dispatch does not drop deliveries:

- [x] `testAllObserversReceiveEventWhenNoReentrantModification`

## Notes for reviewers

- Cross-thread concurrency is intentionally not covered by a new test: the fixed bug is single-thread re-entrancy, and a timing-coordinated concurrency test would risk flakiness (against CLAUDE.md's no-flaky/no-`sleep` rule). The lock still serializes cross-thread structural changes.
- No iOS twin: iOS uses `NSNotificationCenter` for eventing, with no `EventsObservable`/`notifyEvent` equivalent — this is Android-specific.


libs/SalesforceSDK/src/com/salesforce/androidsdk/util/EventsObservable.java
```diff
@@ -26,9 +26,12 @@
  */
 package com.salesforce.androidsdk.util;
 
+import android.database.Observable;
+
 import com.salesforce.androidsdk.util.test.EventsObserver;
 
-import android.database.Observable;
+import java.util.ArrayList;
```
wmathurin, line 33: Import order: Google Java Style groups imports as `java.*` → `javax.*` → `android.*` → own project. The block here puts `java.*` last. Suggested order:
```java
import java.util.ArrayList;
import java.util.List;

import android.database.Observable;

import com.salesforce.androidsdk.util.test.EventsObserver;
```

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/EventsObservable.java
```diff
@@ -88,9 +91,19 @@ public void notifyEvent(EventType type, Object data) {
...
+         * Dispatch over a snapshot taken under the lock. An observer's onEvent
+         * callback may register or unregister an observer on this same thread;
+         * iterating a copy keeps that re-entrant structural change from throwing
+         * ConcurrentModificationException, and invoking callbacks outside the
+         * lock avoids holding the monitor across arbitrary observer code.
+         */
+        final List<EventsObserver> observers;
+        synchronized (mObservers) {
+            observers = new ArrayList<>(mObservers);
+        }
```
wmathurin, line 104: The original code held the lock across the entire iteration, which had the side effect of serializing concurrent cross-thread dispatches — thread B would block until thread A's full iteration finished. The new code releases the lock before iterating, so two threads can now dispatch simultaneously and call `o.onEvent(evt)` concurrently on the same observer instances. This is almost certainly benign in practice (dispatches are normally on the main thread), but worth a brief note in the description so a future reader knows the change in serialization semantics is intentional.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/EventsObservable.java
```diff
@@ -88,9 +91,19 @@ public void notifyEvent(EventType type, Object data) {
...
+        for (EventsObserver o : observers) {
+            o.onEvent(evt);
         }
```
wmathurin, line 107: Nit: the closing brace has 9 leading spaces instead of 8 — off by one indent.

wmathurin (approved):

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/EventsObservable.java
JohnsonEricAtSalesforce: Thanks Wolf. These imports follow the Android Studio default code style (grouping `android`/`androidx` first, `java.*` last), which we've been using across the module for some time — I ran Optimize Imports and it produced this ordering.

Checking the SalesforceSDK module for files that mix `android.*` and `java.*` imports: **39** use the AS-default order (`android` before `java`) and only **1** uses the Google-Java-Style order (`java` first). Module-wide, `android`/`androidx` leads in 85 of 128 files. So this change matches the prevailing convention here.

Did you mean for us to switch this file to Google Java Style, which would diverge from what the rest of the module uses? Happy to change it if you'd prefer we standardize on that ordering — just want to confirm before making our file the outlier.

wmathurin: Not really. It's fine this way I guess.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/EventsObservable.java
JohnsonEricAtSalesforce: Thanks Wolf — I double-checked the leading whitespace byte-by-byte and the closing brace on this line is actually 8 spaces, consistent with the method body. I think what reads as an off-by-one is the block-comment continuation lines just above (lines 95–100): those sit at 9 columns because a `/* */` block aligns its continuation `*` one space in from the opening `/*` (8 spaces of indent + 1 space + `*`), which is the standard alignment. The braces themselves (lines 102, 104, 105, 107, 108) are all at 8. Let me know if you're seeing something different on your end and I'll take another look.

libs/SalesforceSDK/src/com/salesforce/androidsdk/util/EventsObservable.java
JohnsonEricAtSalesforce: Thanks Wolf — and to be straight about it, the change in cross-thread serialization was a *consequence* of dispatching outside the lock rather than something we set out to design. You're right that the old code serialized cross-thread dispatches by holding the monitor across the whole iteration, and that our version lets two threads dispatch concurrently and potentially call `onEvent` on the same observer at once. In practice dispatches originate on the main thread, so that's benign.

We're keeping the dispatch-outside-the-lock form deliberately, though, for one concrete benefit: it never holds the `mObservers` monitor across arbitrary observer callback code. An `onEvent` implementation that blocks or acquires another lock can't then stall every other thread's `register`/`unregister`/dispatch, and it removes a lock-ordering deadlock hazard. The monitor is now confined to the short snapshot copy, while structural changes to `mObservers` stay fully serialized. (Iterating the copy *inside* the lock would also fix the CME and preserve the old serialization, but at the cost of reintroducing that held-across-callbacks risk.) I've updated the PR description's **Fix** section to describe this accurately.

PR #2992, opened 2026-08-13
fix(W-23695429): escape single quotes in SmartSQL IN-clause record IDs

wmathurin:
## Summary
- Escapes single quotes in the ID lists passed to `deleteRecordsFromLocalStore` in `SyncTarget.kt`
- The IDs are joined into a SmartSQL `IN ('id1','id2',...)` clause; without escaping a single quote in an ID could break the query
- Salesforce record IDs are 15/18-char alphanumeric so actual risk is zero, but this satisfies scanner rule BUG-03 (SQL injection, P3) filed as W-23695429

## Test plan
- [ ] Build: `./gradlew :libs:MobileSync:build` — confirmed passing
- [ ] Run tests: `./gradlew :libs:MobileSync:connectedAndroidTest`
- [ ] Verify sync-down + clean-ghosts still work end-to-end in MobileSyncExplorer sample app

### DIFF
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncTarget.kt
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncTarget.kt
@@ -274,9 +274,10 @@ abstract class SyncTarget @JvmOverloads constructor(
         idField: String?
     ) {
         if (ids.isNotEmpty()) {
+            val escapedIds = ids.map { it.replace("'", "''") }
             val smartSql =
                 "SELECT {$soupName:${SmartStore.SOUP_ENTRY_ID}} FROM {$soupName} WHERE {$soupName:$idField} IN ('${
-                    join(/* delimiter = */ "', '", /* tokens = */ ids)
+                    join(/* delimiter = */ "', '", /* tokens = */ escapedIds)
                 }')"
             val querySpec = QuerySpec.buildSmartQuerySpec(smartSql, Int.MAX_VALUE)
             syncManager.smartStore.deleteByQuery(soupName, querySpec)
--- a/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncTargetHelper.kt
+++ b/libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncTargetHelper.kt
@@ -200,7 +200,7 @@ object ParentChildrenSyncTargetHelper {
             childrenInfo.soupName, childrenInfo.parentIdFieldName,
             parentInfo.soupName, parentInfo.idFieldName,
             parentInfo.soupName, parentInfo.idFieldName,
-            "'${join("', '", parentIds)}'"
+            "'${join("', '", parentIds.map { it.replace("'", "''") })}'"
         )
         return QuerySpec.buildSmartQuerySpec(smartSql, Int.MAX_VALUE)
     }
--- a/libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncTargetHelperTest.kt
+++ b/libs/test/MobileSyncTest/src/com/salesforce/androidsdk/mobilesync/target/ParentChildrenSyncTargetHelperTest.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2025-present, salesforce.com, inc.
+ * All rights reserved.
+ * Redistribution and use of this software in source and binary forms, with or
+ * without modification, are permitted provided that the following conditions
+ * are met:
+ * - Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of salesforce.com, inc. nor the names of its contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission of salesforce.com, inc.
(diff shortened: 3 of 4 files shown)


libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncTarget.kt
```diff
@@ -274,9 +274,10 @@ abstract class SyncTarget @JvmOverloads constructor(
         idField: String?
     ) {
         if (ids.isNotEmpty()) {
+            val escapedIds = ids.map { it.replace("'", "''") }
```
JohnsonEricAtSalesforce, line 277: Optional coverage suggestion (not blocking): consider a unit test that passes an ID containing a single quote and asserts the generated SmartSQL doubles it, and that clean-ghosts still deletes the intended rows. The iOS twin (PR 4131) also shipped without a test, so this is not a merge blocker -- just a coverage suggestion for a security-relevant path.

JohnsonEricAtSalesforce, line 277: Follow-up (not blocking): while confirming every caller is covered, I noticed a sibling local-store SmartSQL path with the same trust boundary that this PR does not touch and that is still unescaped on both platforms. Suggest a follow-up work item to apply the same `'` -> `''` escaping there (Android and iOS together):

- **Android:** `ParentChildrenSyncTargetHelper.getQueryForChildren` builds `... AND {parent:idField} IN (%s)` via `"'${join("', '", parentIds)}'"` (around lines 197-203) and feeds `smartStore.deleteByQuery` / `smartStore.query`. `parentIds` originates from `record.getString(idFieldName)` in `ParentChildrenSyncUpTarget` (around line 264) -- server-sync-down-derived, i.e. the same source as the IDs this PR just hardened.
- **iOS:** the analog is `SFParentChildrenSyncHelper.m` `getQueryForChildren:...` (dev, around lines 131-137), `IN (%@)` via `[parentIds componentsJoinedByString:@"', '"]`. PR 4131 did not touch it.

Same low real-world risk as this fix (record IDs come from the Salesforce API), but escaping it too would close the CWE-89 pattern consistently across the sync targets on both platforms.

JohnsonEricAtSalesforce (approved):
Reviewed at HEAD 1c5c37c. This escapes single quotes (`'` -> `''`) in each record ID before they are joined into the SmartSQL `IN (...)` literal in `SyncTarget.deleteRecordsFromLocalStore`. That is the correct SQLite/SmartSQL escaping, and it matches the already-approved-and-merged iOS twin (PR 4131, same W-23695429). Approving.

**What I verified**
- **Correctness:** `String.replace("'", "''")` replaces every literal occurrence; `SmartSqlHelper` only rewrites `{soup:path}` tokens, so quoted string literals (and the doubled quotes) pass through to SQLite unchanged. There is no bind-parameter API on the SmartSQL path, so escaping is the correct and only remediation here.
- **Callers:** all three callers of `deleteRecordsFromLocalStore` (`SyncDownTarget.cleanGhosts`, `ParentChildrenSyncDownTarget.cleanGhosts`, `BriefcaseSyncDownTarget`) flow through the single escaped path.
- **Cross-platform parity:** same escaping semantics as the approved iOS change.
- **Backward compatibility:** internal method-body change only; no public API, signature, string, dependency, permission, or build change.
- **CI:** the two red checks look environmental rather than a regression from this change. The `unit-tests-pr (MobileSync)` failures are dominated by `REQUEST_LIMIT_EXCEEDED` (shared test-org API quota) with NullPointerException / "Creates failed" cascades downstream, concentrated in sync-up tests that this sync-down change does not touch; the same failures appear on unrelated PRs in the same window, and the dev nightly (~9h later, after quota reset) is green (251/251 on API 31/32/33/34/36). The `ui-tests-pr` failures are the known AuthFlowTester Custom-Tab / creds-section flake. Recommend re-running both jobs on a fresh quota window before merge.

One follow-up worth filing (out of scope for this surgical fix): a sibling local-store SmartSQL path with the same trust boundary is still unescaped on both platforms -- see the inline note.

libs/MobileSync/src/com/salesforce/androidsdk/mobilesync/target/SyncTarget.kt
wmathurin: {'reply': 'Done — added `SyncTargetSingleQuoteTest` (`@LargeTest`) in commit `9124c40`. The test upserts two records into the accounts soup (IDs `"001\'Quote"` and `"001Normal"`), calls `deleteRecordsFromLocalStore` for the single-quote ID, and asserts only `"001Normal"` remains. Test runs green on device.', 'resolve': True}

wmathurin: {'reply': 'Done in the same commit `9124c40` — `getQueryForChildren` in `ParentChildrenSyncTargetHelper.kt` now escapes single quotes in `parentIds` (`it.replace("\'", "\'\'")`) before building the IN-clause, matching the pattern applied to `deleteRecordsFromLocalStore` in this PR. A `@SmallTest` (`ParentChildrenSyncTargetHelperTest`) asserts that `"001\'Test"` becomes `"001\'\'Test"` in the generated `smartSql`. Test runs green on device.', 'resolve': True}

JohnsonEricAtSalesforce:
Thanks for the quick turnaround, Wolf — appreciate you folding both notes into `9124c40`. The `getQueryForChildren` escaping and the two regression tests close the loop nicely on the Android side.

One thing to keep on the radar: the iOS twin (PR 4131) merged without the parent-children escaping, so that sibling path is still unescaped on iOS. Might be worth a small follow-up there to keep the two platforms in parity.

PR #2993, opened 2026-08-13
fix(W-23622931): escape single quotes in SmartSQL json_extract path and FTS MATCH value

wmathurin:
## Summary
- **Sink 1 — `json_extract` path** (`SmartSqlHelper.java:176`): escapes `'` → `''` in the soup path before building `json_extract(soup, '$.path')`. A quote in the path broke out of the SQL string literal, allowing SmartSQL injection via a non-indexed query path (CWE-89).
- **Sink 2 — FTS MATCH value** (`QuerySpec.java:384`): escapes `'` → `''` in the qualified match key before inlining into `MATCH 'value'`. Statement arg binding doesn't work for FTS MATCH so inlining is unavoidable; escaping prevents breaking out of the surrounding string literal.
- Both sinks are reachable from the SmartStore Cordova bridge in hybrid apps (W-23622931).

## Test plan
- [ ] Build SmartStore: `./gradlew :libs:SmartStore:build` — confirmed passing
- [ ] Run SmartStore tests: `./gradlew :libs:SmartStore:connectedAndroidTest` — confirmed all passing

### DIFF
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartSqlHelper.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartSqlHelper.java
@@ -173,7 +173,7 @@ public class SmartSqlHelper  {
 
 		if (!indexed) {
 			// Thanks to the json1 extension we can query the data even if it is not indexed
-			columnName = "json_extract(" + SmartStore.SOUP_COL + ", '$." + path + "')";
+			columnName = "json_extract(" + SmartStore.SOUP_COL + ", '$." + path.replace("'", "''") + "')";
 		} else {
 			try {
 				columnName = DBHelper.getInstance(db).getColumnNameForPath(db, soupName, path);
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/QuerySpec.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/QuerySpec.java
@@ -381,7 +381,7 @@ public class QuerySpec {
             case match:
                 pred = computeFieldReference(SmartStore.SOUP_ENTRY_ID) + " IN ("
                         + SELECT + SmartStore.ROWID_COL + " " + FROM + computeSoupFtsReference() + " " + WHERE
-                        + computeSoupFtsReference() + " MATCH '" + qualifyMatchKey(field, matchKey) + "'"
+                        + computeSoupFtsReference() + " MATCH '" + qualifyMatchKey(field, matchKey).replace("'", "''") + "'"
                         // statement arg binding doesn't seem to work so inlining matchKey
                         + ") ";
                 break;
--- a/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
+++ b/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartStore.java
@@ -376,7 +376,7 @@ public class SmartStore  {
             // Column name or expression the db index is on
             String columnName = soupTableName + "_" + i;
             if (TypeGroup.value_indexed_with_json_extract.isMember(indexSpec.type)) {
-                columnName = "json_extract(" + SOUP_COL + ", '$." + indexSpec.path + "')";
+                columnName = "json_extract(" + SOUP_COL + ", '$." + indexSpec.path.replace("'", "''") + "')";
             }
 
             // for create table
--- a/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/QuerySpecTest.java
+++ b/libs/test/SmartStoreTest/src/com/salesforce/androidsdk/smartstore/store/QuerySpecTest.java
@@ -167,6 +167,15 @@ public class QuerySpecTest {
     }
 
     @Test
+    public void testMatchQuerySmartSqlWithSingleQuoteInMatchKey() {
+        // Single quotes in matchKey must be doubled so they don't break the surrounding MATCH '...' literal.
+        QuerySpec querySpec = QuerySpec.buildMatchQuerySpec("employees", "lastName", "O'Brien", "firstName", QuerySpec.Order.ascending, 1);
+        Assert.assertEquals("Wrong smart sql for match query spec with single quote in matchKey",
+            "SELECT {employees:_soup} FROM {employees} WHERE {employees:_soupEntryId} IN (SELECT rowid FROM {employees}_fts WHERE {employees}_fts MATCH '{employees:lastName}:O''Brien') ORDER BY {employees:firstName} ASC ",
+            querySpec.smartSql);
+    }
+
+    @Test
     public void testQualifyMatchKey() {
         Assert.assertEquals("Wrong qualified match query", "abc", QuerySpec.qualifyMatchKey(null, "abc"));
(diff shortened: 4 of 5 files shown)


libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartSqlHelper.java
```diff
@@ -173,7 +173,7 @@ private String getColumnNameForPathForSmartSql(SQLiteDatabase db, String soupNam
 
 		if (!indexed) {
 			// Thanks to the json1 extension we can query the data even if it is not indexed
-			columnName = "json_extract(" + SmartStore.SOUP_COL + ", '$." + path + "')";
+			columnName = "json_extract(" + SmartStore.SOUP_COL + ", '$." + path.replace("'", "''") + "')";
```
JohnsonEricAtSalesforce, line 176: The escaping looks correct. As a follow-on, `SmartSqlTest` currently has no case with a `'` inside a non-indexed path (or a UNION/`--` payload), so this behavior would regress silently if the escape were ever dropped. A small test — a non-indexed path containing `'` asserting the generated SQL doubles the quote (and the FTS `MATCH` equivalent for `QuerySpec.java:384`) — would lock the fix in and matches the project's "failing test first" guidance for bug fixes. Non-blocking.

JohnsonEricAtSalesforce (changes requested):
Thanks for tightening these SmartSQL sinks, Wolfgang. I traced the query-construction path and the fix is correct on both sinks it targets:

- **`SmartSqlHelper.java:176`** (non-indexed `json_extract` path) — every `{soup:path}` reference in smart/exact/like/range/order queries routes through `convertSmartSql` into this line, so escaping here covers those callers. Inside the `'$.<path>'` string literal, `'` is the only metacharacter and `''` is the correct SQLite escape, so the break-out is closed. (The `$`→`\$` handling at line 149 is orthogonal regex-replacement escaping and is unaffected.)
- **`QuerySpec.java:384`** (FTS `MATCH` value) — this is the one query type that inlines its key rather than binding a `?` arg (`getArgs()` returns `null` for `match`; exact/like/range bind params and are safe). Escaping the **output** of `qualifyMatchKey(...)` — i.e. after the `field:` qualifiers are inserted — is the right layer; escaping earlier would let the qualifier regex re-tokenize the doubled quotes.

I confirmed platform parity: the iOS twin (`SalesforceMobileSDK-iOS` PR 4132, W-23622931, merged to `dev` on 2026-08-14) makes the identical change at the same two sinks and the same layer, so this PR is consistent with the shipped iOS fix.

### Requested change — fold in the third `json_extract` sink (`SmartStore.java:379`)

A third, structurally identical sink is left unescaped in `SmartStore.java:379` (index-registration path):

```java
columnName = "json_extract(" + SOUP_COL + ", '$." + indexSpec.path + "')";
```

Here `indexSpec.path` is concatenated unescaped. It's reachable from the same hybrid attack surface: JS → `SmartStorePlugin` `pgRegisterSoup` → `IndexSpec.fromJSON` (which does no validation on `path`, `IndexSpec.java:44`) → this DDL, where the column expression is baked into `CREATE TABLE`/`CREATE INDEX` and persisted in `soup_index_map`. A `'` in the index path breaks out of the literal the same way — same CWE-89 class, just at registration time instead of query time. Since it's the identical fix in the same file, could we fold it into this PR so the sink family is fully closed?

```java
columnName = "json_extract(" + SOUP_COL + ", '$." + indexSpec.path.replace("'", "''") + "')";
```

I confirmed this line isn't being addressed in any other open/merged PR or GUS item. For cross-platform parity, note the iOS twin (PR 4132) fixed the two query-time sinks but left the equivalent `SFSmartStore.m:1120` index-path sink untouched, so a matching iOS follow-up would keep the platforms aligned.

**Non-blocking:** No regression test accompanies the escaping — see the inline note on `SmartSqlHelper.java:176`.

Reviewed at HEAD `7e31f066`.

libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/SmartSqlHelper.java
wmathurin: {'reply': "Done — both regression tests have been added and pass on device:\n\n- `SmartSqlTest.testConvertSmartSqlForNonIndexedColumnWithSingleQuoteInPath` — verifies that a `'` in a non-indexed `json_extract` path is doubled (covers the fix in SmartSqlHelper.java:176)\n- `QuerySpecTest.testMatchQuerySmartSqlWithSingleQuoteInMatchKey` — verifies that a `'` in an FTS MATCH `matchKey` is doubled (covers the fix in QuerySpec.java:384)\n\nBoth tests are on branch `W-23622931-smartsql-injection-escaping` and ran successfully on the emulator.", 'resolve': True}

JohnsonEricAtSalesforce: Thanks — verified both. `testConvertSmartSqlForNonIndexedColumnWithSingleQuoteInPath` asserts the generated SQL contains `json_extract(soup, '$.user''s.address')`, and `testMatchQuerySmartSqlWithSingleQuoteInMatchKey` asserts `MATCH '{employees:lastName}:O''Brien'` — each fails if the escape is ever dropped, so the fix is locked in on both sinks. Good to resolve this thread.

JohnsonEricAtSalesforce:
The regression tests address the non-blocking note — thanks. The one requested change is still open: the third `json_extract` sink at `SmartStore.java:379` (`indexSpec.path`, index-registration path) is unescaped and reachable from the same hybrid bridge via `pgRegisterSoup` → `IndexSpec.fromJSON` (no path validation). It's the same CWE-89 construction you've already closed at the two query-time sinks, so folding in the one-liner there —

```java
columnName = "json_extract(" + SOUP_COL + ", '$." + indexSpec.path.replace("'", "''") + "')";
```

— would close the sink family completely rather than leaving a registration-time sibling. (I confirmed no other open/merged PR or work item covers this line. The iOS twin PR left the equivalent `SFSmartStore.m:1120` sink untouched too, so a matching iOS follow-up would keep the platforms in parity.) Keeping the review as `CHANGES_REQUESTED` on that item.

JohnsonEricAtSalesforce (approved):
Thanks, Wolf — the requested change is in. `7d81420e` escapes `indexSpec.path` at the index-registration sink (`SmartStore.java:379`) with the same `'` → `''` doubling used at the two query-time sinks, so all three `json_extract` / `MATCH` construction points now neutralize the single-quote break-out consistently.

Confirming the full picture on this PR:
- **`SmartSqlHelper.java:176`** (non-indexed `json_extract` path) — escaped, covers every `{soup:path}` reference routed through `convertSmartSql`.
- **`QuerySpec.java:384`** (FTS `MATCH` value) — escaped on the output of `qualifyMatchKey(...)`, the correct layer; the non-FTS query types bind `?` args and were never exposed.
- **`SmartStore.java:379`** (index-registration path via `pgRegisterSoup`) — now escaped, closing the last sibling in the family.
- Regression tests (`50c3c1b3`) lock in the escaping at the two query-time sinks by asserting the doubled quote in the generated SQL.

Platform note for the record: the iOS twin (`SalesforceMobileSDK-iOS` PR for the same work item) fixed the two query-time sinks but not the equivalent `SFSmartStore.m:1120` index-registration sink — a matching iOS follow-up would bring the platforms back to parity on this last one. Not a gate on this Android PR.

Approving. Reviewed at HEAD `7d81420e`.

PR #2997, opened 2026-08-17
Support biometric authentication with advanced authentication

brandonpage:
## Description

Suppress Advanced Authentication (Custom Tab) to show OS Biometric Prompt if locked.  Add a fallback support to the biometric app-lock flow to launch AA.  Previously, when a biometric-locked user encountered AA there was no way to actually use biometric of username / password.

## Changes

- `AppLockManager` — re-lock guard.
- `LoginActivity` / `LoginViewModel` — drive the advanced-auth fallback path when the biometric prompt is dismissed without success.
- `LoginView` / `PickerBottomSheet` (+ test tags) — UI support for the fallback landing/picker.
- Tests for `BiometricAuthenticationManager`, `LoginActivity`, `LoginView`, and `PickerBottomSheet`.

## Review notes

- **Login UI / auth flow**: This touches the login UI and authentication flow — please review with the OAuth/login-flow escalation checklist in mind.
- **Both platforms**: The corresponding iOS change is in progress and will be submitted as a separate PR.

## Testing

- New and updated unit/instrumented tests in `libs/test/SalesforceSDKTest`.


libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
```diff
@@ -885,14 +907,32 @@ open class LoginActivity : FragmentActivity() {
...
+        } else {
+            loadLoginPageInCustomTab(url, customTabLauncher)
+        }
+    }
+
     // endregion
     // region Biometric Authentication Code
 
     private fun presentBiometric() {
         val biometricPrompt = biometricPrompt
```
wmathurin: `suppressInitialCustomTabLaunch` is armed before this capability check, but every terminal `canAuthenticate()` branch here returns without clearing it or starting the Advanced Auth fallback. If capability changed between the activity's initial eligibility check and this call (for example `BIOMETRIC_ERROR_HW_UNAVAILABLE` or `BIOMETRIC_ERROR_NONE_ENROLLED`), `handleBrowserCustomTabReady()` captures the authorization URL and never launches it, leaving the user stranded. Route these terminal outcomes through `onBiometricPromptDismissedWithoutSuccess()` (or equivalent) and add instrumented coverage for hardware-unavailable and no-enrollment cases.

wmathurin (commented):
Inline review comment.

libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
brandonpage: That scenario seems pretty unlikely, but shouldn't be too hard to fix.  👍

JohnsonEricAtSalesforce (approved):
**Approve.** One open question below (finding 1) worth your confirmation; the rest are minor and non-blocking.

### What I checked

- **Lock blocks the auto Advanced Auth launch and routes to biometric first**, including every non-success/error/cancel branch falling back to AA correctly — this looks like it resolves wmathurin's earlier comment about `HW_UNAVAILABLE`/`NONE_ENROLLED` potentially stranding the user.
- **Retry-biometric button** is correctly scoped to the login-server picker only (verified it doesn't leak onto the account-switcher picker) and reuses the existing biometric action rather than duplicating it.
- Tried to break it via rotation, process death, a retry-button race, and the admin-login path — didn't find a bypass in any of them.
- Ran this through the repo's full Mobile SDK review checklist (public API, OAuth/credential safety, SmartStore/SQLCipher, multi-user, localization, platform hygiene, sample apps, test coverage) — nothing else surfaced beyond what's below.

### Comments

1. **Question:** `AppLockManager.onAppForegrounded()`'s new `!locked` guard means `locked` needs to reliably clear once the Advanced Auth fallback completes. I could only find that happening via `handleDuplicateUserAccount`'s `onUnlock()`, which is gated on the refresh token having changed. If a locked user cancels the biometric prompt, falls back to AA, and re-authenticates as the *same* user — does that re-auth always rotate the refresh token? If not, `locked` could stay stuck `true`, which would also block token refresh (`RestClient`'s refresh path checks `shouldAllowRefresh()`). Flag me if server-side rotation is guaranteed here and this is moot.
2. Low-severity/low-reachability: the suppression flag is one-shot, so a second `generateAuthorizationUrl()` completion while the prompt is still up could launch a Custom Tab over it.
3. `LoginView.kt` has the same loading-suppression comment + condition written out twice (outer and inner composable) — could be factored into one shared val.
4. A few comments reference an internal work item ID, and there's an orphaned "Bug A" label with no "Bug B" — worth a quick cleanup pass before merge.
5. `PickerBottomSheetTest.kt`'s `loginPicker_retryBiometric_hiddenWhenBiometricUnavailable` has a body byte-identical to `hiddenWhenNotLocked` right above it — your own comment there already notes it can't independently exercise the "unavailable" scenario. Might be worth removing it or making it clearly a duplicate-coverage note rather than implying it tests something distinct.
6. `LoginActivityTest.kt`'s `test_launchLoginForAdmins_whenLocked_stillLaunchesCustomTab` doesn't actually establish a locked state in its setup, and `launchLoginForAdminsAction()` doesn't consult lock state either — so the test would pass the same whether locked or not. Might be worth renaming it to reflect what it actually covers (that LFA launches regardless of lock state, without needing to simulate locked specifically).

### FYI, not a defect here

Your own test comment already flags that `showBiometricAuthenticationButton` doesn't check hardware/enrollment state via `BiometricManager.canAuthenticate()`. Pre-existing behavior this PR reuses as-is — just flagging in case it's worth tracking separately.

---

brandonpage:
> 1. **Question:** `AppLockManager.onAppForegrounded()`'s new `!locked` guard means `locked` needs to reliably clear once the Advanced Auth fallback completes. I could only find that happening via `handleDuplicateUserAccount`'s `onUnlock()`, which is gated on the refresh token having changed. If a locked user cancels the biometric prompt, falls back to AA, and re-authenticates as the _same_ user — does that re-auth always rotate the refresh token? If not, `locked` could stay stuck `true`, which would also block token refresh (`RestClient`'s refresh path checks `shouldAllowRefresh()`). Flag me if server-side rotation is guaranteed here and this is moot.

The server does not "rotate" the token like a refresh flow with RTR enabled, but we effectively do the same on the app side.  A new username / password login through the WebView/AA of an existing user creates an entirely new session.  We use the guard not only to unlock the user but also to revoke the refresh token from the older session when we set the new one we just received.  

I will address other comments in a separate response.

JohnsonEricAtSalesforce:
Thanks for the explanation — that resolves finding 1 for me. Revoking the prior session's refresh token when the new one is stored achieves the same effect as rotation, so `AppLockManager.locked` won't get stuck on a same-user re-auth through Advanced Auth. No further action needed on my end for that item.

Looking forward to your response on the remaining comments.

brandonpage:
@JohnsonEricAtSalesforce Follow up responses: 

> 2. Low-severity/low-reachability: the suppression flag is one-shot, so a second `generateAuthorizationUrl()` completion while the prompt is still up could launch a Custom Tab over it.

I don't think that can happen since the app is in a locked state.  The user has to bio auth or login again to unlock.  Cold boot will reset the flag.

> 3. `LoginView.kt` has the same loading-suppression comment + condition written out twice (outer and inner composable) — could be factored into one shared val.

Despite being the same comparison I am not sure we should combine them.  They live in different composable scopes, so a shared local val isn't possible.  They could be extracted to a single pure helper instead, but it is sorta goofy at that point:
```
  internal fun showLoadingIndicator(loading: Boolean, biometricPromptShowing: Boolean) =
      loading && !biometricPromptShowing
```

> 4. A few comments reference an internal work item ID, and there's an orphaned "Bug A" label with no "Bug B" — worth a quick cleanup pass before merge.

Thank you!  I thought I had gotten rid of all of those WI references.  

> 5. `PickerBottomSheetTest.kt`'s `loginPicker_retryBiometric_hiddenWhenBiometricUnavailable` has a body byte-identical to `hiddenWhenNotLocked` right above it — your own comment there already notes it can't independently exercise the "unavailable" scenario. Might be worth removing it or making it clearly a duplicate-coverage note rather than implying it tests something distinct.

👍 

> 6. `LoginActivityTest.kt`'s `test_launchLoginForAdmins_whenLocked_stillLaunchesCustomTab` doesn't actually establish a locked state in its setup, and `launchLoginForAdminsAction()` doesn't consult lock state either — so the test would pass the same whether locked or not. Might be worth renaming it to reflect what it actually covers (that LFA launches regardless of lock state, without needing to simulate locked specifically).

👍 

Comments 4, 5 and 6 are addressed in commit [fb0e382b2](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2997/commits/fb0e382b2d9a87c74f407d609cd809c84ada16bf).

PR #3004, opened 2026-08-24
Add downgradeFromDPoP for in-place DPoP→Bearer migration

sfdctaka:
## What

Adds `UserAccountManager.downgradeFromDPoP(userAccount, onSuccess, onFailure)`, the inverse of the existing `upgradeToDPoP`. It rolls a DPoP-bound session back to unbound Bearer **in place** — same connected app, redirect URI, and scopes — independent of the global `useDPoP` flag. It is a **no-op** (invokes `onSuccess` with the unchanged account) when the session is already Bearer.

## Why

Completes the in-place DPoP migration pair so apps can move a user off DPoP without a full logout/login, mirroring `upgradeToDPoP`.

## Changes

- **`UserAccountManagerExtension`** — new `downgradeFromDPoP` extension delegating to `migrateRefreshToken` with `useDPoP = false`; no-op guard when already Bearer. On success it deletes the obsolete DPoP key pair + nonce-cache entries keyed to the pre-downgrade credentials. Also adds a matching no-op guard to the existing `upgradeToDPoP` (returns the unchanged account when already DPoP-bound).
- **AuthFlowTester** — "Downgrade from DPoP" action (enabled only for a DPoP-bound session), downgrade UI test, and a downgrade validation helper. Removes an obsolete DPoP test-server entry.
- **Custom Tab login test fix** — a Salesforce My Domain page renders username + password on one screen; the previous unconditional "advance" tap submitted an empty password and the position-based lookup then typed the password into the username field. The helper now advances only on a genuine two-step page, targets the password field explicitly, and dismisses the soft keyboard before submitting.

## Review notes

- Touches a public API and the credential/login-UI surface — requesting maintainer (Wolf) review before merge; kept as **draft**.
- New user-facing strings are in AuthFlowTester's own `strings.xml` (sample-app labels for the downgrade button), not the SDK's `sf__strings.xml`.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
@@ -35,7 +35,6 @@ import com.salesforce.androidsdk.auth.dpop.DPoPNonceCache
 import com.salesforce.androidsdk.config.OAuthConfig
 import com.salesforce.androidsdk.ui.TokenMigrationActivity
 import com.salesforce.androidsdk.util.SalesforceSDKLogger
-import com.salesforce.androidsdk.util.SalesforceSDKLogger.w
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.Default
 import kotlinx.coroutines.launch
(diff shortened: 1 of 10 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
```diff
@@ -30,9 +30,12 @@ import android.content.Intent
 import com.salesforce.androidsdk.accounts.UserAccountManager.getInstance
 import com.salesforce.androidsdk.app.SalesforceSDKManager
 import com.salesforce.androidsdk.auth.ScopeParser.Companion.toScopeParser
+import com.salesforce.androidsdk.auth.dpop.DPoPKeyManager
+import com.salesforce.androidsdk.auth.dpop.DPoPNonceCache
 import com.salesforce.androidsdk.config.OAuthConfig
 import com.salesforce.androidsdk.ui.TokenMigrationActivity
 import com.salesforce.androidsdk.util.SalesforceSDKLogger
+import com.salesforce.androidsdk.util.SalesforceSDKLogger.w
```
wmathurin: The rest of the file uses `SalesforceSDKLogger.e(TAG, ...)` (class-qualified call) throughout. Importing `w` as a bare top-level function here is slightly inconsistent. Suggest dropping the import and calling `SalesforceSDKLogger.w(TAG, ...)` directly instead — smaller diff, more uniform.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManagerExtension.kt
```diff
@@ -205,6 +216,119 @@ fun UserAccountManager.upgradeToDPoP(
...
+        onFailure(message, null, null)
+        return
+    }
+
+    // Capture the pre-migration credentials identifier now: migration mints a NEW identifier for
+    // the account, so reading userAccount.credentialsIdentifier after the migration completes
+    // would return the wrong (or a null) value. This is what obsolete-DPoP-state cleanup below is
+    // keyed on.
+    val oldCredId = userAccount.credentialsIdentifier
+    val wasDPoPBound = userAccount.tokenType == DPoPKeyManager.DPOP_TOKEN_TYPE
```
wmathurin: `wasDPoPBound` is always `true` here — the function returns early at the top of `downgradeFromDPoP` if `userAccount.tokenType != DPoPKeyManager.DPOP_TOKEN_TYPE`, so by the time execution reaches this line we're guaranteed to be in the DPoP branch. The variable and the `if (wasDPoPBound)` guard two lines below are dead code.

Suggest removing both and simplifying to:

```kotlin
val oldCredId = userAccount.credentialsIdentifier

val onSuccessWithCleanup: (userAccount: UserAccount) -> Unit = { migratedUser ->
    oldCredId?.takeIf { it.isNotEmpty() }?.let { id ->
        runCatching {
            DPoPKeyManager.deleteKeyPair(DPoPKeyManager.aliasForCredentialsIdentifier(id))
            DPoPNonceCache.clear(id)
        }.onFailure { e ->
            w(TAG, "Failed to delete obsolete DPoP state on downgrade", e)
        }
    }
    onSuccess(migratedUser)
}
```

wmathurin (approved):

PR #3005, opened 2026-08-25
fix(dpop): enable pool-server DPoP login; fix identity fetch for all DPoP flows

wmathurin:
## Summary

- Re-enables \`testECAJwtDPoP_ViaLoginPoolServer\` (W-23864247 resolved — server now accepts \`dpop_jkt\` from pool-server logins)
- Fixes DPoP identity-fetch URL strategy so all 13 \`DPoPLoginTests\` pass
- Fixes pre-existing \`LegacyLoginTests\` Bearer baseline failure introduced in \`5d974172\`: CA_OPAQUE all-defaults path now opens Login Options to trigger \`loginDevMenuReload\`, regenerating the OAuth URL without \`dpop_jkt\`

## Root cause (DPoP identity fetch)

Salesforce always returns the pool-server host (e.g. \`login.test1.pc-rnd.salesforce.com\`) in the \`id\` field of every token response, regardless of which server actually issued the token. \`idUrlWithInstance\` corrects this by replacing the \`id\` host with the issuing server's host.

For **My Domain logins**, the token was issued by My Domain. The correct identity URL is \`idUrlWithInstance\` (My Domain host).

For **pool-server DPoP logins**, \`idUrlWithInstance\` points to My Domain, which rejects the pool-server-issued DPoP token with \`Bad_OAuth_Token\` — it does not issue a \`401 use_dpop_nonce\` challenge the way data endpoints do. The correct URL is raw \`idUrl\` (pool-server host). Bearer tokens are unaffected: My Domain accepts pool-server-issued Bearer tokens at the identity endpoint regardless.

This server-side inconsistency is tracked in **W-23992239** (Auth Protocols team).

## Fix (DPoP identity fetch)

\`fetchUserIdentityWithRetry\` uses \`LoginServerManager.isPoolServer\` to select the URL upfront — a single call, no retry:

- **DPoP + pool-server login** → raw \`idUrl\` (pool-server identity endpoint accepts its own token)
- **All other cases** → \`idUrlWithInstance\` (My Domain; correct for My Domain logins and all Bearer tokens)

**Why not the iOS approach?** \`SFIdentityCoordinator\` on iOS handles identity 401/403 by triggering a full credential refresh and retrying. That works because the refresh response resets \`credentials.identityUrl\` to the My Domain host as a side effect. But it is RTR-unsafe — it consumes the refresh token unnecessarily to fix a routing error, and the rotated replacement is discarded. The \`isPoolServer\` selection avoids any token refresh entirely.

**Why not a two-attempt retry?** An earlier iteration tried \`idUrlWithInstance\` first and fell back to \`idUrl\`. The \`isPoolServer\` check makes the intent explicit, eliminates the speculative first attempt for pool-server DPoP logins, and is simpler to reason about.

## Additional changes

- **\`DPoPNonceCache.getAny()\` removed**: was a host-agnostic fallback initially added for the identity endpoint. Salesforce only issues \`DPoP-Nonce\` on token-endpoint responses, so the per-host cache is always correctly populated before the identity call. \`getAny()\` was redundant.
- **RTR safety-net test added** (\`testECAJwtDPoP_ViaLoginPoolServer_Rtr\`): verifies that after a pool-server DPoP + RTR login the refresh token is still valid, confirming the identity fetch never touches the refresh token. See W-23991713 for the equivalent iOS investigation.

## Root cause (LegacyLoginTests Bearer baseline)

SDK 14.0 defaults \`useDPoP=true\`. The CA_OPAQUE all-defaults path skipped Login Options, so the OAuth URL built at app startup (with \`dpop_jkt\`) was used for Bearer tests. Opening and dismissing Login Options triggers \`loginDevMenuReload\`, causing \`LoginActivity.onResume\` to regenerate the URL with \`useDPoP=false\`.

## Test results

All 14 AuthFlowTester UI test suites run on emulator (Pixel 9 Pro XL, API 36). **All non-Beacon tests pass.** Beacon tests (\`BEACON_OPAQUE\` / \`BEACON_JWT\`) fail pre-existing across all suites; the same failures are seen on iOS, pointing to an environment or server-side issue with the Beacon app configuration in the test sandbox rather than a code regression.

In \`MultiUserLoginTests\`, 2 non-Beacon tests (\`testFirstDynamic_SecondStatic_DifferentApps\`, \`testMultiUser_revokeOtherUserRefreshToken\`) showed instrumentation crashes when run as part of the full suite after preceding Beacon failures. Both pass when run in isolation.

## Test plan

- [x] All 13 runnable \`DPoPLoginTests\` pass (1 \`@Ignored\` W-22512846 — server-side RTR+JWT blocker)
- [x] \`testECAJwtDPoP_ViaLoginPoolServer\` passes end-to-end (pool-server login with \`dpop_jkt\`, L1 marker, revoke+refresh with DPoP binding)
- [x] \`testECAJwtDPoP_ViaLoginPoolServer_Rtr\` passes (RTR safety net — refresh token survives the identity fetch)
- [x] \`testECAJwtDPoP_Hybrid\` passes (My Domain DPoP — confirms \`idUrlWithInstance\` path)
- [x] All non-Beacon tests pass across all 14 suites
- [x] Beacon failures confirmed pre-existing and cross-platform; the 2 non-Beacon instrumentation crashes in \`MultiUserLoginTests\` pass when run in isolation

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
@@ -416,21 +416,19 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
 /**
  * Fetches user identity, trying idUrlWithInstance first, then falling back to raw idUrl.
  *
- * Attempt 1 — idUrlWithInstance: the instance / My Domain URL. Works for both My Domain
- * logins (nonce keyed to My Domain host = idUrlWithInstance host) and pool-server logins
- * in the same environment (e.g. login.test1.pc-rnd.salesforce.com tokens are accepted by
- * authflowtesting.test1.my.pc-rnd.salesforce.com; the per-host nonce lookup misses but
- * getAny supplies the pool-server nonce which that environment accepts).
+ * Attempt 1 — idUrlWithInstance: the instance / My Domain URL. Works for My Domain logins
+ * and for pool-server logins where the org's My Domain is in the same environment as the
+ * pool server (the My Domain accepts the token and the pool-server nonce supplied via
+ * DPoPNonceCache.getAny).
  *
  * Attempt 2 — raw idUrl (only when idUrl differs from idUrlWithInstance): fallback for
- * pool-server DPoP logins where idUrlWithInstance points to a different-environment
- * production instance that rejects pool-server-issued tokens. The raw idUrl host matches
- * the pool server and has the correct nonce.
+ * cross-environment pool-server DPoP logins where idUrlWithInstance points to a
+ * production instance that rejects pool-server-issued tokens.
  *
- * Upfront URL selection based on LoginServerManager.isPoolServer was investigated but
- * rejected: pool-server login portals do not serve identity endpoint requests themselves,
- * so routing there would silently break identity fetches for pool-server logins where
- * idUrlWithInstance (the org's My Domain) accepts the token and nonce.
+ * Note: iOS uses raw idUrl exclusively (no idUrlWithInstance substitution). Android needs
+ * idUrlWithInstance first because some pool-server login portals in test environments do
+ * not serve the identity endpoint themselves — routing there fails, while the org's My
+ * Domain (idUrlWithInstance) accepts both the token and the nonce.
  *
  * No token refresh is attempted: the access token was just issued by the login flow,
  * so it is valid by construction. A refresh would also be unsafe under Refresh Token
@@ -452,8 +450,8 @@ private suspend fun fetchUserIdentityWithRetry(
 
 /**
  * Calls the identity service using either [TokenEndpointResponse.idUrlWithInstance]
- * (My Domain / instance host) or the raw [TokenEndpointResponse.idUrl] (pool-server host),
- * as selected by [fetchUserIdentityWithRetry].
+ * (My Domain / instance host) or the raw [TokenEndpointResponse.idUrl], as selected
+ * by [fetchUserIdentityWithRetry].
  */
 private suspend fun fetchUserIdentity(
     tokenResponse: TokenEndpointResponse,
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
@@ -414,15 +414,21 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
 }
 
 /**
- * Fetches user identity with a URL fallback for pool-server DPoP logins.
+ * Fetches user identity, trying idUrlWithInstance first, then falling back to raw idUrl.
  *
- * Attempt 1 — idUrlWithInstance: the normal path. Works for regular My Domain logins
- * where the instance host matches the token-exchange host and the DPoP nonce is found.
+ * Attempt 1 — idUrlWithInstance: works for My Domain logins where the DPoP nonce was
+ * issued at the My Domain host (which idUrlWithInstance resolves to). Also works for
+ * classic sandbox logins (test.salesforce.com idUrl) because the nonce is keyed to the
+ * My Domain / instance host, not the legacy id URL host.
  *
- * Attempt 2 — raw idUrl (only when different from idUrlWithInstance): pool-server DPoP
- * logins where idUrlWithInstance points to the production instance, which rejects
- * pool-server-issued tokens. The raw idUrl uses the pool-server host, which matches
- * the nonce harvested at the token endpoint and accepts the token.
+ * Attempt 2 — raw idUrl (only when idUrl differs from idUrlWithInstance): handles
+ * pool-server DPoP logins where the token and nonce were issued by the pool server.
+ * idUrlWithInstance substitutes the production instance host, which rejects pool-server
+ * tokens and has no matching nonce, so Attempt 1 fails and we retry with the raw idUrl.
+ *
+ * We cannot skip Attempt 1 and select the URL upfront based on idUrl != idUrlWithInstance
+ * because that condition also fires for classic-domain sandboxes (e.g. test.salesforce.com
+ * idUrl with a My Domain instanceUrl), where idUrlWithInstance is the correct URL.
  *
  * No token refresh is attempted: the access token was just issued by the login flow,
  * so it is valid by construction. A refresh would also be unsafe under Refresh Token
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
@@ -137,7 +137,7 @@ internal suspend fun onAuthFlowComplete(
     // Note: Can't use default parameter value for a suspended function parameter.
     val actualFetchUserIdentity: suspend (TokenEndpointResponse) -> OAuth2.IdServiceResponse? =
         fetchUserIdentity ?: { tr: TokenEndpointResponse ->
-            fetchUserIdentityWithRetry(tr)
+            fetchUserIdentityWithRetry(tr, loginServer)
         }
 
     if (blockIntegrationUser) {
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -134,8 +134,15 @@ internal suspend fun onAuthFlowComplete(
     // Reset Dev Support LoginOptionsActivity override
     SalesforceSDKManager.getInstance().debugOverrideAppConfig = null
 
-    // Note: Can't use default parameter value for suspended function parameter fetchUserIdentity
-    val actualFetchUserIdentity = fetchUserIdentity ?: ::fetchUserIdentity
+    // Note: Can't use default parameter value for suspended function parameter fetchUserIdentity.
+    // When no override is provided, use fetchUserIdentityWithRetry, which matches iOS
```
brandonpage: NIT:  Do we need to mention the override?  It's only used for mockk testing, right? 

And why phrase this as "iOS behavior" and "iOS refreshes"?  I get that we are matching that code but if they are both developed during the same release we should just be describing what the code here is doing unless it _differs_ from iOS.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -411,16 +418,76 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
+        withContext(Default) {
+            OAuth2.refreshAuthToken(
+                HttpAccess.DEFAULT,
+                URI(loginServer),
+                consumerKey,
+                refreshToken,
+                null,
+                tokenResponse.credentialsIdentifier,
+                tokenResponse.tokenType,
+            )
```
brandonpage: Doesn't this cause an immediate logout with RTR?  The account creation uses the original `TokenEndpointResponse`, not the new one you created here.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPNonceCache.kt
```diff
@@ -48,6 +48,17 @@ object DPoPNonceCache {
...
+    /**
+     * Returns any cached nonce for the given credentials identifier, regardless of host.
+     *
+     * This mirrors iOS `DPoPNonceCache.latest(forScope:)`. It is used as a fallback when
+     * the per-host lookup misses — for example, when calling the identity endpoint at the
+     * instance URL but the nonce was harvested at the token endpoint on a different host
+     * (e.g. a pool server or a non-My-Domain sandbox login URL).
+     */
+    fun getLatest(credentialsIdentifier: String): String? =
+        cache.entries.firstOrNull { it.key.startsWith("$credentialsIdentifier|") }?.value
```
brandonpage: Does this actually guarantee the latest?  The first result from `startsWith` isn't necessarily the latest, right?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -411,16 +418,76 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
 }
 
 /**
- * Helper method to fetch user identity from token response.
+ * Fetches user identity with up to two fallback attempts.
```
brandonpage: My opinion is that if we need three different attempt methods to fetch user identity, this feature should not be shipped to production.

brandonpage (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPNonceCache.kt
sfdctaka: Isn't the assumption here being $credentialsIndentifier being unique which makes an entry, if it exists, there being one and only one?

But in that case `getAny` is misleading. On the second read of the comment though, it says "returns any cached". So does it matter if it's the first or the latest one?

wmathurin: You're right — `ConcurrentHashMap` iteration order is undefined, so `firstOrNull` returns an arbitrary entry, not the chronologically newest one. Renamed to `getAny` in d31c388 to reflect that. For this use case (DPoP nonce fallback when the per-host lookup misses) any cached nonce for the credential is sufficient — the server will reject a stale nonce with a `use_dpop_nonce` error and supply a fresh one, so we just need something to include rather than sending no nonce at all.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Removed in d31c388 — you are correct on both counts. The access token is valid by construction immediately after login, so Attempt 3 was unreachable. And under RTR it was dangerous: calling the refresh endpoint here would consume the just-issued refresh token and discard the rotated replacement, causing an immediate logout on the next token refresh. The two-attempt URL strategy (Attempt 1: `idUrlWithInstance`, Attempt 2: raw `idUrl` fallback for pool-server logins) is sufficient and neither attempt touches the refresh token.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Agreed, and Attempt 3 is removed in d31c388 — see the RTR danger in the adjacent thread. The remaining two attempts are not error recovery or retry logic; they are a URL selection: Attempt 1 uses `idUrlWithInstance` (normal My Domain logins), Attempt 2 falls back to raw `idUrl` only when the two differ (pool-server logins where the instance URL differs from the token-exchange host). Both use the same freshly-issued access token and no token refresh is performed.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Both addressed in d31c388. The iOS phrasing ("iOS behavior", "iOS refreshes") has been stripped from all inline comments — the docstring now describes the Android behavior on its own terms. The `fetchUserIdentity` override parameter is a standard `@VisibleForTesting` injection point; the only comment near it (line 137) explains the Kotlin suspend-function constraint, not iOS behavior.

sfdctaka (approved):
LGTM!

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPNonceCache.kt
brandonpage: The function name should accurately describe what it does, but I am more concerned with the correctness of the code than the name.  Especially if using the wrong nonce makes API requests fail. 

@sfdctaka Why would that assumption be okay in this situation?  That is an honest question, I may just be uninformed.  

> any cached nonce for the credential is sufficient

@wmathurin Have we actually taken all scenarios into account here?  Say the user was locked out by Bio Auth.  They use a different server to re-login (login pool first or second should not matter).   `startsWith("$credentialsIdentifier|")` can match a nonce for the wrong host used for the first login.  You are now using not a stale nonce, but one from a sessions with a revoked refresh token.  Is that a problem?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Stay tuned - tightening the code in the next commit.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -411,16 +414,52 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
+ * Attempt 2 — raw [TokenEndpointResponse.idUrl] (only when it differs from
+ * [TokenEndpointResponse.idUrlWithInstance]): fallback for pool-server DPoP logins where
+ * [TokenEndpointResponse.idUrlWithInstance] points to a production instance that returns
+ * Wrong_Org for pool-server-issued tokens.
+ *
+ * No token refresh is attempted: the access token was just issued by the login flow,
+ * so it is valid by construction. A refresh would also be unsafe under Refresh Token
+ * Rotation — consuming the fresh token and discarding the rotated replacement.
  */
+private suspend fun fetchUserIdentityWithRetry(
```
wmathurin, line 433: @brandonpage — while investigating why iOS passes the My Domain DPoP test without `idUrlWithInstance`, I found that `SFIdentityCoordinator.m` handles identity 401/403 by triggering a full credential refresh and retrying. The reason that retry succeeds is subtle: the token refresh goes directly to My Domain (since `credentials.domain` = My Domain for a My Domain login), and the **refresh token response sets `credentials.identityUrl` to the My Domain host** — not the pool server host that was in the original login response. So the retried identity call lands on My Domain, which accepts its own token.

Android currently avoids that round trip with the two-attempt URL selection here. But the iOS approach is arguably more general: it does not need to predict the right URL upfront, it just corrects `identityUrl` as a side effect of the refresh. The downside is an extra network round trip on every pool-server DPoP login.

Worth aligning Android with iOS here, or is the two-attempt approach the right tradeoff?

brandonpage: @wmathurin  If the extra refresh is actually a more robust solution then I am not opposed to it.  My concern is RTR.  Since we haven't created the user yet we can refresh outside of our new synchronized code as long as the new `TokenEndpointResponse` is passed back and used instead of the one we got from code exchange.

wmathurin: I don't think the way the refresh was done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3005/changes/0a86e5c2c03d0e438a88d96eeb34a00d67288727) was safe with RTR as you pointed out. Now I wonder if iOS has a latent RTR bug.

wmathurin: I'm going to add a test through login pool server using an eca with dpop + rtr on Android and on iOS. And on iOS, it might surface an issue with the current approach. I'll create a story as well.

wmathurin: Update: we dug deeper and resolved this cleanly.

The iOS credential-refresh approach works because the token refresh response sets `credentials.identityUrl` to the My Domain host (the `id` field in a refresh response points to My Domain, not the pool server). So the retry lands on My Domain with My Domain's own token — correct. But it's RTR-unsafe: burning the refresh token to fix a routing error on login is too expensive, and the rotated token would be discarded.

We also found that the root issue is a server bug: My Domain's `/id/` endpoint returns `Bad_OAuth_Token` for pool-server-issued DPoP tokens instead of issuing a `401 use_dpop_nonce` challenge the way data endpoints do. Bearer tokens are unaffected. Filed as W-23992239 against the Auth Protocols team.

On Android, we've since simplified to a direct `isPoolServer` check (latest commit `0a238ca1e`): DPoP + pool-server login → `idUrl`; everything else → `idUrlWithInstance`. Single call, no retry, no speculative first attempt. Works because `LoginServerManager.isPoolServer` already covers all pool-server patterns including internal environments.

We're not aligning with the iOS credential-refresh pattern — the `isPoolServer` approach is cheaper, RTR-safe, and the server fix (W-23992239) should eventually make the whole workaround unnecessary on both platforms.

wmathurin:
**Observation: Android routing logic could potentially be simplified to match iOS**

`SFIdentityCoordinator` on iOS has always used `credentials.identityUrl` (the raw `id` field from the token response) for all identity calls, routing everything through the pool server. All iOS DPoP tests — including pool-server and My Domain logins — pass with that approach.

The current Android fix introduces an `isPoolServer` check to select between `idUrl` and `idUrlWithInstance`. This works, but raises the question: was `idUrlWithInstance` ever strictly necessary, or was it added as a "more correct" routing improvement that inadvertently exposed the W-23992239 server bug for DPoP pool-server logins?

Since all environments have both iOS and Android clients, and iOS routes everything through the pool server without issues, it's worth asking: **should we simplify Android to always use `idUrl` (matching iOS), and remove the `isPoolServer` conditional entirely?**

brandonpage (approved):

PR #3007, opened 2026-08-25
[Android] NativeLoginManager Basic-Auth header uses URL-safe Base64 instead of standard Base64, breaking login for some username/password combinations

JohnsonEricAtSalesforce:
## Summary

`NativeLoginManager`'s helper for building the Basic-Auth `Authorization` header encoded the colon-concatenated username and password with Android's URL-safe Base64 alphabet and no padding. HTTP Basic Authentication (RFC 7617) requires the standard Base64 alphabet with padding. For most credential pairs the two alphabets produce identical output, which is why this went unnoticed — they only diverge when the encoded bytes would contain `+`, `/`, or need `=` padding under the standard alphabet. When that happens, the org's edge rejects the malformed header with a generic error page instead of a normal OAuth error, making the failure look like an infra issue rather than a client-side encoding bug.

- Changed `generateColonConcatenatedBase64String()` to encode with the standard Base64 alphabet (`NO_WRAP` only, dropping `URL_SAFE` and `NO_PADDING`).
- This helper is shared by both `login()` (username/password) and `submitAuthorizationRequest()` (used by the OTP/passwordless flows), so both call sites are fixed by the one change.
- Removed the now-unused `URL_SAFE`/`NO_PADDING` imports.

## Scope note

This does not touch the other `URL_SAFE` usages elsewhere in the SDK (`SalesforceKeyGenerator`, `DPoPProofBuilder`) — those encode DPoP proofs / SHA-256 hashes / code verifiers, which correctly use URL-safe Base64 per their own specs and are unrelated to this HTTP Basic-Auth header.

## Test plan

- [x] Added a regression test covering `login()`, using a username/password pair whose combined UTF-8 bytes diverge between the standard and URL-safe alphabets (contains `/` and requires `=` padding under the standard alphabet); asserts the `Authorization` header matches standard Base64 output.
- [x] Added a regression test covering `submitPasswordlessAuthorizationRequest()` (the OTP/passwordless call site sharing the same helper), same divergent credential pair.
- [x] Ran `NativeLoginManagerTest` on a connected emulator — 23/23 passed, including both new tests.
- [x] Confirmed `SalesforceKeyGenerator.java` and `DPoPProofBuilder.kt`'s `URL_SAFE` usages are unmodified.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt
@@ -33,9 +33,7 @@ import android.content.pm.PackageManager.FEATURE_FACE
 import android.content.pm.PackageManager.FEATURE_IRIS
 import android.net.Uri
 import android.os.Bundle
-import android.util.Base64.NO_PADDING
 import android.util.Base64.NO_WRAP
-import android.util.Base64.URL_SAFE
 import android.util.Base64.encodeToString
 import android.util.Patterns.EMAIL_ADDRESS
 import androidx.biometric.BiometricManager
@@ -722,7 +720,7 @@ internal class NativeLoginManager(
         value2: String
     ) = encodeToString(
         "$value1:$value2".toByteArray(),
-        URL_SAFE or NO_WRAP or NO_PADDING
+        NO_WRAP
     )
 
     /**
--- a/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/NativeLoginManagerTest.kt
+++ b/libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/auth/NativeLoginManagerTest.kt
@@ -470,7 +470,7 @@ class NativeLoginManagerTest {
         verifyLoginRequestAuthorizationHeader(restClient, "Basic $expectedCreds")
     }
 
-    // region Helpers used by attestation tests
+    // region Helpers used by attestation and Authorization-header tests
 
     /**
      * Verifies that the REST client received a login request with the expected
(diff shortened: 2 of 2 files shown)


wmathurin (approved):

brandonpage:
@JohnsonEricAtSalesforce Did we test the scenario of registering with a non-`URL_SAFE` password in MSDK 13 and attempting a login with 14.0?

JohnsonEricAtSalesforce:
The `ui-tests-pr` failures on this PR are pre-existing and unrelated to this change:

- `BeaconLoginTests.kt:76` and `AdvancedAuthBeaconLoginTests.kt:1` — both fail with `AssertionError: Username field not found in Custom Tab`
- `MultiUserLoginTests.kt:83` — fails with `AssertionError: Expected no 'DP' flag for non-DPoP session`

The same three failure signatures (same files/lines, same assertion messages) appear in the [`Nightly Tests` run](https://github.com/forcedotcom/SalesforceMobileSDK-Android/actions/runs/32818208365) against `dev@a3cf3953f` — the exact commit this branch forked from, run before any commit in this PR existed. `Nightly Tests` on `dev` has failed every scheduled run for at least the last three weeks, so this is a chronic flake in the Custom-Tab/DPoP UI suite, not a regression introduced here.

This PR's diff is limited to `NativeLoginManager.kt`'s `generateColonConcatenatedBase64String()` helper and its imports — none of the three failing test files reference that helper, `NativeLoginManager`, or Base64 encoding.

brandonpage:
> The `ui-tests-pr` failures on this PR are pre-existing and unrelated to this change:
> 
> * `BeaconLoginTests.kt:76` and `AdvancedAuthBeaconLoginTests.kt:1` — both fail with `AssertionError: Username field not found in Custom Tab`

I'm about to look into those because they are strange.  We somehow change the boot config values to an invalid config.  The username field can't be found because the login page does not load.

JohnsonEricAtSalesforce:
Manually tested both directions (observed directly, not from the automated suite):

- Password set via Salesforce web, login via the old native app: login fails today (this is the bug). Fixed by this PR — no compat concern.
- Password set via the old native *registration* flow: it shares the same buggy helper as login, so the account gets created with a corrupted stored password. Those users can't log in today either, on any client version, since the corruption happened at registration and predates this fix. A web password reset is required to recover, same as it would be without this change.

So affected passwords are already broken in both directions pre-fix — this PR doesn't newly break anyone who could previously log in.

PR #3010, opened 2026-08-26
Deprecate useWebServerAuthentication (OAuth user agent flow)

brandonpage:
Deprecates `useWebServerAuthentication` on `SalesforceSDKManager` (14.0, for removal in 15.0). The OAuth user agent flow is being retired; the web server flow will be used for WebView login going forward. The flag will be removed in 15.0, or sooner if the server no longer supports the user agent flow.

- Public property annotated `@Deprecated`; backed by a `@Volatile private var _useWebServerAuthentication` field.
- Internal SDK code reads the backing field directly; the two dev-menu/login consumers use a localized `@Suppress("DEPRECATION")` so no internal usage trips a deprecation warning.
- Tests annotated `@Suppress("DEPRECATION")` with explanatory comments to stay warning-free.

PR #3014, opened 2026-08-27
Deprecate IDP (Identity Provider) login flow

sfdctaka:
## Summary

Marks the **IDP (Identity Provider) login flow** public API as **deprecated** (deprecated in 14.0, targeted for removal in 15.0).

As Mobile SDK moves to advanced (browser-based) authentication as the default and only auth mechanism, the IDP flow — where one app acts as an identity provider so other "service provider" apps on the device log in without their own browser round-trip — is being retired. **There is no drop-in replacement**; apps should move to advanced (browser-based) authentication.

This PR **only adds deprecation annotations**. It does not remove the flow or change any runtime behavior — IDP keeps working in 14.0. App developers now get a compiler deprecation warning one major release ahead of removal.

## What changed

- `@Deprecated(…)` on the public IDP entry points:
  - `SalesforceSDKManager`: `idpManager`, `spManager`, `isIDPLoginFlowEnabled`, `setIDPAppPackageName(String?)`, `setAllowedSPApps(List<SPConfig>)`
  - `auth.idp.interfaces.IDPManager`, `SPManager`, `IDPAuthCodeActivity` (incl. nested `Status` / `StatusUpdateCallback`)
  - `auth.idp.SPConfig` (data class + `forCurrentApp()`); the concrete `IDPAuthCodeActivity`, `IDPReceiver`, `SPReceiver`
  - `ui.LoginActivity.onIDPLoginClick()`
  - `config.RuntimeConfig.ConfigKey.IDPAppPackageName` (annotation + javadoc `@deprecated`)
- The SDK's own internal usages carry scoped `@Suppress("DEPRECATION")` (function/class-level where the deprecated use is localized, `@file:Suppress` on the IDP-flow-specific files) so the SDK still builds **warning-clean**.

## Notes

- Mirrors the `useWebServerAuthentication` (OAuth user agent flow) deprecation landed the same release.
- Annotation-only, no behavior change. `idpManager` / `spManager` keep `private set`; no backing fields introduced.
- The RestExplorer sample app's live `setIDPAppPackageName(...)` call is intentionally left as-is — it demonstrates the still-shipping IDP feature, and the deprecation warning is the correct signal for a developer copying the sample.
- **Public API surface change** — flagging for human review per the repo's escalation policy.
- Companion iOS PR marks the equivalent iOS IDP surface deprecated.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
@@ -677,6 +677,18 @@ open class SalesforceSDKManager protected constructor(
         }.getOrDefault("")
 
 
+    // Backing field for [idpManager].  The SDK reads this directly so its own internal use of the
+    // manager doesn't trigger the deprecation warning on the public property.
+    @Volatile
+    @Suppress("DEPRECATION") // References the deprecated IDPManager type until it is removed in 15.0.
+    private var _idpManager: com.salesforce.androidsdk.auth.idp.interfaces.IDPManager? = null
+
+    // Backing field for [spManager].  The SDK reads this directly so its own internal use of the
+    // manager doesn't trigger the deprecation warning on the public property.
+    @Volatile
+    @Suppress("DEPRECATION") // References the deprecated SPManager type until it is removed in 15.0.
+    private var _spManager: com.salesforce.androidsdk.auth.idp.interfaces.SPManager? = null
+
     /**
      * The Salesforce SDK manager's admin settings manager. Only
      * defined if setAllowedSPApps() was called first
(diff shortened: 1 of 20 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -684,18 +681,27 @@ open class SalesforceSDKManager protected constructor(
      * The Salesforce SDK manager's admin settings manager. Only
      * defined if setAllowedSPApps() was called first
      */
-    var idpManager: IDPManager? = null
+    @Deprecated("The IDP (Identity Provider) login flow is deprecated and will be removed in " +
+            "Salesforce Mobile SDK 15.0. Apps should use advanced (browser-based) authentication.")
+    @Suppress("DEPRECATION") // References the deprecated IDPManager type until it is removed in 15.0.
```
brandonpage: If you want to avoid suppressing the deprecation everywhere in the code you can add a backing filed [like I did for the User Agent flow deprecation](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3010/changes#diff-134eafe1dc580ef45077b9da33e7908bafac1cd2c0e77ae324b2e683be7e3327).

brandonpage (approved):

wmathurin (approved):
The fully-qualified type names on the public `idpManager`/`spManager` property signatures are a bit noisy — e.g. `com.salesforce.androidsdk.auth.idp.interfaces.IDPManager?` in the declaration. The alternative would be to add the import and suppress at the file level with `@file:Suppress("DEPRECATION")`, which is the same pattern used in `IDPManager.kt` and `SPManager.kt`. Minor nit — not a blocker.

PR #3016, opened 2026-08-28
fix(auth): pass LoginActivity's own viewModel into LoginView()

umashankar-sf:
LoginActivity.onCreate() calls setContent { LoginView() } without an argument, so the composable builds a second, independent LoginViewModel via SalesforceSDKManager.loginViewModelFactory instead of reusing the Activity's own viewModel property.

This silently breaks advanced-auth login for any app that customizes LoginActivity.viewModel (e.g. to override singleServerCustomTabActivity): generateAuthorizationUrl() runs on the Compose-built instance and decides launchingCustomTab=true correctly, but the onBrowserCustomTabReady callback - wired in onCreate() onto the Activity's own instance - is null on the instance that actually invokes it, so the Custom Tab never launches. The WebView is left showing the intentional "about:blank" placeholder forever, with no error and no further log output.

Give LoginView() a viewModel parameter (defaulting to the previous factory-based construction, so callers that don't subclass LoginActivity are unaffected) and have LoginActivity pass its own viewModel explicitly.

Verified: reproduced the blank-screen hang on commit 638fbea and on the official v14.0.0-rc.0 tag (0049b454be) with a subclassed LoginActivity/LoginViewModel; confirmed this exact fix resolves it - real Chrome Custom Tab launches and the real login page renders. No automated test added: reproducing this specific defect requires either launching a real LoginActivity subclass (no precedent in this test suite - every existing Compose test here deliberately uses the generic ComponentActivity to avoid exactly that) or exercising onCreate() directly (not covered by the existing mockk-based LoginActivityTest pattern either). Flagging for discussion rather than introducing a new test-infra pattern unilaterally.

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt
@@ -322,7 +322,7 @@ open class LoginActivity : FragmentActivity() {
         // Set content
         setContent {
             MaterialTheme(colorScheme = SalesforceSDKManager.getInstance().colorScheme()) {
-                LoginView()
+                LoginView(viewModel = viewModel)
             }
         }
 
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/components/LoginView.kt
@@ -153,10 +153,10 @@ internal const val LOADING_INDICATOR_SIZE = 50
 internal const val SLOW_ANIMATION_MS = 500
 
 @Composable
-fun LoginView() {
+fun LoginView(
+    viewModel: LoginViewModel = viewModel(factory = SalesforceSDKManager.getInstance().loginViewModelFactory),
+) {
     val activity: LoginActivity = LocalContext.current.getActivity() as LoginActivity
-    val viewModel: LoginViewModel =
-        viewModel(factory = SalesforceSDKManager.getInstance().loginViewModelFactory)
     val frontDoorBridgeUrl = viewModel.frontDoorBridgeUrl.observeAsState()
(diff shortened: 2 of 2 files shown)


brandonpage:
@umashankar-sf We have always avoided passing `ViewModel` between classes as Google explicitly states this should not be done.  Instead we should be relaying on the OS instantiating and using the correct `ViewModel` if we set the same key, which is the root of the issue.  

Follow these steps:
1. Set `sdkManager.loginViewModelFactory = CustomLoginViewModel.Factory` immediately after `initNative(...)`.
2. Delete `override val viewModel: CustomLoginViewModel by viewModels { CustomLoginViewModel.Factory }` since you do not use it, or set the type (the key I mentioned the OS uses earlier) to `LoginViewModel` like this: `override val viewModel: LoginViewModel by viewModels { CustomLoginViewModel.Factory }`.

I am working on a debug log/assert to catch instances where two view models are used in the app to prevent this in the future.

umashankar-sf:
> @umashankar-sf We have always avoided passing `ViewModel` between classes as Google explicitly states this should not be done. Instead we should be relaying on the OS instantiating and using the correct `ViewModel` if we set the same key, which is the root of the issue.
> 
> Follow these steps:
> 
> 1. Set `sdkManager.loginViewModelFactory = CustomLoginViewModel.Factory` immediately after `initNative(...)`.
> 2. Delete `override val viewModel: CustomLoginViewModel by viewModels { CustomLoginViewModel.Factory }` since you do not use it, or set the type (the key I mentioned the OS uses earlier) to `LoginViewModel` like this: `override val viewModel: LoginViewModel by viewModels { CustomLoginViewModel.Factory }`.
> 
> I am working on a debug log/assert to catch instances where two view models are used in the app to prevent this in the future.

Thanks @brandonpage 
this started working after
sdkManager.loginViewModelFactory = CustomLoginViewModel.Factory` immediately after `initNative(...)

PR #3019, opened 2026-08-28
[Android] Capture ui_sid from DPoP token response and use it in mainSid

wmathurin:
## Summary

- Adds `private static final String UI_SID = "ui_sid"` constant to `OAuth2.java`
- Adds `uiSid` field to `OAuth2.TokenEndpointResponse`, read from the token endpoint response and guarded on `tokenType == "dpop"`
- Adds `uiSid` field, getter, and constant to `UserAccount.java`, serialized to/from JSON and Bundle
- Wires `uiSid` through `UserAccountBuilder.kt` (`populateFromTokenEndpointResponse`, `populateFromUserAccount`, `build()`)
- Encrypts and persists `uiSid` in `UserAccountManager` under `AuthenticatorService.KEY_UI_SID`
- Updates `getMainSid()` in `UserAccount.java`: returns `uiSid` if present, otherwise falls back to existing logic (`parentSid` for jwt, `authToken` otherwise)
- Adds 4 unit tests in `UserAccountTest.java`

## Dependencies

**This PR depends on (already merged):**
- [SalesforceMobileSDK-Android#3017](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3017) — [Android] Move mainSid calculation to UserAccount ✅

**Related PRs (same feature):**
- [SalesforceMobileSDK-iOS#4154](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4154) — [iOS] Capture ui_sid from DPoP token response and use it in mainSid
- [SalesforceMobileSDK-iOS-Hybrid#323](https://github.com/forcedotcom/SalesforceMobileSDK-iOS-Hybrid/pull/323) — [iOS-Hybrid] Use creds.mainSid in SalesforceWebViewCookieManager

## Tracking
- W-24024986 — [Android] Capture ui_sid from token endpoint and use it in mainSid
- Fixes W-23984602

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -576,7 +576,6 @@ public class UserAccountManager {
 		final String tokenType = decryptUserData(account, AuthenticatorService.KEY_TOKEN_TYPE, encryptionKey);
 		final String lastTokenRotationTime = decryptUserData(account, AuthenticatorService.KEY_LAST_TOKEN_ROTATION_TIME, encryptionKey);
 		final String uiSid = decryptUserData(account, AuthenticatorService.KEY_UI_SID, encryptionKey);
-		final String uiSid = decryptUserData(account, AuthenticatorService.KEY_UI_SID, encryptionKey);
 		final String featureFlagsRaw = decryptUserData(account, AuthenticatorService.KEY_FEATURE_FLAGS, encryptionKey);
 
 		Map<String, String> additionalOauthValues = null;
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -785,9 +785,10 @@ public class UserAccountManager {
 		if (userAccount.getLastTokenRotationTime() != null) {
 			extras.putString(AuthenticatorService.KEY_LAST_TOKEN_ROTATION_TIME, SalesforceSDKManager.encrypt(userAccount.getLastTokenRotationTime(), encryptionKey));
 		}
-		if (userAccount.getUiSid() != null) {
-			extras.putString(AuthenticatorService.KEY_UI_SID, SalesforceSDKManager.encrypt(userAccount.getUiSid(), encryptionKey));
-		}
+		// Always write KEY_UI_SID: a null value calls setUserData(key, null) which clears any
+		// stale DPoP uiSid left in AccountManager after a DPoP-to-Bearer downgrade.
+		extras.putString(AuthenticatorService.KEY_UI_SID,
+			userAccount.getUiSid() != null ? SalesforceSDKManager.encrypt(userAccount.getUiSid(), encryptionKey) : null);
 		final Set<String> featureFlags = userAccount.getFeatureFlags();
 		if (!featureFlags.isEmpty()) {
 			extras.putString(AuthenticatorService.KEY_FEATURE_FLAGS,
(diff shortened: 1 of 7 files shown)


libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
```diff
@@ -902,6 +902,51 @@ public void test_givenJwtTokenFormat_whenGetMainSid_thenReturnsParentSid() {
...
+        final UserAccount account = UserAccountBuilder.getInstance()
+                .populateFromUserAccount(createTestAccount())
+                .uiSid("test_ui_sid")
+                .build();
+
+        Assert.assertEquals("test_ui_sid", account.getUiSid());
+    }
+
+    @Test
+    public void test_givenNoDPoPTokenType_whenBuildAccount_thenUiSidIsNull() {
```
brandonpage: [Medium] Exercise token parsing and the actual non-DPoP premise

The added tests bypass the new parsing path by calling uiSid directly. Also, createTestAccount sets tokenType to DPoP, so this NoDPoPTokenType test does not establish its stated precondition. Please construct actual TokenEndpointResponse instances with DPoP/Bearer token_type and ui_sid values, and add the AccountManager DPoP-to-Bearer round-trip that catches the stale-value regression.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -783,6 +785,9 @@ private Bundle buildAuthBundle(UserAccount userAccount) {
 		if (userAccount.getLastTokenRotationTime() != null) {
 			extras.putString(AuthenticatorService.KEY_LAST_TOKEN_ROTATION_TIME, SalesforceSDKManager.encrypt(userAccount.getLastTokenRotationTime(), encryptionKey));
 		}
+		if (userAccount.getUiSid() != null) {
```
brandonpage: [High] Clear uiSid during DPoP-to-Bearer refresh-token migration

downgradeFromDPoP uses migrateRefreshToken with useDPoP=false, and the token-migration path calls persistAccount to update the same AccountManager record. The Bearer token response builds an account with uiSid null, so this condition omits KEY_UI_SID from the bundle; updateAccount consequently never calls setUserData for that key, leaving the previous encrypted DPoP UI SID behind. The migration callback receives the correct in-memory Bearer account, but the next buildUserAccount reload restores the stale uiSid and getMainSid prefers it over the Bearer auth token. Please explicitly clear KEY_UI_SID when uiSid is null and add a DPoP-to-Bearer migration persistence round-trip test.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/accounts/UserAccountTest.java
wmathurin: Fixed. Replaced both builder-bypass tests with `TokenEndpointResponse`-based tests that go through the actual parsing path — one with `token_type=DPoP` + `ui_sid` (captures), one with `token_type=Bearer` + `ui_sid` (ignores). Also added a DPoP-to-Bearer AccountManager persistence round-trip test in `UserAccountManagerTest` that would have caught the stale-uiSid regression.

libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
wmathurin: Fixed. Changed the `if (uiSid != null)` guard to always write `KEY_UI_SID` to the bundle — a null value causes `updateAccount` to call `accountManager.setUserData(account, KEY_UI_SID, null)`, which removes the stale encrypted DPoP value from AccountManager. The regression test in `UserAccountManagerTest` (DPoP account → `updateAccount` with Bearer → `buildUserAccount` reload → assert `uiSid` null and `getMainSid` returns `authToken`) confirms the fix.

brandonpage (approved):

PR #3023, opened 2026-09-01
security(hybrid): harden pgSendRequest — add caller-origin check and restrict OAuth tokens to authenticated instance (W-23695430)

wmathurin:
## Summary

- **Vulnerability**: `SalesforceNetworkPlugin.sendRequest()` (the `pgSendRequest` Cordova action) had no caller-origin check. Any JavaScript running inside the WebView — including JavaScript injected via XSS or a compromised third-party page loaded in the view — could call `pgSendRequest` and make fully authenticated REST API calls to Salesforce on behalf of the logged-in user.
- **Fix**: Added an `isTrustedCallerOrigin(url)` check at the top of `sendRequest()`. The check mirrors the `<access origin>` allowlist in `SalesforceMobileSDK-CordovaPlugin`'s `config.xml`:
  - `file://` URLs (local hybrid app assets) — always trusted
  - `localhost` — trusted
  - Hosts ending in `.salesforce.com`, `.force.com`, `.visualforce.com`, `.documentforce.com`, `.salesforce-communities.com` — trusted
  - Everything else — blocked with `callbackContext.error("pgSendRequest blocked: untrusted caller origin")`
- `isTrustedCallerOrigin` is package-private to allow direct unit testing without reflection.

## Test plan

- [ ] Unit tests added in `libs/SalesforceHybrid/unitTest/java/.../SalesforceNetworkPluginTest.java` (12 cases, all passing via `./gradlew :libs:SalesforceHybrid:test`)
- [ ] Verify hybrid app still loads and makes REST calls normally (trusted `file://` origin is allowed)
- [ ] Verify that `build.gradle.kts` unit test source set (`unitTest/java/`) wires up correctly and `testImplementation("junit:junit:4.13.2")` compiles cleanly

## GUS

W-23695430

### DIFF
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -189,8 +189,7 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
     }
 
     // Trusted origins mirror the <access origin> entries in
-    // libs/SalesforceHybrid/res/xml/config.xml (also mirrored in
-    // SalesforceMobileSDK-CordovaPlugin/src/android/libs/mobile_sdk/libs/SalesforceHybrid/res/xml/config.xml).
+    // SalesforceMobileSDK-CordovaPlugin/plugin.xml (source of truth).
     // If that list changes, update this method to match.
     /* package */ boolean isTrustedCallerOrigin(String url) {
         if (url == null) return false;
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -109,6 +109,11 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
      */
     protected void sendRequest(JSONArray args, final CallbackContext callbackContext) {
         try {
+            String callerUrl = webView.getUrl();
+            if (!isTrustedCallerOrigin(callerUrl)) {
+                callbackContext.error("pgSendRequest blocked: untrusted caller origin");
+                return;
+            }
             final RestRequest request = prepareRestRequest(args);
             final boolean returnBinary = ((JSONObject) args.get(0)).optBoolean(RETURN_BINARY, false);
             final boolean doesNotRequireAuth = ((JSONObject) args.get(0)).optBoolean(DOES_NOT_REQUIRE_AUTHENTICATION, false);
@@ -181,6 +186,23 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
             }
             callbackContext.error(errorObject.toString());
         }
+    }
+
+    // Trusted origins mirror the <access origin> entries in
+    // SalesforceMobileSDK-CordovaPlugin/src/android/libs/mobile_sdk/libs/SalesforceHybrid/res/xml/config.xml.
+    // If that list changes, update this method to match.
+    /* package */ boolean isTrustedCallerOrigin(String url) {
+        if (url == null) return false;
+        if (url.startsWith("file://")) return true;
+        try {
+            String host = new java.net.URL(url).getHost();
+            if (host.equals("localhost")) return true;
+            for (String suffix : new String[]{".salesforce.com", ".force.com", ".visualforce.com",
+                                              ".documentforce.com", ".salesforce-communities.com"}) {
+                if (host.endsWith(suffix)) return true;
+            }
+        } catch (Exception e) { /* malformed URL → block */ }
+        return false;
     }
 
     private Object parsedResponse(RestResponse response) throws IOException {
--- a/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
+++ b/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
@@ -199,15 +199,22 @@ public class SalesforceNetworkPlugin extends ForcePlugin {
         }
     }
 
-    // Allow any Salesforce-owned domain, localhost, or file:// (Cordova local app).
+    // Allow localhost, packaged app assets (file:///android_asset/www/), or any Salesforce-owned domain over https.
     /* package */ boolean isTrustedCallerOrigin(String url) {
         if (url == null) return false;
         try {
             java.net.URL parsed = new java.net.URL(url);
             String host = parsed.getHost();
             String scheme = parsed.getProtocol();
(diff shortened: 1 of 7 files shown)


libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -109,6 +109,11 @@ public boolean execute(String actionStr, JavaScriptPluginVersion jsVersion, JSON
      */
     protected void sendRequest(JSONArray args, final CallbackContext callbackContext) {
         try {
+            String callerUrl = webView.getUrl();
```
brandonpage, line 113: **[P1] This is the top-level document URL, not the invoking script origin.** XSS or third-party JavaScript executing inside an otherwise trusted document sees the same `webView.getUrl()` and passes this guard, so the injected-script scenario described by the PR remains open. Please propagate caller/frame provenance through the Cordova bridge or gate bridge access on an exact configured app origin, and add an end-to-end negative test. If the intended scope is only top-level navigation, please narrow the threat model accordingly.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -183,6 +188,23 @@ public void onError(Exception exception) {
...
+    // Trusted origins mirror the <access origin> entries in
+    // SalesforceMobileSDK-CordovaPlugin/plugin.xml (source of truth).
+    // If that list changes, update this method to match.
+    /* package */ boolean isTrustedCallerOrigin(String url) {
+        if (url == null) return false;
+        if (url.startsWith("file://")) return true;
+        try {
+            String host = new java.net.URL(url).getHost();
+            if (host.equals("localhost")) return true;
+            for (String suffix : new String[]{".salesforce.com", ".force.com", ".visualforce.com",
```
brandonpage: **[P1] A global Salesforce suffix is not an application trust boundary.** Domains under `*.force.com` and `*.visualforce.com` can host custom content belonging to other Salesforce orgs. A page from another tenant therefore matches this list and can reach the authenticated native client. Please bind authorization to the exact configured application/org origins rather than the global navigation allowlist, and add a negative test for a different Salesforce tenant.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -183,6 +188,23 @@ public void onError(Exception exception) {
...
     }
 
+    // Trusted origins mirror the <access origin> entries in
+    // SalesforceMobileSDK-CordovaPlugin/plugin.xml (source of truth).
+    // If that list changes, update this method to match.
+    /* package */ boolean isTrustedCallerOrigin(String url) {
+        if (url == null) return false;
+        if (url.startsWith("file://")) return true;
+        try {
+            String host = new java.net.URL(url).getHost();
```
brandonpage: **[P2] Validate the scheme as part of the origin.** This extracts only the host, so `http://myorg.salesforce.com` and even non-HTTP schemes with that host are accepted, while the referenced `plugin.xml` entries are HTTPS-only. Please require a normalized `https` scheme before applying the host allowlist and add HTTP/alternate-scheme tests.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
```diff
@@ -183,6 +188,23 @@ public void onError(Exception exception) {
         }
     }
 
+    // Trusted origins mirror the <access origin> entries in
+    // SalesforceMobileSDK-CordovaPlugin/plugin.xml (source of truth).
+    // If that list changes, update this method to match.
+    /* package */ boolean isTrustedCallerOrigin(String url) {
+        if (url == null) return false;
+        if (url.startsWith("file://")) return true;
```
brandonpage: **[P2] Restrict file trust to packaged web assets.** Every `file:` URL currently passes, including files outside `file:///android_asset/www/`. If downloaded or temporary HTML is ever navigated into this WebView, it gains access to the authenticated plugin. Please canonicalize the path and require the packaged asset root, with a negative test for a file outside that directory.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
wmathurin: Fixed. `isTrustedCallerOrigin` now requires `https` before applying the Salesforce domain suffix checks. Test added for `http://myorg.salesforce.com` being blocked.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
wmathurin: Fixed. `file://` is now restricted to `/android_asset/www/` (the packaged Cordova asset root). Paths outside that directory are blocked. Tests added for both the trusted and blocked cases.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
wmathurin: The broad domain list in `isTrustedCallerOrigin` is intentional and mirrors the `config.xml` allow-navigation list. A single Salesforce org spans multiple subdomains — Lightning (`acme.lightning.force.com`), Visualforce (`acme--c.vf.force.com`), Experience Cloud (`acme.my.site.com`), standard API (`acme.my.salesforce.com`) — so binding the caller check to the exact instance URL would break legitimate hybrid apps. The token-leakage risk is addressed separately by `isInstanceServer`: OAuth tokens are only attached when the endpoint matches the authenticated user's exact instance host. Any other destination — including other orgs on the same shared Salesforce domains — goes unauthenticated.

libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/plugin/SalesforceNetworkPlugin.java
wmathurin: `webView.getUrl()` returns the top-level document URL; a script executing via XSS inside a trusted document passes this check. Accurately scoping per-frame provenance would require propagating equivalent frame context through the Cordova bridge — out of scope for this fix. The caller check guards against unrelated domains navigated into the WebView; the primary token-leakage protection is `isInstanceServer`, which ensures OAuth tokens never leave the authenticated instance regardless of caller.

wmathurin:
This is more involved that I thought it would be. Let's revisit in 14.1.

PR #3026, opened 2026-09-03
Stabilize flaky Android tests and harden API 37 logout handling

brandonpage:
## Summary

- prevent late JavaScript test results from blocking or being attributed to the next test
- increase timeouts for the slower MobileSync and SmartStore JavaScript suites
- wait for Android teardown work to become idle before removing instrumentation state
- add production null-account logout protection and a regression test
- merge Firebase rerun XML into sharded reports so passing retries are classified as flaky successes

## API 37 production hardening

The API 37 failure included a Binder buffer crash while Android was delivering com.salesforce.CLEANUP after repeated teardown sign-out calls. UserAccountManager.signoutCurrentUser now resolves the current account once and returns without invoking logout when no account exists.

This prevents accidental or repeated null-account sign-out calls in production from emitting unnecessary cleanup and logout broadcasts, reducing one-way Binder traffic in the same path that triggered the API 37 crash. Direct SalesforceSDKManager.logout(null) behavior remains unchanged for callers that intentionally request global cleanup.

NativeLoginManagerTest now relies on this production guard instead of duplicating the account check, while retaining waitForIdleSync for real logout cleanup. UserAccountManagerTest verifies that no LOGOUT_COMPLETE broadcast is emitted when there is no current account.

## Verification

- red/green regression: null-account sign-out test failed before the guard because LOGOUT_COMPLETE was broadcast, then passed after the guard
- NativeLoginManagerTest plus UserAccountManagerTest: 44/44 passed on API 36
- targeted API 37 remaining SalesforceSDK shard from the original failure investigation: 838/838 passed after teardown stabilization
- sharded rerun control-flow check failed before the workflow change and passed after it
- exact embedded XML merger passed fixtures for both successful and failed retry outcomes
- workflow YAML parsed successfully and the extracted Copy Test Results shell body passed bash syntax validation
- git diff --check passed

Related CI runs:
- https://github.com/forcedotcom/SalesforceMobileSDK-Android/actions/runs/33600595445
- https://github.com/forcedotcom/SalesforceMobileSDK-Android/actions/runs/33700254428

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
@@ -394,8 +394,11 @@ public class UserAccountManager {
 	 * @param reason The reason for the logout.
 	 */
 	public void signoutCurrentUser(Activity frontActivity, boolean showLoginPage, OAuth2.LogoutReason reason) {
-		SalesforceSDKManager.getInstance().logout(
-				getCurrentAccount(), frontActivity, showLoginPage, reason);
+		final Account currentAccount = getCurrentAccount();
+		if (currentAccount != null) {
+			SalesforceSDKManager.getInstance().logout(
+					currentAccount, frontActivity, showLoginPage, reason);
+		}
 	}
 
 	/**
--- a/.github/workflows/reusable-lib-workflow.yaml
+++ b/.github/workflows/reusable-lib-workflow.yaml
@@ -210,12 +210,13 @@ jobs:
                   for RESULT_FILE in $(gsutil ls "${BUCKET_PATH}/*/test_result_1.xml" 2>/dev/null | grep -v "rerun"); do
                     gsutil cp "${RESULT_FILE}" "firebase_results/api_${LEVEL}_test_result.xml"
                   done
-                  # Pass 2: merge rerun testcases into originals so check_retries detects flaky tests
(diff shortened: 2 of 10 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/accounts/UserAccountManager.java
```diff
@@ -394,8 +394,11 @@ public void signoutCurrentUser(Activity frontActivity, boolean showLoginPage) {
 	 * @param reason The reason for the logout.
 	 */
 	public void signoutCurrentUser(Activity frontActivity, boolean showLoginPage, OAuth2.LogoutReason reason) {
-		SalesforceSDKManager.getInstance().logout(
-				getCurrentAccount(), frontActivity, showLoginPage, reason);
+		final Account currentAccount = getCurrentAccount();
+		if (currentAccount != null) {
+			SalesforceSDKManager.getInstance().logout(
+					currentAccount, frontActivity, showLoginPage, reason);
+		}
```
brandonpage, line 401: This production hardening actually came from CI crashing on API 37 when one of the tests called `signoutCurrentUser` with a null `currentUser`.  The issue is a byproduct of the [Android 17 MessageQueue rewrite](https://developer.android.com/about/versions/17/changes/messagequeue).  Yay for tests (unintentionally) catching real issues.

wmathurin (approved):

PR #3030, opened 2026-09-04
Fix OAuth refresh preserving optional account fields

brandonpage:
## Summary
- preserve existing Lightning, Visualforce, and Content domain/SID values when token refresh responses omit them
- represent omitted optional token endpoint fields as null so `allowUnset(false)` retains stored account values
- add a regression test covering response parsing and production-order account rebuilding

## Testing
- focused regression test passed on Pixel 4 XL API 33
- `./gradlew :libs:SalesforceSDK:assembleAndroidTest`
- `./gradlew :libs:SalesforceSDK:build`

## Work Item
- [W-24103502](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002jyBTLYA2/view)

### DIFF
--- a/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
+++ b/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
@@ -1260,13 +1260,13 @@ public class OAuth2 {
                         }
                     }
                 }
-                idToken = parsedResponse.optString(ID_TOKEN);
-                lightningDomain = parsedResponse.optString(LIGHTNING_DOMAIN);
-                lightningSid = parsedResponse.optString(LIGHTNING_SID);
-                vfDomain = parsedResponse.optString(VF_DOMAIN);
-                vfSid = parsedResponse.optString(VF_SID);
-                contentDomain = parsedResponse.optString(CONTENT_DOMAIN);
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -1260,13 +1260,13 @@ public TokenEndpointResponse(Response response, List<String> additionalOauthKeys
...
                     }
                 }
-                idToken = parsedResponse.optString(ID_TOKEN);
-                lightningDomain = parsedResponse.optString(LIGHTNING_DOMAIN);
-                lightningSid = parsedResponse.optString(LIGHTNING_SID);
-                vfDomain = parsedResponse.optString(VF_DOMAIN);
-                vfSid = parsedResponse.optString(VF_SID);
-                contentDomain = parsedResponse.optString(CONTENT_DOMAIN);
-                contentSid = parsedResponse.optString(CONTENT_SID);
+                idToken = parsedResponse.optString(ID_TOKEN, null);
```
wmathurin, line 1263: Should we keep them if the refresh flow didn't return them? Won't they be expired (since they are tied to the previous access token) ? Also why would the refresh flow not return them ?

PR #3031, opened 2026-09-09
Fix RTR user agent after app restart

wmathurin:
## Summary
- Carry the credential owner through refresh-token requests as request-scoped context so the final User-Agent is computed for the correct account after process restart.
- Preserve the account tag and resulting User-Agent through DPoP nonce retries while retaining existing fallback behavior for context-free callers.
- Add a credential-free interceptor regression and an AuthFlowTester RTR cold-restart UI test that captures the actual final token-request header and verifies RT, A2, and OT.
- Gate the sample-local capture client installation on AuthFlowTester UI-test mode; normal interactive launches retain the standard `HttpAccess.DEFAULT`.
- Document the refresh-token lifecycle and AuthFlowTester coverage.

## Why no timestamp reconciliation
No persisted-state repair is needed. Warm refreshes later in the original process already carry `RT`, and the post-restart regression passes once refresh resolves the credential owner explicitly. The failing request also omitted other per-user markers, confirming request-context loss rather than `RT` persistence loss.

## Validation
- `OAuth2DPoPTest#test_refreshAuthToken_withUserAccount_tagsRequestAndUsesRequestUserAgent`: 1 passed, 0 failed.
- `RTRLoginTests#testECAOpaqueRtr_Hybrid_WithRestart`: 1 passed, 0 failed in 1m55s on Medium Phone API 36.1.
- After capture gating follow-up: AuthFlowTester main + androidTest Kotlin compilation passed; the live RTR restart regression passed again 1/1 in 1m35s.
- `git diff --check`: passed.

## Spec
- Workspace PR: https://git.soma.salesforce.com/SalesforceMobileSDK/SalesforceMobileSDK-Workspace/pull/99

## Work item
- W-24141870


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
```diff
@@ -546,6 +547,23 @@ public static TokenEndpointResponse refreshAuthToken(HttpAccess httpAccessor, UR
...
+                credentialsIdentifier, tokenType, null);
+    }
+
+    /**
+     * Gets a new auth token using the refresh token for a known user. The user is carried as
+     * request-scoped context so the final HTTP interceptor can build the correct per-user agent
+     * without consulting mutable current-user state.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY)
+    public static TokenEndpointResponse refreshAuthToken(HttpAccess httpAccessor, URI loginServer,
```
brandonpage, line 560: I think the `RestrictTo.Scope.LIBRARY` implies this, but should we just remove `refreshAuthToken` from the public API?  With RTR on calling this function will cause the app to logout the next time we make a REST call since our tokens will be out of sync.

Not strictly relevant to this PR but worth a discussion I think.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
```diff
@@ -142,7 +142,8 @@ public Bundle getAuthToken(AccountAuthenticatorResponse response, Account accoun
                 SalesforceSDKLogger.i(TAG, "Initiating token refresh to host: " + tokenServer.getHost());
                 final OAuth2.TokenEndpointResponse tr = OAuth2.refreshAuthToken(HttpAccess.DEFAULT,
                         tokenServer, originalUserAccount.getClientIdForRefresh(), originalUserAccount.getRefreshToken(), addlParamsMap,
-                        originalUserAccount.getCredentialsIdentifier(), originalUserAccount.getTokenType());
+                        originalUserAccount.getCredentialsIdentifier(), originalUserAccount.getTokenType(),
+                        originalUserAccount);
```
brandonpage, line 146: NIT: I count 6 params now that could be rolled into the single new `UserAccount` param.

brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/OAuth2.java
wmathurin: Agreed that the existing public refreshAuthToken API is worth a separate API-design discussion. Removing or restricting the existing overloads would be a source/binary compatibility change, so this PR keeps them unchanged and limits only the new user-aware overload to library scope. The broader RTR token-persistence hazard should be handled as a focused follow-up rather than folded into W-24141870.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java
wmathurin: Good point. AuthenticatorService could derive those values from its account, but ClientManager deliberately passes its coordinated currentRefreshToken, which can be newer than the original UserAccount snapshot during RTR/concurrent refresh. A UserAccount-only overload would risk reintroducing a stale-token problem; an account-plus-token convenience overload would add more surface for little benefit in this targeted fix. I’m keeping the explicit arguments here.

PR #3035, opened 2026-09-14
Harden AuthFlowTester UI login tests

brandonpage:
## Summary
- harden Chrome Custom Tab credential entry and checked submission against keyboard/autofill overlays and API 35/37 accessibility gaps
- retry credential reads against fresh Compose nodes and keep revealed sensitive values hidden after reads
- correct non-hybrid token expectations in two multi-user flag-isolation tests
- temporarily suppress only the username and post-refresh user-agent assertions in the two DPoP pool-server tests, with TODOs to restore them when W-23992239 is fixed

## Validation
- `:native:NativeSampleApps:AuthFlowTester:assembleDebug`
- `:native:NativeSampleApps:AuthFlowTester:assembleDebugAndroidTest`
- Firebase Test Lab multi-user suite: 51/51 on API 31, 35, and 37
  - https://console.firebase.google.com/project/mobile-apps-firebase-test/testlab/histories/bh.d03c10a25c4608c0/matrices/5938361457294963667
  - https://console.firebase.google.com/project/mobile-apps-firebase-test/testlab/histories/bh.d03c10a25c4608c0/matrices/7835346956809571482
  - https://console.firebase.google.com/project/mobile-apps-firebase-test/testlab/histories/bh.d03c10a25c4608c0/matrices/5670936605136682352
- W-23992239 DPoP pool-server cases: 14/14 across API 31-37, with no flaky-test retries
  - https://console.firebase.google.com/project/mobile-apps-firebase-test/testlab/histories/bh.d03c10a25c4608c0/matrices/5865458963675192720
  - https://console.firebase.google.com/project/mobile-apps-firebase-test/testlab/histories/bh.d03c10a25c4608c0/matrices/7232838339838186603

## Scope
This PR changes AuthFlowTester instrumentation tests only. It contains no SalesforceSDK production or unit-test changes.


brandonpage:
The `Setup Android SDK` step seems to be causing all jobs to fail, regardless of tests passing.  This is unrelated to my changes.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/ChromeCustomTabPageObject.kt
```diff
@@ -218,27 +237,154 @@ class ChromeCustomTabPageObject(composeTestRule: ComposeTestRule): LoginPageObje
...
+     * The caller has already established whether it is on the username or password step; requiring
+     * the IME to repeat that subtype is unreliable because API 35 can reset it to an unspecified
+     * Chrome editor while the HTML field visibly remains focused.
+     */
+    private fun setTextInFocusedChromeEditor(text: String): Boolean {
+        val inputMethodState = runCatching {
+            device.executeShellCommand("dumpsys input_method").replace(" ", "")
+        }.getOrDefault("")
+        val chromeHasInputFocus = inputMethodState.contains("packageName=com.android.chrome")
+        if (device.currentPackageName != "com.android.chrome" && !chromeHasInputFocus) {
```
wmathurin: This guard treats Chrome merely being the foreground package as sufficient evidence that an HTML editor is focused. `skipGoogleSignIn()` only waits for the Custom Tab toolbar, so when the server-rendered login form is still loading, `findVisibleChromeInput()` can return null, this fallback can inject Ctrl+A/credential key events into Chrome with no editor, and `.all` below still reports success. That makes `setUsername()` return without having populated the field. Could we require `chromeHasInputFocus` here (and ideally verify the expected editor/value) before accepting the fallback as successful?

brandonpage: Addressed in bac1c35f1. The key-event fallback now requires both Chrome to be foreground and `dumpsys input_method` to report Chrome as the active input connection. It also requires a non-empty set of key-down events before reporting success, so a loading page with no focused editor keeps retrying instead of accepting a no-op injection.

wmathurin (approved):

PR #3038, opened 2026-09-16
W-24185374: Harden concurrent REST and pool login

wmathurin:
## Goal

Follow up on #3037 by bringing Android AuthFlowTester behavior and coverage to parity with iOS, and make DPoP login through a pool server recover safely when the identity endpoint rejects or challenges the freshly issued credential.

## What changed

- Replaced the permission-dependent Limits request with a weighted, low-cost cycle: API Resources, API Resources, Describe Global.
- Hardened the expanded-options UI semantics, scrolling, logout completion, token-request capture, and failure diagnostics used by UI tests.
- Updated DPoP/RTR assertions for valid nonce behavior and improved concurrent-request failure output.
- Updated the AuthFlowTester README and real app screenshots.
- Require a valid identity before account creation regardless of the returned scope list, addressing W-20063511's underlying failure mode.
- Added a documented W-23992239 workaround: the initial pool-issued DPoP credential uses the raw pool identity URL instead of the instance identity URL that currently returns `Bad_OAuth_Token`. This does not claim that W-23992239 covers the separately observed `Wrong_Org` response.
- Reused the shared DPoP request decorator so identity calls harvest a server nonce and retry once with a rebuilt proof before considering token refresh. Repeated nonce challenges remain bounded to one retry.
- Preserve the bounded identity error body and refresh only for 401 or selected 403 credential errors (`Bad_OAuth_Token` and the observed `Wrong_Org`), rather than treating every 403 as refreshable.
- If credential refresh is required, refresh once against the instance token endpoint, merge every token-response field—including hybrid domains/SIDs, cookie/CSRF values, `parentSid`, additional OAuth values, and beacon credentials—and replay identity against the instance identity URL.
- Preserve refresh-token rotation during pre-account recovery, persist `lastTokenRotationTime`, and register the per-user RTR feature only after account persistence.

## Screenshots

| Expanded options | Completed batch |
| --- | --- |
| Concurrent REST request options | Completed concurrent REST requests |

## Validation

- 52/52 focused `AuthenticationUtilitiesTest` and `OAuth2DPoPTest` tests passed, including identity nonce retry/bounding, selective 403 recovery, complete hybrid credential merge, and immediate RTR persistence/registration assertions.
- Affected SDK and AuthFlowTester production/test sources compile successfully.
- 3/3 AuthFlowTester login-pool tests passed together: Bearer/JWT, DPoP, and DPoP+RTR.
- 9/9 Make Many Requests tests passed across normal completion, error details, revoke, logout/restart, RTR, DPoP, cold restart, and multi-user isolation.
- `testECAJwtDPoP_ViaLoginPoolServer_Rtr` passed individually after the fix.
- The full `DPoPLoginTests` class invocation completed 17/18 tests successfully. Android's test orchestrator dropped `testECAJwtDPoP_NoHybrid` at startup as an instrumentation-process crash; that test passes when run individually, so every test in the class has a passing result.

## Root cause

Android did not match iOS identity error handling. A pool identity response of `403 Wrong_Org` was parsed as an empty identity object, and there was no bounded refresh-and-replay path. An instance identity fallback with the original pool-issued DPoP token is invalid (`403 Bad_OAuth_Token`). Recovery now distinguishes DPoP nonce challenges from credential rejection, refreshes at most once when appropriate, carries forward the complete credential state and RTR metadata, and then uses the refreshed instance-issued token at the instance identity URL.


libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -165,7 +168,29 @@ internal suspend fun onAuthFlowComplete(
...
+    if (userIdentity?.username.isNullOrBlank()) {
+        val error = IllegalStateException("Identity service did not return a user identity")
+        w(TAG, "Cannot complete authentication because user identity is missing.", error)
+        onAuthFlowError(
+            context.getString(sf__generic_authentication_error_title),
+            context.getString(sf__generic_authentication_error),
+            error,
+        )
+        return
+    }
```
brandonpage: Does this effectively fix [W-20063511](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002OGYtnYAH/view)?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -425,32 +450,108 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
+                    consumerKey,
+                    checkNotNull(response.refreshToken) {
+                        "Cannot refresh identity without a refresh token"
+                    },
+                    response.additionalOauthValues,
+                    response.credentialsIdentifier,
+                    response.tokenType,
+                )
+            }
+        },
```
brandonpage: I am not sure how I feel about ~50 lines of inline function in this functions parameters.  For one-liners I totally get it, but this seems like bad style.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: I think/hope so - although we should rewrite the story a bit (it's not really about the identity scope). Let me take a shot at it.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -425,32 +450,108 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
+): OAuth2.IdServiceResponse {
+    var refreshAttempted = false
+    while (true) {
```
brandonpage: I really don't like `while (true)` because a small change later can easily cause an infinite loop.  Even a simple max attempts counter could be much better IMO.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -425,32 +450,108 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
+                (statusCode == HTTP_UNAUTHORIZED || statusCode == HTTP_FORBIDDEN) &&
+                !tokenResponse.refreshToken.isNullOrBlank()
+            if (!canRefresh) {
+                w(TAG, "Cannot fetch user identity due to an error.", e)
+                throw e
+            }
+
+            refreshAttempted = true
+            i(TAG, "Identity request returned HTTP $statusCode; refreshing credentials once.")
+            val refreshedResponse = tokenRefresher(tokenResponse)
```
brandonpage: This is perhaps besides the point, but it really feels like we are working around something that should be fixed on the server.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -425,32 +450,108 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
+    refreshed.idUrl?.takeIf(String::isNotBlank)?.let { target.idUrl = it }
+    refreshed.idUrlWithInstance?.takeIf(String::isNotBlank)?.let { target.idUrlWithInstance = it }
+    refreshed.orgId?.takeIf(String::isNotBlank)?.let { target.orgId = it }
+    refreshed.userId?.takeIf(String::isNotBlank)?.let { target.userId = it }
+    refreshed.communityId?.takeIf(String::isNotBlank)?.let { target.communityId = it }
+    refreshed.communityUrl?.takeIf(String::isNotBlank)?.let { target.communityUrl = it }
+    refreshed.scope?.takeIf(String::isNotBlank)?.let { target.scope = it }
+    refreshed.tokenType?.takeIf(String::isNotBlank)?.let { target.tokenType = it }
+    refreshed.tokenFormat?.takeIf(String::isNotBlank)?.let { target.tokenFormat = it }
+    refreshed.uiSid?.takeIf(String::isNotBlank)?.let { target.uiSid = it }
```
brandonpage: Is it possible that the refresh does not contain this sid (or any other value) and we keep an old now invalid one?

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
```diff
@@ -425,32 +450,108 @@ private fun logAddAccount(account: UserAccount?, loginServerManager: LoginServer
...
-        w(TAG, "Cannot fetch user identity due to an error.", throwable)
-    }.getOrNull()
+    }
+}
+
+private fun applyRefreshedCredentials(
+    target: TokenEndpointResponse,
+    refreshed: TokenEndpointResponse,
+) {
+    target.authToken = refreshed.authToken?.takeIf(String::isNotBlank)
```
brandonpage, line 589: This helper does not apply the complete refreshed credential state claimed by the KDoc. It omits `additionalOauthValues`, Lightning/VF/content domains and SIDs, CSRF/cookie fields, `parentSid`, and beacon credentials, although `TokenEndpointResponse` parses them and `UserAccountBuilder.populateFromTokenEndpointResponse()` persists them. If identity recovery runs during hybrid authentication, the refresh can return new session IDs but the account will be built with the pre-refresh values, potentially breaking hybrid bootstrap. Please reuse the canonical account merge semantics from `ClientManager` or copy every account-relevant field, and cover the hybrid recovery case.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
brandonpage: If this refresh returns a different refresh token, the replacement is preserved but the rotation is never recorded. The normal `ClientManager` path stamps `lastTokenRotationTime` and registers `FEATURE_RTR`; this pre-account path therefore creates an account whose RTR metadata and user-agent feature marker remain false until another refresh occurs. Please detect rotation before overwriting the token, carry that fact through account creation, and add an assertion immediately after the pool-login recovery.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/AuthFlowTesterPageObject.kt
```diff
@@ -351,13 +396,19 @@ class AuthFlowTesterPageObject(composeTestRule: ComposeTestRule): BasePageObject
     fun validateMixedSuccessfulRequests(count: Int) {
         repeat(count) { index ->
             waitForManyRequestSquareState(index, "Succeeded")
-            val expectedType = ConcurrentRequestType.entries[index % ConcurrentRequestType.entries.size]
+            val expectedType = concurrentRequestType(index)
```
brandonpage: The test derives `expectedType` from `concurrentRequestType`, the same production helper that selects the actual request. A regression in the intended weighted cycle changes both sides and the test remains green. Please use an explicit test-owned sequence such as `RESOURCES, RESOURCES, DESCRIBE_GLOBAL`, or add a focused unit test that asserts the helper’s outputs independently.

brandonpage (changes requested):

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Agreed. The KDoc now labels the pool URL behavior as a workaround for W-23992239 and deliberately does not claim that story covers Wrong_Org. Identity requests now also handle the expected server-side DPoP nonce challenge directly by harvesting the nonce and replaying once, so the path remains valid after the server fix.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Addressed in d4d7d0f5f. The loop is replaced by an explicit initial attempt, at most one credential refresh, and one replay.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Addressed in d4d7d0f5f. The injectable defaults are now named production functions: fetchIdentityForAuthentication and refreshCredentialsForIdentity.

native/NativeSampleApps/AuthFlowTester/src/androidTest/java/com/salesforce/samples/authflowtester/pageObjects/AuthFlowTesterPageObject.kt
wmathurin: Addressed. The UI test now owns the explicit expected cycle RESOURCES, RESOURCES, DESCRIBE_GLOBAL and no longer imports the production selector.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Addressed. Rotation is detected before the response is merged, lastTokenRotationTime is included in the account before its primary persistence, and FEATURE_RTR is registered afterward. The test verifies this ordering and the timestamp immediately.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Yes. The identity-scope gate is removed: authentication now requires a valid identity regardless of the returned scope list. testOnAuthFlowComplete_withoutIdScopeWithIdentity_shouldSucceed covers the W-20063511 behavior.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Addressed. The merge uses allowUnset(false)-equivalent semantics: null preserves the prior value, while a returned empty string clears stale session data. This applies consistently to every TokenEndpointResponse field.

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticationUtilities.kt
wmathurin: Addressed in d4d7d0f5f. One canonical merge now carries every TokenEndpointResponse field, including additional OAuth values, hybrid domains and SIDs, cookie and CSRF values, parentSid, and beacon credentials. The hybrid recovery test asserts preservation, replacement, and explicit clearing behavior.

brandonpage (approved):

PR #3040, opened 2026-09-17
MSDK 14.x dropped account-resolution caching — ~44x AccountManager IPC increase drives Android ColdStart/WarmBootstrap regression

JohnsonEricAtSalesforce:
## Summary

Restores account-resolution caching that was dropped, which was driving a large increase in `AccountManager` IPC calls on cold-start and warm-bootstrap paths.

- Adds a per-user, single-slot cache for `SalesforceSDKManager.clientManager`, keyed by org/user ID, so repeated access for the same current user re-resolves the account once instead of on every call.
- Routes `getRestClient()` through that cache instead of constructing a `ClientManager` directly. `getRestClient()` is called from `SalesforceActivityDelegate.onResume()`, so it's on the Activity-resume hot path — found during review that it was bypassing the new cache entirely and still paying full `AccountManager` IPC cost on every resume.
- Adds a no-op guard to `registerUsedAppFeature(code, user)` and `unregisterUsedAppFeature(code, user)` so calls that don't actually change the per-user feature set skip the `AccountManager` persistence round-trip. `LoginActivity`'s per-login marker-clearing sweeps call the unregister path for markers that are already unset on every login, so this was a guaranteed no-op write on every login completion before the fix.
- Invalidates the `clientManager` cache in `cleanUp(userAccount)`, the existing choke point every account-removal path (logout, malformed-account purge, corrupt-account cleanup) already funnels through, closing a gap where the cache could return a manager bound to an account that was removed while remaining the "current" identity.
- Adds `@CallSuper` to `cleanUp(UserAccount?)`. Kotlin has no language-level "must call super" enforcement; this makes Android Lint's `MissingSuperCall` check enforce it at build time for any subclass (e.g. `SmartStoreSDKManager`) that overrides it, since skipping the super call would silently reintroduce the invalidation gap above.
- Removes an unused import (`AuthenticatorService.KEY_INSTANCE_URL`) found while working in this file; pre-existing and unrelated to this fix.

## A note on scope

The `clientManager` cache (this PR's core change) touches multi-user account resolution, which is a correctness-sensitive area. It has not yet been reviewed by the original author of that multi-user contract — flagging that explicitly here rather than assuming silence means no concern.

## Test plan

- [x] New/updated unit tests for the no-op guards on `registerUsedAppFeature`/`unregisterUsedAppFeature` (verified fail-without/pass-with the fix)
- [x] New unit tests for `clientManager` cache invalidation on account removal (verified fail-without/pass-with the fix)
- [x] New unit test for `getRestClient()` reusing the cached `clientManager` instead of bypassing it (verified fail-without/pass-with the fix)
- [x] Verified `@CallSuper` actually triggers Android Lint's `MissingSuperCall` at build time by temporarily removing a subclass's `super.cleanUp()` call and confirming the lint error, then restoring it
- [x] `SalesforceSDKManagerTests` + `SalesforceSDKManagerClientManagerTest` run together locally: full pass, no failures
- [x] Confirmed clean compile after the unused-import removal
- [ ] Full sharded `SalesforceSDK` instrumented test suite — deferred to CI


libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1017,13 +1017,36 @@ open class SalesforceSDKManager protected constructor(
...
+         * This is event-driven invalidation, not per-access re-validation:
+         * the cache is only checked for validity here, when an account is
+         * actually removed, not on every clientManager read. Re-validating
+         * on every access would reintroduce the AccountManager IPC cost
+         * this whole cache exists to avoid. Every account-removal path
+         * (logout, purgeMalformedPersistedAccount, and the corrupt-account
+         * path in getRestClient) funnels through this method, so it is a
+         * complete invalidation point for the cache's lifetime.
+         */
+        cachedClientManager = null
```
wmathurin, line 1049: `perUserFeatures` should be cleaned up here as well for the supplied user (for example, remove the `orgId/userId` entry). I verified that this overload is reached on normal logout: `logout()` resolves the `UserAccount`, `removeAccount()` removes the persisted account (or clears credentials), then calls the private `cleanUp(...)`, which invokes this method. The malformed/corrupt-account paths also funnel through it. Leaving `perUserFeatures` alive means logging out and back in as the same user starts the new session with the previous session's login/session markers; the new no-op guards can then skip copying an unchanged marker set into the fresh `UserAccount`, leaving in-memory and persisted flags inconsistent. Please add logout/re-login coverage showing that the same identity starts with a clean feature set and can select a different login type without carrying markers forward.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1593,32 +1629,71 @@ open class SalesforceSDKManager protected constructor(
...
+     * The manager is cached per current-user identity to avoid re-resolving the
+     * AccountManager-backed account on every access; the cache is invalidated
+     * whenever the current user's identity changes.
      */
     val clientManager: ClientManager?
-        get() = userAccountManager.currentUser?.let { user ->
-            ClientManager(appContext, user)
-        }?.takeIf { manager -> manager.account != null }
+        get() {
+            val user = userAccountManager.currentUser ?: return null
```
wmathurin: This still calls `UserAccountManager.getCurrentUser()` on every access. That method does not consume `cachedCurrentUserAccount`; it calls `getCurrentAccount()` and `buildUserAccount()` and then refreshes the cache. Only `cachedCurrentUser` uses the existing cached value. Thus a cache hit avoids the subsequent `ClientManager.buildAccount(user)` resolution, but it does not reduce this path to the 'single cheap key comparison' described above or avoid all account-resolution IPC. Could this use `cachedCurrentUser` for the identity-only fast path (with the existing switch/cleanup invalidations), or should the comments and test be narrowed to the one lookup actually avoided? A lookup-count test would prove the intended performance behavior better than object identity alone.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1593,32 +1629,71 @@ open class SalesforceSDKManager protected constructor(
             }
         """.trimIndent()
 
+    /**
+     * Cache for [clientManager], keyed by the current user's org/user ID pair.
+     *
+     * Holds at most one `(key, manager)` pair, not a per-account map — a
```
wmathurin: The bounded single-slot policy seems reasonable for the stable-current-user hot path, but could we name it explicitly as a 'single-entry current-user cache'? 'Cached per current-user identity' can read as one retained manager per user. The actual A -> B -> A behavior replaces A with B and constructs a new manager when returning to A; it intentionally does not maintain a per-user map.

wmathurin (commented):
Review feedback on the cache lifecycles and documented performance semantics.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Good catch — fixed in `f4b8bda25`. `cleanUp()` now removes this identity's `perUserFeatures` entry (`perUserFeatures.remove("${orgId}/${userId}")`) alongside the existing `cachedClientManager` invalidation. Added `clientManager_afterLogoutAndReloginAsSameIdentity_startsWithClearedFeatureMarkers`, which logs out, re-adds the same identity, and asserts the old login-type marker is gone and a different login type can be selected cleanly. Confirmed it fails against the pre-fix code and passes with the fix.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: You're right — `getCurrentUser()` was re-resolving via `AccountManager` unconditionally, so the doc overstated what the cache avoided. Fixed in `b64e0080c`: the getter now uses `userAccountManager.cachedCurrentUser` for the identity lookup instead. This is safe because `cachedCurrentUser` only needs to be correct for identity (org/user ID) to compute the cache key — `storeCurrentUserInfo` invalidates it on every user switch, so it's never stale across a switch, even though its OAuth fields can lag. Added `clientManager_repeatedAccess_doesNotForceFreshCurrentUserResolution`, which seeds the cache once and asserts the reference is unchanged after repeated `clientManager` access — `getCurrentUser()` always allocates a new `UserAccount`, so an unchanged reference proves the fresh-resolution path wasn't taken. Confirmed fail-without/pass-with. Also updated the KDoc to state which lookup is avoided and why the switch is safe.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Reworded in `a3672cdfa` — comment now leads with "single-entry current-user cache" and spells out the A -> B -> A behavior explicitly (replaces A with B, then constructs a new manager when A becomes current again, rather than retaining A's manager across the trip through B).

wmathurin (approved):
Reviewed the follow-up commits. The per-user feature cache is now cleared on account cleanup, clientManager uses the cached identity path, and the single-entry cache semantics are documented and covered by passing focused tests. The remaining SalesforceSDK CI failures are unrelated expired test credentials.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1593,32 +1641,83 @@ open class SalesforceSDKManager protected constructor(
...
+     * fields are fine here; `storeCurrentUserInfo` invalidates this
+     * underlying cache on every user switch, so the identity itself is
+     * never stale across a switch.
      */
     val clientManager: ClientManager?
-        get() = userAccountManager.currentUser?.let { user ->
-            ClientManager(appContext, user)
-        }?.takeIf { manager -> manager.account != null }
+        get() {
+            val user = userAccountManager.cachedCurrentUser ?: return null
```
brandonpage, line 1682: Using `cachedCurrentUser` here introduces a switch race. `UserAccountManager.storeCurrentUserInfo()` clears `cachedCurrentUserAccount` before publishing the new org/user IDs. A concurrent getter in that window can repopulate the cache from the old IDs; the later preference write does not clear it again. After an A → B switch, this property can therefore continue returning the manager for A until some unrelated invalidation. Please make the identity publication/cache invalidation atomic, or clear again after the IDs are published, and add a deterministic concurrent-switch regression test.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
```diff
@@ -1593,32 +1641,83 @@ open class SalesforceSDKManager protected constructor(
...
      */
     fun getRestClient(
         activityContext: Activity,
         restClientCallback: RestClientCallback,
     ) {
         val account = userAccountManager.currentAccount
         if (account != null) {
-            val user = userAccountManager.buildUserAccount(account)
-            val client = user?.let { ClientManager(appContext, it).peekRestClient() }
+            val client = clientManager?.peekRestClient()
```
brandonpage: `account` and `client` are resolved from separate current-user snapshots. If the current user switches after line 1718, this line can return the client for B while the method either delivers it after selecting A or, if no client is available, calls `logout(account = A)` for a failure from the later snapshot. The old implementation derived both values from the captured account. Please use one identity snapshot for both client creation and any cleanup/logout target, with a switch-race test.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerClientManagerTest.kt
```diff
@@ -111,6 +111,130 @@ class SalesforceSDKManagerClientManagerTest {
         assertManagerBoundTo(retainedManagerA, userA)
     }
 
+    @Test
+    fun clientManager_repeatedAccessForSameCurrentUser_returnsSameCachedInstance() {
```
brandonpage, line 116: These tests cover manager reuse and a retained A client refreshing after switching to B, but they do not exercise the RTR regression this cache change is intended to preserve: rotate RT1 → RT2, read `sdkManager.clientManager` again and confirm the cached manager is reused, then refresh again and assert that the request uses RT2, persists RT3, and does not log out. Because that guarantee currently depends on `peekRestClient()` rehydrating the user and the token provider re-reading persisted state, please add this cached-manager RTR test so a future change in either layer cannot silently reintroduce the logout.

JohnsonEricAtSalesforce:
**Deep Review:** `cleanUp(userAccount: UserAccount?)` is `protected open`, so it's part of the subclassing surface any consuming app extending `SalesforceSDKManager` relies on. Adding `@CallSuper` here means any existing downstream override that skips `super.cleanUp(userAccount)` will start failing Android Lint's `MissingSuperCall` check the next time that consumer runs lint against this updated SDK. This isn't a source/binary-breaking signature change, and any consumer actually hit by it already had a real latent bug (skipping base cleanup), so the annotation is doing its job — but it's a new build-time constraint surfacing across the module boundary, and it isn't called out in the PR body. Worth a sentence there so third-party subclassers aren't left to discover it via a lint failure after upgrading.

JohnsonEricAtSalesforce:
**Deep Review:** `getRestClient()`'s corrupt-account detection changed resolution direction, not just cache-vs-no-cache. Before, it resolved identity via `buildUserAccount(account)` directly off the exact `currentAccount` handle already in hand. After this PR, it goes through the `clientManager` property, which independently resolves identity via `cachedCurrentUser` and then reverse-looks-up an `Account` via `UserAccountManager.buildAccount(user)`. I traced this and the two paths can't diverge under current invariants — `storeCurrentUserInfo`/`clearStoredCurrentUserInfo` are the only writers of the stored current-user identity and both invalidate `cachedCurrentUserAccount`, so `cachedCurrentUser` and `currentAccount` always agree at any single point read. Functionally correct and already covered by the new `getRestClient_withCurrentUser_populatesClientManagerCacheWithDeliveredInstance` test plus the existing corrupt-account tests — but the PR body's technical description only frames this as "reuses the cache," not that the resolution direction flipped. Worth a sentence if the body gets revised again.

JohnsonEricAtSalesforce:
@brandonpage - I've got your review topics in-progress, but I'm taking a little longer this morning to really look them over.  My first pass yesterday uncovered a couple other topics yours helped discover.  Thanks!

JohnsonEricAtSalesforce:
While addressing your review comment on the current-user cache race, I found the same clear-then-write race also existed on the logout path: `clearStoredCurrentUserInfo()` did a locked cache clear followed by an **unlocked** `SharedPreferences.clear().apply()`. A concurrent cache-populate call landing in that unlocked window could rebuild `cachedCurrentUserAccount` from the not-yet-cleared stored IDs and publish a stale result after logout completes. Fixed in `4afc82148` with the same `synchronized(currentUserLock)` shape as the fix for your original comment — flagging separately since it's a derivative of your finding, not something you called out directly.

JohnsonEricAtSalesforce:
Followed up on the `currentUserLock` approach with an alternatives check + call-site impact analysis: considered a generation-counter/seqlock, `ReentrantReadWriteLock`, a lock-free `AtomicReference` swap, and shrinking the critical section to exclude the `AccountManager` IPC. All either reintroduce the original race or don't hold up structurally for a "read that can become a write" shape (`getCachedCurrentUser()` can fall through to `getCurrentUser()` on a cache miss). Traced ~17 call sites for `getCurrentUser()`/`.currentUser` and ~22 for `getCachedCurrentUser()`/`.cachedCurrentUser`: the Activity-resume hot path goes through the separate `@Volatile cachedClientManager`, not `currentUserLock` directly, so lock contention only bites during genuine concurrent identity switches — rare and user-initiated, not steady-state ColdStart/WarmBootstrap. Kept `synchronized(currentUserLock)` as shipped.

JohnsonEricAtSalesforce:
One more finding from self-review: `getRestClient()` built a `UserAccount` from the current account, then `ClientManager.peekRestClient()` rebuilt an independent one internally via `getValidatedUser()` — decrypting the same ~40 `AccountManager` fields twice per call on the Activity-resume hot path. Confirmed this is pre-existing on `upstream/dev`, not introduced by this PR. Fixed in `602046a20` by adding a `peekRestClient(UserAccount)` overload that reuses the caller's already-resolved account while preserving the same liveness re-check and field validation.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Fixed in `ce60caaaf` — `storeCurrentUserInfo()` now clears the cache and writes the new IDs under `synchronized(currentUserLock)`, and `getCurrentUser()`/`getCachedCurrentUser()`/`clearCachedCurrentUser()` synchronize on the same lock, so a concurrent getter can no longer land in the window between the clear and the publish. Added a deterministic test (`testStoreCurrentUserInfoBlocksUntilConcurrentCacheReadCompletes` in `UserAccountManagerTest.java`, currently in `1fa8db6c0` after a later revision — see note below) that pins a reader mid-critical-section via a blocking seam and asserts a concurrent writer genuinely blocks until the reader releases, rather than inferring correctness from many racing trials.

Note: this PR picked up 3 more commits after the original fix landed. One of them (`4afc82148`) found and fixed the same race on the logout path (`clearStoredCurrentUserInfo()` had an equivalent unlocked window); another (`1fa8db6c0`) replaced this test and the logout-path test with the deterministic design described above, since the original stress-test version could theoretically pass without the fix under an unlucky schedule.

libs/SalesforceSDK/src/com/salesforce/androidsdk/app/SalesforceSDKManager.kt
JohnsonEricAtSalesforce: Fixed in `5031a9d44` — `getRestClient()` now reads `userAccountManager.currentAccount` exactly once and derives both the client and the logout target from that same account, via a new `resolveClientManager(user)` helper shared with the `clientManager` property. Added `getRestClient_resolvesCurrentAccountExactlyOnce` (`SalesforceSDKManagerClientManagerTest.kt`) asserting the identity read happens exactly once.

Since then, a related finding from self-review turned up a second, independent identity read on this same path: `ClientManager.peekRestClient()` was internally rebuilding a second `UserAccount` from the same `Account` rather than reusing the one `getRestClient()` already built. Fixed separately in `602046a20` with a new `peekRestClient(UserAccount)` overload.

libs/test/SalesforceSDKTest/src/com/salesforce/androidsdk/app/SalesforceSDKManagerClientManagerTest.kt
JohnsonEricAtSalesforce: Added in `fcb13b568` — `clientManager_survivesTwoSequentialRefreshTokenRotationsWithoutReconstructionOrLogout` rotates RT1→RT2→RT3 against the cached manager, asserting the cache is reused across both refreshes, the second refresh uses the rotated token, and no logout occurs.

brandonpage (approved):

JohnsonEricAtSalesforce:
**Follow-up: resolved a self-caught API-surface regression before merge**

While re-checking CI ahead of merge, I found that an earlier commit on this branch (`602046a20`) had made the new `peekRestClient(UserAccount)` overload `public`, which tripped the `MobileSdk14ApiSurfaceTest` guard that deliberately keeps that overload off `ClientManager`'s public API surface.

Fixed in commit https://github.com/forcedotcom/SalesforceMobileSDK-Android/commit/66f1a24fe369ff87567ecf6f75b1e762802b8b75: the overload is package-private again, and `SalesforceSDKManager` reaches it through a module-internal Kotlin extension (`ClientManagerInternal.kt`) so the API surface stays unchanged while the caller still avoids re-resolving the `UserAccount`. CI on the new head is green apart from the known dev-wide expired-test-credential flake (unrelated to this change).

PR #3042, opened 2026-09-17
Fix developer support for release SDK consumers

brandonpage:
## Summary

- determine developer-support defaults from the consuming application's `FLAG_DEBUGGABLE` state instead of the SDK AAR's `BuildConfig.DEBUG`
- package non-exported developer-support activities in the release AAR while keeping `TestAuthenticationActivity` debug-only
- add regression coverage for debuggable and non-debuggable host applications

## Testing

- `./gradlew :libs:SalesforceSDK:connectedAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.salesforce.androidsdk.app.SalesforceSDKManagerTests` (68 tests)
- `./gradlew :libs:SalesforceSDK:lint`
- `./gradlew :libs:SalesforceSDK:assembleRelease`
- verified a temporary Maven consumer using the release `SalesforceSDK:14.0.0` AAR with both debuggable and non-debuggable host APKs on API 31 and 37
- confirmed the release AAR has `BuildConfig.DEBUG=false`, includes `DevInfoActivity` and `LoginOptionsActivity`, and excludes `TestAuthenticationActivity`

## Compatibility and security

- no public API signatures changed
- developer-support activities remain `android:exported="false"`
- login-screen developer affordances retain their existing debug-app gates
- no new strings, permissions, or dependencies


libs/SalesforceSDK/AndroidManifest.xml
```diff
@@ -79,6 +79,15 @@
...
 
+        <!-- Activities launched from the developer support menu. -->
+        <activity android:name="com.salesforce.androidsdk.ui.DevInfoActivity"
+            android:theme="@style/SalesforceSDK"
+            android:exported="false" />
+
+        <activity android:name="com.salesforce.androidsdk.ui.LoginOptionsActivity"
+            android:theme="@style/SalesforceSDK"
+            android:exported="false" />
+
```
brandonpage, line 90: This needed to be moved from the debug manifest to prevent the app from crashing if they were launched from a release SDK.  

However, I did not move the `POST_NOTIFICATIONS` permission because it is redundant.  Our Firebase dependency already requests it so we do not need to.  Should I remove it from the debug manifest?

libs/SalesforceSDK/src/debug/AndroidManifest.xml
```diff
@@ -3,16 +3,6 @@
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
wmathurin: This declaration is redundant because `firebase-messaging:25.0.2` already declares `POST_NOTIFICATIONS`. I’d remove it from this manifest.

libs/SalesforceSDK/src/debug/AndroidManifest.xml
```diff
@@ -3,16 +3,6 @@
...
-            android:exported="false" />
-
-        <!-- Login Options activity -->
-        <activity android:name="com.salesforce.androidsdk.ui.LoginOptionsActivity"
-            android:theme="@style/SalesforceSDK"
-            android:exported="false" />
-
         <!-- Test Authentication Activity For Automated Testing. This activity is only functional for debug builds of the app using Salesforce Mobile SDK -->
         <activity
             android:name="com.salesforce.androidsdk.util.test.TestAuthenticationActivity"
```
wmathurin, line 5: Minor PR-description correction: the release AAR still contains the `TestAuthenticationActivity` class; only its manifest declaration is excluded. I’d change “excludes `TestAuthenticationActivity`” to “excludes its activity declaration.”

brandonpage: You’re right: the release AAR still contains `TestAuthenticationActivity`; only its manifest declaration is excluded. An R8-minified consuming app may subsequently strip the unreachable class from its final APK.

libs/SalesforceSDK/src/debug/AndroidManifest.xml
brandonpage: Removed the redundant debug-manifest declaration and its comment in 070799aa7. I also verified the RestExplorer debug and release merged manifests still include POST_NOTIFICATIONS from firebase-messaging:25.0.2.

wmathurin (approved):

PR #3045, opened 2026-09-22
@W-24257702 fix: Cache Android DPoP key pairs

wmathurin:
## Summary

- cache Android Keystore-backed DPoP `KeyPair` handles by alias for the SDK process lifetime
- serialize cold load/generation and eviction so concurrent callers converge on one cached object
- evict the in-memory handle when the persistent key is deleted
- document cache-hit, process-restart, and logout behavior

## Performance

On a Pixel 9 Pro XL API 36 emulator, the pre-cache warm lookup median was 743 µs (`n=4`, range 544–8,259 µs). Post-cache warm hits were 2 µs (`n=2`), approximately 371× faster / 99.7% lower and 741 µs saved per proof at the observed median. Cold misses remained 6–11 ms as expected.

The focused AuthFlowTester DPoP UI test reached token exchange but was blocked before its 20-request phase by the existing `Login form remained on screen after password submission` assertion. The measurements therefore use the token-exchange samples captured before that blocker. Temporary instrumentation is not included in this PR.

## Testing

- `DPoPKeyManagerTest`, `DPoPProofBuilderTest`, `DPoPRequestDecoratorTest`, and `OAuth2DPoPTest`: 48/48 passed on the emulator
- `:native:NativeSampleApps:AuthFlowTester:assembleDebug`: passed
- `:native:NativeSampleApps:AuthFlowTester:assembleDebugAndroidTest`: passed

## API changes

No public API changes. The cache reset helper is `internal` and exists for process-restart testing.

## Related

- [W-24257702](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002lFqsOYAS/view)
- [Spec PR](https://git.soma.salesforce.com/SalesforceMobileSDK/SalesforceMobileSDK-Workspace/pull/107)


brandonpage (approved):

libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/dpop/DPoPKeyManager.kt
```diff
@@ -55,8 +56,20 @@ object DPoPKeyManager {
         tokenType?.equals(DPOP_TOKEN_TYPE, ignoreCase = true) == true
 
     private const val TAG = "DPoPKeyManager"
+    private val keyPairCache = ConcurrentHashMap<String, KeyPair>()
+    private val keyStoreLock = Any()
 
     fun generateOrLoadKeyPair(alias: String): KeyPair {
+        keyPairCache[alias]?.let { return it }
```
JohnsonEricAtSalesforce, line 63: This fast path returns a cached `KeyPair` unconditionally. If the underlying `AndroidKeyStore` entry for this alias is ever removed out-of-band (partial app-data clear, KeyStore reset) while the process stays alive, this returns a handle whose private key material no longer exists in the KeyStore. The only caller-side handling today is `DPoPRequestDecorator.attachProof()`'s catch-all, which logs and drops the DPoP header rather than evicting this entry and retrying — see the review body for the full trace. Worth a follow-up: detect a stale handle (e.g. on a `sign()` failure) and re-run `loadOrGenerateKeyPair()` once before giving up on the proof.

JohnsonEricAtSalesforce (approved):
Verified the cache lifecycle claims by tracing the runtime paths rather than just reading the lock declarations:

- **Concurrency**: `generateOrLoadKeyPair()`'s fast path reads `keyPairCache` without a lock; on a miss it re-checks inside `synchronized(keyStoreLock)` before calling the renamed `loadOrGenerateKeyPair()`. Traced the eight-thread test added in this PR — all eight callers converge on the same `KeyPair` instance (`assertSame`), consistent with the double-checked-locking pattern in the diff.
- **Eviction**: `deleteKeyPair()` removes the in-memory entry and deletes the KeyStore entry inside the same synchronized block, so a delete-then-immediately-re-request path can't observe an eviction without the corresponding KeyStore deletion (or vice versa).
- **Public API**: confirmed by inspecting every declaration in `DPoPKeyManager.kt` — the only new symbol, `clearInMemoryCache()`, is `internal`; nothing else changes visibility or signature.
- **AuthFlowTester blocker**: confirmed via `git blame` that the "Login form remained on screen after password submission" assertion predates this branch (authored by a different contributor before this PR's first commit) — the PR description's framing of it as pre-existing and unrelated holds up.

One gap worth a follow-up: the underlying bug report explicitly calls out that a stale cached handle should fail cleanly and fall back to `generateOrLoadKeyPair()` if the underlying KeyStore entry is ever removed out-of-band (partial app-data clear, KeyStore reset) while the process stays alive. Tracing the call path — `DPoPRequestDecorator.attachProof()` → `generateOrLoadKeyPair()` → cached `KeyPair` → `DPoPProofBuilder.sign()` → `Signature.initSign(keyPair.private)` — I don't see that fallback. `attachProof()`'s catch-all (`DPoPRequestDecorator.kt`, not touched by this diff) logs and silently omits the DPoP header rather than detecting the stale handle, evicting it, and retrying with a freshly loaded key. That degrades to "request goes out without a DPoP proof" for a DPoP-bound account, rather than the clean-failure-and-recovery the bug asked for, and no test exercises this path. This doesn't block the fix — the common cache hit/miss/evict path is correct, and this is a narrower edge case than the hot-path problem the PR solves — but it's the one acceptance-criteria item I can't confirm is handled, so flagging it for a follow-up rather than letting it pass silently.

Also worth surfacing, not a defect: the measured warm-hit savings (743µs → 2µs) are noticeably smaller than the ~3–10ms of avoidable overhead the underlying bug estimated from static analysis alone. Both numbers can be correct at once — a warm `KeyStore.getKey`/`getCertificate` lookup landing sub-millisecond while a cold load/generate lands in single-digit milliseconds is plausible — but since the original estimate was explicitly flagged as unprofiled, the two are worth reconciling before citing either externally.

